PR #7, opened 2010-10-02
If function

chriseppstein:
This branch adds an if(bool, when_true, when_false) function.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -6,6 +6,9 @@
 ## 3.2.0 (Unreleased)
 
 * Add an {Sass::Script::Functions#invert `invert` function} that takes the inverse of colors.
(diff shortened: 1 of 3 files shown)


chriseppstein:
I've been using this in compass for a couple weeks now and it's incredibly useful. I've made a couple tweaks to it, for instance, the false value is optional now.

ChristianPeters:
This function deserves its place in the reference. I needed a ternary operator and remembered that I have read something similar somewhere on github some time ago. It took me quite a while to find this pull request...

kflorence:
This doesn't seem to work for variable assignment:

``` sass
$some-variable: if ( true, true, false ); // => "if true, true, false"
```

chriseppstein:
@kflorence you can't put a space after `if`

kflorence:
Yep, that's the problem. Silly.

PR #14, opened 2010-10-11
Perf fixes #2

### LINKED ISSUE #2: CSS @import directives
If a CSS @import is specified anywhere except the start of a document, it should be moved to the before the first selector or else it will be ignored by the browsers.

thedarkone:
Hey Nathan,

I fixed the remaining commits from the previous pull request.


nex3:
Sorry for the long turn-around time on this. We've been seeing some slowdowns on the latest master; we'd welcome your eye for speed increases.

PR #27, opened 2010-11-22
Sass::Plugin::Rack dwell time

AE9RB:
I ran into a performance problem while making https://github.com/dturnbull/googlyscript

I'm shipping googlyscript with custom sass middleware, but the dwell option is well suited for sass core.

### DIFF
--- a/lib/sass/plugin/rack.rb
+++ b/lib/sass/plugin/rack.rb
@@ -27,8 +27,12 @@ module Sass
       # Initialize the middleware.
       #
       # @param app [#call] The Rack application
-      def initialize(app)
+      # @param dwell [Float] A delay, in seconds, which Sass will not check
+      #  for updates.
+      def initialize(app, dwell = 1.0)
         @app = app
+        @dwell = dwell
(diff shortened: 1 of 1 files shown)


nex3:
This seems reasonable; I'll merge it in when I get home.

I do think a better solution here would be to only insert the middleware in the first place in development mode, but a performant default for when people don't do that is a good idea.

AE9RB:
The performance hit is during development.  Google Closure has unique needs.  I'm not deploying Plugin::Rack.  Just a single small .sass file can add 10% to the page load time and a commercial project was slowing down about 300%.  Adding several seconds to every page load during javascript development is really quite painful.  The googlyscript server is capable of 450 request/second on my laptop with a single mongrel and I'm continuing to look for ways to improve that.

nex3:
Oh, I see, the issue was that the large number of imported JS files were each causing an update check? That does sound unpleasant.

AE9RB:
Thanks for adding this.  I'm well over 1000 requests/sec and the browser can't eat any faster so mission accomplished.

PR #38, opened 2010-12-28
fssm, indenting, version fix

ttilley:
This contains the latest FSSM, a new option for specifying what to indent generated css with, and a fix for creating gems with the right version info when you don't have a local copy of the stable branch (switches from using local 'stable' to 'origin/stable' for determining alpha version number).


ttilley:
i'll re-open this at a later date, but it would be convenient if you ganked the commit_count versioning fix. ;)

PR #45, opened 2011-02-17
Allow pathnames with template locations

parndt:
This line of code raises an exception when sass is supplied an array containing one or more pathnames as locations for templates. A simple to_s seems to resolve the issue as it is expecting a string reference.


parndt:
Three weeks and no response, is there something that needs to be improved with my patch?

chriseppstein:
Nope. We've just been busy. Thanks for the patch.

Issue #33, opened 2010-12-10
3.1.0.alpha.50: Sass::Plugin.options[:always_check] = false doesn't work in Rails 2.3.10


agibralter:
`Sass::Plugin.options[:always_check] = false` doesn't work but `Sass::Plugin.options[:never_update] = true` seems to do the trick... all my production servers were running almost 10x slower since I launched them with Sass 3.1.0.alpha.50. Added `:never_update` and things went back to normal.

nex3:
#35 confirms this on Rails 3 as well. It's also worth noting that this dramatically affects test times.

nex3:
I'm having trouble reproducing this in either production or test. You're not spinning up a new server or otherwise causing global state to get reset for each request, are you? Can you provide me a sample rails app in which running a production server or tests locally demonstrates the issue?

To make it clear whether or not it's happening, include this snippet somewhere in your app:

```
module Sass::Plugin
  def update_stylesheets_with_instrumentation(*args)
    puts "Updating stylesheets"
    update_stylesheets_without_instrumentation(*args)
  end
  alias_method_chain :update_stylesheets, :instrumentation
end
```

This will print "Updating stylesheets" whenever Sass is going to touch the filesystem.

agibralter:
Ok -- so I have
    Sass::Plugin.options[:always_check] = false
    Sass::Plugin.options[:never_update] = true
in `config/environments/development.rb` and I put
    RAILS_GEM_VERSION = '2.3.10' unless defined? RAILS_GEM_VERSION
    require File.join(File.dirname(**FILE**), 'boot')
    Rails::Initializer.run do |config|
      # ...
      config.after_initialize do
        # ...
        module Sass::Plugin
          def update_stylesheets_with_instrumentation(_args)
            ::Rails.logger.info "\n\n\n---Updating stylesheets---\n\n\n"
            update_stylesheets_without_instrumentation(_args)
          end
          alias_method_chain :update_stylesheets, :instrumentation
        end
      end
    end
in `config/environment.rb`. I restarted and ran `tail -f log/development.log` only to find `---Updating stylesheets---` appearing in my log...

nex3:
Did you see it more than once? Sass always checks your stylesheets once when the app starts up, but if :always_check is false it shouldn't appear again.

If you do see it more than once, could you produce a minimal Rails app that demonstrates the issue and send it to me?

agibralter:
Yeah so I see it on every request.

[...]

PR #50, opened 2011-03-10
Fix for issue #33

mocoso:
Fix for issue #33 with a failing test

I found it hard to follow your plugin tests and so you may wish to refactor the test that I added


chriseppstein:
Thanks for the patch!

PR #77, opened 2011-05-09
Added --flip support for transformation of stylesheets to right-to-left equivalents

tav:
Hey guys,

I've added a `--flip` option to sass to automatically convert properties/values, e.g. `float`, `background-position`, etc. into their right-to-left equivalents for languages like Hebrew and Arabic.

Despite it not following the usual sass behaviour of not being catering to CSS specifics, it would be nice to have this as a core feature of sass rather than as a plugin...

### DIFF
--- a/lib/sass/environment.rb
+++ b/lib/sass/environment.rb
@@ -21,12 +21,17 @@ module Sass
     attr_writer :options
 
     # @param parent [Environment] See \{#parent}
-    def initialize(parent = nil)
+    def initialize(parent = nil, flip = false)
       @parent = parent
       unless parent
         @stack = []
         @mixins_in_use = Set.new
         set_var("important", Script::String.new("!important"))
+        if flip
+          set_var("dir", Script::String.new("rtl"))
+        else
+          set_var("dir", Script::String.new("ltr"))
+        end
       end
     end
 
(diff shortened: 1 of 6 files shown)


nex3:
This is not something that belongs in the core of Sass. The Sass core doesn't have any specific knowledge of which CSS properties exist or what their semantics are, nor will it apply a transformation that will affect the meaning of plain CSS. This is something I'd encourage you to write as a plugin, preferably using mixins rather than modifying the core Sass code.

tav:
@nex3, yup and thus why I also submitted this feature request: https://github.com/nex3/sass/issues/78

If plugins could register custom visitors during the compilation phases, then that would be great.

tav:
As for using mixins and variables, I've already tried that approach and it fails quite dramatically. It's very easy to miss properties — especially when you're working with a team of developers who are not all familiar with the intricacies of internationalising CSS.

Thanks to Chris for the custom visitors idea btw. That really should solve both of our issues =)

nex3:
It sounds like missing properties is more of a process issue than a technical one. It wouldn't be hard to use the Sass parser to do a simple search over the documents to raise an error for any properties that don't use the internationalization functions. I can see how an automatic transformation would be easier than that, but adding infrastructure for parse-tree transformations is a big step with lots of implications (as I mentioned in #78), and I want more solid use-cases before I go for it.

PR #87, opened 2011-05-19
:cache_location option is ignored

ordinaryzelig:
I used the instructions in the docs for Sass::Plugin::Rack to set the :cache_location option to something other than '.sass-cache'. But the cache_location was being ignored and the default '.sass-cache' was being used. I dug into the code and found out why: Sass::Plugin::Configuration#options was prematurely instantiating a new Sass::CacheStores::Filesystem object with the value of options[:cache_location]. The problem with this is that when I set the :cache_location like in the Sass::Plugin::Rack docs, calling options() immediately instantiates the Filesystem cache store even before I have a chance to change :cache_location. So setting :cache_location is useless.

To fix this, I set the default_proc in the default_options hash to set and instantiate the Sass::CacheStores::Filesystem object. This effectively delays the instantiation until it is needed. By that time, the :cache_location option can be set to something else.

I added the default_proc to the Rails and Merb plugins as well. I wrote tests for them in separate files for each. They were just copies of plugin_configuration_test.rb, but requiring the respective plugin file beforehand. But the Merb plugin required the Merb constant to be defined, which I didn't know how to do, so I removed that test.

So that's 1 commit. The other commit is a small one in the Sass::Plugin::Rack docs where 'merge' should be 'merge!'.

### DIFF
--- a/lib/sass/plugin/configuration.rb
+++ b/lib/sass/plugin/configuration.rb
@@ -10,13 +10,18 @@ module Sass
       #
       # @return [{Symbol => Object}]
       def default_options
-        @default_options ||= {
+        return @default_options if @default_options
+        @default_options = {
           :css_location       => './public/stylesheets',
           :always_update      => false,
           :always_check       => true,
           :full_exception     => true,
-          :cache_location     => ".sass-cache"
-        }.freeze
+          :cache_location     => ".sass-cache",
+          # This is bad. See Sass::Plugin::Configuration#default_options_default_proc.
+          # :cache_store      => Sass::CacheStores::Filesystem.new(options[:cache_location]),
+        }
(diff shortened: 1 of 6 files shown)


josepjaume:
I'm having the same issue. Will this be merged?

kud:
+1

my sass_options = {:cache_location => "./cache/.sass-cache"} isn't used by compass.

MoOx:
Use
`sass_options = {
    :cache_location => "#{Compass.configuration.project_path}/.cache/.sass-cache"
}`

kud:
I've used cache_path = 'cache/sass_cache' to work.

h2non:
Why this pull request was ignored?? I have the same issue
Please, consider to merge it!

nex3:
I believe this has been since fixed. The default cache store is now initialized based on `:cache_location` within the Engine constructor.

PR #94, opened 2011-05-27
Adds back support for ruby 1.8.6

tropperstyle:
This commit broke support for 1.8.6

This reverts commit 5cc60685b2d05050711fc6bb0edb393512a53dd1.


nex3:
Ruby 1.8.6 is officially unsupported. More things than this break, many of which are difficult to notice and difficult to fix.

PR #119, opened 2011-06-14
Revert change that breaks ruby 1.8.6 support

retoo:
Ruby 1.8.6 doesn't have the Hash[array] constructor,
Therefore the cosmetical change in 5cc606 breaks support
for 1.8.6 without providing any advantage (beside the neater
looking syntax).

This reverts commit 5cc60685b2d05050711fc6bb0edb393512a53dd1.


nex3:
[See my response to the Haml pull request](https://github.com/nex3/haml/pull/407).

PR #120, opened 2011-06-15
fixed hash initialization error on ruby 1.8.6

leonardehrenfried:
Even though I'm aware that ruby 1.8.6 is not supported anymore I'm sending this patch, which fixes the following issue: https://github.com/nex3/haml/issues/388

Background: We are running an old version of ruby where the array-to-hash conversion doesn't seem to work so well. It is replaced by a workaround that emulates the desired behaviour.

All tests pass.

### DIFF
--- a/lib/sass/util.rb
+++ b/lib/sass/util.rb
@@ -38,7 +38,12 @@ module Sass
     # @param arr [Array<(Object, Object)>] An array of pairs
     # @return [Hash] A hash
     def to_hash(arr)
-      Hash[arr.compact]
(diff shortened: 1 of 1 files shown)


nex3:
Ruby 1.8.6 is unsupported due to some deep and subtle incompatibilities in the core data structures. While these don't always show up in the tests, they can cause serious and hard-to-track-down bugs. We don't want to make it look as though Sass works with Ruby 1.8.6 when these bugs exist, so we aren't interested in changing the Hash initialization code.

PR #125, opened 2011-06-28
Removed warnings and whitespace

joevandyk:


nex3:
The warning changes are by and large contrary to the Sass coding style, so I'm not accepting them.

I would accept a whitespace cleanup patch, but this one no longer applies cleanly.

PR #137, opened 2011-07-14
X11 color mappings

hudsonhyunlim:
This change is for mapping X11 colors to existing color mapping.  I tried a pull request almost a year ago, but it never got pulled into master due to some code formatting issues.  Here is the previous discussion thread: https://github.com/nex3/haml/pull/258. I noticed that sass split off of haml since last year, so I've forked sass and worked on a up-to-date branch instead of continuing that.

One thing that is still omitted is the camel case naming, as the key lookup normalizes all names to lower case on look up on parser.rb line 291.  I think it's inefficient to have camelcase definition, then lowering all the case by code.  Also, there is quite a bit of ambiguity in color naming; for example, words like 'cornsilk' and 'goldenrod' are one word color names, 'seagreen' is both one word and two words, and 'lavenderblush' is two words.  In the end, browser uses one word only, so camelcasing with this ambiguity in place could be quite a mess.


chriseppstein:
look ok to me. Might cause some regressions for some users who are passing these words as unquoted strings to functions expecting a string object. Seems like an acceptable risk.

@nex3 do you +1?

kflorence:
We are currently experiencing errors due to this issue. When will this make it into the stable branch?

nex3:
When 3.2 releases, which should be soon. I'm afraid I don't have a specific time estimate.

kflorence:
Good enough for me :) Thanks for the quick reply.

PR #138, opened 2011-07-15
Glob args

pointlessone:
In CSS2 functions have simple signatures: fixed number of arguments or optional
trailing arguments. This was implemented in SASS for a long time now. But CSS3
intruduces some new features that require variable arguments number. For
example, linear-gradient function. It expects an optional point or angle
arguments and a set of color stops. The number of color stops is limited only on
the lower end.

The need for variable argument number is expressed in Compass
framework. For example, take a lok at gradients implementation in v0.11.
Currently mixins have signaures of 10 arguments for color stops. And you're in
trouble if you want to have more.

This commit implements variable arguments number only for _functions_. It's more a
proof of concept. I plan to implement it for mixins too.

WIP

### DIFF
--- a/lib/sass/script/parser.rb
+++ b/lib/sass/script/parser.rb
@@ -311,11 +311,20 @@ RUBY
 
         res = []
         must_have_default = false
+        has_glob = false
         loop do
           line = @lexer.line
           offset = @lexer.offset + 1
+          if tok = try_tok(:times)
+            if has_glob
+              raise SyntaxError.new("Function can have only one glob argument.")
+            end
+            var_class = Script::GlobVariable
+          else
+            var_class = Script::Variable
+          end
           c = assert_tok(:const)
-          var = Script::Variable.new(c.value)
+          var = var_class.new(c.value)
           if tok = try_tok(:colon)
             val = assert_expr(:space)
             must_have_default = true
--- a/lib/sass/script/funcall.rb
+++ b/lib/sass/script/funcall.rb
@@ -132,23 +132,51 @@ module Sass
 
       def perform_sass_fn(function, args, keywords)
         # TODO: merge with mixin arg evaluation?
-        keywords.each do |name, value|
-          # TODO: Make this fast
-          unless function.args.find {|(var, default)| var.underscored_name == name}
-            raise Sass::SyntaxError.new("Function #{@name} doesn't have an argument named $#{name}")
+
+        glob_arg = function.args.detect{|arg| arg.first.is_a? Script::GlobVariable }
+        glob_position = function.args.index glob_arg
+        has_glob = !!glob_arg
+
+        unless has_glob
+          if (unknown_args = keywords.keys - function.args.map {|(var, default)| var.underscored_name }).any?
+            raise Sass::SyntaxError.new("Function #{@name} doesn't have named argument#{'s' if unknown_args.length > 1}: #{unknown_args.map{|arg| "$#{arg}" }.join(', ')}")
           end
-        end
 
-        if args.size > function.args.size
-          raise ArgumentError.new("Wrong number of arguments (#{args.size} for #{function.args.size})")
-        end
+          if args.size > function.args.size
+            raise ArgumentError.new("Wrong number of arguments (#{args.size} for #{function.args.size}#{'+' if has_glob})")
+          end
 
-        environment = function.args.zip(args).
-          inject(Sass::Environment.new(function.environment)) do |env, ((var, default), value)|
-          env.set_local_var(var.name,
-            value || keywords[var.underscored_name] || (default && default.perform(env)))
-          raise Sass::SyntaxError.new("Function #{@name} is missing parameter #{var.inspect}.") unless env.var(var.name)
-          env
+          environment = function.args.zip(args).
+            inject(Sass::Environment.new(function.environment)) do |env, ((var, default), value)|
+            env.set_local_var(var.name,
+              value || keywords[var.underscored_name] || (default && default.perform(env)))
+            raise Sass::SyntaxError.new("Function #{@name} is missing parameter #{var.inspect}.") unless env.var(var.name)
+            env
+          end
+        else
+          pre_glob_args = function.args[0...glob_position]
+          post_glob_args = function.args[(glob_position + 1)..-1]
+
+          environment = Sass::Environment.new(function.environment)
+          pre_glob_args.each do |(var, default)|
+            unless kw_value = keywords.delete(var.underscored_name)
+              value = args.shift
+            end
+            environment.set_local_var(var.name,
+              kw_value || value || (default && default.perform(environment)))
+            raise Sass::SyntaxError.new("Function #{@name} is missing parameter #{var.inspect}.") unless environment.var(var.name)
+          end
+
+          post_glob_args.reverse.each do |(var, default)|
+            unless kw_value = keywords.delete(var.underscored_name)
+              value = args.pop
+            end
+            environment.set_local_var(var.name,
+              kw_value || value || (default && default.perform(environment)))
+            raise Sass::SyntaxError.new("Function #{@name} is missing parameter #{var.inspect}.") unless environment.var(var.name)
+          end
+
+          environment.set_local_var(glob_arg[0].name, args + keywords.values)
         end
 
         val = catch :_sass_return do
--- a/lib/sass/script/variable.rb
+++ b/lib/sass/script/variable.rb
@@ -50,5 +50,9 @@ module Sass
         return val
       end
     end
+
+    # A special variable class for glob arguments in function deginitions
+    class GlobVariable < Variable
+    end
   end
 end
--- a/lib/sass/tree/visitors/perform.rb
+++ b/lib/sass/tree/visitors/perform.rb
@@ -171,30 +171,66 @@ class Sass::Tree::Visitors::Perform < Sass::Tree::Visitors::Base
     passed_args = node.args.dup
     passed_keywords = node.keywords.dup
 
-    raise Sass::SyntaxError.new(<<END.gsub("\n", "")) if mixin.args.size < passed_args.size
+    glob_arg = mixin.args.detect{|arg| arg.first.is_a? Sass::Script::GlobVariable }
+    glob_position = mixin.args.index glob_arg
+    has_glob = !!glob_arg
+
+    if has_glob
+      pre_glob_args = mixin.args[0...glob_position]
+      post_glob_args = mixin.args[(glob_position + 1)..-1]
+
+      environment = Sass::Environment.new(mixin.environment)
+      pre_glob_args.each do |(var, default)|
+        val = if kw_value = passed_keywords.delete(var.underscored_name)
+          kw_value.perform(@environment)
+        elsif value = passed_args.shift
+          value.perform(@environment)
+        else
+          default.perform(environment)
+        end
+        environment.set_local_var(var.name, val)
+        raise Sass::SyntaxError.new("Mixin #{@name} is missing parameter #{var.inspect}.") unless environment.var(var.name)
+      end
+
+      post_glob_args.reverse.each do |(var, default)|
+        val = if kw_value = passed_keywords.delete(var.underscored_name)
+          kw_value.perform(@environment)
+        elsif value = passed_args.pop
+          value.perform(@environment)
+        else
+          default.perform(environment)
+        end
+        environment.set_local_var(var.name, val)
+        raise Sass::SyntaxError.new("Mixin #{@name} is missing parameter #{var.inspect}.") unless environment.var(var.name)
+      end
+
+      environment.set_local_var(glob_arg[0].name, passed_args + passed_keywords.values)
+    else
+      raise Sass::SyntaxError.new(<<END.gsub("\n", "")) if mixin.args.size < passed_args.size
 Mixin #{node.name} takes #{mixin.args.size} argument#{'s' if mixin.args.size != 1}
  but #{node.args.size} #{node.args.size == 1 ? 'was' : 'were'} passed.
 END
 
-    passed_keywords.each do |name, value|
-      # TODO: Make this fast
-      unless mixin.args.find {|(var, default)| var.underscored_name == name}
-        raise Sass::SyntaxError.new("Mixin #{node.name} doesn't have an argument named $#{name}")
+      passed_keywords.each do |name, value|
+        # TODO: Make this fast
+        unless mixin.args.find {|(var, default)| var.underscored_name == name}
+          raise Sass::SyntaxError.new("Mixin #{node.name} doesn't have an argument named $#{name}")
+        end
       end
-    end
 
-    environment = mixin.args.zip(passed_args).
-      inject(Sass::Environment.new(mixin.environment)) do |env, ((var, default), value)|
-      env.set_local_var(var.name,
-        if value
-          value.perform(@environment)
-        elsif kv = passed_keywords[var.underscored_name]
-          kv.perform(@environment)
-        elsif default
-          default.perform(env)
-        end)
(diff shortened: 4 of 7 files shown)


lib/sass/script/parser.rb
```diff
@@ -311,11 +311,20 @@ RUBY
 
         res = []
         must_have_default = false
+        has_glob = false
         loop do
           line = @lexer.line
           offset = @lexer.offset + 1
+          if tok = try_tok(:times)
```
pointlessone, line 318: I'm not sure this is the right thing to do. It works for now but probably I need to use a distinct entity here.

chriseppstein:
I'm very excited to review this soon. Thanks for making the patch.

pointlessone:
I have a question on argument precedence.

In the original implementation positional arguments have higher priority than keyword arguments. I think, that's not really right thing to do. For example:

``` sass
@function plus($a, $b)
  @return $a + $b

 a
  a: plus($a: 1, 2)
```

Against the expectations, you'll get error: _"Function plus is missing parameter $b"_.

I think that if I specifically name arguments I want them to have specified values. And all other arguments should be distributed among unnamed arguments. Thus, keyword arguments should take precedence before unnamed ones.

This pull request actually does that for functions/mixins with glob arguments. I think that should be done for functions/mixins without globs. It should not break anything as atm you can pass named arguments only on the tail anyway.

pointlessone:
@chriseppstein This is more a proof of concept. There's a room for cleanup and probably a bit of optimization for speed.

chriseppstein:
related, but distinct we need a corollary syntax for receiving arguments as a glob:

```
@mixin foo($args...)
```

Which would result in `$args` being a comma delimited list of the remaining arguments.

chriseppstein:
regarding argument passing precedence of named arguments I'm worried that it encourages less readable code. I would like to see some real-world examples where this would make the argument passing less confusing.

nex3:
I'm pretty skeptical about the Ruby-style syntax for these. It seems quite incomprehensible to anyone without Ruby familiarity. I'd prefer something along the lines of "foo(bar, baz, @rest bip)" in the manner of Lisp.

chriseppstein:
-1 on using a directive within an argument list to a function or mixin -- that seems very strange. Though I don't have a solid suggestion about what should be the syntax. A few ideas:

An analog to the varargs accepting syntax I mentioned above:

```
@include foo($a, ...$args);
```

A trailing variable _after_ the arglist:

```
@include foo($a) $args;
```

A special function (name is placeholder):

```
@include foo($a, arguments($args))
```

chriseppstein:
@nex3  Do any of my proposed syntaxes feel right to you? I'd really like to see this in 3.2.

Another option:

```
@include foo($a, $args !arguments);
```

@cheba are you still interested in working on this?

pointlessone:
@chriseppstein I'd like to. But I have no time to at least 6th of January. If you guys are short on time for 3.2 you can take it.

Of all the proposed variants I'm for the ellipsis:

``` sass
@include foo($a, ...$args);
```

The last one is even better if the comma is mistakenly missed. Otherwise it's just inconsistent.

``` sass
@include foo($a, $args, !arguments);
```

Also there are a few questions I have in my head. Here are a few of them.
1. Can glob be only the last in signature?
2. What do we do with unknown named arguments?
3. What's the precedence of arguments? At the time of this pull request it was positioned arguments had precedence of named arguments. I believe, it should be the other way: named over positioned.

chriseppstein:
1. I don't see why they have to come last.
2. named arguments always have defaults. I'm not sure what you mean by "unknown named arguments"
3. I'm not sure how I feel about named arguments overridding position (or the other way around). Maybe it should have been an error in the current codebase.

When I see this:

``` scss
@mixin foo($a, $b: 1, $c: 2) {
  .foo {
    a: $a;
    b: $b;
    c: $c;
  }
}

@include foo(x, y, $b: 999);
```

I think there are three fairly reasonable interpretations:
1. positional precedence: `.foo { a: x; b: y; c: 2; }`
2. keyword precedence: `.foo { a: x; b: 999; c: 2; }`
3. keyword inserted into positional location: `.foo { a: x; b: 999; c: y; }` (error if sum of position + keyword > total)

I'd really like to have the ability to pass keyword style var-args as well, but I think this requires a map-like data structure first.

I'm leaning towards a special function `args()`. I think it's the most consistent with CSS.

pointlessone:
The third variant looks right to me (`.foo { a: x; b: 999; c: y; }`). When I specifically name some parameter I care for it to be what I specify and other values just distributed on the parameters left unnamed. Losing passed arguments (2) feels wrong to me.

On `args($arguments)` syntax. It may look closer to CSS but semantically it doesn't feel right. All functions in CSS take values and return values. In this syntax function takes parameter name (not value) and doesn't return anything.

My favorite syntax for glob arguments is the one we have in Ruby (`*args`). Mostly because it allows programmer to specify both position of glob argument and its name. Other languages (such as PHP and JavaScript) have no specific syntax for glob arguments and accept all arguments passed to the function dumping them all into special variable or providing some methods to access all, not only globbed, arguments. In C you just specify that function accepts variable number of arguments and access them by means of macros (function-like syntax).

chriseppstein:
@nex3 I'd really like to get this feature into 3.2 Can you weigh in on syntax please.

nex3:
I'm still not liking the Ruby-style syntax here. The `!`-flag is probably the best way to go if we can come up with a name for it that makes it clear what's going on to someone reading the stylesheet without a ton of familiarity with programming languages. Otherwise maybe `...$foo` is right.

With respect to argument ordering, I feel like these ambiguous cases should all be errors. In Chris's example, `$b` is being given two values, which should cause an error.

I'd like to restrict glob arguments to appearing only at the end of the argument list for simplicity.

richthegeek:
I'm in favour of a prefix-syntax (like !var or *var, or maybe $$var) if this is implemented.

pointlessone:
I'm once again working on this. I've faced a problem with order of parameters.

Currently error is raised if positional (unnamed) parameter passed after named one.

``` sass
@function a($a, $b, $c)
  @return 1
a
  a: a(1, 2, $b: 3)
```

This will raise "Syntax error: Function a was passed argument $b both by position and by name"

On the other hand, mixins report missing parameter $c.

``` sass
@mixin a($a, $b, $c)
  a: a

a
  +a(1, 2, $b: 3)
```

The code above will raise "Syntax error: Mixin a is missing argument $c". I think, this should be unified. At least in terms of consistent message.

Next problem I have can be illustrated by the next example:

``` sass
@function a($a, $b, $c)
  @return 1
a
  a: a(1, $b: 2, 3)
```

This code raises "Syntax error: Positional arguments must come before keyword arguments".

This exception in combination with previous one (on functions) and the desire to have glob argument only at the end makes it impossible to use both glob arguments and named parameters.

``` sass
@function a($a, $b, $c, ...$d)
  @return 1
a
  a: a(1, 2, $c: 3, 4, 5, 6)
  b: a(1, 2, 4, 5, 6, $c: 3)
```

The `a` property will raise "positional before keyword" and the `b` property will result in "position and by name" error.

Possible solutions:
- Drop keyword after position rule
- Give keyword a preference
- Forbid using keywords wherever glob argument is present

#### Drop keyword after position rule

This one I see as the least effective.

``` sass
@function a($a, $b, $c, ...$d)
  @return 1
a
  a: a(1, 2, $c: 3, 4, 5, 6)
```

You have to keep in its position to avoid "name and position" error. This almost annihilates any benefit having named parameter. Maybe just adding a bit of readability for functions and mixins with lots of arguments.

#### Give keyword a preference

``` sass
@function a($a, $b, $c, ...$d)
  @return 1
a
  a: a(1, 2, 4, 5, 6, $c: 3)
  // Same as calling a(1, 2, 3, 4, 5, 6)
```

This is probably the most logical solution. At least I see it as such. This variant eliminates "name and position" error entirely.

Algorithm is following:
1. Pick argument from function/mixin definition
2. Check if we have named parameter for it in the call
3. Take it if we have and go to next argument (step 1)
4. Check if we have any positional parameters left in the call
5. Take it if we have and go to next argument (step 1)
6. Check if we have default value for this argument
7. Take if if we have and go to next argument (step 1)
8. Panic if we don't

After all arguments have been processed pile all positional parameters that left into glob argument if there's one or raise an error about too many parameters.

This solution is compatible with current implementation. But not vice versa, obviously.

The key difference to current implementation if that if there's a named parameter all positional parameters are shifted to the right while current implementation doesn't do that.

It may be illustrated like this:

``` sass
@function a($a, $b, $c)
  @return 1

// Current
   a(1, 3, $b: 2)
   1      3
          $b: 2

// Proposed
   a(1, 3, $b: 2)
   1      $b: 2  3
```

I know that this was proposed before. I've been proposing it. But @nex3 was for simpler Python-like syntax and it was turned down.

#### Forbid using keywords wherever glob argument is present

This is the easiest of all but I wouldn't go for it. It's confusing. It makes programmer chose if he wants to use named parameters or glob args, not both. Also it may brake old code if you add glob arg to a function. All of a sudden you can not use named parameters in calls/includes where you could. Thus instead of improving API we could make it worse.

On the other hand, this is exactly what Python does.

``` python
def a(a, b, *c):
    pass

a(1, 2, 3, 4, 5, 6)  # OK
a(1, b=2, 3, 4, 5, 6) # SyntaxError: non-keyword arg after keyword arg
a(1, 3, 4, 5, 6, b=2) # TypeError: a() got multiple values for keyword argument 'b'
```

---

PS: As there was no consensus or any strong opinions, or decisions from maintainers, I took a liberty to choose `...$arg` syntax. `!arg` looks to similar to `!important` keyword from CSS and may be misleading about its purpose and function.

nex3:
@chriseppstein and I were brainstorming syntaxes for this and ways to integrate var args with keyword arguments... Chris, do you have a link to the gist handy? I believe the syntax we settled on was `$argname...` for both defining and calling the mixin. Also, we specified how that would interact with variable keyword arguments, both in the short and long term.

I'm not a fan of allowing the re-ordering of parameters under any circumstances. I don't think adding variable arguments to a mixin or function will actually break any callers of it. For example, the following would work:

``` scss
@mixin a($a, $b, $c...) { ... }

@include a(1, $b: 2);
```

Since existing callers wouldn't be using the variable arguments anyway, it would work out fine.

pointlessone:
@nex3 I'd like to take a look at the gist. Any pointer to where I can find it?

``` sass
@mixin a($a, $b, $c...) { ... }

@include a(1, $b: 2);
```

Would definitely work. I didn't say that it will break existing code. I said that it will feel awkward for new code.

For example:

``` sass
@include a(1, $b: 2, 3);
```

This will not work as positional argument can not be after named one. That's how it's currently implemented.

``` sass
@include a(1, 3, $b: 2);
```

This again will not work (in current argument handling implementation) as `$b` gets two values. Also it rearranges arguments and you don't want that.

So the only valid (in current implementation) variant is the one without named arguments:

``` sass
@include a(1, 2, 3);
```

chriseppstein:
The write up is here: https://gist.github.com/ce0af5c1886f952978c6

pointlessone:
@chriseppstein @nex3 
Well, that's a great doc.

It looks like arglist doesn't map to any basic data type (like list of hash). I see it acting like a list but this way you can only get positioned arguments out of it. For named arguments you need to extract those out of arglist. I'm not sure that such kind of object orientation is a good idea for stylesheet declaration language such as CSS or SASS. In any case I can't find any other example of such behavior.

The least I could think of to ease this object orientation is to treat positional and named arguments in more symmetric way. Say, treating it like a special data type that doesn't work as a list but a list can be extracted out of it. As well as a hash of named arguments.

Next, I'd prefer to use same syntax for arglist declaration and call expansion. I think, that simplifies syntax a bit. That would eliminate the need to think if I have to place ellipsis after or before var name and instead do something useful. There's simple no way that one can be used in places where the other can. I'm more in favor of trail ellipsis.

BTW, this is illustrated nicely in the section [Expanding a map into keyword arguments](https://gist.github.com/ce0af5c1886f952978c6#expanding-a-map-into-keyword-arguments).

Section [Fixing the semantics of calling with mixed positional and keyword arguments](https://gist.github.com/ce0af5c1886f952978c6#fixing-the-semantics-of-calling-with-mixed-positional-and-keyword-arguments) basically describes what I've written about in [my comment from a few days ago](https://github.com/nex3/sass/pull/138#issuecomment-6835200). Specifically the variant of giving preference to named arguments. I salute you on this decision.

Also I must point that example of current implementation is not correct.

> This means that the following calls are currently equivalent:
> 
> ``` sass
> @include some-mixin(1, $b: 2, 3);
> @include some-mixin(1, 3, $b: 2);
> ```

They are only equivalent in that that they both will produce an error. The first one will give you "positional after named" error and the second one will give "by position and by name" error.

pointlessone:
Commented too soon...

But I think it should stay unedited as the comment only on the document itself.

Now I see some of my concerns have been addressed in comments.

On the other hand, @nex3 basically says that [Fixing the semantics of calling with mixed positional and keyword arguments](https://gist.github.com/ce0af5c1886f952978c6#fixing-the-semantics-of-calling-with-mixed-positional-and-keyword-arguments) section is not the way to go. Which brings us back to "what should we do with arguments?" question.

nex3:
> It looks like arglist doesn't map to any basic data type (like list of hash). I see it acting like a list but this way you can only get positioned arguments out of it. For named arguments you need to extract those out of arglist. I'm not sure that such kind of object orientation is a good idea for stylesheet declaration language such as CSS or SASS. In any case I can't find any other example of such behavior.

I agree that it seems weird to have this be a special type, but I think that's just an emergent property of the inherent weirdness involved in trying to shoehorn powerful function-signature constructs into a language that's designed to have a minimal programming surface area. I don't think there's a good way of handling this case without introducing some sort of new type.

> The least I could think of to ease this object orientation is to treat positional and named arguments in more symmetric way. Say, treating it like a special data type that doesn't work as a list but a list can be extracted out of it. As well as a hash of named arguments.

Case in point: this is also a special type. I don't think requiring the user to write `arguments-list($args)` really provides any value to the user. I suppose it's conceptually a little cleaner, but I think the vast majority of use-cases of variable arguments will either be passing them directly through to another mixin/function or accessing them as a list. I suspect that actually accessing variable keyword arguments is an edge case, and I'm fine with making it weirder/more difficult if that allows us to make the list accessing simpler and easier.

> Next, I'd prefer to use same syntax for arglist declaration and call expansion. I think, that simplifies syntax a bit. That would eliminate the need to think if I have to place ellipsis after or before var name and instead do something useful. There's simple no way that one can be used in places where the other can. I'm more in favor of trail ellipsis.
> 
> BTW, this is illustrated nicely in the section [Expanding a map into keyword arguments](https://gist.github.com/ce0af5c1886f952978c6#expanding-a-map-into-keyword-arguments).

Agree.

> On the other hand, @nex3 basically says that [Fixing the semantics of calling with mixed positional and keyword arguments](https://gist.github.com/ce0af5c1886f952978c6#fixing-the-semantics-of-calling-with-mixed-positional-and-keyword-arguments) section is not the way to go. Which brings us back to "what should we do with arguments?" question.

I'm happy with the current behavior of keyword arguments. As to how they'll interact with variable arguments, I envision it working just as though the arguments were provided long-form. Some examples:

``` scss
$args: foo, bar, baz;
@include mixin(zip, zap, ...$args); // identical to mixin(zip, zap, foo, bar, baz)

@include mixin($kwd: val, ...$args); // Equivalent to mixin($kwd: zip, foo, bar, baz), which is an error

$kwargs: (kw1: val1, kw2: val2);
@include mixin(zip, zap, ...$kwargs); // Equivalent to mixin(zip, zap, $kw1: val1, $kw2: val2)

@include mixin($kw1: val, ...$kwargs); // Equivalent to mixin($kw1: val, $kw1: val1, $kw2: val2), which is an error
```

Note that the latter two examples won't work in v3.2, since we're not supporting an actual Map data type immediately.

PR #151, opened 2011-08-08
Add stylesheet_updated callback

ChristianPeters:
Compass supports a stylesheet_saved callback which used to rely on Sass' updating_stylesheet. The first one expects the file to be written, but the second one is invoked beforehand.

This commit adds the missing callback in order to ensure a consistent behavior of the Compass callback.

Please check whether the added documentation is correct - I am not sure about the part in the middle.


ChristianPeters:
Link to pull request for Compass that depends on this one: https://github.com/chriseppstein/compass/pull/508

chriseppstein:
I don't understand why @nex3 chose to fire the event before the file was written. Having both seems unnecessary. Can we just move when the `updating_stylesheet` event runs to ensure the file has been written?

ChristianPeters:
Okay, how can we make progress in this issue?

I added an alternative pull request, which simply moves the callbacks: https://github.com/nex3/sass/pull/181

Generally, I would prefer the name "stylesheet_updated" instead of "updating_stylesheet", but the latter is consistent to "not_updating_stylesheet" and does not produce any compatibility problems.

So, please, make a choice. :)

chriseppstein:
@nex3 any preference here?

ChristianPeters:
Actually, nex3 has already decided to introduce new callbacks stating: "I agree that "updating" should be changed to "updated" now"

He also said that the old callback should be deprecated, so https://github.com/nex3/sass/pull/181 is the way to go.

PR #159, opened 2011-08-24
Added decimal() function to convert percentage value into decimal value

beausmith:
While a percentage can be converted into a decimal value in plain SASS, a dedicated function is more explicit.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ## 3.2.0 (Unreleased)
 
+* Added `decimal` function to convert percentage value into decimal value
+  (thanks to [Beau Smith](http://about.me/beau))
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -1135,6 +1135,21 @@ module Sass::Script
     end
     declare :percentage, [:value]
 
+    # Converts a percentage to a decimal number.
+    #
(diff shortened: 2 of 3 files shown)


nex3:
I think writing `$number/100%` is easy enough that it's not worth adding an extra function to do it.

beausmith:
Sure, it's easy enough, but far from obvious. Having a function explicitly for this purpose is definitely useful.

nex3:
If users are doing reasonably complex things with units, they'll need to learn how unit arithmetic works eventually. I don't think it's sustainable to try to build in functions for all the common things you might want to do with units.

If a user has to read the docs or ask for help because there isn't a function for this, and they end up getting a better understanding of how units work as a result, I think that's a net benefit.

beausmith:
Thanks for the further explanation Nathan. I completely agree with your logic from educational and sustainability perspective.

From another perspective, if percentage() exists to convert decimals to percentages, it seems reasonable to have a function to provide the inverse action; after all, percentage() can also be substituted with unit arithmetic: `$decimal*100%`

Thanks for all the work you put into making SASS awesome.

philipwalton:
For what it's worth, I had this exact same issue and spent a significant amount of time trying to figure it out before finding this thread.

percent/100% wasn't obvious to me initially.

metaskills:
I found this helpful since I was looking for a clean way to implement a sass extension as part of a gem that provided a few less'ish functions similar to bootstrap-sass or flat-ui-sass.

Anahkiasen:
I'd like a decimal function.

PR #162, opened 2011-08-31
Shouldn't install on Ruby 1.8.6

ericmason:
It's broken so why let it install? :)


chriseppstein:
Thanks for the patch.

PR #167, opened 2011-09-11
Support for a block of styles to mixins.

chriseppstein:
This patch allows passing a ruleset to mixins for placement by the new `@children` directive. It is useful when abstracting a concept associated with selectors or directives.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -5,12 +5,12 @@
 
 ## 3.2.0 (Unreleased)
 
-* A mixin include can now accept a block of styles ({file:SASS_REFERENCE.md#mixin-children Reference Documentation}).
-  The style block will be passed to the mixin and can be placed at the point @children is used. E.g.:
+* A mixin include can now accept a block of content ({file:SASS_REFERENCE.md#mixin-content Reference Documentation}).
+  The style block will be passed to the mixin and can be placed at the point @content is used. E.g.:
   
       @mixin iphone {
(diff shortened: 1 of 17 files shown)


canweriotnow:
Wow... I can already think of multiple use cases for my current project. Awesome!

nheinrich:
FANTASTIC

chriseppstein:
Here's a gist where I first explored this concept. some of the comments there might be relevant: https://gist.github.com/407318

In particular, @hagenburger suggested using `@content` instead of `@children`.

I like `@content` but I'm worried that it might get used by the w3c someday.

smt:
My favorite is `@yield`, but I would prefer `@children` over `@content`.

doctyper:
Awesome!

In the context of `@keyframes`, I think it makes more sense to name the feature `@content` over `@children` (If I understand the syntax correctly):

```
@mixin keyframes {
    @-webkit-keyframes {
        @content;
    }

    @-moz-keyframes {
        @content;
    }
}

@include keyframes {
    0% {
        opacity: 1;
    }

    25% {
        opacity: 0.5;
    }

    50% {
        opacity: 0.25;
    }

    75% {
        opacity: 0;
    }
}
```

`@yield` is interesting though...

chriseppstein:
Ah yes, this is a great use for `@keyframes`!

nex3:
I think we should go for `@content`, with the usual assumption that the w3c moves slowly enough that we can migrate the syntax to avoid a conflict should they add support for it.

PR #175, opened 2011-09-24
Sass::Plugin#force_update_stylesheets mutates options hash

jpterry:
Sass::Plugin#force_update_stylesheets modified compiler.options, but didn't set it back correctly.

If Sass::Plugin.options[:never_update] = true and
force_update_stylesheets was called it could result in options[:never_update] being set to false and later requests could cause compiler.update_stylesheets to be called.

Test included.


nex3:
Closed by cc73198e29547c170aa61365456597c6846a3f5e

PR #181, opened 2011-09-29
Run updating_stylesheet callbacks after files have been written

ChristianPeters:
Compass supports a stylesheet_saved callback which relies on the updating_stylesheet of Sass. The former one expects the file to be written, but the latter one is invoked beforehand.

The changes proposed here ensure that the callback is run after the files have been written.

This PR is an alternative to https://github.com/nex3/sass/pull/151, which adds another callback instead of moving the invocation of updating_stylesheet(s).

The third commit of this PR adds an additional invocation of the updating_stylesheets callback. It is now also run after files found in the template_location are updated. I think this was simply missing beforehand.


nex3:
I agree that "updating" should be changed to "updated" now that these are run after the stylesheet is created. Be sure to alias `#on_updating_stylesheet`, though.

Also, please document this in the Changelog as a non-backwards-compatible change.

ChristianPeters:
Okay, I will add that.

I am currently rethinking my changes to the **updating_stylesheets** callback and need further input.

What is the intention behind this mass update callback? I can see 2 alternative implementations:
1. Notification about n update _candidate_ files that have been detected and are processed in the following
   - The callback is run anyway and yields all files, regardless whether they are stale or fresh
   - The callback is run before any stylesheet is processed
   - The name should stay **updating_stylesheets**
2. Notification about m <= n files that have actually been updated
   - The callback yields only updated files
   - The callback is run after all files are written
   - The name should be changed to **updated_stylesheets**
   - Another callback should be added: **not_updated_stylesheets**, yielding all fresh and therefore untouched stylesheets. Otherwise, we would be inconsistent to **updated_stylesheet** and **not_updated_stylesheet**.

From my point of view, the second option is preferred, as it enables postprocessing to the written css files. Can you think of use cases that rely on the former implementation?

ChristianPeters:
> I agree that "updating" should be changed to "updated" now that these are run after the stylesheet is created. Be sure to alias #on_updating_stylesheet, though.

Hmm. I think the only clean way of doing this (without a hard cut) is to leave the old callback with its old behavior and extend it with a deprecation warning. Aliasing the old callback with the new one could lead to unexpected behavior. I won't write into the deprecation warning: "Hey, this callback is deprecated... actually, it does something different now, which might lead to confusing results!"

So, despite @chriseppstein's concerns about adding another callback, this is what I am going to do now. In difference to  https://github.com/nex3/sass/pull/151, I am going to add a deprecation warning to **updating_stylesheets**, though.

PR #205, opened 2011-10-30
Use same path:line convention as in ruby backtrace, rdebug, rspec etc

sergeykish:
Using same convention everywhere makes it easier to write support in editor.
Path without spaces easier to select.

```
- /* line #{node.line}, #{relative_filename} */
+ /* #{relative_filename}:#{node.line} */
```

This syntax is already adopted by major ruby worlds players.


nex3:
Closing due to unresponsiveness.

Issue #106, opened 2011-06-09
Allow interpolation in static directives


chriseppstein:
[Orginally filed](https://github.com/nex3/haml/issues/229) against haml by @nex3 and commented on by @endymion & @tbranyen.

This should work for `@media` and unrecognized selectors. It may also be worth allowing variables directly at some places in `@media`, although this may create confusion if they aren't allowed directly in other selectors.

chriseppstein:
Patch for review:

https://github.com/chriseppstein/sass/commit/1af19978312e528c32e30e5b99a2175eab284665

micha149:
What about this useful change?!

torkiljohnsen:
+1

nevir:
+1, found myself wanting this as well

PR #217, opened 2011-11-26
Allow interpolation in @media directives.

nevir:
This is a cherry-pick and merge of @chriseppstein's commit in #106 in easy pull request form

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ## 3.2.0 (Unreleased)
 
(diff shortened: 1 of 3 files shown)


nex3:
This commit is more of a proof-of-concept than a full-fledged implementation. Rest assured that I do have an implementation in the works.

chriseppstein:
Yep. that totally didn't work in scss :(

rotev:
Bummers. I really need that :(

chriseppstein:
It already exists in sass 3.2 prereleases. `gem install sass --pre`

PR #220, opened 2011-12-02
Added --indent option to sass-convert

### LINKED ISSUE #107: Make SCSS output style from sass-convert configurable like CSS output
I'd like to have an expanded and nested SCSS output like:

```
#main {
    #header {
        font: {
            size: 12px;
        };
    }
}
```

I hardcoded this behaviour in lib/sass/tree/node.rb of haml 3.0.13 for myself with:

```
--- node_3.0.13.rb  2010-07-01 15:31:02.321652147 +0200
+++ node.rb 2010-07-01 15:22:56.670401567 +0200
@@ -395,7 +395,7 @@ module Sass

         (fmt == :sass ? "\n" : " {\n") +
           children.map {|c| c.send("to_#{fmt}", tabs + 1, opts)}.join.rstrip +
-           (fmt == :sass ? "\n" : " }\n")
+           (fmt == :sass ? "\n" : "\n#{'  '  * tabs}}\n")
       end

      # Converts a selector to a Sass or SCSS string.
```

[Originally filed](https://github.com/nex3/haml/issues/205) by @topaxi and commented on by @nex3 and @graywh

hlb:
It should fix https://github.com/nex3/sass/issues/107
- sass-convert --indent 4 uses 4 spaces
- sass-convert --indent t uses real tabs

There is no option to make the close parenthesis sitting in the same line of last rule, because I think it is hard to use and should be avoid.

### DIFF
--- a/lib/sass/css.rb
+++ b/lib/sass/css.rb
@@ -25,6 +25,9 @@ module Sass
     #     (`:color blue` as opposed to `color: blue`).
     #     This is only meaningful when generating Sass code,
     #     rather than SCSS.
+    # @option options :indent [String] (2)
+    #     Customize the indent size in written files. 2 by default.
+    #     Use "t" if you want to use real tab.
     def initialize(template, options = {})
       if template.is_a? IO
         template = template.read
--- a/lib/sass/exec.rb
+++ b/lib/sass/exec.rb
@@ -515,6 +515,17 @@ END
           @options[:for_tree][:dasherize] = true
         end
 
+        opts.on('--indent indent',
+          'Customize the indent size in written files. 2 by default.',
+          'Use "t" if you want to use real tab.') do |indent|
+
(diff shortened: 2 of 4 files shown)


chriseppstein:
This change also moves the closing `}` to the next line. I think this is a very welcome change. 

Also would like to see the same option in the sass engine and CLI.

I wonder whether a `--oneline` option should also be added that would reformat the properties to be on a single line.

hlb:
@chriseppstein, I do not understand what "reformat the properties to be on a single line" is. Do you mean that we should adding `--oneline` to make close parentheses behave like the old way?

nex3:
I've left a couple comments; I'll be happy to merge this once they're addressed.

hlb:
@nex3 Thank you for the feedback. I have split them into two commits.

PR #223, opened 2011-12-08
make sure Number.round outputs a float instead of rational when mathn is included

twalpole:
When sass is used in a rails project that also uses the mathn standard library and ruby 1.9.3 Sass::Script::Number.round output a Rational when fed a float.  The result of this was rgba(0,0,0,0.5) producing  rgba(0,0,0,500/1000) in the output css.   Also a divide by zero error occurred during the tests in the mix function.   Ensuring floats are used rather than Rationals fixes these issues
I'm not sure how to have tests of the functions run with and without mathn included so havent updated the test cases


nex3:
The fact that this can even happen, and in the standard library no less, is extremely obnoxious. I'll cherry-pick this into stable.

PR #226, opened 2011-12-14
Use on_updated_stylesheet instead of on_updating_stylesheet in Exec

Marzelpan:
Following the instructions in the deprecation warning you get on every call of `sass --update` or `sass --watch`...


Marzelpan:
Argh, this should be merged to the stable branch but applies to master as well :)

PR #238, opened 2012-01-04
Fixes potential issue when parsing comments

pnomolos:
As far as I can tell, using `line` rather than `line.text` is going to result in an error as the `line` variable will cascade down and attempted to be passed to `StringScanner` where it will blow up.

I was having this issue, then fixed it and can't seem to cause it to blow up anymore by reverting my code, but I thought commenting on it still may be helpful.

(I also hope I got the formatting correct above).


excid3:
+1

PR #242, opened 2012-01-08
Extend warnings

chriseppstein:
In porting from base classes to placeholder selectors, I really needed this.

In 3.2 we will warn if a selector being extended is not found. In 3.3/4.0 it will become an error.

The warning can be silenced by placing `!optional` at the end of an `@extend statement`. This maybe useful for libraries and partial re-use.

### DIFF
--- a/lib/sass/engine.rb
+++ b/lib/sass/engine.rb
@@ -676,7 +676,12 @@ WARNING
         raise SyntaxError.new("Illegal nesting: Nothing may be nested beneath extend directives.",
           :line => @line + 1) unless line.children.empty?
         offset = line.offset + line.text.index(value).to_i
-        Tree::ExtendNode.new(parse_interp(value, offset))
+        optional = false
+        if value =~ /\!optional\s*$/
+          optional = true
+          value = value.gsub(/\s*!optional\s*$/,'')
+        end
+        Tree::ExtendNode.new(parse_interp(value, offset), optional)
       when 'warn'
         raise SyntaxError.new("Invalid warn directive '@warn': expected expression.") unless value
         raise SyntaxError.new("Illegal nesting: Nothing may be nested beneath warn directives.",
--- a/lib/sass/scss/parser.rb
+++ b/lib/sass/scss/parser.rb
@@ -289,7 +289,7 @@ module Sass
       end
 
       def extend_directive
-        node(Sass::Tree::ExtendNode.new(expr!(:selector_sequence)))
+        node(Sass::Tree::ExtendNode.new(expr!(:selector_sequence), !!tok(OPTIONAL)))
       end
 
       def import_directive
--- a/lib/sass/selector/abstract_sequence.rb
+++ b/lib/sass/selector/abstract_sequence.rb
@@ -82,6 +82,16 @@ module Sass
         _specificity(members)
       end
 
+      # Marks this selector as extended.
+      def extended!
+        @extended = true
+      end
+
+      # Checks whether this selector has been extended
+      def extended?
+        @extended
+      end
+
       protected
 
       def _specificity(arr)
@@ -89,6 +99,7 @@ module Sass
         arr.map {|m| spec += m.is_a?(String) ? 0 : m.specificity}
         spec
       end
+
     end
   end
 end
--- a/lib/sass/selector/abstract_sequence.rb
+++ b/lib/sass/selector/abstract_sequence.rb
@@ -82,6 +82,16 @@ module Sass
         _specificity(members)
       end
 
+      # Marks this selector as not requiring the base selector to exist.
+      def extension_optional!
+        @extension_optional = true
+      end
+
+      # Checks whether this selector has been extended
+      def extension_optional?
+        @extension_optional
+      end
+
       # Marks this selector as extended.
       def extended!
         @extended = true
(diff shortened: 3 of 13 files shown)


lib/sass/engine.rb
```diff
@@ -671,7 +671,12 @@ def parse_directive(parent, line, root)
         raise SyntaxError.new("Illegal nesting: Nothing may be nested beneath extend directives.",
           :line => @line + 1) unless line.children.empty?
         offset = line.offset + line.text.index(value).to_i
-        Tree::ExtendNode.new(parse_interp(value, offset))
+        optional = false
+        if value =~ /\!optional\s*$/
+          optional = true
+          value = value.gsub(/\s*!optional\s*$/,'')
+        end
```
chriseppstein, line 683: Should this use the `RX::OPTIONAL` constant?

lib/sass/selector/abstract_sequence.rb
```diff
@@ -64,6 +64,43 @@ def has_placeholder?
...
+
+      # Marks this selector as not requiring the base selector to exist.
+      def extension_optional!
+        @extension_optional = true
+      end
+
+      # Checks whether this selector has been extended
+      def extension_optional?
+        @extension_optional
+      end
```
chriseppstein, line 93: I thought about refactoring these four methods into a shared module for `Selector::AbstractSequence` and `Selector::Simple`. Wasn't sure where to put that or what to call it. Alternatively there could be a shared base class for the two kinds of selectors.

lib/sass/selector/sequence.rb
```diff
@@ -20,6 +21,7 @@ def line=(line)
       # @param filename [String, nil]
       # @return [String, nil]
       def filename=(filename)
+        @filename = filename
```
chriseppstein, line 24: this bug baffled me for quite a while. >_<

lib/sass/tree/root_node.rb
```diff
@@ -20,8 +20,24 @@ def render
...
         result, extends = Visitors::Cssize.visit(result)
-        result = result.do_extend(extends) unless extends.empty?
-        result.to_s
+        result.do_extend(extends).to_s
+      end
+
+      def do_extend(extends)
+        return self if extends.empty?
+        result = super
+        extends.each do |e, by|
```
chriseppstein: It seems like the way that this data structure works, it's not possible for this part of the code to properly handle a missing `@extend .foo.bar` because `.foo` and `.bar` are stored as separate keys in the map. As a result, the user gets two warnings in this case. There might be a way to change how the selectors are stored or how each works that would allow for the correct error message in this case. But since this is not a common case, I'm willing to ship with this issue in place. Having slightly confusing warnings in this case is still better than not knowing.

lib/sass/util/subset_map.rb
```diff
@@ -96,6 +95,14 @@ def get(set)
       def [](set)
         get(set).map {|v, _| v}
       end
+
+      def each
+        @hash.each do |k,v|
+          v.each do |(_,_,i)|
```
chriseppstein: do you think it would it be correct if I only yielded once here with v being a map of indexes to values.

chriseppstein:
I have dusted this branch off and am very pleased with it now.

It only gives a single warning for a single unused extend even when multiple selectors are being extended.

You had given a case that you were concerned about (that comment was lost when I rebased against master) but it amounted to:

``` scss
.foo { a: b; }
.bar { c: d; }
.baz { @extend .foo.bar; }
```

Which you claimed wouldn't warn, but it does -- the extend didn't apply and the warn occurs like so:

```
WARNING on line 3:
  Missing selector .foo.bar not found for @extend of .baz.
  This will become an error in a future release.
  To allow this condition, change to:
    @extend .foo.bar !optional
```

I'd like to merge this in.

chriseppstein:
Rebased on master so that it applies cleanly.

lib/sass/engine.rb
nex3, line 683: I don't think it needs to, although `optional = value.gsub!(/\s*!optional\s*$/, '')` would be cleaner.

lib/sass/scss/parser.rb
```diff
@@ -289,7 +289,7 @@ def else_directive
       end
 
       def extend_directive
-        node(Sass::Tree::ExtendNode.new(expr!(:selector_sequence)))
+        node(Sass::Tree::ExtendNode.new(expr!(:selector_sequence), !!tok(OPTIONAL)))
```
nex3, line 292: This should consume any whitespace after `!optional`.

lib/sass/selector/abstract_sequence.rb
```diff
@@ -82,6 +82,26 @@ def specificity
         _specificity(members)
       end
 
+      # Marks this selector as not requiring the base selector to exist.
+      def extension_optional!
+        @extension_optional = true
+      end
+
+      # Checks whether this selector has been extended
```
nex3, line 90: Wrong docstring

lib/sass/selector/sequence.rb
nex3, line 24: Should this change be separated into its own commit?

nex3:
This patch still has false positives. Example:

``` scss
.foo {@extend .bar.baz}
.bar {a: b}
.baz {c: d}
```

This shouldn't trigger a warning, because both `.bar` and `.baz` exist in the document. I think this was adequately established in our last discussion about this behavior.

This is a fundamental issue with the approach in this patch. Effectively what you're doing is marking each individual `@extend` when it matches a selector, but that's never going to account for valid extendees that just empirically match nothing.

I think we have to fall back on a placeholder approach.

PR #271, opened 2012-02-09
Add specs for mixin color-arguments with compressed output style

LTe:
In an earlier version of sass compressed output does not replace color argument to hex value in mixins.

### DIFF
--- a/test/sass/scss/scss_test.rb
+++ b/test/sass/scss/scss_test.rb
@@ -708,6 +708,25 @@ CSS
 SCSS
   end
 
+  def test_mixin_arguments_color_without_quote
+    assert_equal <<CSS, render(<<SCSS, :style => :compressed)
+.foo{color:#ff0}
+CSS
+@mixin foo($var: yellow) {color: $var}
+.foo {@include foo;}
+SCSS
+  end
+
+  def test_mixin_arguments_color_with_quote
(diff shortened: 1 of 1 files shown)


nex3:
I don't understand the purpose of these tests.

LTe:
Version 3.1.3

``` ruby
gem 'sass', '3.1.3'
require 'sass'

@scss = <<SCSS
  @mixin foo($var) { background-image: image_url($var+"_color.png") }
  #main { @include foo(yellow) }
SCSS

puts Sass::Engine.new(@scss, :syntax => :scss, :style => :compressed).render
# => #main{background-image:image_url("yellow_color.png")}
```

But version >=3.1.4 produce

``` ruby
gem 'sass', '3.1.4'
require 'sass'

@scss = <<SCSS
  @mixin foo($var) { background-image: image_url($var+"_color.png") }
  #main { @include foo(yellow) }
SCSS

puts Sass::Engine.new(@scss, :syntax => :scss, :style => :compressed).render
# => #main{background-image:image_url("#ff0_color.png")}
```

Because the behavior has changed, it should be test for it.
What do you think?

nex3:
This has nothing to do with mixins; it's just a question of the format a color uses when it's converted to a string. This behavior change was tested in `SassScriptTest#test_compressed_colors` when the change was made in 0c47835ca389fac1e185c2aefdcec0fb375c4e5f.

PR #275, opened 2012-02-13
added directory support for cli

yitsushi:
I added directory support for cli because we use a lot of file in the same subdirectory (eg: src/ and want to make a single file like application.css and application.min.css) and we felt us naked without this =) If I added this then I think send a pull request.

If the given "file" is a directory
  then :syntax option will be skipped and guessing from file extension.

Example: sass src/ assets/application.css

I know not a perfect and nice additional but works. If i will have time (and someone else don't make it) I make a better solution. If this solution is not ok then please tell me what is the problem.

Thanks: Balazs Nadasdi

### DIFF
--- a/lib/sass.rb
+++ b/lib/sass.rb
@@ -76,7 +76,9 @@ module Sass
   def self.compile_file(filename, *args)
     options = args.last.is_a?(Hash) ? args.pop : {}
     css_filename = args.shift
+
     result = Sass::Engine.for_file(filename, options).render
+
     if css_filename
       options[:css_filename] ||= css_filename
       open(css_filename,"w") {|css_file| css_file.write(result)}
--- a/lib/sass/engine.rb
+++ b/lib/sass/engine.rb
@@ -212,6 +212,22 @@ module Sass
 
       Sass::Engine.new(File.read(filename), options.merge(:filename => filename))
     end
+    
+    def self.for_directory(directory, options)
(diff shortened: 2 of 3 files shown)


nex3:
How is this different than `sass --update src:assets`?

yitsushi:
this command creates separated files
$ sass --update src:assets
$ ls assets
one.css   three.css two.css

and this patch make just one file (or one output stream to stdout)
$ sass src assets/app.css
$ ls assets
app.css

Nothing more =) So i don't need to merge these files with an other app or i don't need to make a main.sass what contains only @imports's

nex3:
Use [partials](http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#partials).

yitsushi:
Oh yeah before i modified this our base.sass was 63 lines thats only @import's 

@import 'page'
@import 'main'
@import 'editor'
@import 'enterpage'
@import 'galleries'
@import 'images'
@import 'members'
@import 'messages'

And i we created a new file we need to add this it to somewhere. =/ So partial is NOT a nice solution for it. =) because a little annoying.

nex3:
Handling many imports has been discussed to death elsewhere. The upshot is that we don't support it because it makes the ordering of the imports unclear, but if you really want it you can define a custom importer.

yitsushi:
Thx =)

Issue #286, opened 2012-02-21
Syntax improvements related to the `&` selector


chriseppstein:
See #282 for the discussion leading to the creation of this umbrella issue.

nex3:
If we do something smart with `& + &`, how should we handle `& > &` or `& &`?

chriseppstein:
I don't think we should try to get into the business of understanding how to interpret combinators with &. Instead I think we should give better expressiveness about what is meant by the author.

For instance:
- `last-selector(&) + last-selector(&)`
- `all-but-last-selector(&) > .foo > last-selector(&)`

bad naming above is on purpose. Also, could consider using pseudo selectors instead of functions like `&::last-component` but this creates potential namespace collisions.

nex3:
I'm not opposed on principle to adding some way of using functions along with `&`, but we'd need to find a good syntax for it. This means something that has a low probability of conflicting with CSS, and ideally something that integrates well with our existing syntax.

Here's one possibility that has a small delta on our existing syntax: we allow `&` as a first-class object in SassScript that evaluates to the parent selector. This allows you to write standard functions that operate on selectors; e.g. `& + #{last-selector(&)}`.

chriseppstein:
It seems like if we are going to a make a literal that represents a selector in SassScript that we're close to allow storing a selector into a variable and allowing variables of type "selector" to be used without interpolation in selectors. I think this could be very attractive.

nex3:
Having selectors be a first-class type (that is, storing them in variables) is very, very different than allowing variables without interpolation in selectors. Once variables are allowed without interpolation somewhere, people will expect all of SassScript to work; see for example #284. There's no good way to support all of SassScript in a selector.

chriseppstein:
agree that these are separable concepts.

Trying to decide if I want to make a case for allowing a simple subset of SassScript that is Selector Safe™.

chriseppstein:
Adding #285 to the list of things to consider here.

ZeeAgency:
I may be wrong but what about a syntax like &(1) where 1 is the number of steps back in the selector nesting chain (not sure how to call it, hopes you understand what I mean...).

For instance :

```
.foo {
    .bar {
        & &(1) {
            a: b;
        }
    }
}
```

Would output :

```
.foo .bar .foo {
    a: b;
}
```

nex3:
[...]

PR #310, opened 2012-03-06
EvaluationContext takes the environment as argument and then extracts options from it

jede:
Hi there!

I encountered a similiar example to the one shown here, and realized that there really wasn't an easy way to do this in ruby as a SASS extension. So I thought it could be a good idea to make the environment accessible to the EvaluationContext.

``` scss
$variable: 5;

@function more() {
  @return $variable + 3;
}
```

With my addition this would now be possible:

``` ruby
def more
  environment.var('variable')
end
```

### DIFF
--- a/lib/sass/script/funcall.rb
+++ b/lib/sass/script/funcall.rb
@@ -103,7 +103,7 @@ module Sass
         unless Functions.callable?(ruby_name)
           opts(to_literal(args))
         else
-          opts(Functions::EvaluationContext.new(environment.options).send(ruby_name, *args))
+          opts(Functions::EvaluationContext.new(environment).send(ruby_name, *args))
         end
       rescue ArgumentError => e
         # If this is a legitimate Ruby-raised argument error, re-raise it.
--- a/lib/sass/script/funcall.rb
+++ b/lib/sass/script/funcall.rb
@@ -103,7 +103,8 @@ module Sass
         unless Functions.callable?(ruby_name)
           opts(to_literal(args))
         else
-          opts(Functions::EvaluationContext.new(environment).send(ruby_name, *args))
+          local_environment = Environment.new(environment)
+          opts(Functions::EvaluationContext.new(local_environment).send(ruby_name, *args))
         end
       rescue ArgumentError => e
         # If this is a legitimate Ruby-raised argument error, re-raise it.
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -309,14 +309,21 @@ module Sass::Script
     class EvaluationContext
       include Functions
 
+
+      # The environment of the {Sass::Engine}
+      #
+      # @return [Environment]
+      attr_reader :environment
+
       # The options hash for the {Sass::Engine} that is processing the function call
       #
       # @return [{Symbol => Object}]
       attr_reader :options
 
-      # @param options [{Symbol => Object}] See \{#options}
-      def initialize(options)
-        @options = options
+      # @param environment [Environment] See \{#environment}
+      def initialize(environment)
+        @environment = environment
+        @options = environment.options
       end
 
       # Asserts that the type of a given SassScript value
--- a/test/sass/engine_test.rb
+++ b/test/sass/engine_test.rb
@@ -11,6 +11,11 @@ module Sass::Script::Functions::UserFunctions
   def option(name)
     Sass::Script::String.new(@options[name.value.to_sym].to_s)
   end
+
+  def set_a_local_variable
+    environment.set_var('variable', Sass::Script::Number.new(5))
+    return Sass::Script::Null.new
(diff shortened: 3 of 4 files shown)


nex3:
What's the use case for this? It seems like it would make it difficult for users to tell where variables are being used.

jede:
In my case I want to be able to set a config variable which is then used in functions. This works well today using SCSS.

Don't you already have that issue with mixins, that it is hard to know when a variable is used?

chriseppstein:
+1 this brings ruby-based functions to parity with sass-based functions.

chriseppstein:
@nex3 In the past you have rejected this functionality claiming that it makes optimization harder. Do you still feel this way? This functionality would enable the dynamic construction of variable names -- not a feature that exists in sass functions.

chriseppstein:
Another thing to note is that having the environment accessible technically allows the creation of variables too.

jede:
Which type of optimization do you mean would be harder?

Imo dynamically created variables would be quite useless since you couldn't access them in the templates (you don't know their names). That would however allow them to be accessed by other ruby implemented functions, but that is already possible since you could, theoretically, create a singleton class to hold all of your custom global variables.

chriseppstein:
A use of dynamically constructed variables would be like what we do for compass sprites: according to a naming convention of the names of folders/files on disk.

nex3:
The optimizations I was concerned about in the past are looking increasingly unlikely, so I'm not particularly opposed to this feature anymore.

christophehurpeau:
I'd like that too, please !

chriseppstein:
@nex3 can we get this in 3.2?

nex3:
I'm still skeptical of allowing Sass functions access to random variables that happen to exist in the scope where the function is called. I'd feel better about this if it only allowed access to the top-level environment that contains global definitions.

chriseppstein:
talked to @nex3 about this, and I agree that ruby functions shouldn't be able to do things that sass functions can't. This would allow access to local variables in the calling context and also setting variables in the local scope.

Snugug:
I'm in need of this too, but I'd like environment to be able to access parent (the full selector string in which it is being called from, or nil if it's the global environment). So, I could have something like the following:

``` scss
$test: 4px;

.bar {
  content: '#{parent()}';

  .qux {
    content: '#{parent()}';
  }
}
.baz {
  width: width();
}
```

``` ruby
def parent
  Sass::Script::String.new(environment.parent)
end
def width
  Sass::Script::Number.new(environment.var('test')
end
```

``` css
.bar {
  content: '.bar';
}

.bar .qux {
  content: '.bar .qux';
}

.baz {
  width: 4px;
}
```

chriseppstein:
@Snugug this use case is covered by #286.

Snugug:
Kinda sorta. While the use case is the same, the methodologies are slightly different. For the usecase and outcome there, he's looking to be able to use the parent selector even if one doesn't exist. In this one, I'm trying to grab the parent selector through the SassScript enviro as a string/list. His is entirely through front-facing Sass, my usecase is backend Ruby, both with the same end result.  Well similar results but with much different ways of getting there which produce different usecases and different potential code.

chriseppstein:
My goal with this patch is that Ruby functions can do all the things that Sass functions can do. Once #286 lands, that would imply that ruby would have a function it could call that would be equivalent to `&` in sass script.

Snugug:
Ahh, all makes sense now.

chriseppstein:
@jede any interest in taking another whack at this?

jede:
@chriseppstein sure! So the problem is, if I understand @nex3 correctly, that ruby methods should not be able to access everything? It would be really good with an example of this (i.e. something this would allow that you cannot do with sass).

nex3:
Here's an example:

``` ruby
def evil(varname, value)
  environment.set_var(varname.value, value)
end
```

``` scss
.foo {
  -no-op: evil('varname', 123px);
  width: $varname;
}
```

The problem is that this would allow functions access to all levels of the environment, including local variables, which should not be accessible by functions, whether Ruby or Sass.

jede:
Great! I'll have a go at it.

jede:
This should work now. I added tests to cover the case @nex3 provided, please have a look and see what you think! Don't know why @travisbot hasn't commented, but the tests pass: http://travis-ci.org/#!/nex3/sass/builds/2377423

nex3:
This does address the issue I brought up, but it's still not completely correct. The code is still able to _read_ variables in the local scope, even if it can't _write_ them.

I think the correct way to handle this is to keep track of the global-scope `Environment` object and pass that to the functions.

chriseppstein:
I took care of the last remaining point here (https://github.com/nex3/sass/commit/576aec61ba7639315494c0b55cd086aaf68b7f40) and merged this to master.

PR #327, opened 2012-03-20
A bit faster (for most cases) check for unknwon named argments in function call

pointlessone:
### Benchmark

**Before**

```
                user       system     total       real
All keywords   21.000000   0.070000  21.070000 ( 21.905063)
No keywords     5.050000   0.020000   5.070000 (  5.534620)
First keyword   4.950000   0.030000   4.980000 (  5.182199)
Last keyword    4.620000   0.020000   4.640000 (  4.815920)
```

**After**

```
                user       system     total       real
All keywords    6.870000   0.040000   6.910000 (  7.220139)
No keywords     5.000000   0.020000   5.020000 (  5.467534)
First keyword   6.210000   0.030000   6.240000 (  7.025294)
Last keyword    4.680000   0.020000   4.700000 (  4.738899)
```

Benchmark setup: function with 100 arguments. Run 10,000 times.

It's about 3x faster for case with all keyword arguments. For no keyword arguments and only one last it's about the same. For first argument case it's a bit slower but it's virtually impossible case. In current implementation, it seems, one can only pass last args as named. I believe that to be a bug but that should be discussed elsewhere.


chriseppstein:
Neat. Any noticeable speedup to the file as a whole?

pointlessone:
@chriseppstein Not that much, just under 100ms on my Compass-based stylesheets. That's very small portion of the overal time and can be dismissed as a measurement error. I assume, that's because most funcalls don't use named arguments.

lib/sass/script/funcall.rb
```diff
@@ -145,10 +145,10 @@ def construct_ruby_args(name, args, environment)
...
       def perform_sass_fn(function, args, keywords)
         # TODO: merge with mixin arg evaluation?
-        keywords.each do |name, value|
-          # TODO: Make this fast
-          unless function.args.find {|(var, default)| var.underscored_name == name}
-            raise Sass::SyntaxError.new("Function #{@name} doesn't have an argument named $#{name}")
+        if keywords.any?
+          unknown_args = keywords.keys - function.args.map {|var| var.first.underscored_name }
+          if unknown_args.any?
+            raise Sass::SyntaxError.new("Function #{@name} doesn't have an arguments: #{unknwon_args.map{|name| "$#{name}"}}")
```
chriseppstein: grammar issue here. "an arguments", and we should use plural if there's just one.

pointlessone: Do we have any tools for pluralization? Or something simple will do? Like this:

``` ruby
raise Sass::SyntaxError.new("Function #{@name} doesn't have an argument#{'s' if unknown.args.length > 1}: #{unknwon_args.map{|name| "$#{name}"}}")
```

Also it looks, like I've forgot to join unknown args into string…

chriseppstein: We have not written any such utility and we don't depend on active support. So that will have to do.

Also, seems strange that you didn't have to update any tests regarding our error messages. Maybe we can add an error message test case for multiple unknown args.

pointlessone: Well, all tests pass. I assume, there's no tests for this specific error message. Probably, should add, though.

PR #361, opened 2012-04-21
Add `nil` literal

petebrowne:
Here's a common use use case for a `nil` literal:

``` scss
@mixin block-colors($background-color, $border-color: nil) {
  $border-color: darken($background-color, 10%) !default;

  background-color: $background-color;
  border-color: $border-color;
}
```

This is semantically different than using `false` in conjunction with an `@if` directive. Plus, guarded assignment can be used to set a dynamic default value.

### DIFF
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -627,12 +627,13 @@ this still works, but it's deprecated and prints a warning.
 
 ### Data Types
 
-SassScript supports four main data types:
+SassScript supports six main data types:
 
 * numbers (e.g. `1.2`, `13`, `10px`)
 * strings of text, with and without quotes (e.g. `"foo"`, `'bar'`, `baz`)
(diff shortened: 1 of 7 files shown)


nex3:
I think this is a good thing to add, but I think "null" is a better name for this. It's used by a wider variety of languages, including in particular Javascript.

I'll add more comments on the patch.

petebrowne:
I agree with the name change - I'll work on incorporating the changes you're suggesting. Thanks

Anahkiasen:
I agree with the idea that null == 0 (but null !== 0), seems simpler and more logical to me.

petebrowne:
What do you think @nex3 @chriseppstein? Should `1 + null` return `1` or raise a SyntaxError?

chriseppstein:
Syntax error. We can relax it later if we feel it's wrong, but changing to syntax error later is very hard.

nex3:
I'm not at all fond of Javascript's aggressive type coercion. `1 + null` should be an error, and `null != 0`.

petebrowne:
I've opened a separate pull request, #374, with the suggested changes.

PR #374, opened 2012-05-02
Add null literal.

petebrowne:
Modified pull request from #361 with suggested changes.

I've left string operations with `null`s as they were:

``` scss
selector {
  property: 'foo' + null; // 'foo'
  property: 'foo' - null; // 'foo-'
  // etc...
}
```

But added a SyntaxError whenever a `null` is involved in other operations:

``` scss
selector {
  property: null + 1; // Sass::SyntaxError
  property: black / null; // Sass::SyntaxError
  // etc...
}
```

Also, we discussed properties not printing when they have empty values. Currently, a SyntaxError is thrown in that case. Is there any reason to change from this?

### DIFF
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -627,12 +627,13 @@ this still works, but it's deprecated and prints a warning.
 
 ### Data Types
 
-SassScript supports four main data types:
+SassScript supports six main data types:
 
 * numbers (e.g. `1.2`, `13`, `10px`)
 * strings of text, with and without quotes (e.g. `"foo"`, `'bar'`, `baz`)
 * colors (e.g. `blue`, `#04a3f9`, `rgba(255, 0, 0, 0.5)`)
 * booleans (e.g. `true`, `false`)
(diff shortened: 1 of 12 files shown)


chriseppstein:
@petebrowne The reason to omit properties that are set to null, an empty string, or an empty list is that it optimizes a very common case of omitting a property that not explicitly passed without a bunch of `@if` boilerplate. It's also not obviously a conceptual error because if you set a property to nothing has a very similar feeling to not setting it.

petebrowne:
The following will still throw a SyntaxError:

``` scss
selector {
  property: ;
}
```

But properties with _resolved_ values that are empty are omitted:

``` scss
$null-value: null;

selector {
  a: null-value;
  b: $null;
  c: 'foo';
}
```

Would produce:

``` css
selector {
  c: 'foo'; }
```

nex3:
I merged this in, with the change that `null` now throws an error for all non-boolean operations, including with strings.

petebrowne:
Awesome! @chriseppstein, @nex3 Thanks for the help & suggestions

Issue #279, opened 2012-02-15
Variables in @import


Snugug:
Currently when using @import, you need to hardcode in uri. I'd like the ability to pass in a variable/function call to get the uri and use that for the @import call, something like the following:

```
@function drupal-theme($project, $append: '') {
  @return /sites/all/themes/#{$project}/#{$append};
}
…
@import #{drupal-theme('my-theme', 'sass/style)};
```

Which would be the same as:

```
@import '/sites/all/themes/my-theme/sass/style';
```

This would also make it easier for a master import for sub-partials a la Compass framework stuff:

```
$compass-css3: 'border-radius' 'inline-block' 'opacity' 'box-shadow' 'text-shadow'…;
@each $css3 in $compass-css3 {
  @import 'css3/#{$css3};
}
```

nex3:
Allowing dynamic imports would completely break Sass's ability to quickly determine which files import which other files, and thus what needs re-compiling when.

julkue:
This is not unnecessary because the output file of the compiled .css file could be everywhere. I think a dynamic import would be great!

jslegers:
I agree with @Snugug. IMO, dynamic imports are essential for flexible theming.

I'm thinking especially of the following feature :

![File structure](http://i59.tinypic.com/5nodph.png)

``` scss
$basetheme : 'default';
$subtheme : 'blueoyster';

//----------------------------------------------------
// DO A LOT OF OTHER STUFF HERE
//----------------------------------------------------

@import 'themes/' + $basetheme + '/__styles';
@import 'themes/' + $subtheme + '/__styles';
```

Another obvious use case would be polyfilling (https://github.com/sass/sass/issues/739).

johanroug:
agree.. just tried, but then I stumpled on this post

nod:
From a deployment standpoint, being able to dynamically set different import paths during compile is pretty huge.

uxtx:
+1

spoeken:
+1

joaomilho:
:+1:

jlukic:
Currently this missing feature is blocking the SCSS port of Semantic UI

In LESS we use this simple [theme bootstrapper](https://github.com/Semantic-Org/Semantic-UI/blob/next/src/theme.less) to process theme inheritance, but it requires the ability to use variables inside `import` statements.

smichaelsen:
+1

mdix:
+1

niklas-r:
[...]

PR #395, opened 2012-05-19
Mention ability to import a directory

alanhogan:


nex3:
This is a Rails feature, not a Sass feature. Nor will it ever be (see e.g. #279).

alanhogan:
Hmm, what if we noted that you can’t import a directory, except using sass-rails?

alanhogan:
Because personally, I wouldn’t have thought to check sass-rails documentation on how what I perceived as a sass feature works.

PR #406, opened 2012-06-01
Don't die if a UTF encoding isn't supported.

agarbutt:
sha: f1a7936a025c145363041dcb939aa6909e67432b incorrectly introduced a missing constant. It should be Encoding::ConverterNotFoundError.


agarbutt:
Poke. 

Can anyone suggest how SASS is being tested against multiple rubies so that i may be able to add tests to this particular change.

nex3:
This has been merged.

PR #439, opened 2012-07-04
Add `adjust-by` as a built-in number function

StanAngeloff:
This function allows any number, incl. those with units, to be
manipulated by adding to/subtracting from them. Example:

```
adjust-by(10px, 10%) => 11px
```

In template design, it is very often the case a unit is selected
and all computations then cast between the selected unit. A very
hacky approach already exists:

```
$base: 10px;
($base / 1px) * 1.1 * 1px => 11px
```

but is less than ideal.

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -115,6 +115,9 @@ module Sass::Script
   # \{#percentage percentage($value)}
   # : Converts a unitless number to a percentage.
   #
+  # \{#adjust_by adjust-by($value, $amount)}
+  # : Adjusts a number by the given amount in percentage.
+  #
   # \{#round round($value)}
   # : Rounds a number to the nearest whole number.
   #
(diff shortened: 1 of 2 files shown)


chriseppstein:
Actually it's much simpler than that. The 1px in the numerator and denominator cancel each other out, so it's just:

```
$base * 1.1 => 11px
```

If the % unit was present you'd need to do:

```
$factor: 10%;
$base * (100% + $factor) / 100% => 11px
```

Wrapping that up into a single sass function we get:

``` scss
@function adjust-by($base, $factor) {
   @if unitless($factor) {
    @return $base * (1 + $factor);
   }
   @else {
    @return $base * (100% + $factor) / 100%;
   }
}
```

Given that this is fairly trivial to author in sass and that the name might have non-numeric connotations, I think it shouldn't be added to Sass.

PR #441, opened 2012-07-06
Support @import inline(file.css) for sass 3.2

pwnall:
Currently, `@import "file.scss"` inlines the contents of the SCSS file into the output, but `@import "file.css"` generates an `@import url(file.css)` CSS directive.

This inconsistency is surprising, and the surprise is compounded by the fact that Chrome/Firefox and IE resolve relative @import URLs differently (unsurprisingly, IE gets it wrong), so a developer might not even realize what's going on when testing in Chrome/Firefox.

Although renaming the offending file to use the `.scss` or `.sass` extension seems to be the obvious solution, renaming can be undesirable for  3rd-party files, such as jquery.js, as it slightly obscures the origin of the code.

This change adds a new syntax, `@import inline(file.css)`, that can be used to express the desire of inlining the imported file. `@import inline()` has the behavior no matter what the imported file's extension is. Together with `@import url()`, it offers a decent path for migrating away from `@import "*.css"`, which can be deprecated later down the line.

I don't have a good understanding of the sass codebase, and I look forward to your feedback on how I can make this change better. I hope you'll consider including it, so other developers won't bump into this problem like I did.

Thank you very, very much for the awesomeness that is sass!

### DIFF
--- a/lib/sass/importers/filesystem.rb
+++ b/lib/sass/importers/filesystem.rb
@@ -73,7 +73,15 @@ module Sass
       #
       # @return [{String => Symbol}]
       def extensions
-        {'sass' => :sass, 'scss' => :scss}
+        {'sass' => :sass, 'scss' => :scss, 'css' => :scss}
+      end
+
+      # The keys of the extensions hash, ordered by their import priority.
+      #
+      # @return [Array{string}] all the keys of the extensions hash, sorted in
+      #   the order used by import to resolve names without extensions.
+      def sorted_extensions
+        ['sass', 'scss', 'css']
       end
 
       # Given an `@import`ed path, returns an array of possible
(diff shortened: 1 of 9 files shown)


pwnall:
This is roughly the same as https://github.com/nex3/sass/pull/442 but rebasing wasn't trivial, so I figured it's better to submit a separate change.

nex3:
Duplicate of #442.

windyGex:
well.i support the 'inline' syntax.

windyGex:
@pwnall but i find this syntax not support relative path:

```
@import inline("../../core.css")
```

will be error ~

pwnall:
@windyGex Sorry, but @nex3 has already rejected this proposal in #442

PR #442, opened 2012-07-06
Support @import inline(file.css) for sass 3.1.x

pwnall:
Currently, `@import "file.scss"` inlines the contents of the SCSS file into the output, but `@import "file.css"` generates an `@import url(file.css)` CSS directive.

This inconsistency is surprising, and the surprise is compounded by the fact that Chrome/Firefox and IE resolve relative @import URLs differently (unsurprisingly, IE gets it wrong), so a developer might not even realize what's going on when testing in Chrome/Firefox.

Although renaming the offending file to use the `.scss` or `.sass` extension seems to be the obvious solution, renaming can be undesirable for  3rd-party files, such as jquery.js, as it slightly obscures the origin of the code.

This change adds a new syntax, `@import inline(file.css)`, that can be used to express the desire of inlining the imported file. `@import inline()` has the behavior no matter what the imported file's extension is. Together with `@import url()`, it offers a decent path for migrating away from `@import "*.css"`, which can be deprecated later down the line.

I don't have a good understanding of the sass codebase, and I look forward to your feedback on how I can make this change better. I hope you'll consider including it, so other developers won't bump into this problem like I did.

Thank you very, very much for the awesomeness that is sass!

### DIFF
--- a/lib/sass/importers/filesystem.rb
+++ b/lib/sass/importers/filesystem.rb
@@ -73,7 +73,15 @@ module Sass
       #
       # @return [{String => Symbol}]
       def extensions
-        {'sass' => :sass, 'scss' => :scss}
+        {'sass' => :sass, 'scss' => :scss, 'css' => :scss}
+      end
+
+      # The keys of the extensions hash, ordered by their import priority.
+      #
+      # @return [Array{string}] all the keys of the extensions hash, sorted in
+      #   the order used by import to resolve names without extensions.
+      def sorted_extensions
+        ['sass', 'scss', 'css']
       end
 
       # Given an `@import`ed path, returns an array of possible
@@ -86,11 +94,10 @@ module Sass
       def possible_files(name)
         name = escape_glob_characters(name)
         dirname, basename, extname = split(name)
-        sorted_exts = extensions.sort
         syntax = extensions[extname]
 
-        return [["#{dirname}/{_,}#{basename}.#{extensions.invert[syntax]}", syntax]] if syntax
-        sorted_exts.map {|ext, syn| ["#{dirname}/{_,}#{basename}.#{ext}", syn]}
+        return [["#{dirname}/{_,}#{basename}.#{extname}", syntax]] if syntax
+        sorted_extensions.map {|ext| ["#{dirname}/{_,}#{basename}.#{ext}", extensions[ext]]}
       end
 
       def escape_glob_characters(name)
--- a/lib/sass/scss/parser.rb
+++ b/lib/sass/scss/parser.rb
@@ -283,7 +283,7 @@ module Sass
       end
 
       def import_arg
-        return unless arg = tok(STRING) || (uri = tok!(URI))
+        return unless arg = tok(STRING) || (inline = tok(INLINE)) || (uri = tok!(URI))
         path = @scanner[1] || @scanner[2] || @scanner[3]
         ss
 
@@ -293,7 +293,11 @@ module Sass
           return node(Sass::Tree::DirectiveNode.new("@import #{arg} #{media}".strip))
         end
 
-        node(Sass::Tree::ImportNode.new(path.strip))
+        if inline
+          return node(Sass::Tree::ImportNode.new(path.strip, true))
+        end
+
+        node(Sass::Tree::ImportNode.new(path.strip))    
       end
 
       def use_css_import?; false; end
--- a/lib/sass/scss/rx.rb
+++ b/lib/sass/scss/rx.rb
@@ -99,6 +99,7 @@ module Sass
       NUMBER = /#{NUM}(?:#{IDENT}|%)?/
 
       URI = /url\(#{W}(?:#{STRING}|#{URL})#{W}\)/i
+      INLINE = /inline\(#{W}(?:#{STRING}|#{URL})#{W}\)/i
       FUNCTION = /#{IDENT}\(/
 
       UNICODERANGE = /u\+(?:#{H}{1,6}-#{H}{1,6}|#{RANGE})/i
--- a/lib/sass/tree/import_node.rb
+++ b/lib/sass/tree/import_node.rb
@@ -9,9 +9,15 @@ module Sass
       # @return [String]
       attr_reader :imported_filename
 
+      # True if the imported file is wrapped by an inline() directive.
+      # 
+      # @return [Boolean]
+      attr_reader :inline_directive
+
       # @param imported_filename [String] The name of the imported file
-      def initialize(imported_filename)
+      def initialize(imported_filename, inline_directive = false)
         @imported_filename = imported_filename
+        @inline_directive = inline_directive
         super(nil)
       end
 
@@ -29,6 +35,8 @@ module Sass
       #
       # @return [Boolean] Whether or not this is a simple CSS @import declaration.
       def css_import?
+        return false if @inline_directive
+
         if @imported_filename =~ /\.css$/
           @imported_filename
         elsif imported_file.is_a?(String) && imported_file =~ /\.css$/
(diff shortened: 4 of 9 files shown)


nex3:
This is not something that we're interested in supporting. It's easy to rename your CSS files to SCSS, enough so that it's not worth adding new syntax to avoid it.

kflorence:
While that is true, in cases where you can't rename the files (because they are owned by a 3rd party, for example), this is a little annoying.

MattKetmo:
+1 for with feature, renaming a "vendor" css file is not a good practice imo.

LoicMahieu:
+1 for this feature.

Like said before, renaming a file for this reasons is not advisable.

SpeCT:
+1 for this feature

paprikka:
+1 for this feature

Renaming vendor files isn't considered a good practice.
By the way, it would play really nicely with Bower modules.

paprikka:
Sorry, I've used a wrong expression here, I should have said bower **components**. 

Basically these are just folders downloaded from different paths (bower directory, github repos etc...).

What do you exactly mean by _submodule_? Git submodule?

http://bower.io/

LoicMahieu:
Sorry I was completely out of subject. I thought reply to another comment. I was too fast :)

And yes, I'm completely +1
It would be very useful for all vendor files.

devinfd:
+1. I am in need of this feature right now for this very reason. I need to import some bower component style sheets but I am not willing to change their file names to .scss because I'll have to do that every time I update that component. @import inline(file.css) is a necessary feature.

LoicMahieu:
@devinfd To solve this temporarily we use https://github.com/suisho/grunt-cssjoin

paprikka:
Thanks for the advice, I'm using grunt-contrib-concat at the moment, but I think it would be nice not to pollute gruntfile with actual project data.

Anyway, for those, who would like to use this approach, I suggest putting imports in a different JSON file and then read it with: `grunt.file.readJSON` if you're masochistic enough to care:)

``` json
{
  "scripts": [
    "vendor/angular-ui-utils/modules/ie-shiv/ie-shiv.js",
    "vendor/jquery/jquery.js",
    "vendor/es5-shim/es5-shim.js",
    "vendor/lodash/lodash.js",
    "vendor/angular/angular.js",
    "vendor/sass-bootstrap/js/bootstrap-modal.js",
    "vendor/sass-bootstrap/js/bootstrap-transition.js",
    "vendor/angular-bootstrap/ui-bootstrap-tpls.js",
    "vendor/angular-bootstrap/ui-bootstrap.js",
    "vendor/angular-cookies/angular-cookies.js",
    "vendor/angular-resource/angular-resource.js",
    "vendor/angular-sanitize/angular-sanitize.js",
    "vendor/angular-polyfill/angular-polyfill.js",
    "vendor/toastr/toastr.js"
  ]
}
```

and then in gruntfile you could just use (I'm using coffeescript here):

``` coffeescript
module.exports = (grunt)->
  imports = grunt.file.readJSON 'app/imports.json'
```

``` coffeescript
    concat:
      vendorScripts:
        # Put specific bower-related vendor imports here. Don't include complete vendor dir.
        src: imports.scripts
        dest: 'public/js/vendor.js'
```

IainIsCreative:
> This is not something that we're interested in supporting. It's easy to rename your CSS files to SCSS, enough so that it's not worth adding new syntax to avoid it.

I highly disagree with this. I get it that your CSS files can become SCSS/SASS files, but what if I have to import a url from a seperate domain? For example; what if my font face needs an import link for the font I'm adding to the project? What do I do then?

I'd rather keep all my CSS down to one file but the fact I can't import an external url for something like a font license is a bit painful.

cimmanon:
> I get it that your CSS files can become SCSS/SASS files, but what if I have to import a url from a seperate domain?

Unless I've missed something, you can't do this with Sass files either.  I tried this yesterday when someone asked about it on SO just to make sure.

https://github.com/nex3/sass/issues/712

paprikka:
The point is - renaming css files manually is not a problem when you deal with small or mid-sized projects, but it a horrible thing to maintain if you want to use components and separate vendor files from your own. On the other hand such separation of resources is gradually becoming a standard approach, so at some point SASS **will just have to** embrace it. 

Take bower, polymer and angular js for an example, think of typical Brunch.io or Yeoman project structure - ability to import a CSS file would help a lot here.

chriseppstein:
This is a use case I want to support. I like the idea of importing css files without having to rename them. We plan to revamp our importing system in 4.0, I want to add this ability at that time. It should be possible to implement a CSS importer as a ruby extension to sass.

IainIsCreative:
> I want to add this ability at that time. It should be possible to implement a CSS importer as a ruby extension to sass.

Please do. I really like SASS and Compass, but the only drawback is linking other CSS files (or anything that can be an imported url, i.e. MyFonts license links), so here's hoping a solution is coming. :-)

kflorence:
We've solved this temporarily by scraping our Sass files for import URI's and including them manually from the file system. Looking forward to a nicer solution :)

nex3:
I agree with Chris that this is something we're likely to support with the import revamp.

pwnall:
@nex3, @chriseppstein Please let me know if I can help, once you've figured out the syntax and how this would fit in the codebase!

thom4parisot:
So what are the plans currently? It's kinda weird to rename a file coming from another build process just to inline it in Sass.

chriseppstein:
I wrote a plugin to provide this capability for now: https://github.com/chriseppstein/sass-css-importer

jarrodpayne:
+1 for this. I went to import some **Topcoat** .css files only to find out it wasn't possible. Sure, I could abandon my workflow using **Bower Components** for this, but it seems like this would be something on the roadmap unless there is some reason why this shouldn't be possible.

shahin8r:
+1 for this feature.

Just started with Sass and the first thing I noticed is that I can no longer import fonts from external domains.

xrado:
+1

hoeni:
+1 renaming CSS files in bower_components does not seem a good practice to me

rmariuzzo:
+1 for this feature! Y U NO include this!?

cdCorey:
+1 to this feature. CSS != SCSS, and we shouldn't be required to _deceptively rename our files_ to make this work. We're literally being told that we should work around this arbitrary limitation of Sass by lying to the Sass compiler -- and by extension, to our fellow developers! It may be easy, but it's clearly bad practice. It's disgusting that such a basic, obvious, useful, intuitive feature is not being supported when it could be so easily.

jarrodpayne:
@nex3 Could you please explain why this is something your not interested in supporting given the context of needing vendor CSS files? Perhaps a reasonable explanation would help those of us who :+1: this PR.

nex3:
As I've mentioned earlier in this thread, we're planning to support this with our upcoming revamp of the `@import` syntax. Given that, it doesn't make sense to make a backwards-incompatible change to the semantics of `@import` or to introduce a new syntax that will only be applicable for a relatively short time.

jarrodpayne:
@nex3 Your right, you did mention this previously. Certainly, this seems reasonable enough for me to drop it :)

Thanks for the explanation.

pwnall:
@nex3 My offer (to help out with the grunt work once you have the design figured out) still stands :heart:

pthrasher:
@nex3 Is there any timeline for the import revamp?

This is a feature I need now, and I see this is an old issue. Any updates? I'm also more than happy to provide help where needed as I'm using sass professionally.

nex3:
We generally don't provide explicit timelines because time estimation is difficult and circumstances change frequently.

Right now, my first priority is getting performance up to snuff by compiling Sass to Ruby before executing it. Once that's finished, the import revamp is next on the list.

AndrewDryga:
+1

phoenixgao:
+1

Issue #25, opened 2010-11-22
Change equality operators


chriseppstein:
I think `==` is very programmery and out of place in a sass file.
`is` matches `and`, `or`, and `not`. Similarly `is not` should replace `!=`.

Example:

```
@if $border-style is dashed { ... }
```

nex3:
I feel like we had this discussion long ago when we were first introducing `@if` and booleans, but I don't remember what our reasoning was. I'm not opposed to the idea of using `is` and `is not`, but I do have two hesitations. First, I feel like having a more programmery punctuation-heavy operator calls attention to the fact that this is doing something different than CSS (although since boolean operations universally happen in the context of `@if` or possibly variable assignment, that's somewhat less relevant). The other worry I have is that this isn't a big enough issue that the cost of migrating existing stylesheets is outweighed by the benefit of the new syntax. I'm certainly not convinced of that, though.

chriseppstein:
All I can say is something inside me cringes when I type == and != in a sass file. It's an asthetic reason and I'd like to change it sooner rather than later. It's a relatively easy thing to search and replace and it improves the overall readability of a file.

adrusi:
I don't think you'd have to remove support for `==` and `!=` I doubt you'd want to do anything else with them anytime soon, so just leave them as they are for backwards compatability

twoolie:
Coffeescript has shown that `is` and `==` can live in harmony.

zh4ngx:
I like where this is going and would like to help. Are we working off of pull 447?

nex3:
@ZhangBanger You could, or you could start a new PR.

maxw3st:
I'm adrusi on leaving: == and != left alone. If you want to add that functionality to <code>is</code> and <code>is not</code> that's fine, but no reason to remove the operators. ( the currently proposed 'eq' and 'ne' should work fine, but please leave the math operators, as they have it in Coffeescript.)

borodean:
CoffeeScript uses `isnt` operator as a synonym to `!==` which I find very pretty. I think if Sass would choose to follow a humanistic naming convention this one would be a more perfect option than an untidy `is not`. (However, considering Sass supports Boolean operators both of them should work.)

ghost:
I personally would prefer `is not` because it adds fewer keywords.

chriseppstein:
@nex3 I'd Iike to introduce `is` and `is not` as the new equality operators in our next release. I'd like to deprecate `==` and `!=` at that time. Objections?

[...]

PR #447, opened 2012-07-17
Implemented 'eq' and 'ne' operators as a first draft for Issue 25

zofrex:
Issue 25 requested "is" and "is not" to be able to be used in addition to == and !=. While adding "is" support was simple enough, adding "is not" support is probably much harder (as the not token is already in use, so it's not as simple as adding a new token to the lexer).

As a draft proposal to achieve the same aims as those proposed in #25, while at the same time being lazy, I have added 'eq' and 'ne' operators. Thoughts?


nex3:
I don't think `eq` and `ne` really solve the same problem. The issue isn't just not using punctuation, it's using something that feels more natural-language-y, in line with `and` and `or`. I don't think `eq` and `ne` achieve that.

zofrex:
Obviously it would be easy to change `eq` and `ne` to anything else, the only issue is that if they overlap with other operators (such as `not`) then it might be trickier to implement. Have you got any other suggestions for something that feels natural-language-y?

nex3:
The point of #25 is that `is` and `is not` are better than `==` and `!=`. If those specific operators can't be used, then it's not really worth adding additional operators.

ianstormtaylor:
I personally don't care about not using `==` and `!=`, but if you wanted a quick win, `isnt` doesn't contain `not` and is more english-y.

chriseppstein:
I'm ok with `isnt` or `isnot`. Every time I type `==` and `!=` I cry a little inside.

chriseppstein:
I think this patch should deprecate `==` and `!=` in favor of the new operators.

nex3:
I'm not okay with `isnot`. We should do this correctly (`is not`) or not at all. I don't think the parsing issues are insurmountable.

chriseppstein:
`$foo is not $bar or $baz` -- this seems ambiguous. is it `$foo is (not $bar) or $baz` or is it `$foo is not ($bar or $baz)`? I am worried that this will be a gotcha for users.

chriseppstein:
maybe if we _only_ have equality check and negation is just negation?

nathggns:
Well just have an order of importance like any other programming language, even basic math has it. You're trying to make SASS more like a programming language, that isn't going to come without "gotcha"s. I believe it should be $foo is not $bar or $baz = $foo is not ($bar or $baz) as that is how it should work if you just replaced is not with "!=="

chriseppstein:
@nex3 thoughts -- the parsing issues might not be a big deal, but the user issues might be.

nex3:
Doing equality comparisons on booleans isn't a very common operation, so I'm not too concerned about adding a gotcha for it. Since `$foo is not $bar` should obviously _not_ parse as `$foo is (not $bar)`, it seems like `$foo is not $bar or $baz` should parse as `$foo is not ($bar or $baz)`. That also optimizes for the case of people not doing boolean comparisons.

ianstormtaylor:
coffeescript uses a combination of `isnt` and `not`. if you're only implementing `isnt`, there isn't any confusion.

robwierzbowski:
> I think this patch should deprecate == and != in favor of the new operators.

Please don't. For some of us, reading symbol operators is quicker and easier than language operators. I would appreciate if this is left up to author choice.

ianstormtaylor:
I'd agree that symbols are actually better for legibility—quick scanning.

nathancrank:
I'd agree on not deprecating == and !=, I find them to be much clearer, legible and easily scannable than the proposed replacements. `is` and `isnot` or `isnt` are far more ambiguous to me. I don't really see why it has to be either or option, there are plenty of other languages using both without issue.

chriseppstein:
This same argument could be made for using `&&`, `||`, and `!` over `and`, `or`, and `not`. We've already made the decision to prefer words over symbols in Sass and this is reflected in CSS as well (see: media queries).

ianstormtaylor:
I personally never use words over symbols, regardless of whether the language makes them available. It's definitely harder to parse operators that are words when they are surrounded by other words.

nathancrank:
That's the best argument on this whole thread.

On Monday, January 14, 2013, Chris Eppstein wrote:

> This same argument could be made for using &&, ||, and ! over and, or, andnot`.
> We've already made the decision to prefer words over symbols in Sass and
> this is reflected in CSS as well (see: media queries).
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/nex3/sass/pull/447#issuecomment-12245760.

## 

Nathan Crank
me@nathancrank.com

chriseppstein:
@ianstormtaylor Please show me a code example where words are a markedly worse experience for you.

ianstormtaylor:
``` javascript
if (happy && knowsIt)
  clapsHands()
else
  showIt()
```

That is the Javascript example taken from the Coffeescript docs (with alterations, to keep the argument focused). Here is that same code in Coffeescript:

``` coffeescript
if happy and knowsIt
  clapsHands()
else
  showIt()
```

I'd argue that there isn't that much of a difference in readability there. Actually, no I think the Coffeescript here is nicer (especially since it removes parens). And in that case it might be nicer to have the `and` to help beginners. Here's the example right below it in the docs:

``` javascript
date = friday ? sue : jill
```

And the Coffeescript equivalent: 

``` coffeescript
date = if friday then sue else jill
```

Here I'd argue that the Coffeescript is much harder to parse. You could argue that it's better for beginners if you wanted, which might be true, but as soon as you know what a ternary is the Javascript variant is easier to read.

You could also argue that in Sass most of the cases are in if statements where words don't add that much parsing difficulty (as opposed to more complicated cases like ternaries).

chriseppstein:
I meant using Sass. We're not designing Coffeescript ;)

Anahkiasen:
Well Sass is kind of the Coffeescript of CSS so that holds up. But it's all a matter of preference in the end, I find the Coffeescript syntax above much more readable and clean that the horrible base syntax of ternary operators. There will always be two schools, some people prefer to read their code like sentences, other prefer a more cyrptic approach but that is probably faster to parse.

That's why I favor those who say we should offer the two syntaxes.

chriseppstein:
There are obvious corollaries, and we can draw on our intuitions from other languages when thinking about Sass. But making syntax decisions about sass code can only be made my looking at hypothetical sass code. I have looked at Sass code written with symbols vs. words and decided that my preference is towards words. But maybe my examples missed some critical case to be considered.

robwierzbowski:
Using vanilla CSS as an example, there are plenty of symbols used in favor of words. With selectors we use "," instead of "or", spacing rules instead of "and", +, ~, and soon ! to declare specific relationships instead of writing them out (and @media instead of at media, etc., etc.). In any language the important thing is to strike the best balance between symbol and natural language for understanding and usability. 

I'm all for clear rules, but the language/symbol line has some wiggle room. We're using curly brackets `{}`, not `if x then ... fi`. If the argument for words is "a single rule for clarity and understanding" it seems `>` would have to be changed to `greaterthan` and `<=` to `lessthanorequalto`.

Sass is already the programmy version of CSS. To use it you have to be familiar with some basic pieces of programming structure, so I don't think an argument of "novices won't understand" is as important as "is it usable?". For authors coming in from javascript and php, the symbols are familiar and expected.

Reduced examples like

``` scss

if $this isnot 'a good example' {
  // ...
}

// vs

if $this != 'a  good example' {
  // ...
}
```

aren't that different to my eyes, but a whole page of code is very different, very much easier to parse with symbols (x1000 without code highlighting). Mentally switching between `==`, `!=`, an `>=` is especially easy comparatively.  I'm not sure typing out a whole page example would help though—what is easy for you might not be easy for me, and vice versa. There are some people in this issue who have voiced their opinions as to what they prefer, and I think that's a good reason to give both operators a trial period, then reevaluate after real world use. But, if it goes the other way I'm sure we'll live.

PR #474, opened 2012-08-07
Per the CSS spec, There are 96px in 1in.

### LINKED ISSUE #471: Make px convertable to other absolute units
CSS now defines `1px` as `1/96th` of `1in` instead of as one pixel on the screen. As such, it can be converted and compared to other units like `in`, `cm`, etc. We need to add this fact to our conversion tables.

Reference documentation:
- https://developer.mozilla.org/en-US/docs/CSS/length
- http://dev.w3.org/csswg/css3-values/#absolute-lengths

chriseppstein:
http://dev.w3.org/csswg/css3-values/#absolute-lengths

Closes #471

### DIFF
--- a/lib/sass/script/number.rb
+++ b/lib/sass/script/number.rb
@@ -422,12 +422,13 @@ module Sass::Script
     end
 
     # A hash of unit names to their index in the conversion table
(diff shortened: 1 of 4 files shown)


nex3:
Can you also add a mention of this to the changelog?

chriseppstein:
will do.

Issue #476, opened 2012-08-08
Conversion error in sass-convert from css to scss


NielsDoucet:
When converting the css fragment below[1] using the SassConvert executable, it generates [2]this scss

[1]

``` css
div.class div.class2 div {
    width: 100px;
}

body .class *.test .test1 {
    width: 200px;
}

.test1 p.class .test2 {
    width: 200px;
}

.test1 p.class .test3 {
    height: 100px;
}
```

[2]

``` scss
div.class div {
  width: 100px; }

body .test1 {
  width: 200px; }

.test1 p.class {
  .test2 {
    width: 200px; }
  .test3 {
    height: 100px; } }
```

The problem arises when the css rule defines more than 2 selectors.
All selectors between the 2 outer ones are simply stripped and not included in the generated scss, unless they can be included in a nested structure.
This problem occurs in version 3.1.19 and 3.1.20 (I haven't tested this in other versions).

MoOx:
Your `body .class *.test .test1`is just weird. Why do you use `*`in the middle of a selector ?

chriseppstein:
@nex3, I can reproduce this. Seems like a real bug.

NielsDoucet:
I don't see what is weird about it, but ok. \* is just a wildcard, so *_.class_ is identical to _.class_.
I was just testing whether it would make any difference to add the wildcard, that's all.

NielsDoucet:
@chriseppstein I added a [pull request](https://github.com/nex3/sass/pull/479) for a possible solution.
I'm not sure if this covers all the bases, but it seems to be doing what it's supposed to do.

My commit seems to involve a lot of unintended whitespace changes. Could this be related to the fact that I'm developing on a windows machine?

nex3:
This is fixed in `master`, which is close enough to release that it's probably not worth worrying about 3.1.

nex3:
This is fixed in `master`, which is close enough to release that it's probably not worth worrying about 3.1.

PR #479, opened 2012-08-10
https://github.com/nex3/sass/issues/476

NielsDoucet:
Instead of only returning the first member of the current rule, we return the entire parsed rule


chriseppstein:
Thank @github for the ?w=1 parameter:

https://github.com/nex3/sass/pull/479/files?w=1

chriseppstein:
Fixes #476.

nex3:
Closing since #476 is already fixed.

PR #527, opened 2012-09-25
Change eval calls to include file name and line number

tenderlove:
This commit changes the calls to `eval` to include file names and line numbers.  It is nice for debugging.

Thanks!


tenderlove:
oops, wrong merge head.

PR #540, opened 2012-10-09
Rebuild sourcemaps in the watch/update modes

crdev:
This patch adds the capability to rebuild sourcemaps in the --watch and --update modes.


crdev:
I've addressed both items. Sourcemap filename is now passed into `render_with_sourcemap` in the form of a non-encoded URI, and into the `updating_stylesheet` and `updated_stylesheet` callbacks as plain name (like `foo.css.map`), notifications about sourcemap updates printed next to the css updates.

PR #549, opened 2012-10-23
allow people to opt-out of sass initialization -> save 1s load time and not apply monkeypatches/middleware

grosser:
atm I'm forced to load all of sass, just because I added it to my Gemfile, which is kind of painful since it takes long and adds a middleware/monkeypatches.

I'd like to skip that, since my asset loading is handled by sprockets.

### DIFF
--- a/init.rb
+++ b/init.rb
@@ -1,18 +1,20 @@
-begin
-  require File.join(File.dirname(__FILE__), 'lib', 'sass') # From here
-rescue LoadError
+unless defined?(Sass::RAILS_LOADED)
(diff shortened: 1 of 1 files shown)


nex3:
I thought modern Rails versions didn't even use the top-level `init.rb`.

grosser:
I'm on rails 2 :dancers: 
I'll see if it's still an issue on our rails 3 branch, if not I'll close it :)

grosser:
Yep, no longer an issue :D

PR #550, opened 2012-10-24
Allow un-/defining custom unit conversions

fd:
Define a new unit `foo` in terms of `px`:

``` ruby
# sass/scss config
Sass::Script::Number.define_unit 1, "foo", 42, "px"
# signifying 1foo == 42px
```

Then use `foo`:

``` scss
div {
  height: 2foo;
}
```

Which produces:

``` css
div {
  height: 84px;
}
```

The idea is that (eventually) one could define a custom unit like so:

``` scss
@unit 1foo == 42px;
```

### DIFF
--- a/lib/sass/script/number.rb
+++ b/lib/sass/script/number.rb
@@ -428,6 +428,7 @@ module Sass::Script
         if i && !BASE_UNITS[i]
           @value /= conversion_factor("px", u)
           @numerator_units[0] = "px"
+          @original = nil
         end
       end
     end
--- a/lib/sass/script/number.rb
+++ b/lib/sass/script/number.rb
@@ -17,7 +17,7 @@ module Sass::Script
 
     # A list of units in the numerator of the number.
     # For example, `1px*em/in*cm` would return `["px", "em"]`
-    # @return [Array<String>] 
+    # @return [Array<String>]
     attr_reader :numerator_units
 
     # A list of units in the denominator of the number.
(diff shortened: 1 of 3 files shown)


Anahkiasen:
This sounds very interesting, +1
Plugins such as modular scale and other could take advantage of such a feature.

fd:
yes, especially when you're using grid systems:

``` scss
@unit 1g == 20px;

#sidebar {
  width: 7g;
  padding: 1g;
}

#content {
  margin-left: 7g;
  padding: 1g;
}
```

nex3:
First of all, I'm not a big fan of changing the semantics of a stylesheet by external means like Ruby customizations. As much as possible, Sass stylesheets should have the same meaning regardless of external factors.

As for the `@unit` proposal, I'm generally wary of extensions that look too much like plain CSS. This runs a real risk of allowing users to define units that are later added to the spec.

@chriseppstein, what are your thoughts on this?

fd:
> First of all, I'm not a big fan of changing the semantics of a stylesheet by external means like Ruby customizations. As much as possible, Sass stylesheets should have the same meaning regardless of external factors.

The external (Ruby) API is just a simple way to get this working (for now) as I'm unsure about changing the SCSS/SASS parsers.

> As for the @unit proposal, I'm generally wary of extensions that look too much like plain CSS. This runs a real risk of allowing users to define units that are later added to the spec.

The proposed syntax is the first thing that came to mind. Here is another proposed syntax.

``` scss
@define unit(1u == 20px)
```

Or:

``` scss
@define unit(1u, 20px)
```

nex3:
The issue isn't the syntax of `@unit` itself, but rather the usage of the new units. A user seeing `1u` doesn't have any way of knowing if that's a real CSS unit or a custom unit.

robwierzbowski:
I would actually like this not to be included. As a superset of css, when I look at a scss file I want to be able to see the sass additions quickly and easily, which custom native-looking units would kind of ruin.

You can do sass units with variables or functions already, in a way that maintains the knowledge that the unit you're using is an author/sass construct.

```
width: $g * 3;
width: g(3);
```

Still very terse, very efficient syntax. 

 +1 to Nex3's point about new native units as well.

PR #569, opened 2012-11-13
Source range computation implementation for the SASS parser

crdev:
NB: This branch is based off of `sourcemap_fix_relpath`, otherwise it would not make any sense.

Sourcemaps can now be constructed for nodes built while parsing SASS (`engine.rb`). A few fixes to the existing code have been made. All SCSS sourcemap tests have received their SASS counterparts.

### DIFF
--- a/lib/sass/engine.rb
+++ b/lib/sass/engine.rb
@@ -452,7 +452,7 @@ END
           raise SyntaxError.new(message, :line => index)
         end
 
-        lines << Line.new(line.strip, line_tabs, index, tab_str.size, @options[:filename], [])
+        lines << Line.new(line.strip, line_tabs, index, line_tab_str.size, @options[:filename], [])
       end
       lines
     end
@@ -496,6 +496,7 @@ MSG
 
     def build_tree(parent, line, root = false)
       @line = line.index
+      @offset = line.offset
       node_or_nodes = parse_line(parent, line, root)
 
       Array(node_or_nodes).each do |node|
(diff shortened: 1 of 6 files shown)


paulirish:
:+1: this provides a LOT of power. can't wait to see it.

Anahkiasen:
What does it do exactly ?

paulirish:
Well the sourcemap can map all lines/columns from original sources into their destination CSS.

What this enables on the tool side: ctrl-click any property/value/selector and go to the location in the sass/scss where that originates: the expression, mixin, variable where it was birthed or computed.  This makes the editing experience of working with sass and tweaking CSS much more enjoyable.  

The UI side of the above is ready, and this PR enables the sourcemap that can deliver the mapping from sass to the tool.

Down the line I can imagine editing values on the right styles pane of Chrome DevTools and changing the original scss source.

matthew-dean:
Note that @paulirish says this feature is not SASS / SCSS specific and should work with any CSS pre-processing language. https://github.com/cloudhead/less.js/pull/1038

ruudk:
When I run this command from my homedir (~): 

``` command
sass --load-path ~/Downloads/ --scss --cache-location --sourcemap ~/Downloads/test.scss ~/Downloads/output.css
```

It creates the `--sourcemap` directory inside my homedirectory instead of the directory of `~/Downloads/output.css`.

Is it possible to configure the location of the sourcemap?

crdev:
@ruudk: You did not specify the actual cache location after `--cache-location`, thus sass thinks that `--sourcemap` IS your desired `--cache-location` value.

PR #572, opened 2012-11-18
Fixes to respond_to? and warning tests for Ruby 2.0.0-preview1

sax:
In Ruby 2.0, respond_to? no longer returns true for protected methods unless you explicitly pass `true` as a second argument. This causes try_send to fail for @content blocks, keeping sass from knowing that a mixin supports blocks.

Also, Kernel.warn behaves slightly differently. Since the difference is in number of newlines, I figure it's a cosmetic change and made the most expedient fix I could think of.

Let me know if you see any issues!

### DIFF
--- a/lib/sass/tree/visitors/check_nesting.rb
+++ b/lib/sass/tree/visitors/check_nesting.rb
@@ -140,7 +140,7 @@ class Sass::Tree::Visitors::CheckNesting < Sass::Tree::Visitors::Base
   end
 
   def try_send(method, *args)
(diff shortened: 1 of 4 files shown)


ansman:
Any progress on this? 2.0 is blazingly fast and I've love to be able to use it.

sax:
Had a ton of emergency work that I had to finish in December, so I have not been able to refactor this. Hopefully I'll be able to look at this with the new year!

PR #582, opened 2012-11-29
Templates in jars @importing each other under JRuby breaks

argv-minus-one:
When a Sass template (such as the ones in Compass) is inside a jar file, and Sass is running on JRuby, that template importing another inside the same jar file will currently fail.

It appears that this is because JRuby doesn't properly handle a jar path containing `./`, and `Sass::Importers::Filesystem#possible_files` was generating them. This patch (or, rather, doing the same thing by reopening the class in my own code) appears to fix the issue in my test environment.


argv-minus-one:
Oh no! I chose to merge with the wrong branch! Sorry about that. Closing this pull request and made another.

PR #595, opened 2012-12-13
Fix for when template_location_array is a String

glebm:
Do not call to_a on Strings:

```
NoMethodError (undefined method `to_a' for #<String:0x639002>)
```


nex3:
Hold on, I don't think this actually does anything. `template_location_array` is normalized in `Sass::Plugin::Configuration#normalize_template_location!`, so it should never return a string.

PR #618, opened 2013-01-05
Bug#594 @import 'file' on windows fails with more than 1 subdir in the p...

dlmiles:
...ath

This commit changes lib/sass/importers/filesystem.rb and the method
find_real_file(dir, name, options) to ensure the 'dir' argument is
canonialized to rubys standard "/" use for path separators.


dlmiles:
Superseded by a version that works on both Windows and Linux.

PR #623, opened 2013-01-11
replace inheritence with deligation in MultibyteStringScanner

tank-bohr:
MultibyteStringScanner doesn't work properly whith Rubinius

### DIFF
--- a/lib/sass/util/multibyte_string_scanner.rb
+++ b/lib/sass/util/multibyte_string_scanner.rb
@@ -7,25 +7,26 @@ else
   # multibyte character encodings. The native class deals only in bytes, not
   # characters, for methods like [#pos] and [#matched_size]. This class deals
   # only in characters, instead.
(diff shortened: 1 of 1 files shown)


nex3:
Can you give a quick explanation of why this inheritance doesn't work in Rubinius?

nex3:
Also, all my tests pass locally on Rubinius 2.0.0.rc1, so please provide a test that fails without your patch.

tank-bohr:
> Also, all my tests pass locally on Rubinius 2.0.0.rc1, so please provide a test that fails without your patch.

probably your tests pass in 1.8 mode.

> Can you give a quick explanation of why this inheritance doesn't work in Rubinius?

There is a problem with Sass::Util::MultibyteStringScanner#getch method
In rubinius this method is implemented as `scan(/./m)` see https://github.com/rubinius/rubinius/blob/master/lib/strscan.rb#L77 But `scan` method is overridden. So the additional call of `_forward _match` occur

PR #629, opened 2013-01-16
Adds :staleness_check => true to options passed to importer when that's the case

svallory:
As explained in https://github.com/nex3/sass/issues/625#issuecomment-12334541 this change allows to create custom importers which only require a file once.


nex3:
There's a `:_from_import_node` option in master which should solve this use case.

PR #636, opened 2013-01-22
Code changes to reduce number of warning messages.

rurounijones:
Code changes to reduce the number of warnings that are generated
when sass code is run as part of the sinatra test suite.

This should also reduce the number of warnings generated in other
softwares' test suites.

Although some of the changes may make the code a bit less 
'clean' I think it is important for well-used libraries to try and be good 
citizens and keep the number of warnings down.

This pull request fixes a lot of the low-hanging stuff.

### DIFF
--- a/lib/sass/cache_stores/chain.rb
+++ b/lib/sass/cache_stores/chain.rb
@@ -21,9 +21,9 @@ module Sass
 
       # @see Base#retrieve
       def retrieve(key, sha)
-        @caches.each_with_index do |c, i|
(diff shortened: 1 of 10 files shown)


rurounijones:
Not sure what to do with doc-src/SASS_CHANGELOG.md since there isn't a section for the "next version" from what I can tell and I am not sure what it will be (3.2.6? 3.3.0 etc)

rurounijones:
If this is ok I will squash and rebase on current master if you want.

PR #689, opened 2013-03-20
Native Sass function to return list separator

Snugug:
This should return the separator of a list or false if the input isn't a list. Solves https://github.com/nex3/sass/issues/688

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -147,6 +147,9 @@ module Sass::Script
   # \{#append append($list1, $val, \[$separator\])}
   # : Appends a single value onto the end of a list.
   #
+  # \{#list-separator list-separator(#list)}
+  # : Returns the separator of a list.
+  #
   # ## Introspection Functions
   #
   # \{#type_of type-of($value)}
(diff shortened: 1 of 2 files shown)


Snugug:
Not entirely sure how I would go about writing this test (`functions_test.rb`, right?). I'll take a look in the morning when I'm less out of it.

Snugug:
Updated with tests, naming, and incorporated your comments

nex3:
I don't see any tests... did you remember to add them to the commit?

Snugug:
Anything else I've got to do for this? I've run into another instance where I need this working.

robwierzbowski:
Shouldn't this function return `Sass::Script::String.new()` instead of `String.new`? On my machine the function silently fails if the result returned isn't a sass string.

robwierzbowski:
Also, if you're using variable arguments the list class for the variable argument is `ArgList`, which is always comma delimited. Right now `list-separator()` is only checking for the `List` class and then defaulting to space. A quick example:

``` scss
@mixin arglist-test(args...) {
  @debug list-separator(args);
}

.test {
  @include arglist-test(this, is, comma, separated); // DEBUG: space
}

```

Changing

```
if list.class == Sass::Script::List
```

to

```
if list.class == Sass::Script::List or list.class == Sass::Script::ArgList
```

should produce the correct `ArgList` result.

nex3:
> _@robwierzbowski wrote:_
> Shouldn't this function return `Sass::Script::String.new()` instead of `String.new`? On my machine the function silently fails if the result returned isn't a sass string.

In this context `String` resolves to `Sass::Script::String`, since we're in `Sass::Script`. It could be more explicit, but it works as-is. If it returned a `::String`, it would break when `Funcall` tried to call `#options=`. I'll add some explicit validation to the test as well, though (b26d2699c8ff84faaa5b55189222d6f9d7991fcf).

> Also, if you're using variable arguments the list class for the variable argument is `ArgList`, which is always comma delimited. Right now `list-separator()` is only checking for the `List` class and then defaulting to space.

Good point. Fixed by 7e512381422e5d3bc60b0aec36f61a743adffc70.

robwierzbowski:
Thanks. Re: `String.new()`, it's working fine for me now. Not sure what the problem was before.

PR #695, opened 2013-03-22
Fix Warnings

tenderlove:
Hi,

I've read [previous pull requests](https://github.com/nex3/sass/issues/160) that tried to eliminate warnings.  However, I think the Sass codebase contains some great examples on why running with `-w` is a good thing.

I will enumerate the examples here and link to commits that deal with the issue.
## Unused variable warnings

Sass assigns values to variables which are unused.  In some cases, the code has no side effects and can be eliminated.  I was able to delete code:
- 8badbab218f73190eadbbd0f0a54ef0f288465f8
- 58c57a8e390b67f6e4c6d52454bcde1e1bdf0e3e
## Calls to deprecated code

Sass calls methods on URI that will eventually be deleted.  I fixed that [here](https://github.com/tenderlove/sass/commit/c94257db653c6bbe9706b973481a78d705d3a32f).
## Redefined methods

Sass defines tests that never run because the methods were redefined.  I changed the test name [here](https://github.com/tenderlove/sass/commit/3df278a3ff5b6140e18ffdc57c80e1f701e6213f), [here](https://github.com/tenderlove/sass/commit/ff954d1d39e7915724a25c9a943d87d949d9f27d), and could even remove completely duplicated code [here](https://github.com/tenderlove/sass/commit/bd9a56ca2b632bce751c9f3da12377c9aa27f380).
## Uninitialized instance variables

Uninitialized instance variables led me to find an instance variable [that is probably misspelled](https://github.com/sass/sass/blob/bcde8752a704ceaa5ef9405d71b4f8fa3c6146bf/lib/sass/scss/parser.rb#L1154).  `@throw_err` is never set anywhere, and the positive branch of that if statement is never executed in the tests.
## Circular requires

Circular requires could possibly lead to a deadlock, so I [fixed those](https://github.com/tenderlove/sass/commit/ff0d98a0536687b7d2d6f918a198ec81e1494077).
## Shadowed variable warnings

Shadowed variables could refer to variables in the outside scope, so I fixed those [here](https://github.com/tenderlove/sass/commit/c909d1572637dbcf8aad6bc4c02cadce4a3d3e75) and [here](https://github.com/tenderlove/sass/commit/f75d7d2b8b2178d5b7f9af1375b0e7151e3771d5).
## Setting external encodings

I eliminated a warning around [setting the external encoding](https://github.com/tenderlove/sass/commit/220d5c7e966fd3b3dd755d5f827950510246d745).  Does this mean that Sass requires people to set the external_encoding to UTF-8?  I suspect it's possible to eliminate this requirement.
## Literals in conditionals

This is probably my most dubious commit.  I know team Sass uses literals in conditionals as coding style, but it causes warnings (since it will always evaluate the same way).  I changed the literals [to constants](https://github.com/tenderlove/sass/commit/283d8aa7924bf2661be2fc606ebc7a9bcedf5d86) as a compromise between readability and eliminating warnings.
## Duplicate character class warnings

I left these.  I think it's a bug in Ruby, so I [filed a ticket](https://bugs.ruby-lang.org/issues/8151).

Anyway, all of these are great examples of why I like to run `-w` on my code.  I realize Team Sass may not care for using `-w`, so I've purposely not squashed these commits in the hope that some (if not all) of them are merged to mainline.  Please feel free to cherry pick what you do want, and leave what you don't.

Also, I really really want to say thanks for making the tests super easy to get running.  It's not often I can clone a repo and run `rake` and it does the right thing.  THANK YOU! :smile:

:heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart:


nex3:
You make a pretty compelling case for the utility of some of these warnings. That said, I still don't like the idea of warnings that make the code less readable, like the warnings for `!:foo` and to some degree uninitialized instance variables (although the `@throw_err` issue is worrisome).

I'll go through the individual commits and leave review comments. Once those are addressed, I'll merge in everything other than the literal-in-conditional and uninitialized-ivars. I'll fix the `@throw_err` thing myself.

Thanks for the PR, @tenderlove.

tenderlove:
> You make a pretty compelling case for the utility of some of these warnings. That said, I still don't like the idea of > warnings that make the code less readable, like the warnings for !:foo and to some degree uninitialized instance variables (although the @throw_err issue is worrisome).

I prefer my libraries to run warning free so that app developers are free to make the choice to use `-w` or not, but I definitely see where you're coming from.  `-w` helps me track down bugs (like some of these) and it's nice to have the option to use it.

> I'll go through the individual commits and leave review comments. Once those are addressed, I'll merge in everything other than the literal-in-conditional and uninitialized-ivars. I'll fix the @throw_err thing myself.

Awesome, thank you! :smile:

> Thanks for the PR, @tenderlove.

No problem, I'm glad to help out.  I've pushed a commit that fixes all the stuff you commented on except for the circular require issue.  If you feel strongly about that one, then I'll change it as well.

Thanks for your time!

josh:
:heart: 

Hopefully I can remove this line from my tests - https://github.com/sstephenson/sprockets/blob/master/test/test_sass.rb#L20-L22

PR #696, opened 2013-03-22
Added zip and index functions to Sass::Script::Functions Overview

robwierzbowski:
...And edited function descriptions for clarity.


robwierzbowski:
Yay.

PR #703, opened 2013-04-05
Add suggestion to use a partial to "Undefined variable" error message

x3ro:
So I'm fairly new to Sass, and I didn't know about partials. I was trying to import a Sass file, and suddenly my variables weren't working. I started googling (which was successful in the end) but a pointer such as the one I'm suggesting in the commit would've helped quite a bit, since I would've had a keyword to google for.

I know that this is a definite case of RTFM, but I imagine that this might not be an uncommon issue for peope new to Sass, since it is not super-intuitive, imho.

Ideally, the error message mentioning the partial should only be displayed when actually inside an import, but I didn't find a way to determine if that is the case. If someone would be so kind as to give me a hint on how I can detect that, I'd be glad to change the PR :smile:

### DIFF
--- a/lib/sass/script/variable.rb
+++ b/lib/sass/script/variable.rb
@@ -46,7 +46,7 @@ module Sass
       # @return [Literal] The SassScript object that is the value of the variable
       # @raise [Sass::SyntaxError] if the variable is undefined
       def _perform(environment)
-        raise SyntaxError.new("Undefined variable: \"$#{name}\".") unless val = environment.var(name)
+        raise SyntaxError.new("Undefined variable: \"$#{name}\". Did you forget to use a partial?") unless val = environment.var(name)
         if val.is_a?(Number)
           val = val.dup
           val.original = nil


robwierzbowski:
I'm all for the spirit of this issue, but I don't think adding a more specific message would be helpful. There are about a million reasons that a variable can be undefined, and we don't want to send people looking for the wrong problem.

x3ro:
As I mentioned, I'd of course like this error message only to be displayed when actually inside a partial. I'm not so familiar with SASS as to be able to judge if this is a common mistake people make, but I can imagine that it might be. What other, frequent reasons are there for a variable to be undefined? Off the top of my head I can only think of scoping (of which partials are _kind of_ a part) and typo :smile:

robwierzbowski:
Unfortunately that's not possible because Sass runs once over the stylesheets, compiling as it goes along. Creating a complete list of variables, extends, etc., then going back and seeing if a variable was undefined at one point but defined somewhere else or later on would require two passes.

Some reasons variables can be undefined: they have never been defined (:smile:), the call is misspelled, a person forgot to pass them into a mixin/function and the mixin/function argument has no default values, they are in a separate partial, they're defined after the call to use them.

x3ro:
That's unfortunate indeed :frowning: Well ^^ I guess this can be closed then ;)

Issue #371, opened 2012-04-27
sass-convert places brackets after line comments


scottkellum:
Open brackets are commented out when converting sass with a line comment after a selector.

``` sass
.foo // comment
  width: 20em
```

converts to

``` scss
.foo // comment {
  width: 20em;
}
```

The expected conversion looks like this:

``` scss
.foo { // comment
  width: 20em;
}
```

PR #713, opened 2013-04-16
Fix issue #371: sass-convert outputs invalid scss when a .sass file's selector has a single-line comment

LeeMallabone:


nex3:
I don't want to delete these comments. I want to extract them and put them after the `{` in the SCSS output (or failing that at least above the selector).

ryanburnette:
When using sass-convert I'm getting invalid Sass output when comments are formatted certain ways. Here's an example.

_images removed_

Is this the same issue or should I open a new one?

nex3:
@ryanburnette If you're getting invalid output, that's a different issue. It's hard to tell what's going on from those screenshots, though. Can you open a new issue with minimal, non-screenshot examples of input that produces invalid output?

ryanburnette:
@nex3 Sure I'd be happy to. I'll ping you in the initial comment.

LeeMallabone:
@nex3 As requested, I've updated the commit to retain the comments when converting from sass to scss, rather than throwing them away. I've set the type on the (previously discarded) comments to `:quiet`. I'm not sure if that's correct or a mistake: let me know if it's important to support the `/*! loud comment */` type for this use case.

nex3:
`:quiet` isn't a valid comment type; see the documentation of [`Comment#type`](http://sass-lang.com/docs/yardoc/Sass/Tree/CommentNode.html#type-instance_method). It should always be `:silent`, so it behaves the same in SCSS as it did in Sass.

LeeMallabone:
Thanks, I've updated the commit to change the comment type to `:silent`. Let me know if this change needs any further work before it can be merged.

PR #716, opened 2013-04-19
Specify the correct module for the sourcemap_name method

crdev:
Running with the --sourcemap argument, I got
lib/sass/plugin/compiler.rb:174:in `block (2 levels) in update_stylesheets':
    uninitialized constant Sass::Plugin::Compiler::Util (NameError)


nex3:
I believe this is already fixed on master: https://github.com/nex3/sass/blob/master/lib/sass/plugin/compiler.rb#L174.

PR #730, opened 2013-04-30
--unix-newlines/--in-place are not working on Windows

ranvis:
- `sass --unix-newlines` doesn't produce LF-eol files.
  
  Caused because output files are opened with 'w' not 'wb.'
- `sass-convert --in-place` doesn't do anything.
  
  Haven't investigated why actually...

these are confirmed on ruby 1.9.3p392 (2013-02-22) [i386-mingw32]


ranvis:
Updated test for Ruby 1.8

PR #742, opened 2013-05-09
Prevent built .gem files from accidentally being version controlled

mcandre:
Building sass with `gem build sass.gemspec` results in a file `sass-x.y.z.gem`, which does not belong in version control. I added `*.gem` to `.gitignore` to prevent this from happening.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.gem
 /.yardoc
 /coverage
 /doc


nex3:
You should really run `rake package` instead of `gem build sass.gemspec`. Running `gem build` can result in the wrong version.

mcandre:
Even so, it's nice to keep `.gem` files out of version control.

chriseppstein:
`gem build` is the ruby standard way of making a .gem file. I've got this working for compass so that they get a dev version based on the git sha if they build one this way.

Also, I don't see any harm in adding gem files to the `.gitignore` file.

nex3:
If you can get `gem build` to work correctly, I'll be okay with merging this in. Until then, I don't want to gitignore an output file that shouldn't be generated in the first place.

chriseppstein:
sounds good.

nex3:
Closing this again as no work seems to be happening on getting `gem build` working, and it's a one-line change that won't be difficult to reproduce once that work happens.

PR #761, opened 2013-05-21
Update to_css.rb

wankyu:
Squeeze some spaces for "compact" mode.


nex3:
Compact mode explicitly doesn't try to eliminate as much whitespace as possible. It just tries to put all properties of a given rule on a single line. For whitespace-elimination, use compressed mode.

PR #777, opened 2013-05-30
Upgrade listen

chriseppstein:
This upgrades the watcher to use listen 1.1.x and has a substantial rewrite of the watcher code.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -94,6 +94,11 @@ Thanks to Alexander Pavlov for implementing this.
 * `grey` and `transparent` are no longer interpreted as strings; they're now
   interpreted as colors, as per the CSS spec.
 
+* The automatic placement of the current working directory onto the Sass
+  load path is now deprecated as this causes unpredictable build
+  processes.  If you need the current working directory to be available,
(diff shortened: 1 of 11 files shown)


nex3:
Done reviewing.

chriseppstein:
@nex3 I've addressed your code review comments. I've also refactored the watching code further to delegate watched directories and files to the importer objects and I've added tests for the watcher.

nex3:
Other than a couple small comments, I'm happy with where this is up through 5f47ae305a4f110a8ecb6f728e09b3ff0be86be8. As I mentioned in 1fa0c477bca21307e09267535ef53bb6a2a3bf55, I feel like half the commit should be a separate pull request and the other half should be integration tests rather than unit tests.

Writing integration tests involving file watching is really gross, though, so if you don't want to do that I'm fine with just rebasing and merging up through 5f47ae305a4f110a8ecb6f728e09b3ff0be86be8.

chriseppstein:
Integration tests are good, let's create an issue to track making them and address that if we find an instance where the current testing approach is not good enough -- at least we have some tests now ^_^

nex3:
Done reviewing 1fa0c477bca21307e09267535ef53bb6a2a3bf55.

PR #782, opened 2013-05-31
Expand the target path on --watch

HersonHN:
I was having an issue when I use relative paths with tilde on "--watch" mode, so I put a File.expand_path on paramter's parsing.


chriseppstein:
Pretty sure this is take care of by #777

nex3:
The shell should expand any paths you pass in on the command line. If you're passing paths around via Ruby, you should assume that shell expansion won't happen.

Issue #784, opened 2013-05-31
Don't remove exclamation marks from loud comments


lboecker:
I don't know who came up with this, but it doesn't make much sense to me.
When working with Middleman, the comment on top of my main stylesheet looks like this

``` css
/*!! Copyright ... */
```

and both exclamation marks end up removed. The loud comment at the top of normalize.css gets removed completely because it contains only 1 exclamation mark rather than 3 (!). I'm not sure if saving a couple of bytes is worth the hassle. Thoughts?

**UPDATE:** https://github.com/nex3/sass/issues/784#issuecomment-18773351

chriseppstein:
I came up with it. It's not about saving bytes. It's about supporting the use case of producing a comment that does not start with an exclamation mark. Not everyone is this excited about their copyright notices. Sass only removes one exclamation mark.

nex3:
It's worth noting that it looks like most other minifiers don't remove the `!` from `/*!`-style comments.

lboecker:
@chriseppstein Just to be clear: I don't care much about my own copyright either, but I believe the library authors deserve some credit and I'm legally required to leave their copyright in there anyway.

chriseppstein:
I'm open to alternative suggestions that preserve the use case of enabling a person to have a comment that does not have a `!` in the output.

lboecker:
@chriseppstein The use-case you're describing is relevant for all output styles but compressed, right? Maybe changing the behaviour based on the output style is the simplest solution. [Most](https://github.com/GoalSmashers/clean-css#how-to-use-clean-css-programmatically) [other](https://github.com/gruntjs/grunt-contrib-uglify#preservecomments) minifiers take some kind of options hash, but I'm not sure if thats practical since there's only one option in the first place.

chriseppstein:
@superlukas it applies specifically to compressed mode. That is the point of loud comments.

nex3:
I'm not sure what the problem with including `!` in the output is. It's a widely-recognized indicator that the comment shouldn't be compressed away; the desire not to compress the comment isn't something that goes away after Sass's compression pass.

lboecker:
@nex3 has a point. I just did some testing though and @chriseppstein is right too, Sass doesn't remove the comments. It seems to be a problem with Sprockets and/or Sprockets-Sass. I guess the content is piped through Sass more than just once and the comments are lost somewhere on the way.

chriseppstein:
[...]

PR #786, opened 2013-06-02
Don't strip bangs from loud comments (closes: #784 and #786)

lboecker:
First of all I'm not a Ruby dev, and all I did was `ack -i loud` so I might be missing something. I'm not sure about the newline in `test_loud_comment_is_evaluated`, but it looks fine to me. Comments are welcome :smile:


chriseppstein:
This was merged to master instead of stable.

PR #789, opened 2013-06-04
Fix doc: @extend is a directive, not selector

hlb:


chriseppstein:
Maybe I'm too close to the problem, but "after the directive" could be misinterpreted as `@extend !optional .foo;` or `@extend .foo; !optional`.

Maybe a better phrasing would be "after the selector portion of the directive"?

hlb:
maybe just "the end of the line"?

chriseppstein:
It's only the end of the line in .sass files.

hlb:
"in the end of directive"?

robwierzbowski:
I believe the current wording is correct. After the **selector** part (`.this`, `%that`) of the **directive** statement.

hlb:
ok :)

PR #790, opened 2013-06-05
Make sass work when symlinked by encap.

pwnall:
The sass binaries don't work for when I set them up as encap packages, because of the odd way they handle symlinks. This change makes them work again.

FWIW, this is what I use for my setup. It worked with Sass 3.1.19 on top of MRI 1.9.3p194, but doesn't work on 3.2.9 on top of MRI 2.0.0p195.
https://gist.github.com/pwnall/3017424#file-build_ruby-sh
https://gist.github.com/pwnall/3017424#file-build_sass-sh

Thank you!

### DIFF
--- a/bin/sass
+++ b/bin/sass
@@ -2,7 +2,11 @@
 # The command line Sass parser.
 
 THIS_FILE = File.symlink?(__FILE__) ? File.readlink(__FILE__) : __FILE__
-require File.dirname(THIS_FILE) + '/../lib/sass'
+begin
+  require File.dirname(THIS_FILE) + '/../lib/sass'
+rescue LoadError
+  require 'sass'
+end
 require 'sass/exec'
 
 opts = Sass::Exec::Sass.new(ARGV)
--- a/bin/sass-convert
+++ b/bin/sass-convert
@@ -1,7 +1,11 @@
 #!/usr/bin/env ruby
 
 THIS_FILE = File.symlink?(__FILE__) ? File.readlink(__FILE__) : __FILE__
-require File.dirname(THIS_FILE) + '/../lib/sass'
+begin
+  require File.dirname(THIS_FILE) + '/../lib/sass'
+rescue LoadError
+  require 'sass'
+end
 require 'sass/exec'
 
 opts = Sass::Exec::SassConvert.new(ARGV)
--- a/bin/scss
+++ b/bin/scss
@@ -2,7 +2,11 @@
 # The command line Sass parser.
 
 THIS_FILE = File.symlink?(__FILE__) ? File.readlink(__FILE__) : __FILE__
-require File.dirname(THIS_FILE) + '/../lib/sass'
+begin
+  require File.dirname(THIS_FILE) + '/../lib/sass'
+rescue LoadError
+  require 'sass'
+end
 require 'sass/exec'
 
 opts = Sass::Exec::Scss.new(ARGV)
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -3,6 +3,11 @@
 * Table of contents
 {:toc}
 
+## 3.2.13 (Unreleased)
+
+* Numbers returned by user-defined functions now trigger division, just like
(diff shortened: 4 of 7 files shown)


pwnall:
Alternatively, I can catch `LoadError` and fall back to `require 'sass'`. That approach also works in my setup. Please let me know if you prefer that to what I'm doing right now, and I'll change the pull request.

Thank you!

nex3:
Can you explain in more detail what the problem is? What's being symlinked where, and how is it failing?

pwnall:
This is the error.

``` bash
$ scss -v
/usr/local/encap/ruby-2.0.0p195/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:110:in `require': cannot load such file -- ../../../../../../../encap/ruby-sass-3.2.9/lib/ruby/gems/2.0.0/gems/sass-3.2.9/bin/../lib/sass (LoadError)
    from /usr/local/encap/ruby-2.0.0p195/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:110:in `rescue in require'
    from /usr/local/encap/ruby-2.0.0p195/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:35:in `require'
    from /usr/local/lib/ruby/gems/2.0.0/gems/sass-3.2.9/bin/scss:5:in `<top (required)>'
    from /usr/local/bin/scss:23:in `load'
    from /usr/local/bin/scss:23:in `<main>'
```

This is the filesystem setup.

``` bash
$ which scss
/usr/local/bin/scss
$ ls -l /usr/local/bin/scss
/usr/local/bin/scss -> ../encap/ruby-sass-3.2.9/bin/scss
$ ls -l /usr/local/encap/ruby-sass-3.2.9/bin/scss
/usr/local/encap/ruby-sass-3.2.9/bin/scss
```

The gem is installed at `/usr/local/encap/ruby-sass-3.2.9/lib/ruby/gems/2.0.0/gems/sass-3.2.9/`

This is the Rubygems-generated file at `/usr/local/encap/ruby-sass-3.2.9/bin/scss`

``` ruby
#!/usr/bin/env ruby
#
# This file was generated by RubyGems.
#
# The application 'sass' is installed as part of a gem, and
# this file is here to facilitate running it.
#

require 'rubygems'

version = ">= 0"

if ARGV.first
  str = ARGV.first
  str = str.dup.force_encoding("BINARY") if str.respond_to? :force_encoding
  if str =~ /\A_(.*)_\z/
    version = $1
    ARGV.shift
  end
end

gem 'sass', version
load Gem.bin_path('sass', 'scss', version)
```

pwnall:
After more thinking and testing, I think the `rescue LoadError`-based patch is better. I updated the pull request.

nex3:
I'm still not completely understanding the failure. It looks like `/usr/local/encap/ruby-sass-3.2.9/bin/scss` will load `/usr/local/encap/ruby-sass-3.2.9/lib/ruby/gems/2.0.0/gems/sass-3.2.9/bin/scss`, within which `__FILE__` will refer to `/usr/local/encap/ruby-sass-3.2.9/lib/ruby/gems/2.0.0/gems/sass-3.2.9/bin/scss`. Then it will try to require `/usr/local/encap/ruby-sass-3.2.9/lib/ruby/gems/2.0.0/gems/sass-3.2.9/bin/../lib/sass.rb`, which should work.

pwnall:
I'm not sure where the loading breaks, I'll get back to looking at it soon. However, here is a reasonably simple argument for this patch.

All the other Ruby gems that I worked with use something like `require 'sass'` in their `bin/` files, and rely on Rubygems to set up the load paths correctly. All those gems work. Sass deviates from the standard pattern by doing the `readlink` trick. If the trick fails for any reason, I think it makes sense to fall back to the standard pattern, and work in any environment where the other gems work.

varenc:
this patch would fix an issue I am having with encap as well. Can this be reopened and reexamined?

pwnall:
I rebased the PR against `stable`.

pwnall:
@nex3 Thank you! <3

Issue #766, opened 2013-05-24
Sass feature detection


chriseppstein:
There's a number of sass features that we will add in the future which will be syntactically valid in the old version but still not working as intended.

Sass libraries that need to support multiple versions of Sass may be able to do so if we provide a function like `sass-supports($feature-name)` which returns true if the feature is present in the sass implementation.

Hat tip: this idea came from @eoneill.

nex3:
This is reasonable. We could add this in the current master as a function that always returns false (since there are no usefully-detectable features until we add more in a future release).

c-lliope:
#796 resolves this issue -- can we close it?

chriseppstein:
Yep!

PR #796, opened 2013-06-06
Feature detection

chriseppstein:
Implementation for #766.

You can now call sass-supports($feature) in Sass files to check if Sass supports
some feature by name. This only returns true for the one major feature
added in the upcoming release so far: `sourcemaps`.


nex3:
Done reviewing.

chriseppstein:
@nex3 One reason why we might want to return true for features that we implement in 3.3 even though `sass-supports` is a new feature is that it would give other Sass implementations the ability to ship syntactic compatibility with 3.3 without needing to delay the release while they implement the other features. E.g. libsass or sassc might return false for `sourcemaps` while we return true.

/cc @akhleung @hcatlin

HamptonMakes:
This feature would be extremely helpful for libsass/sassc! Though, just for clever naming, I like `can-do`

chriseppstein:
@nex3 I'm done addressing your comments to-date on this PR.

nex3:
Done responding.

chriseppstein:
@nex3 I've addressed all remaining comments. Good to merge?

nex3:
I think you missed [this comment](https://github.com/nex3/sass/commit/3f4b81a1305f7ec226335fb9f9245667b7514257#test-sass-functions_test-rb-P5). Other than that, go ahead.

PR #797, opened 2013-06-06
Use Bool::TRUE and Bool::FALSE constants to reduce GC needs.

chriseppstein:
Instead of generating a different sass object for each boolean defined
and returned during the processing of Sass files, we now define TRUE and
FALSE constants and return those instead.


chriseppstein:
In case you already review this patch, I realized that overriding `Bool.new` was a much more elegant approach to this and all the ruby API users will automatically get the benefit of this patch so I rewrote this patch.

PR #804, opened 2013-06-11
Use the new "# sourceMappingURL=..." comment format when referencing source maps

crdev:
According to https://docs.google.com/a/chromium.org/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k,
the preferred way is to use "#", since "@" conflicts with IE's conditional comments.


nex3:
This was fixed by f308e2cf7c997bb05d627ed008fd65d39f43699f.

PR #808, opened 2013-06-17
Allow negative indices in `nth` list function

ggilder:
Allow negative indices between -1 and (-1 \* list length) in the `nth`
function.


ggilder:
@nex3 Just added some explanation. Let me know if you think it's clear enough...

PR #812, opened 2013-06-22
Call function

chriseppstein:
This patch lets a function call be invoked where the function itself is provided by a script expression. A special built-in function `call` is handled by the when the function call is performed and the function is named `call`.

This implements the functions half of #626.


nex3:
Could we try doing this review via Gerrit?

nex3:
I was a little weirded out by the special-case handling of `call` in `Funcall`, so I tried making it a normal Ruby function and it seems to work. Let me know what you think.

HamptonMakes:
+1 on this feature.

Issue #561, opened 2012-11-02
Test to see if Function or Mixin Exists


Snugug:
While I know that the general consensus whenever I want more program language-ie stuff in Sass is that it's not a programming language, with the complexity of Sass systems that currently exist and being developed, I truly feel as if we need to add some more advanced program language esque stuff into Sass to make everything work well together. To this extent, I'd really like to see `function-exists()` and `mixin-exists()` functions, or simply an `exists()` function, that will return a boolean `true` or `false` depending on whether or not that function or mixin is indeed available to use when that function gets called.

There are, IMO, countless usecases for this, but the one very specifically that I'm running in to right now is that Compass's `$base-font-size` only accepts single values, where as the grid system Singularity's `$base-font-size` accepts multiple values to allow for font scaling and has a companion function to find which base font size to use. If I'm writing a function to return an `rem` value, I need to know the proper base font size. If Singularity's base font size finder function exists, I can safely use that function to find the base font size regardless of if it's a single value or multiple values, but if I assume one or the other, the results for the function won't always be accurate, which is a problem. The solution to this is to have a check in place to see if that exists.

robwierzbowski:
+1 from me. I don't think there's going to be any shortage of use cases for an `exists()` function, just the decision whether this type of more programmy construct belongs in Sass. Right now I'm writing code that uses a variable as a flag to tell a wrapper mixin to use either a user-set mixin or a fallback mixin -- with `exists()` I could just test for the user-set mixin and do away with the variable entirely. 

If not as part of Sass itself, this would be a really useful extension.

boram:
+1 from me as well. I have a ruby gem that needs a mixin named clearfix. Since its defined in Compass and Bourbon, it'd be nice to check if its already been defined by those frameworks.

edwardoriordan:
> To this extent, I'd really like to see function-exists() and mixin-exists() functions, or simply an exists() function, that will return a boolean true or false depending on whether or not that function or mixin is indeed available to use when that function gets called.

I created an extension that lets you do this -> https://github.com/edwardoriordan/sass-utilities. It requires sass 3.3.x alpha but that is for variable-exists type functionality (and variable interpolation that is also provides). The code to test for functions and mixins, however, would work fine in sass 3.2.x.

It works by simply querying the environment object to see if the named function or mixin exists. Like

``` ruby
def mixin_exists(mixin_name)
    if(environment.mixin(mixin_name.value))
      Sass::Script::Bool.new(true)
    else
[...]

PR #821, opened 2013-06-25
Existence checks

chriseppstein:
This pull request provides 4 sass functions which all for determining the existence of:
- variables (global or local)
- global variables
- functions
- mixins

This addresses the feature request in issue #561.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -68,6 +68,8 @@ Thanks to Alexander Pavlov for implementing this.
   through `feature-exists` by calling `Sass.add_feature(feature_name)`. Features
   exposed by plugins must begin with a dash to distinguish them from
   official features.
+* The function `variable-exists($named)` can be used to check if a
+  variable exists.
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -68,10 +68,16 @@ Thanks to Alexander Pavlov for implementing this.
   through `feature-exists` by calling `Sass.add_feature(feature_name)`. Features
   exposed by plugins must begin with a dash to distinguish them from
   official features.
-* The function `variable-exists($named)` can be used to check if a
-  variable exists and `global-variable-exists($named)` can be used to
-  check if a global variable exists. To check if a function exists you
-  can use `function-exists($named)`.
+
+* It is now possible to determine the existence of different Sass
+  constructs using these new functions:
+
+  * `variable-exists($named)` check if a variable resolves in the
+    current scope.
+  * `global-variable-exists($named)` check if a global variable of the
+    given name exists.
+  * `function-exists($named)` check if a function exists.
+  * `mixin-exists($named)` check if a mixin exists.
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -72,12 +72,12 @@ Thanks to Alexander Pavlov for implementing this.
 * It is now possible to determine the existence of different Sass
   constructs using these new functions:
 
-  * `variable-exists($named)` check if a variable resolves in the
+  * `variable-exists($name)` checks if a variable resolves in the
     current scope.
-  * `global-variable-exists($named)` check if a global variable of the
+  * `global-variable-exists($name)` checks if a global variable of the
     given name exists.
-  * `function-exists($named)` check if a function exists.
-  * `mixin-exists($named)` check if a mixin exists.
+  * `function-exists($name)` checks if a function exists.
+  * `mixin-exists($name)` checks if a mixin exists.
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -70,7 +70,8 @@ Thanks to Alexander Pavlov for implementing this.
   official features.
 * The function `variable-exists($named)` can be used to check if a
   variable exists and `global-variable-exists($named)` can be used to
-  check if a global variable exists.
+  check if a global variable exists. To check if a function exists you
+  can use `function-exists($named)`.
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -69,7 +69,8 @@ Thanks to Alexander Pavlov for implementing this.
   exposed by plugins must begin with a dash to distinguish them from
   official features.
 * The function `variable-exists($named)` can be used to check if a
-  variable exists.
+  variable exists and `global-variable-exists($named)` can be used to
+  check if a global variable exists.
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/lib/sass/environment.rb
+++ b/lib/sass/environment.rb
@@ -2,67 +2,63 @@ require 'set'
 
 module Sass
 
-  # Provides class methods for declaring a class has hashes that inherit from a parent instance.
-  module HashInheritance
-    private
-    UNDERSCORE, DASH = '_', '-'
+  # The abstract base class for lexical environments for SassScript.
+  class BaseEnvironment
+    class << self
+      # Note: when updating this,
+      # update sass/yard/inherited_hash.rb as well.
+      def inherited_hash_accessor(name)
+        inherited_hash_reader(name)
+        inherited_hash_writer(name)
+      end
 
-    # Note: when updating this,
-    # update sass/yard/inherited_hash.rb as well.
-    def inherited_hash_accessor(name)
-      inherited_hash_reader(name)
-      inherited_hash_writer(name)
-    end
-
-    def inherited_hash_reader(name)
-      class_eval <<RUBY, __FILE__, __LINE__ + 1
-        def #{name}(name)
-          _#{name}(name.tr(UNDERSCORE, DASH))
-        end
-
-        def _#{name}(name)
-          (@#{name}s && @#{name}s[name]) || @parent && @parent._#{name}(name)
-        end
-        protected :_#{name}
-RUBY
-    end
-
-    def inherited_hash_writer(name)
-      class_eval <<RUBY, __FILE__, __LINE__ + 1
-        def set_#{name}(name, value)
-          name = name.tr(UNDERSCORE, DASH)
-          @#{name}s[name] = value unless try_set_#{name}(name, value)
-        end
-
-        def try_set_#{name}(name, value)
-          @#{name}s ||= {}
-          if @#{name}s.include?(name)
-            @#{name}s[name] = value
-            true
-          elsif @parent
-            @parent.try_set_#{name}(name, value)
-          else
-            false
+      def inherited_hash_reader(name)
+        class_eval <<-RUBY, __FILE__, __LINE__ + 1
+          def #{name}(name)
+            _#{name}(name.tr('_', '-'))
           end
-        end
-        protected :try_set_#{name}
 
-        def set_local_#{name}(name, value)
-          @#{name}s ||= {}
-          @#{name}s[name.tr(UNDERSCORE, DASH)] = value
-        end
-RUBY
+          def _#{name}(name)
+            (@#{name}s && @#{name}s[name]) || @parent && @parent._#{name}(name)
+          end
+          protected :_#{name}
+        RUBY
+      end
+
+      def inherited_hash_writer(name)
+        class_eval <<-RUBY, __FILE__, __LINE__ + 1
+          def set_#{name}(name, value)
+            name = name.tr('_', '-')
+            @#{name}s[name] = value unless try_set_#{name}(name, value)
+          end
+
+          def try_set_#{name}(name, value)
+            @#{name}s ||= {}
+            if @#{name}s.include?(name)
+              @#{name}s[name] = value
+              true
+            elsif @parent
+              @parent.try_set_#{name}(name, value)
+            else
+              false
+            end
+          end
+          protected :try_set_#{name}
+
+          def set_local_#{name}(name, value)
+            @#{name}s ||= {}
+            @#{name}s[name.tr('_', '-')] = value
+          end
+        RUBY
+      end
     end
-  end
-
-  # The base class for lexical environments for SassScript.
-  class BasicEnvironment
-    extend HashInheritance
 
     # The options passed to the Sass Engine.
     attr_reader :options
+
     # [Environment] The caller environment
     attr_writer :caller
+
     # [Environment] The content environment
     attr_writer :content
 
@@ -116,14 +112,12 @@ RUBY
   #
   # Environment also keeps track of the {Engine} options
   # so that they can be made available to {Sass::Script::Functions}.
-  class Environment < BasicEnvironment
+  class Environment < BaseEnvironment
     # The enclosing environment,
     # or nil if this is the global environment.
     #
     # @return [Environment]
     attr_reader :parent
-
-    extend HashInheritance
 
     # variable
     # Script::Value
@@ -139,10 +133,10 @@ RUBY
   end
 
   # A read-only wrapper for a lexical environment for SassScript.
-  class ReadOnlyEnvironment < BasicEnvironment
+  class ReadOnlyEnvironment < BaseEnvironment
     # The read-only environment of the caller of this environment's mixin or function.
     #
-    # @see BasicEnvironment#caller
+    # @see BaseEnvironment#caller
     # @return {ReadOnlyEnvironment}
     def caller
       return @caller if @caller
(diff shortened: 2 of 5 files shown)


Snugug:
Want. Is this ready to be merged? I would love to start using it.

kmiyashiro:
:+1: :crying_cat_face:

jslegers:
What about placeholders? To avoid duplication, it would be very useful to test whether a placeholder exists.

r4j4h:
Holy crap, +1 to @jslegers question!

robwierzbowski:
Opened a new issue for the placeholder discussion.

doc-src/SASS_CHANGELOG.md
```diff
@@ -69,6 +69,16 @@ Thanks to Alexander Pavlov for implementing this.
   exposed by plugins must begin with a dash to distinguish them from
   official features.
 
+* It is now possible to determine the existence of different Sass
+  constructs using these new functions:
+
+  * `variable-exists($named)` check if a variable resolves in the
```
nex3: "check" -> "checks" for all of these.

Also, the parameter should be `$name`, so it's a noun.

lib/sass/script/functions.rb
```diff
@@ -179,6 +179,18 @@ module Sass::Script
   # \{#feature_exists feature-exists($feature)}
   # : Returns whether the named feature exists in the current sass runtime.
   #
+  # \{#variable_exists variable-exists($named)}
+  # : Returns whether a variable with the given name resolves in the current scope.
```
nex3: "resolves" -> "exists"

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
     end
     declare :counter, [], :var_args => true
 
+    # Check whether a variable of the given name resolves in the current
+    # scope.
```
nex3: Mention that global variables are also considered to be in the current scope.

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
     end
     declare :counter, [], :var_args => true
 
+    # Check whether a variable of the given name resolves in the current
```
nex3: "of" -> "with"

Same goes for other function docs.

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
...
+    # scope.
+    #
+    # When this method returns false, guarded assignment (assigning a
+    # default value) to that variable name will succeed.
```
nex3: "variable name" -> "variable"

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
...
+    #
+    # @example
+    #   $a-false-value: false; variable-exists(a-false-value) => true
+    #   variable-exists(nonexistent) => false
+    #
+    # @param named [Sass::Script::String] The name of the variable to
+    #   check.
```
nex3: Mention that this doesn't include the `$`. Same for `global-variable-scope`.

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
...
+    # @return [Sass::Script::Bool] Whether the variable is defined in
+    #   the current scope.
+    def variable_exists(named)
+      assert_type named, :String
+      Sass::Script::Value::Bool.new(environment.var(named.value))
+    end
+    declare :variable_exists, [:named]
+
+    # Check whether a variable of the given name exists in the global
+    # scope.
```
nex3: Mention that this means that the variable is defined at the top level of the file, in case people don't understand "global scope".

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
...
     declare :counter, [], :var_args => true
 
+    # Check whether a variable of the given name resolves in the current
+    # scope.
+    #
+    # When this method returns false, guarded assignment (assigning a
+    # default value) to that variable name will succeed.
+    #
+    # @example
+    #   $a-false-value: false; variable-exists(a-false-value) => true
```
nex3: Putting these on the same line makes them difficult to read. Split them across lines and add a newline to separate this example from the next one. Same goes for other examples.

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
...
+      assert_type named, :String
+      Sass::Script::Value::Bool.new(environment.var(named.value))
+    end
+    declare :variable_exists, [:named]
+
+    # Check whether a variable of the given name exists in the global
+    # scope.
+    #
+    # When this method returns false, guarded assignment (assigning a
+    # default value) to that variable name will succeed.
```
nex3: This isn't strictly true; guarded assignment could still fail if there's a local variable with the given name.

I'd just get rid of this paragraph. It's a useful way of thinking about `variable-exists`, but I don't think it helps clarify `global-variable-exists`.

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
...
+    #
+    # @example
+    #   $a-false-value: false; global-variable-exists(a-false-value) => true
+    #   .foo { $a-false-value: false; @if global-variable-exists(a-false-value) { /* false, doesn't run */ } }
+    #   variable-exists(nonexistent) => false
```
nex3: `global-variable-exists`

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
...
+      exists = Sass::Script::Functions.callable?(named.value.tr("-","_"))
+      exists ||= environment.function(named.value)
+      Sass::Script::Value::Bool.new(exists)
+    end
+    declare :function_exists, [:named]
+
+    # Check whether a mixin of the given name exists.
+    #
+    # @example
+    #   mixin-exists(lighten) => true
```
nex3: `lighten` is not a pre-defined mixin.

test/sass/functions_test.rb
```diff
@@ -1255,6 +1257,59 @@ def test_assert_unit
     end
   end
 
+  def test_variable_exists
+    assert_equal "false", evaluate("variable-exists(foo)")
+    assert_equal "true", evaluate("variable-exists(foo)", env("foo" => Sass::Script::Value::Null.new))
```
nex3: I think I'd rather have more integration-y tests for these, like you have for `function-exists` and `mixin-exists`.

test/sass/functions_test.rb
```diff
@@ -1255,6 +1257,59 @@ def test_assert_unit
...
+    assert_equal "true", evaluate("variable-exists($named: foo)", env("foo" => Sass::Script::Value::Null.new))
+  end
```
nex3, line 1279: Add tests that the argument type is checked for all of these. See other tests for examples.

nex3:
Done reviewing.

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,80 @@ def counter(*args)
...
+    #   check.
+    # @return [Sass::Script::Bool] Whether the variable is defined in
+    #   the global scope.
+    def global_variable_exists(named)
+      assert_type named, :String
+      Sass::Script::Value::Bool.new(environment.global_env.var(named.value))
+    end
+    declare :global_variable_exists, [:named]
+
+
```
nex3, line 1753: Nit: extra newline.

doc-src/SASS_CHANGELOG.md
chriseppstein: changed.

lib/sass/script/functions.rb
chriseppstein: changed.

lib/sass/script/functions.rb
chriseppstein: mentioned.

lib/sass/script/functions.rb
chriseppstein: changed.

lib/sass/script/functions.rb
chriseppstein: changed.

lib/sass/script/functions.rb
chriseppstein: mentioned.

lib/sass/script/functions.rb
chriseppstein: mentioned.

lib/sass/script/functions.rb
chriseppstein: Split them up.

lib/sass/script/functions.rb
chriseppstein: Removed from the docs of both methods.

lib/sass/script/functions.rb
chriseppstein: fixed.

lib/sass/script/functions.rb
chriseppstein: fixed.

test/sass/functions_test.rb
chriseppstein: This found a problem for variable-exists. After #934 lands, I can get access to the caller environment and fix it.

test/sass/functions_test.rb
chriseppstein, line 1279: added.

lib/sass/script/functions.rb
chriseppstein, line 1753: removed.

chriseppstein:
There's one test failing. Landing #934 will allow me to fix. @nex3 if you have alternate suggestions for how to fix that test, please suggest.

test/sass/functions_test.rb
```diff
@@ -1255,6 +1257,98 @@ def test_assert_unit
...
+  $foo: has-value;
+  true: variable-exists(foo);
+  true: variable-exists($name: foo);
+  true: variable-exists(global);
+  true: variable-exists($name: global);
+}
+SCSS
+    #assert_equal "false", evaluate("")
+    #assert_equal "true", evaluate("", env("foo" => Sass::Script::Value::Null.new))
+    #assert_equal "true", evaluate("", env("foo" => Sass::Script::Value::Null.new))
```
nex3: Remove commented-out tests.

test/sass/functions_test.rb
```diff
@@ -1255,6 +1257,98 @@ def test_assert_unit
...
+$false: global-variable-exists(foo);
+$true: global-variable-exists(g);
+.test {
+  $foo: locally-defined;
+  false: global-variable-exists(foo);
+  false: global-variable-exists(foo2);
+  true: global-variable-exists(g);
+  true: global-variable-exists(h);
+  false: $false;
+  true: $true;
```
nex3, line 1304: One of these should try using a named argument.

nex3:
I'd fix the issue with `global-variable-exists` by adding an `Environment#local_environment` attribute that exposes an `Environment` subclass that wraps the local environment but throws an exception for any attempts to modify it.

test/sass/functions_test.rb
chriseppstein: removed.

test/sass/functions_test.rb
chriseppstein, line 1304: used.

chriseppstein:
@nex3 moar review please.

lib/sass/environment.rb
```diff
@@ -1,27 +1,83 @@
...
-  # so that they can be made available to {Sass::Script::Functions}.
-  class Environment
-    # The enclosing environment,
-    # or nil if this is the global environment.
-    #
-    # @return [Environment]
-    attr_reader :parent
+
+  # Provides class methods for declaring a class has hashes that inherit from a parent instance.
+  module HashInheritance
```
nex3: I don't like making this a separate module if it's not written in a way that's reusable by other classes. Either make it more generic or go back to making these class methods on `BasicEnvironment`.

If you go the more-generic route, move this into its own file under `lib/sass/util` and have some way to configure the parent (making it assume the existence of a `#parent` accessor is fine, I just don't like it assuming an instance variable).

lib/sass/environment.rb
```diff
@@ -1,27 +1,83 @@
...
+    private
+    UNDERSCORE, DASH = '_', '-'
```
nex3: I know these were here before, but they're gross. If we're modifying this code, might as well change these to just use the literal characters.

lib/sass/environment.rb
```diff
@@ -1,27 +1,83 @@
...
-  #
-  # Environment also keeps track of the {Engine} options
-  # so that they can be made available to {Sass::Script::Functions}.
-  class Environment
-    # The enclosing environment,
-    # or nil if this is the global environment.
-    #
-    # @return [Environment]
-    attr_reader :parent
+
```
nex3, line 4: Nit: extra newline

lib/sass/environment.rb
```diff
@@ -1,27 +1,83 @@
...
+    end
+  end
+
+  # The base class for lexical environments for SassScript.
+  class BasicEnvironment
+    extend HashInheritance
+
+    # The options passed to the Sass Engine.
     attr_reader :options
+    # [Environment] The caller environment
```
nex3, line 59: Nit: newlines between these attributes and the doc comments of the next attribute.

lib/sass/environment.rb
```diff
@@ -47,59 +103,62 @@ def global_env
...
+    attr_reader :parent
 
-          def set_local_#{name}(name, value)
-            @#{name}s ||= {}
-            @#{name}s[name.tr(UNDERSCORE, DASH)] = value
-          end
-RUBY
-      end
-    end
+    extend HashInheritance
```
nex3: You don't need to extend this again here; class methods are preserved through inheritance.

lib/sass/environment.rb
```diff
@@ -1,27 +1,83 @@
...
+
+        def set_local_#{name}(name, value)
+          @#{name}s ||= {}
+          @#{name}s[name.tr(UNDERSCORE, DASH)] = value
+        end
+RUBY
+    end
+  end
+
+  # The base class for lexical environments for SassScript.
```
nex3: "abstract base class"

lib/sass/environment.rb
```diff
@@ -47,59 +103,62 @@ def global_env
...
+    # The read-only content passed to this environment.
+    #
+    # @see BasicEnvironment#content
+    # @return {ReadOnlyEnvironment}
+    def content
+      return @content if @content
+      env = super
+      @content ||= env.is_a?(ReadOnlyEnvironment) ? env : ReadOnlyEnvironment.new(env, env.options)
+    end
+
```
nex3: Nit: extra newline

lib/sass/environment.rb
```diff
@@ -1,27 +1,83 @@
...
+        def set_local_#{name}(name, value)
+          @#{name}s ||= {}
+          @#{name}s[name.tr(UNDERSCORE, DASH)] = value
+        end
+RUBY
+    end
+  end
+
+  # The base class for lexical environments for SassScript.
+  class BasicEnvironment
```
nex3: Since this is abstract (as opposed to a simpler but still instantiable version of Environment), I'd call it `BaseEnvironment` or something like that.

lib/sass/script/functions.rb
```diff
@@ -1700,6 +1712,79 @@ def counter(*args)
...
+    #   variable-exists(a-false-value) => true
+    #
+    #   variable-exists(nonexistent) => false
+    # @param name [Sass::Script::String] The name of the variable to
+    #   check. The name should not include the `$`.
+    # @return [Sass::Script::Bool] Whether the variable is defined in
+    #   the current scope.
+    def variable_exists(name)
+      assert_type name, :String
+      Sass::Script::Value::Bool.new(environment.caller.var(name.value))
```
nex3, line 1729: It seems weird to me that `#environment` is documented as returning the global environment, but the caller of the global environment is now the local environment. I think it would be clearer and more discoverable to have a separate `#local_environment` method in `EvaluationContext`.

nex3:
Done reviewing.

lib/sass/environment.rb
chriseppstein: it's class methods now.

lib/sass/environment.rb
chriseppstein: removed.

lib/sass/environment.rb
chriseppstein, line 59: added.

lib/sass/environment.rb
chriseppstein: it's not a module anymore.

lib/sass/environment.rb
chriseppstein: changed.

lib/sass/environment.rb
chriseppstein: Removed.

lib/sass/environment.rb
chriseppstein: Renamed.

lib/sass/script/functions.rb
chriseppstein, line 1729: The premise of this comment is false. I'm not sure where the documentation says it's the global environment, but that documentation is not technically correct. The environment passed to the evaluation context of the function is a read/write local environment that inherits directly from the global environment. This is the same as you'd have in a function defined in Sass. That that environment has a read only reference to the caller's environment doesn't seem strange at all to me and avoids introducing two distinct ways of accessing the caller environment (`Environment#caller` is not a new accessor).

I need to land this pull request so that I can use it in compass. We can refactor this on master if you feel it's important.

chriseppstein:
Merged as 4e515ab6a46507013fcbe426a5ecfa061b4ae571 after rebasing.

lib/sass/script/functions.rb
nex3, line 1729: The documentation of `EvaluationContext#environment` is "the global environment". I'll correct it.

PR #828, opened 2013-06-27
Threadsafety

### LINKED ISSUE #799: Staleness Checker is not threadsafe
the @dependencies ivar on the StalenessChecker class uses a hash implementation in jruby that is not threadsafe.

chriseppstein:
This PR addresses two concurrency issues we've identified by trying to compile Sass using JRuby where requests are handled by threads instead of processes. I've also created a small framework for running scripts that can be used to reproduce threadsafety issues.

This patch adds a dependency on the `thread_safe` gem so that we can use the `ThreadSafe::Cache` class.

This fixes #799.


nex3:
It looks like the Travis build is failing for this. That should definitely be fixed.

I don't like that there isn't a way to run all the thread safety tests at once. The flexibility of the test runner scripts is nice when designing tests to run, but it seems like the scripts in `test/threadsafety/issues` are each designed to be run in a certain configuration, and those configurations are currently unavailable anywhere. Having a single way to run everything will also allow us to test this on Travis, which I believe currently isn't happening.

One possibility would be to have a threadsafety_test.rb in `test/threadsafety` that uses `Test::Unit` to have a bunch of test cases that just invoke the threadsafety scripts and asserts that they complete with the correct status.

I'm nervous about merging this into stable. Have you run this against enough real-world code that you're confident it won't break?

test/threadsafety/README.md
```diff
@@ -0,0 +1,48 @@
+Reproducing Threadsafety Issues
+===============================
+
+Write a script that exercises the code with the threadsafety issue and
+place it in the `issues` directory. Write code that is as targetted as
```
nex3: "targetted" -> "targeted"

test/threadsafety/README.md
```diff
@@ -0,0 +1,48 @@
...
+number of retries can be set with the `-r` option (default 100). If you
+do not need a fresh ruby runtime to reproduce this, you can call the
+`multi_threaded_runner.rb` script as an executable instead.
+
+In order to handle deadlocks and hangs, the code running in each thread
+will only be terminated after 10 seconds. If you need longer, use the
+`-w` command line switch. When this occurs, the test fails with an
+error.
+
+If the issue is some kind of data corruption, the script should detect
```
nex3, line 21: What does "data corruption" mean? How is it detected?

test/threadsafety/README.md
```diff
@@ -0,0 +1,48 @@
...
+For example:
+
+    # Start ruby 1,000 times and load sass 4 times in each startup.
+    bundle exec ./test_threadsafety.sh -w 60 -t 4 -r 1000 issues/sass_boot.rb
+    
+    # In a single ruby vm, run issues/dependencies_cache.rb in 4 threads after executing the sass_boot.rb script.
+    bundle exec ./test_threadsafety.rb -w 60 -t 4 -s issues/sass_boot.rb issues/dependencies_cache.rb
+
+
+
```
nex3: Lots of extra newlines.

test/threadsafety/Gemfile
```diff
@@ -0,0 +1 @@
+gem 'sass', :path => "../../"
```
nex3, line 2: When I run stuff with this, I get the following nastygram:

```
Your Gemfile has no gem server sources. If you need gems that are not already on your machine, add a line like this to your Gemfile:
source 'https://rubygems.org'
```

We should add the `source` annotation here.

test/Gemfile
```diff
@@ -1,3 +1,5 @@
 source :gemcutter
 
 gem 'rake'
+
+gemspec :path => "../"
```
nex3, line 5: I don't understand why we have a Gemfile in `test`, a different Gemfile in `test/threadsafety`, but no Gemfile at the root. master currently has a Gemfile at the root. Can't we just use that for everything?

test/threadsafety/test_threadsafety.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
nex3, line 1: I really don't want this to be a bash file. Ruby can start up child processes just as well, and it works on Windows and is comprehensible.

The UI for running this script is also very painful. It shouldn't need to be run from `test/threadsafety`, and running it with no arguments or with `--help` or `-h` should print useful help output rather than crashing. It would also be nice to merge it into `multi_threaded_runner.rb`; just have it run itself as a subprocess if the user passes some sort of `--fresh-runtime` flag.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+      wait = [0.01, wait_time - (Time.now - start_time)].max
+      t.join(wait)
+    end
+    if threads.any?
+      raise RuntimeError, "#{threads.size} threads failed to complete in #{@wait_time} seconds"
+    end
+  end
+end
+
+class MultiThreadedScriptRunner < MultiThreadedRunner
```
nex3, line 25: Why is this a separate class from `MultiThreadedRunner`? The latter is never instantiated.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+  DEFAULT_THREAD_COUNT = 4
+  DEFAULT_WAIT_TIME = 10
+  def initialize(thread_count = DEFAULT_THREAD_COUNT, &work_to_do)
+    @thread_count = thread_count
+    @work_to_do = work_to_do
+  end
+
+  def run(wait_time = DEFAULT_WAIT_TIME)
+    start_time = Time.now
+    threads = (0...@thread_count).map{|threadnumber| Thread.new(&@work_to_do) }
```
nex3: Style nit: space before `{` but not before `}`. Also, the `threadnumber` variable is unused.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+    threads.reject! do |t|
+      wait = [0.01, wait_time - (Time.now - start_time)].max
+      t.join(wait)
+    end
+    if threads.any?
+      raise RuntimeError, "#{threads.size} threads failed to complete in #{@wait_time} seconds"
```
nex3: Style nit: `RuntimeError.new(...)`.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+    @thread_count = thread_count
+    @work_to_do = work_to_do
+  end
+
+  def run(wait_time = DEFAULT_WAIT_TIME)
+    start_time = Time.now
+    threads = (0...@thread_count).map{|threadnumber| Thread.new(&@work_to_do) }
+    threads.reject! do |t|
+      wait = [0.01, wait_time - (Time.now - start_time)].max
+      t.join(wait)
```
nex3, line 17: I'm confused by this wait time calculation. For the first thread, you'll wait until `wait_time` expires, but after that `wait_time - (Time.now - start_time)` will always be negative so you'll always just wait 0.01 seconds. If that's the behavior you want, this seems like a roundabout way to get it; if it's not, what are you going for?

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+  DEFAULT_WAIT_TIME = 10
+  def initialize(thread_count = DEFAULT_THREAD_COUNT, &work_to_do)
+    @thread_count = thread_count
+    @work_to_do = work_to_do
+  end
+
+  def run(wait_time = DEFAULT_WAIT_TIME)
+    start_time = Time.now
+    threads = (0...@thread_count).map{|threadnumber| Thread.new(&@work_to_do) }
+    threads.reject! do |t|
```
nex3, line 15: Add a comment explaining that you're rejecting threads that complete on time.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+class MultiThreadedScriptRunner < MultiThreadedRunner
+  def initialize(script_path, thread_count = DEFAULT_THREAD_COUNT)
+    script_contents = File.read(script_path)
+    super(thread_count) do
+      instance_eval script_contents, script_path
+    end
+  end
+end
+
+if File.expand_path($0) == File.expand_path(__FILE__)
```
nex3, line 34: This file is never required. Assume it will always be run as an executable and get rid of this.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+      options[:wait_time] = seconds
+    end
+    opts.on("-s", "--setup START_UP_SCRIPT", String, "A ruby script to load before starting the threads.") do |path|
+      load path
+    end
+    opts.on("-c", "--[no-]clear-cache", "Clear the cache before running (default true).") do |clear|
+      options[:clear_cache] = clear
+    end
+    opts.on_tail("-h", "--help", "Print this message") do
+      puts opts
```
nex3, line 61: Exit here.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+      instance_eval script_contents, script_path
+    end
+  end
+end
+
+if File.expand_path($0) == File.expand_path(__FILE__)
+  require 'optparse'
+  options = {
+    :threads => MultiThreadedRunner::DEFAULT_THREAD_COUNT,
+    :script => nil,
```
nex3, line 38: It's weird that this is in the options hash, given that it's always set to `ARGV.first` and it's never set by the option parser.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  parser = OptionParser.new do |opts|
+    opts.banner = "Usage: test_threadsafety.rb [options] path/to/ruby/script.rb"
+    opts.on("-t", "--threads NUMBER_OF_THREADS", Integer, "How many threads to run concurrently.") do |threads|
+      options[:threads] = threads
+    end
+    opts.on("-w", "--wait NUMBER_OF_SECONDS", Float, "How long to wait for each script to complete (in seconds).") do |seconds|
+      options[:wait_time] = seconds
+    end
+    opts.on("-s", "--setup START_UP_SCRIPT", String, "A ruby script to load before starting the threads.") do |path|
```
nex3: "ruby" -> "Ruby"

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+  parser.parse!
+
+  options[:script] ||= ARGV.first
+
+  if options[:clear_cache]
+    require 'fileutils'
+    FileUtils.rm_rf(".sass-cache")
+  end
+
+  runner = MultiThreadedScriptRunner.new(options[:script], options[:threads])
```
nex3, line 84: If the path to a script isn't provided, this should print the options and exit rather than crashing.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,76 @@
...
+  begin
+    runner.run(options[:wait_time])
+  rescue RuntimeError => e
```
nex3, line 87: Why is this useful? It seems much more useful to let this crash and print the stack trace as well as the exception message.

test/threadsafety/issues/sass_boot.rb
```diff
@@ -0,0 +1,3 @@
+require 'sass'
+require 'sass/plugin'
+puts "Sass Booted"
```
nex3, line 3: Is this file intended to be tested on its own, as the README suggests? If so, add a comment explaining what that test is exercising.

lib/sass/cache_stores/filesystem.rb
```diff
@@ -11,6 +11,7 @@ class Filesystem < Base
 
       # @param cache_location [String] see \{#cache\_location}
       def initialize(cache_location)
+        require 'tempfile'
```
nex3, line 14: Just require this at the top level.

lib/sass/cache_stores/filesystem.rb
```diff
@@ -47,6 +48,17 @@ def _store(key, version, sha, contents)
 
       private
 
+      def atomic_create_and_write_file(filename)
```
nex3, line 58: Explain how and why this is atomic in a doc comment here.

Is it important that this be safe for multiple Sass processes concurrently using the same cache? If so, explain that as well. If not, it's probably cleaner to use a mutex here than rely on the presumptive atomicity of the  filesystem.

lib/sass/cache_stores/filesystem.rb
```diff
@@ -47,6 +48,17 @@ def _store(key, version, sha, contents)
...
+        tmpfile = Tempfile.new(File.basename(filename), File.dirname(filename))
+        begin
+          tmpfile.binmode if tmpfile.respond_to?(:binmode)
+          yield tmpfile
+        ensure
+          tmpfile.close
+        end
+        File.rename tmpfile.path, filename
+      end
```
nex3, line 74: Ensure that the tempfile is deleted if something goes wrong.

test/threadsafety/issues/dependencies_cache.rb
```diff
@@ -0,0 +1,10 @@
+# bundle exec ./multi_threaded_runner.rb -c -s issues/dependencies_cache_setup.rb issues/dependencies_cache.rb
+checker = Sass::Plugin::StalenessChecker.new({:load_paths => [$importer]})
```
nex3, line 2: Explain what this is testing for.

It feels weird to me that this just blithely refers to `$importer` on the assumption that you'll load `dependencies_cache_setup` first. I'd prefer a system where this sort of tight coupling was more obvious. Even if the setup and body had to be defined as strings in a Ruby test file, that would probably be better.

nex3:
Done reviewing.

test/threadsafety/README.md
chriseppstein, line 21: Data corruption can occur at many different levels and in many different ways. It could be that the in-memory consistency of a data structure gets corrupted by concurrent manipulation. Or it could by simultaneous writes to a file on disk. How it is detected is determined on a case-by-case basis.

test/threadsafety/README.md
chriseppstein: removed

test/Gemfile
chriseppstein, line 5: we can. I will do this cleanup in a subsequent pull request.

test/threadsafety/Gemfile
chriseppstein, line 2: added.

test/threadsafety/multi_threaded_runner.rb
chriseppstein, line 25: I like building small things and then composing them. It felt like a natural delineation in my mind. I can merge them together, but I don't really see the point.

test/threadsafety/multi_threaded_runner.rb
chriseppstein: picked.

test/threadsafety/multi_threaded_runner.rb
chriseppstein: picked.

test/threadsafety/multi_threaded_runner.rb
chriseppstein, line 17: The goal is to wait for `wait_time` for all the threads to complete.

The outcome you describe only happens in the case where the timeout is reached on the first thread. This is unlikely. If the timeout is 10 seconds, and threads usually take 5 seconds to complete then if there are four threads and they complete  in 5s, 6s, 4s, and the last one hangs, we'd end up waiting `5 + 1 + 0 + 10 => 16s` instead of the desired `10s`. To accomplish that, we should only wait for the remaining 4 seconds on the last two threads. Worse, if all the threads hang, and we just wait for `wait_time` on each, we'd end up waiting `40s`.

If you have a more clear way to express this same logic, let me know.

test/threadsafety/multi_threaded_runner.rb
chriseppstein, line 15: added.

test/threadsafety/multi_threaded_runner.rb
chriseppstein, line 61: good catch. added.

test/threadsafety/multi_threaded_runner.rb
chriseppstein, line 38: Doesn't seem weird to me. not all options to the program must be passed as command-line switches. The result of inspecting `ARGV` is `options`.

test/threadsafety/multi_threaded_runner.rb
chriseppstein: picked.

test/threadsafety/multi_threaded_runner.rb
chriseppstein, line 84: added.

test/threadsafety/multi_threaded_runner.rb
chriseppstein, line 87: A stack trace is warranted for _unexpected_ errors. However, some failure conditions are expected by running this script (A hung thread) and a stack trace in this situation appears to be a script bug instead of part of the normal operation. I have added a `--trace` option so that unexpected errors can get the full trace.

test/threadsafety/issues/sass_boot.rb
chriseppstein, line 3: This is both a test to make sure that the sass boot process doesn't have any threadsafety issues, as well as being useful as a startup script that only runs once. The `puts` documents what it is testing.

lib/sass/cache_stores/filesystem.rb
chriseppstein, line 14: I prefer to not load libraries at startup unless it's inevitable that they will be loaded downstream. The filesystem cache is only instantiated if `:cache_store` is not set for a `Sass::Engine`. So delaying this require is appropriate.

lib/sass/cache_stores/filesystem.rb
chriseppstein, line 58: > Explain how and why this is atomic in a doc comment here. 

Added.

> Is it important that this be safe for multiple Sass processes concurrently using the same cache?

Yes. People leaving multiple watchers running on accident is a real and common occurrence.

>  If not, it's probably cleaner to use a mutex

An in-memory Mutex only works for a single process. Using `File.flock` doesn't work on windows. Also, a mutex introduces delays and deadlock possibilities. Lock-free implementations like the one we are using here are the "cleaner" implementation to handle concurrency.

> rely on the presumptive atomicity of the filesystem

This is a posix guarantee and it is also true on windows' filesystems. It is a very commonly employed strategy. If there are filesystems that do not make this guarantee and need to be supported, we can address those when a user comes to us with that need. I don't know of such a filesystem that would ever by used by a Sass developer.

lib/sass/cache_stores/filesystem.rb
chriseppstein, line 74: Good catch. ensured.

chriseppstein:
> I don't like that there isn't a way to run all the thread safety tests at once. The flexibility of the test runner scripts is nice when designing tests to run, but it seems like the scripts in test/threadsafety/issues are each designed to be run in a certain configuration, and those configurations are currently unavailable anywhere. Having a single way to run everything will also allow us to test this on Travis, which I believe currently isn't happening.

This would be amazing. Let me explain why I think it's not a reasonable goal and what the goal I was hoping to achieve is and maybe we can find common ground about what is the best way to accomplish this.

Thread safety issues can be notoriously hard to reproduce. The bad one we have here at work happens on one machine approximately once a week. It may be that this happens only once in a hundred thousand compiles or only under high load. When this particular issue happens, the process starts using all available CPU and we found the issue by attaching a remote debugger and seeing what the code was doing. I believe that I have written a script that would eventually reproduce this issue, but after letting it run many hours, even in a very tight loop, I did not encounter the issue and I gave up. As far as I know, this issue is only reproducible on JRuby. Other issues may be reproducible only on Rubinius. MRI is much more thread safe than other implementations because of the GIL.

Given all this, you might ask why bother writing a script at all? Mostly, I wanted to provide a set of tools for creating a reproducible test case while developing a fix. Even if running the code for hours is the only way to reproduce it, it adds to the level of confidence in the fix. I'd like each fix to come with detailed instructions for how to reproduce the issue manually using these tools.

It's not clear to me that we need to worry about regressions in the same code, preventing concurrency bugs are more about knowing what not to do, and what best practices to follow.

> I'm nervous about merging this into stable. Have you run this against enough real-world code that you're confident it won't break?

What makes you nervous? It's hard to say whether my own testing would satisfy you. I will be deploying a monkey patched version next week -- it will take a 2-4 weeks to know for sure whether the dependencies cache issue is fixed. But I don't think we should wait to merge this, theoretically, it fixes the issue. Our code is doing something that the docs say explicitly it should not. If this happens again, I will follow up with a subsequent adjustment -- perhaps there are other hashes shared across threads that have escaped my notice.

test/threadsafety/multi_threaded_runner.rb
chriseppstein, line 34: This will no longer be true when I convert the bash script to a ruby script.

test/threadsafety/test_threadsafety.sh
chriseppstein, line 1: > It shouldn't need to be run from `test/threadsafety`

Not sure where you got that, this script runs fine from any directory.

> I really don't want this to be a bash file.

I will rewrite it in ruby so that it is using a technology you find more comprehensible.

> It would also be nice to merge it into multi_threaded_runner.rb;

I agree. This is how I started, I tried to do it using fork, but ran into issues in jruby where fork is not supported. After spinning my wheels for a while I decided to change tack. But I will try again using a `system()` call instead of fork.

test/threadsafety/README.md
chriseppstein: fixed.

test/threadsafety/issues/dependencies_cache.rb
chriseppstein, line 2: I'll think about this.

thedarkone:
> I'm nervous about merging this into stable. Have you run this against enough real-world code that you're confident it won't break?

Aren't the changes (except adding the new `thread_safe` dependency) trivial or only adding more test cases? `thread_safe` gem can be avoided by wrapping `StalenessChecker` global cache in a mutex... or is this about (post merge) confidently declaring `sass` to be thread safe? Or are you not sure that the `StalenessChecker`'s global `Hash` is a problem?

ad concurrent `Hash` modification:
https://gist.github.com/thedarkone/559673 <- this should almost 99% break when run on JRuby and Rubinius.

lib/sass/cache_stores/filesystem.rb
```diff
@@ -47,6 +48,31 @@ def _store(key, version, sha, contents)
 
       private
 
+      # This creates a temp file and yields it for writing. When the
+      # write is complete, the file is moved into the desired location
+      # the atomicity of this operation is provided by the filesystem's
```
nex3, line 53: "location the" -> "location. The"

lib/sass/cache_stores/filesystem.rb
```diff
@@ -47,6 +48,31 @@ def _store(key, version, sha, contents)
...
+      # rename operation.
+      #
+      # @param filename the file to write to.
+      # @yieldparam tmpfile the tmpfile that can be written to.
```
nex3, line 57: Type-annotate these parameters, and add `@yield [tmpfile]` above.

lib/sass/cache_stores/filesystem.rb
```diff
@@ -47,6 +48,31 @@ def _store(key, version, sha, contents)
...
+      def atomic_create_and_write_file(filename)
+        tmpfile = Tempfile.new(File.basename(filename), File.dirname(filename))
+        tmp_path = tmpfile.path
+        begin
+          begin
+            tmpfile.binmode if tmpfile.respond_to?(:binmode)
+            yield tmpfile
+          ensure
+            tmpfile.close
```
nex3, line 66: This is opened outside of this `begin`/`ensure` block. Shouldn't it be closed even if we don't reach the `begin`? Same goes for calling `FileUtils.rm_f`.

test/threadsafety/multi_threaded_runner.rb
```diff
@@ -0,0 +1,94 @@
...
+  parser.parse!
+
+  if ARGV.size == 1
+    options[:script] ||= ARGV.first
+  elsif ARGV.size == 0
+    $stderr.puts "A script to run is required."
+    $stderr.puts parser
+    exit 1
+  else
+    $stderr.puts "Only one script may be ran at a time. Got: #{ARGV.join(', ')}"
```
nex3, line 74: "ran" -> "run"

test/threadsafety/test_threadsafety.sh
nex3, line 1: It looks like test_threadsafety.sh still exists. Are there more changes coming down the pipe for this PR?

test/threadsafety/multi_threaded_runner.rb
nex3, line 25: I like doing that when it makes the code clearer or serves some other purpose. In this case, all the test cases are going to exist outside of this file, so you're never going to instantiate MultiThreadedRunner. Its existence just serves to confuse the reader.

test/threadsafety/multi_threaded_runner.rb
nex3, line 38: But it's not really an option in any sense. At one point you `||=` it as though it is, but it's always `nil` at that point so it's always set to `ARGV.first`. It would be a lot clearer if you just passed `ARGV.first` in to `MultiThreadedScriptRunner.new` directly without all this indirection that ultimately does nothing.

nex3:
If this is a toolkit for creating reproductions of thread safety issues, that needs to be documented more clearly in the README. Keep in mind that people will be coming in with the assumption that they'll be looking at regression tests for past thread safety issues; you need to explain what it actually is in that context.

It also doesn't make much sense to me that there are test cases under `test/threadsafety/issues` checked in to the repository if they aren't going to be run regularly as regression tests. I'd like to see those removed entirely. If you want to have a record of issues we've encountered before, I'd rather have that in prose than in code if the code can't easily be run as regression tests.

As for my issues with merging this, it's mostly that `atomic_create_and_write_file` is a reasonably large function that does some filesystem stuff, which is always an avenue for potential instability. I'm fine with trying it, but I want to be on alert to see if anyone experiences any issues.

zaius:
I'm running into what appears to be concurrency issues with sass when rendering templates concurrently while under high load. The error that comes out of it is:

```
ActionView::Template::Error: No such file or directory - ./.sass-cache/cfe593394911620d188a2e3f090247688848d599/(__TEMPLATE__)c
```

Here's the stacktrace: https://gist.github.com/zaius/6b9db85d50d4e408d187

Not sure if I should post this as a new issue or not, but it seems to be related to lack of thread safety. So let me know if it makes sense to merge this into my branch and do some testing.

nex3:
@zaius that's plausibly caused by stuff this change addresses. Try merging it and let us know if that helps.

chriseppstein:
I'm abandoning this pull request. The fixes provided here are now are important and need to land ASAP so I've made them available as individual pull requests against stable:
- https://github.com/nex3/sass/pull/938
- https://github.com/nex3/sass/pull/939

We'll continue to address threadsafety issues without providing any sort of special testing utilities for contributors.

PR #831, opened 2013-06-28
Allow modulo arithmetric for numbers with units

icedtoast:
For example, 24px % 13px will return 11px. Removal of previous exception
that was used when units were present "Cannot modulo by a number with
units: 13px"


chriseppstein:
:thumbsup:

nex3:
Merged in c8086d75437610142f33427b1f6af6b9b34baba8.

PR #834, opened 2013-07-02
Value helpers

chriseppstein:
It turns out you can only type `Sass::Script::Value::<type>.new` so many times before you want to :cry:. I recently passed that limit and coded this up while sitting in a pool of tears.

This is a refactor to make constructing sass values from ruby extensions more convenient. The API is tailored to how developers work with the Sass values instead of how the parser wants to construct the same objects. This refactor significantly cleaned up our own ruby-based Sass functions.

P.S. Yes, I am aware that I can include `Sass::Script::Value` in order to make object creation less typing, but even without that, the creation patterns for developers differs significantly from that of the parser. And I find a function call to be more aesthetically pleasing in this context than a call to `.new` on a class.

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -1779,8 +1779,7 @@ module Sass::Script
         value.slice!(length)
       end
       new_list_value = values.first.zip(*values[1..-1])
-      new_list_arry = new_list_value.map {|list| Sass::Script::Value::List.new(list, :space)}
-      Sass::Script::Value::List.new(new_list_arry, :comma)
+      list(new_list_value.map{|list| list(list, :space)}, :comma)
     end
     declare :zip, [], :var_args => true
 
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -1,3 +1,5 @@
+require 'sass/script/value/helpers'
+
 module Sass::Script
   # Methods in this module are accessible from the SassScript context.
   # For example, you can write
@@ -373,6 +375,7 @@ module Sass::Script
     # are available to use in functions.
     class EvaluationContext
       include Functions
+      include Value::Helpers
 
       # The human-readable names for [Sass::Script::Value::Base]. The default is
       # just the downcased name of the type. The default is the downcased type
@@ -1494,7 +1497,7 @@ module Sass::Script
     def comparable(number_1, number_2)
       assert_type number_1, :Number, :number_1
       assert_type number_2, :Number, :number_2
-      Sass::Script::Value::Bool.new(number_1.comparable_to?(number_2))
+      bool(number_1.comparable_to?(number_2))
     end
     declare :comparable, [:number_1, :number_2]
 
@@ -1803,7 +1806,7 @@ module Sass::Script
       if index
         Sass::Script::Value::Number.new(index + 1)
       else
-        Sass::Script::Value::Bool::FALSE
+        bool(false)
       end
     end
     declare :index, [:list, :value]
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -586,7 +586,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$saturation` or `$lightness` are out of bounds
     #   or any parameter is the wrong type
     def hsl(hue, saturation, lightness)
-      hsla(hue, saturation, lightness, Sass::Script::Value::Number.new(1))
+      hsla(hue, saturation, lightness, number(1))
     end
     declare :hsl, [:hue, :saturation, :lightness]
 
@@ -638,7 +638,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def red(color)
       assert_type color, :Color, :color
-      Sass::Script::Value::Number.new(color.red)
+      number(color.red)
     end
     declare :red, [:color]
 
@@ -654,7 +654,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def green(color)
       assert_type color, :Color, :color
-      Sass::Script::Value::Number.new(color.green)
+      number(color.green)
     end
     declare :green, [:color]
 
@@ -670,7 +670,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def blue(color)
       assert_type color, :Color, :color
-      Sass::Script::Value::Number.new(color.blue)
+      number(color.blue)
     end
     declare :blue, [:color]
 
@@ -688,7 +688,8 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def hue(color)
       assert_type color, :Color, :color
-      Sass::Script::Value::Number.new(color.hue, ["deg"])
+      assert_type color, :Color
+      number(color.hue, "deg")
     end
     declare :hue, [:color]
 
@@ -706,7 +707,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def saturation(color)
       assert_type color, :Color, :color
-      Sass::Script::Value::Number.new(color.saturation, ["%"])
+      number(color.saturation, "%")
     end
     declare :saturation, [:color]
 
@@ -724,7 +725,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def lightness(color)
       assert_type color, :Color, :color
-      Sass::Script::Value::Number.new(color.lightness, ["%"])
+      number(color.lightness, "%")
     end
     declare :lightness, [:color]
 
@@ -744,13 +745,13 @@ module Sass::Script
              a.value =~ /^[a-zA-Z]+\s*=/
          end
         # Support the proprietary MS alpha() function
-        return Sass::Script::Value::String.new("alpha(#{args.map {|a| a.to_s}.join(", ")})")
+        return identifier("alpha(#{args.map {|a| a.to_s}.join(", ")})")
       end
 
       raise ArgumentError.new("wrong number of arguments (#{args.size} for 1)") if args.size != 1
 
       assert_type args.first, :Color, :color
-      Sass::Script::Value::Number.new(args.first.alpha)
+      number(args.first.alpha)
     end
     declare :alpha, [:color]
 
@@ -763,10 +764,10 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def opacity(color)
       if color.is_a?(Sass::Script::Value::Number)
-        return Sass::Script::Value::String.new("opacity(#{color})")
+        return identifier("opacity(#{color})")
       end
       assert_type color, :Color, :color
-      Sass::Script::Value::Number.new(color.alpha)
+      number(color.alpha)
     end
     declare :opacity, [:color]
 
@@ -869,7 +870,7 @@ module Sass::Script
     def saturate(color, amount = nil)
       # Support the filter effects definition of saturate.
       # https://dvcs.w3.org/hg/FXTF/raw-file/tip/filters/index.html
-      return Sass::Script::Value::String.new("saturate(#{color})") if amount.nil?
+      return identifier("saturate(#{color})") if amount.nil?
       _adjust(color, amount, :saturation, 0..100, :+, "%")
     end
     declare :saturate, [:color, :amount]
@@ -929,7 +930,7 @@ module Sass::Script
     def ie_hex_str(color)
       assert_type color, :Color, :color
       alpha = (color.alpha * 255).round.to_s(16).rjust(2, '0')
-      Sass::Script::Value::String.new("##{alpha}#{color.send(:hex_str)[1..-1]}".upcase)
+      identifier("##{alpha}#{color.send(:hex_str)[1..-1]}".upcase)
     end
     declare :ie_hex_str, [:color]
 
@@ -1143,7 +1144,7 @@ module Sass::Script
     # @return [Sass::Script::Value::Color]
     # @raise [ArgumentError] if `$weight` is out of bounds or any parameter is
     #   the wrong type
-    def mix(color_1, color_2, weight = Sass::Script::Value::Number.new(50))
+    def mix(color_1, color_2, weight = number(50))
       assert_type color_1, :Color, :color_1
       assert_type color_2, :Color, :color_2
       assert_type weight, :Number, :weight
@@ -1178,7 +1179,7 @@ module Sass::Script
 
       rgb = color_1.rgb.zip(color_2.rgb).map {|v1, v2| v1 * w1 + v2 * w2}
       alpha = color_1.alpha * p + color_2.alpha * (1 - p)
-      Sass::Script::Value::Color.new(rgb + [alpha])
+      rgb_color(*rgb, alpha)
     end
     declare :mix, [:color_1, :color_2]
     declare :mix, [:color_1, :color_2, :weight]
@@ -1193,9 +1194,9 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def grayscale(color)
       if color.is_a?(Sass::Script::Value::Number)
-        return Sass::Script::Value::String.new("grayscale(#{color})")
+        return identifier("grayscale(#{color})")
       end
-      desaturate color, Sass::Script::Value::Number.new(100)
+      desaturate color, number(100)
     end
     declare :grayscale, [:color]
 
@@ -1208,7 +1209,7 @@ module Sass::Script
     # @return [Sass::Script::Value::Color]
     # @raise [ArgumentError] if `$color` isn't a color
     def complement(color)
-      adjust_hue color, Sass::Script::Value::Number.new(180)
+      adjust_hue color, number(180)
     end
     declare :complement, [:color]
 
@@ -1221,7 +1222,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$color` isn't a color
     def invert(color)
       if color.is_a?(Sass::Script::Value::Number)
-        return Sass::Script::Value::String.new("invert(#{color})")
+        return identifier("invert(#{color})")
       end
 
       assert_type color, :Color, :color
@@ -1244,8 +1245,8 @@ module Sass::Script
     # @return [Sass::Script::Value::String]
     # @raise [ArgumentError] if `$string` isn't a string
     def unquote(string)
-      if string.is_a?(Sass::Script::Value::String)
-        Sass::Script::Value::String.new(string.value, :identifier)
+      if string.is_a?(Sass::Script::Value::String) && string.type != :identifier
+        identifier(string.value)
       else
         string
       end
@@ -1283,7 +1284,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$string` isn't a string
     def str_length(string)
       assert_type string, :String, :string
-      Sass::Script::Value::Number.new(string.value.size)
+      number(string.value.size)
     end
     declare :str_length, [:string]
 
@@ -1343,7 +1344,7 @@ module Sass::Script
       assert_type string, :String, :string
       assert_type substring, :String, :substring
       index = string.value.index(substring.value) || -1
-      Sass::Script::Value::Number.new(index + 1)
+      number(index + 1)
     end
     declare :str_index, [:string, :substring]
 
@@ -1373,7 +1374,7 @@ module Sass::Script
       assert_type string, :String, :string
       assert_unit start_at, nil, "start-at"
 
-      end_at = Sass::Script::Value::Number.new(-1) if end_at.nil?
+      end_at = number(-1) if end_at.nil?
       assert_unit end_at, nil, "end-at"
 
       s = start_at.value > 0 ? start_at.value - 1 : start_at.value
@@ -1432,7 +1433,7 @@ module Sass::Script
     # @return [Sass::Script::Value::String] The unquoted string name of the
     #   value's type
     def type_of(value)
-      Sass::Script::Value::String.new(value.class.name.gsub(/Sass::Script::Value::/, '').downcase)
+      identifier(value.class.name.gsub(/Sass::Script::Value::/, '').downcase)
     end
     declare :type_of, [:value]
 
@@ -1448,7 +1449,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$feature` isn't a string
     def feature_exists(feature)
       assert_type feature, :String, :feature
-      Sass::Script::Value::Bool.new(Sass.has_feature?(feature.value))
+      bool(Sass.has_feature?(feature.value))
     end
     declare :feature_exists, [:feature]
 
@@ -1483,7 +1484,7 @@ module Sass::Script
     # @raise [ArgumentError] if `$number` isn't a number
     def unitless(number)
       assert_type number, :Number, :number
-      Sass::Script::Value::Bool.new(number.unitless?)
+      bool(number.unitless?)
     end
     declare :unitless, [:number]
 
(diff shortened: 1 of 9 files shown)


ggilder:
:+1:

lib/sass/script/functions.rb
```diff
@@ -1167,7 +1170,11 @@ def unquote(string)
     #   quote(foo) => "foo"
     def quote(string)
       assert_type string, :String
-      Sass::Script::Value::String.new(string.value, :string)
+      if string.type != :string
+        string(string.value)
+      else
+        string
```
nex3, line 1272: Why is it worth returning the same object here, as opposed to just doing `string(string.value)` everywhere?

lib/sass/script/value/color.rb
```diff
@@ -248,6 +248,18 @@ def initialize(attrs, allow_both_rgb_and_hsl = false)
       @attrs[:alpha] = Sass::Util.check_range("Alpha channel", 0..1, @attrs[:alpha])
     end
 
+    def self.from_hex(hex_string, alpha = nil)
```
nex3, line 256: This needs a doc comment.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
+module Sass::Script::Value
+  # Provides helper functions for creating sass values from within ruby methods
+  module Helpers
+
```
nex3: Nit: extra newline

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
+module Sass::Script::Value
+  # Provides helper functions for creating sass values from within ruby methods
```
nex3: Nit: period

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+  module Helpers
+
+    # Construct a Sass Boolean value.
```
nex3: "Sass Boolean value" -> "Sass::Script::Value::Bool". Same for other doc comments.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    #
+    # This method returns boolean constants instead of creating a new value object
+    # with each call.
```
nex3: This is an implementation detail that people calling this won't care about. If it needs to be documented, it should be documented in the class documentation for `Bool`.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    #
+    # @since `3.3.0`
```
nex3: It seems weird to me that you're using this annotation here, but nowhere else. If we're going to use it, we should allow readers of the API to expect it to be used thoroughly, at least for stuff that's relevant to creating custom functions.

Also, can't we just put this on the module itself rather than each of the functions?

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    # @return [Sass::Script::Value::Bool] whether the ruby value is truthy.
```
nex3, line 8: Needs `@param` documentation for `value`.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    def bool(value)
+      Bool.new(value)
+    end
+
+    # Construct a Sass Color from a hex color string.
+    #
+    # @since `3.3.0`
+    # @param value [::String] A string representing a hex color.
+    #                         The leading hash ("#") is optional.
```
nex3: This should only be indented two spaces. Same goes for other parameter docs.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    # @return [Sass::Script::Value::Color] the color object
+    def hex_color(value, alpha = nil)
+      Color.from_hex(value, alpha)
+    end
+
+    # Construct a Sass Color from hsl values.
+    #
+    # @since `3.3.0`
+    # @param hue [::Number] The hue of the color in degrees. A number generally
+    #                       between 0 and 360, but it must only be non-negative.
```
nex3: I'd word this "a non-negative number, usually less than 360."

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+      numerator_units = Sass::Script::Value::Number::NO_UNITS
+      denominator_units = Sass::Script::Value::Number::NO_UNITS
+      if unit_string
+        num_over_denominator = unit_string.split(/ *\/ */)
+        unless (1..2).include?(num_over_denominator.size)
+          raise ArgumentError.new("Malformed unit string: #{unit_string}")
+        end
+        numerator_units = num_over_denominator[0].split(/ *\* */)
+        denominator_units = (num_over_denominator[1] || "").split(/ *\* */)
+        unless numerator_units.size >= 1 && numerator_units.all? {|unit| unit =~ VALID_UNIT }
```
nex3: Nit: no space before `}`.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    #   `numeral_unit1 * numeral_unit2 ... / denominator_unit1 * denominator_unit2 ...`
+    #   this is the same format that is returned by {Sass::Script::Value::Number#unit_str the `unit_str` method}
+    #
+    # @see Sass::Script::Value::Number#unit_str
+    #
+    # @return [Sass::Script::Value::Number] The sass number representing the given ruby number.
+    def number(number, unit_string = nil)
+      numerator_units = Sass::Script::Value::Number::NO_UNITS
+      denominator_units = Sass::Script::Value::Number::NO_UNITS
+      if unit_string
```
nex3: Short-circuit here if there's no unit string.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+        num_over_denominator = unit_string.split(/ *\/ */)
+        unless (1..2).include?(num_over_denominator.size)
+          raise ArgumentError.new("Malformed unit string: #{unit_string}")
+        end
+        numerator_units = num_over_denominator[0].split(/ *\* */)
+        denominator_units = (num_over_denominator[1] || "").split(/ *\* */)
+        unless numerator_units.size >= 1 && numerator_units.all? {|unit| unit =~ VALID_UNIT }
+          raise ArgumentError.new("Malformed numerator in unit string: #{unit_string}")
+        end
+        unless denominator_units.all? {|unit| unit =~ VALID_UNIT }
```
nex3: Nit: no space before `}`.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    # @overload space_list(*elements)
+    #   Create a space-separated list from the arguments given.
+    #   @param elements [Array<Sass::Script::Value::Base>] Each argument will be a list element.
+    #   @return [Sass::Script::Value::List] The space separated list.
+    #
+    # @overload space_list(array)
+    #   Create a space-separated list from the array given.
+    #   @param array [Array<Sass::Script::Value::Base>] A ruby array of Sass values to make into a list.
+    #   @return [Sass::Script::Value::List] The space separated list.
+    def space_list(*elements)
```
nex3: I feel like `list(..., :comma)`/`list(..., :space)` would be a cleaner API. You can just detect if the last element is a symbol, and default to `:comma` if the user doesn't specify.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    #
+    # @since `3.3.0`
+    # @return [Sass::Script::Value::Null]
+    def null
+      Sass::Script::Value::Null.new
+    end
+
+    # Create a sass string.
+    #
+    # In most contexts, a sass string outputs with quotes surrounding it.
```
nex3: Instead of this, call it a "quoted string". That's the terminology that all the documentation uses, and we should be consistent.

lib/sass/script/value/helpers.rb
```diff
@@ -0,0 +1,158 @@
...
+    # @since `3.3.0`
+    # @param str [::String] A ruby string.
+    # @return [Sass::Script::Value::String] A sass string.
+    def string(str)
+      Sass::Script::String.new(str, :string)
+    end
+
+    # Create a sass identifier.
+    #
+    # Identifiers are like strings, but they do not have quotes.
```
nex3: As above, just call this an "unquoted string". Calling the function `identifier` is fine for brevity, but the docs should line up with docs elsewhere.

lib/sass/script/value/null.rb
```diff
@@ -1,9 +1,18 @@
...
-    def initialize
-      super nil
+
+    # The null value in SassScript.
+    #
+    # This is assigned before new is overridden below so that we use the default implementation.
+    NULL = new(nil)
+
+    # We override object creation so that users of the core API
+    # will not need to know that null is a specific constant.
```
nex3, line 11: This reads weird in the public documentation for the function. I'd put it in the body of the method.

lib/sass/script/value/number.rb
```diff
@@ -60,6 +60,9 @@ def self.precision_factor
     def initialize(value, numerator_units = NO_UNITS, denominator_units = NO_UNITS)
       numerator_units = [numerator_units] if numerator_units.is_a?(::String)
       denominator_units = [denominator_units] if denominator_units.is_a?(::String)
+      unless value.is_a?(::Numeric)
+        raise ArgumentError.new("#{value.inspect} is not a number.")
+      end
```
nex3: Ruby is dynamically-typed; I don't like manually adding type checks.

test/sass/script_test.rb
```diff
@@ -613,4 +613,8 @@ def test_number_printing
     assert_equal "-Infinity", eval("-1.0/0.0")
     assert_equal "NaN", eval("0.0/0.0")
   end
+
+  def test_null_is_a_singleton
+    assert_same Sass::Script::Value::Null.new, Sass::Script::Value::Null.new
+  end
```
nex3, line 687: I don't like this test. It's just testing an implementation detail.

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+
+class ValueHelpersTest < Test::Unit::TestCase
+  include Sass::Script
+  include Sass::Script::Value::Helpers
+
+  def test_bool
+    assert_same Value::Bool::TRUE, bool(true)
+    assert_same Value::Bool::FALSE, bool(false)
+    assert_same Value::Bool::FALSE, bool(nil)
+    assert_same Value::Bool::TRUE, bool(Object.new)
```
nex3, line 12: Same comment here as for `test_null_is_a_singleton`. This should just use `assert_equal`.

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  def test_hex_color
+    color_without_hash = hex_color("FF007F")
+    assert_equal 255, color_without_hash.red
+    assert_equal 0, color_without_hash.green
+    assert_equal 127, color_without_hash.blue
+    assert_equal 1, color_without_hash.alpha
+
+    color_with_hash = hex_color("#FF007F")
```
nex3, line 32: I'd rather have separate functions with more specific names for each of these test cases (e.g. `test_hex_color_without_hash`, `test_hex_color_with_alpha`, etc.)

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    assert_equal 0, color_with_hash.green
+    assert_equal 127, color_with_hash.blue
+    assert_equal 1, color_with_hash.alpha
+
+    color_with_alpha = hex_color("FF007F", 0.5)
+    assert_equal 0.5, color_with_alpha.alpha
+
+    assert_raises ArgumentError do
+      hex_color("FF007F", 50)
+    end
```
nex3, line 57: What about argument errors from ill-formed colors? What about three-digit hex codes?

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    assert_equal 0.5, no_alpha.saturation
+    assert_equal 1, no_alpha.lightness
+    assert_equal 1, no_alpha.alpha
+
+    has_alpha = hsl_color(1, 0.5, 1, 0.5)
+    assert_equal 1, has_alpha.hue
+    assert_equal 0.5, has_alpha.saturation
+    assert_equal 1, has_alpha.lightness
+    assert_equal 0.5, has_alpha.alpha
+  end
```
nex3, line 74: Test for argument errors for out-of-bounds arguments.

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    assert_equal 0, no_alpha.green
+    assert_equal 0, no_alpha.blue
+    assert_equal 1, no_alpha.alpha
+
+    has_alpha = rgb_color(255, 255, 255, 0.5)
+    assert_equal 255, has_alpha.red
+    assert_equal 255, has_alpha.green
+    assert_equal 255, has_alpha.blue
+    assert_equal 0.5, has_alpha.alpha
+  end
```
nex3, line 90: Out-of-bounds argument tests here as well.

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+
+  def test_number
+    n = number(1);
```
nex3: Nit: unnecessary semicolon. Also several times below.

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    assert_equal 1, n.value
+    assert_equal "1px", n.to_sass
+
+    # single numerator and denominator units
+    ratio = number(1, "px/em");
+    assert_equal "1px/em", ratio.to_sass
+
+    # many numerator and denominator units
+    complex = number(1, "px*in/em*%");
+    assert_equal "1#{['px','in'].sort.join("*")}/#{['em','%'].sort.join("*")}", complex.to_sass
```
nex3: This `.sort.join` stuff just obfuscates the code. We know what the exact expected output is; just include it.

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+
+    # many numerator and denominator units with spaces
+    complex = number(1, "px * in / em * %");
+    assert_equal "1#{['px','in'].sort.join("*")}/#{['em','%'].sort.join("*")}", complex.to_sass
+  end
```
nex3, line 129: Definitely test malformed arguments here.

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    assert_equal %q{"sassy string"}, s.to_sass
+  end
+
+  def test_identifier
+    s = identifier("a-sass-ident")
+    s.options = {}
+    assert_kind_of Sass::Script::Value::String, s
+    assert_equal "a-sass-ident", s.value
+    assert_equal :identifier, s.type
+    assert_equal %q{a-sass-ident}, s.to_sass
```
nex3: Why is this using `%q`?

test/sass/value_helpers_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    assert_kind_of Sass::Script::Value::Null, null
+  end
+
+  def test_string
+    s = string("sassy string")
+    s.options = {}
+    assert_kind_of Sass::Script::Value::String, s
+    assert_equal "sassy string", s.value
+    assert_equal :string, s.type
+    assert_equal %q{"sassy string"}, s.to_sass
```
nex3: `%q` is less clear than just using single quotes here.

nex3:
Done reviewing.

lib/sass/script/functions.rb
chriseppstein, line 1272: I don't like making new objects unnecessarily.

lib/sass/script/value/color.rb
chriseppstein, line 256: Added.

lib/sass/script/value/helpers.rb
chriseppstein: De-newlinified.

lib/sass/script/value/helpers.rb
chriseppstein: Added.

lib/sass/script/value/helpers.rb
chriseppstein: The exact class returned is specified in the `@return`. I want this to read nicely.

lib/sass/script/value/helpers.rb
chriseppstein: Removed.

lib/sass/script/value/helpers.rb
chriseppstein: > If we're going to use it, we should allow readers of the API to expect it to be used thoroughly

I don't ascribe to the idea that unless something can be done right everywhere there is no point in doing the right thing in this case. But I agree it would be good to document the public API for writing Sass functions with release information.

> Also, can't we just put this on the module itself rather than each of the functions?

Maybe? I don't think that's as useful when reading the actual code instead of the generated docs.

lib/sass/script/value/helpers.rb
chriseppstein, line 8: added

lib/sass/script/value/helpers.rb
chriseppstein: Changed.

lib/sass/script/value/helpers.rb
chriseppstein: changed.

lib/sass/script/value/helpers.rb
chriseppstein: denitted.

lib/sass/script/value/helpers.rb
chriseppstein: I don't think that makes this method better.

lib/sass/script/value/helpers.rb
chriseppstein: denitted.

lib/sass/script/value/helpers.rb
chriseppstein: I didn't want to have a default separator. I feel like this is a detail that is easy for many developers to overlook and I wanted them to be forced to say what separator is in use in a way that made them think about it.

lib/sass/script/value/helpers.rb
chriseppstein: Changed. I also renamed the method to `quoted_string`.

lib/sass/script/value/helpers.rb
chriseppstein: Changed. I also renamed the method to `unquoted_string` and added a method alias for `identifier` -- i feel both are more readable in their particular use cases.

lib/sass/script/value/null.rb
chriseppstein, line 11: Then there's really no point in calling the method out in the docs at all. I have marked it `@private` and left the doc string where it was.

lib/sass/script/value/number.rb
chriseppstein: Removed.

test/sass/script_test.rb
chriseppstein, line 687: Some implementation details are more important than others. It's testing that we don't create a bunch of garbage objects. This is an important performance characteristic. The point of not testing implementation details is because you will change them in the course of normal development and don't want to update the tests when you do. I don't want this implementation detail to change.

test/sass/value_helpers_test.rb
chriseppstein, line 12: Same reply.

test/sass/value_helpers_test.rb
chriseppstein, line 32: split.

test/sass/value_helpers_test.rb
chriseppstein, line 57: added both.

test/sass/value_helpers_test.rb
chriseppstein, line 74: This is covered by [an existing test](https://github.com/nex3/sass/blob/stable/test/sass/functions_test.rb#L102-L103).

test/sass/value_helpers_test.rb
chriseppstein, line 90: This is covered by [an existing test](https://github.com/nex3/sass/blob/stable/test/sass/functions_test.rb#L241-L244)

test/sass/value_helpers_test.rb
chriseppstein: Removed.

test/sass/value_helpers_test.rb
chriseppstein: simplified.

test/sass/value_helpers_test.rb
chriseppstein, line 129: Added. found bugs O_o.

test/sass/value_helpers_test.rb
chriseppstein: It was copypasta. Fixed.

test/sass/value_helpers_test.rb
chriseppstein: changed.

chriseppstein:
I will rebase this against master once #850 lands and squash it into a single commit. If you'd like to review my follow on patch, it's here as a separate commit until then.

lib/sass/script/functions.rb
nex3, line 1272: In that case, we should push this logic down into `Helpers#quoted_string`.

lib/sass/script/value/helpers.rb
nex3: > Maybe? I don't think that's as useful when reading the actual code instead of the generated docs.

I'm happy enough if it's accessible when they're looking for it; I don't think it needs to be staring them in the face. I'm okay with keeping this in if you move it to the module level and file an issue to use `@since` more thoroughly.

lib/sass/script/value/helpers.rb
nex3: It removes indentation for many lines, which is better than not duplicating a constructor.

lib/sass/script/value/helpers.rb
nex3: I feel like the correct answer the substantial majority of the time is "comma separated". Space-separated lists exist because we need them to express the full range of CSS values; they usually shouldn't be used for list-processing stuff. I think people who want space-separated lists will know they want them, and people who don't know which they want will want comma-separated lists.

test/sass/script_test.rb
nex3, line 687: Okay.

nex3:
Done re-reviewing. I added several responses to comments above; I don't know if GitHub's awful review system provides an easy way of finding those, so you may have to look through each one individually.

lib/sass/script/value/helpers.rb
chriseppstein: > I feel like the correct answer the substantial majority of the time is "comma separated".

Right. But, I don't. And the fact that we disagree about this tells me that it's the exact reason why the type should be specified. While I agree that if you're returning a list that is not going to be output to the stylesheet, the type doesn't matter, but if it is going to be output, there's no obvious default.

I have changed this function to the signature you suggested so that it is easier to select the type at runtime -- this is a very common pattern needed for writing list functions that accept either kind of list. However, I am still keeping the type argument required.

lib/sass/script/value/helpers.rb
chriseppstein: Solution: extract method refactor

lib/sass/script/value/helpers.rb
chriseppstein: I moved the `@since` declaration. Filed: https://github.com/nex3/sass/issues/942

lib/sass/script/functions.rb
chriseppstein, line 1272: That doesn't make sense. that method accepts a ruby object, not a sass value.

chriseppstein:
@nex3 do you :+1: now?

nex3:
A couple more comments. If you agree with all my suggestions, :+1:.

Issue #470, opened 2012-08-06
Make the if() function a parser builtin.


chriseppstein:
As expected when we first added this, people occasionally expect this function to have conditional evaluation of the 2nd and 3rd arguments based on the 1st argument.

We should make this function a built-in to the parser so that it behaves correctly.

chriseppstein:
This is simple to do and has bothered me several times recently. Going to try to get it implemented soon.

PR #836, opened 2013-07-03
Make if() a builtin function so that only one of the two outcomes are evaluated.

chriseppstein:
It's super annoying, that it doesn't work like a proper if.

This addresses #470.


chriseppstein:
addressed my own code comments. ready for your review @nex3.

KittyGiraudel:
This would definitely be a very cool fix. I've been bothered by this behaviour a couple of days ago; had to wrap the whole thing into a conditional @if statement. Not great.

Can we expect this patch for v3.3 maybe? Or even before?

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
+require 'delegate'
+
+module Sass
+  module Util
+    # A hash that normalizes its string keys while still allowing you to get back
+    # to the original keys that were stored. If several different values normalize
+    # to the same value, whichever is stored last wins.
+    class NormalizedMap < DelegateClass(Hash)
+
```
nex3, line 9: Style nit: extra newline.

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
+require 'delegate'
+
+module Sass
+  module Util
+    # A hash that normalizes its string keys while still allowing you to get back
+    # to the original keys that were stored. If several different values normalize
+    # to the same value, whichever is stored last wins.
+    class NormalizedMap < DelegateClass(Hash)
```
nex3, line 8: This'll need to be integrated with `OrderedHash` in the maps branch somehow. Probably just `DelegateClass(ruby1_8? OrderedHash : Hash)`.

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
...
+      # @param hash {String => Object} A hash to normalize and use
+      #   as the initial values for this map.
+      def initialize(hash = nil)
+        @key_strings = {}
+        @map = {}
+
+        # We delegate all hash methods that are not overridden here to @map.
+        super(@map)
+
+        hash.each {|k, v| self[k] = v } if hash
```
nex3, line 21: The native Hash will only do this if its argument is itself a Hash. Otherwise it'll just ignore the argument. This should do the same. Alternately, since we never actually construct this with arguments, we could just only support the 0-arg constructor.

Style nit: no space before `}`.

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      # Specifies how to transform the key.
+      #
+      # This can be overridden to create other normalization behaviors
+      def normalize(key)
+        key.tr("-","_")
```
nex3, line 28: It's a small thing, but it feels like the underlying standard should be dashes, since that's the suggested style.

Style nit: space after `,`.

lib/sass/script/tree/funcall.rb
```diff
@@ -29,8 +29,8 @@ class Funcall < Node
 
     # @param name [String] See \{#name}
     # @param args [Array<Node>] See \{#args}
+    # @param keywords [Sass::Util::NormalizedMap<{String => Node}>] See \{#keywords}
```
nex3, line 32: `Sass::Util::NormalizedMap<Node>`, since the key is always a string for NormalizedMap.

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
...
+      # @private
+      def delete(k)
+        normalized = normalize(k)
+        @key_strings.delete(normalized)
+        super(normalized)
+      end
+
+      # Returns the hash with the keys as they were stored (before normalization)
+      def as_stored
+        Sass::Util.map_keys(@map) {|k| @key_strings[k] }
```
nex3, line 57: Style nit: no space before `}`.

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
...
+
+        # We delegate all hash methods that are not overridden here to @map.
+        super(@map)
+
+        hash.each {|k, v| self[k] = v } if hash
+      end
+
+      # Specifies how to transform the key.
+      #
+      # This can be overridden to create other normalization behaviors
```
nex3, line 26: Nit: period

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      # @private
+      def delete(k)
+        normalized = normalize(k)
+        @key_strings.delete(normalized)
+        super(normalized)
+      end
+
+      # Returns the hash with the keys as they were stored (before normalization)
```
nex3, line 55: Nit: period

Also, return type documentation.

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
...
+      # this is magically invoked by ruby, not sure why DelegateClass doesn't take care of it.
+      # @private
+      def initialize_dup(other)
+        super
+        @map = other.instance_variable_get("@map").dup
+        __setobj__(@map)
+      end
+
+      # @private
+      def deep_copy
```
nex3, line 69: I'm not a fan of defining `#deep_copy` on this just because it happens to be a class we control. I feel like `#deep_copy` should only be defined on objects that are part of Sass proper, whereas this is more of a generic utility object. I also don't like that we need to do `respond_to?` to make it work in general.

lib/sass/util/normalized_map.rb
```diff
@@ -0,0 +1,80 @@
...
+        copy = self.class.new
+        copy.instance_variable_set("@key_strings", @key_strings.dup)
+        new_map = Sass::Util.map_vals(@map) {|v| v.respond_to?(:deep_copy) ? v.deep_copy : (v.dup rescue v)}
+        copy.instance_variable_set("@map", new_map)
+        copy.__setobj__(new_map)
+        copy
+      end
+
```
nex3, line 77: Style nit: extra newline.

test/sass/conversion_test.rb
```diff
@@ -1692,6 +1692,35 @@ def test_comment_indentation
 SCSS
   end
 
+  def test_keyword_arguments
```
nex3, line 1695: Also test that with dasherize, underscores become dashes.

test/sass/conversion_test.rb
```diff
@@ -1692,6 +1692,35 @@ def test_comment_indentation
...
+    assert_renders(<<SASS, <<SCSS, :dasherize => true)
+$foo: foo($dash-ed: 2px)
+SASS
+$foo: foo($dash-ed: 2px);
+SCSS
+    assert_renders(<<SASS, <<SCSS, :dasherize => false)
```
nex3, line 1701: There's no difference between setting dasherize to false and just not passing it. I don't think this needs both this and the following assertion.

test/sass/util/normalized_map_test.rb
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env ruby
+require File.dirname(__FILE__) + '/../../test_helper'
+require 'sass/util/normalized_map'
+
+class NormalizedMapTest < Test::Unit::TestCase
+
```
nex3, line 6: Style nit: extra newline.

doc-src/SASS_REFERENCE.md
```diff
@@ -1747,17 +1747,29 @@ Usage Example:
...
+**Note:** Control directives are an advanced feature, and are uncommon
+ in day-to-day styling.  They exist mainly for use in [mixins](#mixins),
+particularly those that are part of libraries like
+[Compass](http://compass-style.org), and so require substantial
+flexibility.
+
+### `if()`
+
+The builtin `if()` function allows you to branch on a condition and
+evaluates only one of two possible outcomes. It can be used as a value
```
nex3, line 1765: "evaluates" -> "returns", then mention later that unlike a function it only evaluates the argument that gets returned. Also explain why that's useful (e.g. you can safely refer to a variable that may not exist).

I don't like "as a value" here; I'd just say "It can be used in any script context."

lib/sass/script/functions.rb
```diff
@@ -194,10 +194,6 @@ module Sass::Script
   #
   # ## Miscellaneous Functions
   #
-  # \{#if if($condition, $if-true, $if-false)}
-  # : Returns one of two values, depending on whether or not `$condition` is
-  #   true.
```
nex3, line 199: I'd keep this here, since this is a place people will look if they see `if()` in some Sass and want to look it up. Then link to the reference instead of the method doc.

lib/sass/script/parser.rb
```diff
@@ -343,10 +343,15 @@ def ident
 
       def funcall
         return raw unless tok = try_tok(:funcall)
-        args, keywords, splat = fn_arglist || [[], {}]
+        args, keywords, splat = fn_arglist
         assert_tok(:rparen)
-        node(Script::Tree::Funcall.new(tok.value, args, keywords, splat),
-          tok.source_range.start_pos, source_position)
+        fn_node = if tok.value == "if"
+                    raise SyntaxError.new("Cannot use ... with if()") if splat
```
nex3, line 349: I don't think this is a necessary restriction. We can't have special evaluation behavior with a splat, but we don't have to bail on it.

The easiest way to implement this would probably be to keep the old if function, mark it `@private`, and use that if a splat exists.

lib/sass/script/tree/if_function.rb
```diff
@@ -0,0 +1,64 @@
+require 'sass/script/functions'
+
+module Sass::Script::Tree
+  # A SassScript parse node representing an inline if function.
+  class IfFunction < Funcall
+
```
nex3, line 6: Style nit: extra newline.

lib/sass/script/tree/if_function.rb
```diff
@@ -0,0 +1,64 @@
...
+    # Create the IfFunction node
```
nex3, line 7: Nit: period

lib/sass/script/tree/if_function.rb
```diff
@@ -0,0 +1,64 @@
...
+    #
+    # @param args [Array<Node>] The arguments passed to the function.
+    # @param keywords [{String => Node}] The keyword arguments passed to the function.
+    # @raise ArgumentError when the three required arguments (condition, if_true, if_false) are not present.
+    def initialize(args, keywords)
+      super("if", args, keywords.dup, nil)
+      if args.size > 3
+        raise Sass::SyntaxError.new("#{args.size} arguments provided to if() but only 3 are accepted.")
```
nex3, line 15: This error should look like Ruby argument errors: `wrong number of arguments (#{args.size} for 3) for `if``

lib/sass/script/tree/if_function.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      raise Sass::SyntaxError.new("A condition is required for if()") unless condition
+      raise Sass::SyntaxError.new("A true value is required for if()") unless true_expression
+      raise Sass::SyntaxError.new("A false value is required for if()") unless false_expression
```
nex3, line 20: These should also look like normal argument errors. For example, `if()` should produce `wrong number of arguments (0 for 3) for `if``, while `if($foo: 12)` should produce `Function if requires an argument named $condition`.

lib/sass/script/tree/if_function.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      if (unknown_keys = keywords.keys - %w(condition if_true if_false)).any?
+        raise Sass::SyntaxError.new("$#{unknown_keys.first} is not allowed as an argument to if()")
+      end
+    end
```
nex3, line 25: There should be error-checking for an argument passed both positionally and by name (e.g. `if(true, 1, 2, $condition: false)`).

lib/sass/script/tree/if_function.rb
```diff
@@ -0,0 +1,64 @@
...
+      args[0] || keywords["condition"]
+    end
+
+    # The expression that is performed if the condition is true
+    # @return [Node]
+    def true_expression
+      args[1] || keywords["if-true"]
+    end
+
+
```
nex3, line 39: Style nit: extra newline

lib/sass/script/tree/if_function.rb
```diff
@@ -0,0 +1,64 @@
+require 'sass/script/functions'
+
+module Sass::Script::Tree
+  # A SassScript parse node representing an inline if function.
+  class IfFunction < Funcall
```
nex3, line 5: I don't think this gets enough value from Funcall to warrant extending it. It seems like it should be its own class, with straightforward ivars for condition, true_expression, and false_expression.

lib/sass/script/tree/if_function.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    # The expression that is performed if the condition is false
+    # @return [Node]
+    def false_expression
+      args[2] || keywords["if-false"]
+    end
+
+    # @see Node#deep_copy
+    def deep_copy
+      IfFunction.new(@args.map{|a| a.deep_copy}, @keywords.deep_copy)
```
nex3, line 48: Style nit: space before `{`.

nex3:
Done reviewing.

chriseppstein:
After reading many of your comments, I feel like a better implementation may be to add a generic ability to delay performing certain arguments and pass them to the ruby function as a node that can be performed inside the function's implementation instead of being performed before calling.

lib/sass/util/normalized_map.rb
chriseppstein, line 28: While dashes are the preferred style in Sass files, our preferred notation for implementation details has been underscores because it integrates more easily with ruby names. I am simply maintaining that convention.

Space added.

lib/sass/util/normalized_map.rb
chriseppstein, line 8: I am going to land this commit on master and then merge it into the maps branch where I will address this.

lib/sass/util/normalized_map.rb
chriseppstein, line 9: Removed.

lib/sass/util/normalized_map.rb
chriseppstein, line 57: removed.

lib/sass/util/normalized_map.rb
chriseppstein, line 26: added.

lib/sass/util/normalized_map.rb
chriseppstein, line 77: removed

lib/sass/util/normalized_map.rb
chriseppstein, line 69: Agree. I've removed it and the deep copy is handled in the Funcall now:

``` ruby
copied_keywords = Sass::Util::NormalizedMap.new
@keywords.as_stored.each {|k,v| copied_keywords[k] = v.deep_copy}
```

test/sass/conversion_test.rb
chriseppstein, line 1695: added.

test/sass/conversion_test.rb
chriseppstein, line 1701: I like this because it was central to the test and so it made the intent more clear for a reader. But whatever, removed.

test/sass/util/normalized_map_test.rb
chriseppstein, line 6: removed.

chriseppstein:
Closing this in favor of #934.

PR #849, opened 2013-07-13
Code style enforcement

chriseppstein:
This is a work-in-progress. Don't review it yet.

This branch adds a style checker to our code base and then does a bunch of code cleanup based on the warnings it gave us.

I'd like to make running this part of our TravisCI tests so that pulls will get notified if they break any of the rules.


chriseppstein:
wrong base branch

PR #850, opened 2013-07-13
Coding style enforcement

chriseppstein:
This pull request adds a style checker to our project (Rubocop) and configures it according to our coding style.

I've cleaned up all the non-conforming code. There was a lot of it.

`rake test` will now check rubocop unless in ruby 1.8, but it is not in our default `Gemfile` because it only installs on ruby 1.9.3 or greater. It would be nice if there was a way to accomplish a travisci build that didn't involve an extra gemfile, but I couldn't find one.

Since the result is not ruby version dependent, Rubocop will run in just one of our TravisCI build matrix rows. Here's an [example of it passing](https://travis-ci.org/nex3/sass/jobs/11714793).

Each code style change that I made/enforced is in a separate commit for your code reviewing pleasure.


chriseppstein:
This PR is ready for review.

.rubocop.yml
```diff
@@ -0,0 +1,105 @@
...
+  Enabled: false
+LineLength:
+  Enabled: true
+  Max: 100
+Loop:
+  # This isn't a big deal.
+  Enabled: false
+MethodLength:
+  # This max should actually be 25.
+  # But that will require significant refactoring.
```
nex3: Include "TODO:" in here somewhere so this comes up if we ever run out of things to do and start searching for TODOs (ha ha ha).

.rubocop.yml
```diff
@@ -0,0 +1,105 @@
...
+  Max: 50
+  CountComments: false
+ModuleFunction:
+  # module function makes things private so it means you can't use the module as a module.
```
nex3: "module function" -> "module_function"

.rubocop.yml
```diff
@@ -0,0 +1,105 @@
...
+  Enabled: false
+NumericLiterals:
+  # Fails for hex literals. A patch for this should be released soon.
```
nex3: Is there an issue number for this? If so, mention it here.

.rubocop.yml
```diff
@@ -0,0 +1,105 @@
+AlignParameters:
+  # Nathan doesn't like to align parameters to the method call
+  Enabled: false
+CharacterLiteral:
+  # Character literals are pretty useful when working with text like we do.
```
nex3, line 8: Nit: some of these comments have periods and some don't. I don't care which we do, but it should be consistent.

Rakefile
```diff
@@ -19,6 +19,22 @@ Rake::TestTask.new do |t|
   t.verbose = true
 end
 
+# ----- Code Style Enforcement -----
+
+$: << scope("lib")
```
nex3: Nit: I prefer using the more explicit `$LOAD_PATH` to `$:`.

Rakefile
```diff
@@ -19,6 +19,22 @@ Rake::TestTask.new do |t|
...
+if !Sass::Util.ruby1_8? && (ENV.has_key?("RUBOCOP") && ENV["RUBOCOP"] == "true" || !ENV.has_key?("RUBOCOP"))
+  require 'rubocop/rake_task'
+  require "#{File.dirname(__FILE__)}/test/rubocop_extensions.rb"
+  Rubocop::RakeTask.new do |t|
+    t.patterns = FileList["lib/**/*"]
+  end
+
+  task :test => :rubocop
+else
+  puts "Skipping rubocop style check."
```
nex3: I don't think it makes sense to print this when doing _anything_ with Rake. Can we just print it when running the test task instead?

gemfile_with_rubocop
```diff
@@ -0,0 +1,6 @@
+source "http://rubygems.org"
```
nex3: This file should definitely be capitalized. I'd probably call it "Gemfile.rubocop", but "Gemfile_with_rubocop" is also fine.

Also, if possible, we should require/load the normal Gemfile to reduce duplication.

lib/sass/cache_stores/chain.rb
```diff
@@ -22,7 +22,8 @@ def store(key, sha, obj)
       # @see Base#retrieve
       def retrieve(key, sha)
         @caches.each_with_index do |c, i|
-          next unless obj = c.retrieve(key, sha)
+          obj = c.retrieve(key, sha)
+          next unless obj
```
nex3, line 26: I kind of like this pattern. Can we make it work if we wrap the assignment in parens?

lib/sass/engine.rb
```diff
@@ -196,9 +196,11 @@ def self.normalize_options(options)
       # Remove any deprecated importers if the location is imported explicitly
       options[:load_paths].reject! do |importer|
         importer.is_a?(Sass::Importers::DeprecatedPath) &&
-          options[:load_paths].find {|other_importer| other_importer.is_a?(Sass::Importers::Filesystem) &&
-                                                      other_importer != importer &&
-                                                      other_importer.root == importer.root}
+          options[:load_paths].find do |other_importer|
+            other_importer.is_a?(Sass::Importers::Filesystem) &&
+            other_importer != importer &&
+            other_importer.root == importer.root
```
nex3: These last two lines should be indented, since they're a continued expression.

lib/sass/engine.rb
```diff
@@ -331,7 +335,8 @@ def dependencies
     #
     # @private
     def _dependencies(seen, engines)
-      return if seen.include?(key = [@options[:filename], @options[:importer]])
+      key = @options[:filename]
+      return if seen.include?([key, @options[:importer]])
```
nex3: Okay this one is legit cleaner.

lib/sass/engine.rb
```diff
@@ -366,7 +371,7 @@ def _render_with_sourcemap(sourcemap_uri)
       rendered << Sass::Util.escape_uri(sourcemap_uri)
       rendered << " */"
       rendered = encode_and_set_charset(rendered)
-      return rendered, sourcemap
+      [rendered, sourcemap]
```
nex3: I dislike this change. I feel that `return foo, var` makes it clear that the returned array is intended to mean "multiple return values" rather than "a list of values".

gemfile_with_rubocop
nex3: Actually, as discussed online, let's just put `gem 'rubocop' unless ruby1_8?` in the normal Gemfile.

lib/sass/engine.rb
```diff
@@ -651,6 +660,7 @@ def parse_line(parent, line, root)
     end
 
     def parse_property_or_rule(line)
+      # rubocop:disable UselessAssignment
```
nex3: We can safely remove the last `offset += trailing.length` line here.

lib/sass/environment.rb
```diff
@@ -74,7 +74,9 @@ def stack
     private
 
     class << self
+
```
nex3, line 77: If it's easy to remove this space, I'd prefer that, but if it's not then whatever.

lib/sass/exec.rb
```diff
@@ -17,6 +17,7 @@ def initialize(args)
       #
       # @see #parse
       def parse!
+        # rubocop:disable RescueException
```
nex3, line 20: It feels like this is a check we can disable. Similar to HandleException, abuses of `rescue Exception` are easy to catch in code review.

lib/sass/script/tree/funcall.rb
```diff
@@ -142,7 +148,9 @@ def _perform(environment)
         # `_perform`.
         if e.message =~ /^method '([^']+)': given (\d+), expected (\d+)/
           error_name, given, expected = $1, $2, $3
+          # rubocop:disable BlockNesting
```
nex3: I think we should disable this check. I'm no fan of deep nesting, but I don't think it's completely avoidable either, and there are cases where I don't think it even looks that bad.

lib/sass/selector/simple_sequence.rb
```diff
@@ -107,26 +107,33 @@ def resolve_parent_refs(super_seq)
       #   by extending this selector with `extends`.
       # @see CommaSequence#do_extend
       def do_extend(extends, parent_directives, seen = Set.new)
-        Sass::Util.group_by_to_a(extends.get(members.to_set)) {|ex, _| ex.extender}.map do |seq, group|
+        groups = Sass::Util.group_by_to_a(extends.get(members.to_set)) {|ex, _| ex.extender}
+        groups.map! do |seq, group|
```
nex3, line 111: I hate this no-multiline-block-chaining rule. I think setting up a pipeline like that is way more readable than doing a bunch of imperative modifications.

lib/sass/tree/rule_node.rb
```diff
@@ -131,7 +132,9 @@ def try_to_parse_non_interpolated_rules
         # We don't use real filename/line info because we don't have it yet.
         # When we get it, we'll set it on the parsed rules if possible.
         parser = Sass::SCSS::StaticParser.new(@rule.join.strip, '', nil, 1)
+        # rubocop:disable RescueModifier
```
nex3, line 135: This is another sometimes-valid thing that's easy to catch in code review. `rescue nil` sticks out like a sore thumb; I don't think we're at risk of adding it accidentally.

lib/sass/scss/parser.rb
```diff
@@ -1072,10 +1116,13 @@ def node(node, start_pos, end_pos = source_position)
       @sass_script_parser = Class.new(Sass::Script::Parser)
       @sass_script_parser.send(:include, ScriptParser)
       # @private
```
nex3: This should be moved into the `class << self`.

test/rubocop_extensions.rb
```diff
@@ -0,0 +1,47 @@
+require "rubocop/cop/style/surrounding_space"
+
+module Rubocop
+  module Cop
+    module Style
+      # Common functionality for checking surrounding space.
+      class SpaceAroundBlockBraces < Cop
```
nex3, line 8: Mention in the doc comment that this is monkey-patching an existing class.

nex3:
Done reviewing. Overall I love the style improvements here.

.rubocop.yml
chriseppstein: Added.

.rubocop.yml
chriseppstein: Changed.

.rubocop.yml
chriseppstein: Mentioned.

.rubocop.yml
chriseppstein, line 8: normalized. Also: capitalization of first letter. Because we can't be too anal when it comes to the comments in our configuration files.

Rakefile
chriseppstein: It's now so explicit I had to mark the commit as nsfw.

Rakefile
chriseppstein: done.

gemfile_with_rubocop
chriseppstein: this was pushed before you commented :) https://github.com/nex3/sass/commit/b91b55b3562f93f17e4aa569f158545cbe722399

lib/sass/cache_stores/chain.rb
chriseppstein, line 26: I am against this pattern. I think that the variables that exist in a method should be clearly scannable at the beginning of the line. Definitions in a `if/while/unless/etc` statement are ok because there's just one token to parse before the eye sees the definition. But defining them at the end of the statement is really hard for the casual reader.

lib/sass/engine.rb
chriseppstein: indented.

lib/sass/engine.rb
chriseppstein: ಠ_ಠ  Except it changed the logic. Which I have fixed now.

lib/sass/engine.rb
chriseppstein: sweet.

lib/sass/engine.rb
chriseppstein: I do not want to disable this check. It caught a bunch of redundant use of return statements. And I do not want to add a bunch of disable comments.

So I monkeypatched rubocop to allow us to configure returns with multiple return values as ok.

lib/sass/environment.rb
chriseppstein, line 77: It's not :(

lib/sass/exec.rb
chriseppstein, line 20: This caught several violations so I don't agree with your assertion. This is a fairly common mistake and I don't there couple places where we have to disable the check are warranted, IMO. The goal for me is to not have to iterate a code review for these simple sorts of things that can be caught before we ever see the code.

lib/sass/script/tree/funcall.rb
chriseppstein: disabled.

lib/sass/selector/simple_sequence.rb
chriseppstein, line 111: Chaining encourages the creation of useless objects that need to be GC'd and this code here is especially bad because it is invoked many times.

Performance and memory management trump readability in cases like this.

lib/sass/tree/rule_node.rb
chriseppstein, line 135: Same reply. we only do this once. So it's not a common enough pattern to justify letting the contribution get to the code-review step. If you don't like this comment, I'd rather just rewrite this code to not use `rescue nil`.

lib/sass/scss/parser.rb
chriseppstein: Moved.

test/rubocop_extensions.rb
chriseppstein, line 8: Mentioned.

PR #854, opened 2013-07-17
Add tests and documentation for JSON error and warning output.

eriwen:
This PR addresses a couple of @nex3's [requests](https://github.com/nex3/sass/issues/593#issuecomment-11631340) on nex3/sass#593. 

I will submit a different pull request to add metadata to SyntaxErrors, but I'm less sure of how that work will end up. 

I'm just adding tests and docs here. There is a persistent test failure `test_extend_sourcemap_scss` unrelated to this change. 

Hope you approve :)

### DIFF
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -319,6 +319,48 @@ Available options are:
   for displaying the Sass filename and line number.
   Automatically disabled when using the `:compressed` output style.
 
+{#trace-option} `:trace`
+: Show a full traceback on error.
+
+{#json_err-option} `:json_err`
+: When set to true, errors are written to stderr in JSON format.
+  This implies --trace.
+
+#### JSON format
+: All warnings and errors have a `type`, `message`, and `backtrace`. Some may
+  have some additional fields pertinent to the particular warning.
+
+    {
+        "type": ["warning", "WarningType"],
+        "message": "Warning message",
+        "backtrace": ["bar.sass:3:in `foo'", "qux.sass:72", ...],
+        "additional_field1": "value"
+    }
+
+: In addition to `type`, `message`, and `backtrace`, all `Sass::SyntaxError`s
+ have a `sass_backtrace`, `filename`, `line`, and a `mixin` (if the error
+ occurred within a mixin).
+
+    {
+        "type": ["error", "Sass::SyntaxError"],
+        "message": "Undefined mixin 'foo'.",
+        "backtrace": ["bar.sass:3:in `foo'", "qux.sass:72", ...],
+        "filename": "bar.sass",
+        "line": 3,
+        "mixin": "foo",
+        "sass_backtrace": [
+            {
+                "filename": "bar.sass"
+                "mixin": "foo",
+                "line": 3
+            },
+            {
+                "filename": "qux.sass"
+                "line": 72
+            }
+        ]
+    }
+
 {#custom-option} `:custom`
 : An option that's available for individual applications to set
   to make data available to {Sass::Script::Functions custom Sass functions}.
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -319,16 +319,9 @@ Available options are:
   for displaying the Sass filename and line number.
   Automatically disabled when using the `:compressed` output style.
 
-{#trace-option} `:trace`
-: Show a full traceback on error.
-
-{#json_err-option} `:json_err`
-: When set to true, errors are written to stderr in JSON format.
-  This implies --trace.
-
 #### JSON format
-: All warnings and errors have a `type`, `message`, and `backtrace`. Some may
-  have some additional fields pertinent to the particular warning.
+All warnings and errors have a `type`, `message`, and `backtrace`. Some may
+have some additional fields pertinent to the particular warning.
 
     {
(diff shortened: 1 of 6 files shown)


test/sass/cache_test.rb
```diff
@@ -66,6 +67,40 @@ def test_arbitrary_objects_can_go_into_cache
...
+    end
+
+    def path_to(key)
+      "PATH"
+    end
+  end
+
+  def test_storage_error_gives_metadata_in_json_warning
+    cache = FaultyStore.new
+    with_json_warnings do
```
nex3: You're calling `#with_json_warnings`, `#collect_stderr`, and `JSON.parse` in all of these tests. Maybe instead of `#with_json_warnings` there should just be a `#collect_json_warnings` that collects and parses all the JSON that's emitted in a block.

test/sass/cache_test.rb
```diff
@@ -66,6 +67,40 @@ def test_arbitrary_objects_can_go_into_cache
...
+      json = JSON.parse(collect_stderr {cache.store("_", "_", "_")})
+      assert_equal json["type"], ["warning", "cache"]
```
nex3: These assertions are in the wrong order; the expected value should be on the left, and the actual value should be on the right.

test/sass/cache_test.rb
```diff
@@ -66,6 +67,40 @@ def test_arbitrary_objects_can_go_into_cache
     assert_equal an_object, cache.retrieve("an_object", "")
   end
 
+  class FaultyStore < Sass::CacheStores::Base
```
nex3: I like to put dummy classes at the top of the file regardless of where they're used.

test/sass/cache_test.rb
```diff
@@ -66,6 +67,40 @@ def test_arbitrary_objects_can_go_into_cache
...
+
+  def test_storage_error_gives_metadata_in_json_warning
+    cache = FaultyStore.new
+    with_json_warnings do
+      json = JSON.parse(collect_stderr {cache.store("_", "_", "_")})
+      assert_equal json["type"], ["warning", "cache"]
+      assert_equal json["message"], "Warning. Error encountered while saving cache PATH: LOAD_ERROR"
+      assert_not_nil json["backtrace"]
+    end
+  end
```
nex3, line 90: This and the test below should assert that the "path" parameter is also as expected.

test/sass/importer_test.rb
```diff
@@ -189,4 +190,26 @@ def test_absolute_files_across_template_locations
...
+      assert_equal json["type"], ["warning", "ambiguous_import"]
+      assert_equal json["message"], <<MESSAGE
+WARNING: In #{expected_dir}:
+  There are multiple files that match the name "#{template_name}":
+    #{template_name}.sass
+    #{template_name}.scss
+MESSAGE
+      assert_equal expected_dir, json["dir"]
+      assert_equal template_name, json["name"]
+      assert_equal [[absolutize("templates/#{template_name}.sass"), "sass"], [absolutize("templates/#{template_name}.scss"), "scss"]], json["candidates"]
```
nex3: Since you wrote this, we've started enforcing a 100 character line length limit, so this will have to be wrapped.

doc-src/SASS_REFERENCE.md
```diff
@@ -319,6 +319,48 @@ Available options are:
   for displaying the Sass filename and line number.
   Automatically disabled when using the `:compressed` output style.
 
+{#trace-option} `:trace`
```
nex3: `--trace` is only used by the executable, so it shouldn't be documented here.

doc-src/SASS_REFERENCE.md
```diff
@@ -319,6 +319,48 @@ Available options are:
...
+: Show a full traceback on error.
+
+{#json_err-option} `:json_err`
```
nex3: This section is used for options passed via the Sass options hash. The way I set it up, JSON errors would be controlled globally via `Sass.json_err`. It's possible that controlling them via the options hash is better, but if so, you should add support for it in the code before documenting it here.

doc-src/SASS_REFERENCE.md
```diff
@@ -319,6 +319,48 @@ Available options are:
...
+: When set to true, errors are written to stderr in JSON format.
+  This implies --trace.
+
+#### JSON format
+: All warnings and errors have a `type`, `message`, and `backtrace`. Some may
```
nex3: This colon prefix is for definition lists. If you're just writing a normal paragraph, you don't need it.

doc-src/SASS_REFERENCE.md
```diff
@@ -319,6 +319,48 @@ Available options are:
...
+  have some additional fields pertinent to the particular warning.
+
+    {
+        "type": ["warning", "WarningType"],
+        "message": "Warning message",
+        "backtrace": ["bar.sass:3:in `foo'", "qux.sass:72", ...],
+        "additional_field1": "value"
+    }
+
+: In addition to `type`, `message`, and `backtrace`, all `Sass::SyntaxError`s
```
nex3: "`Sass::SyntaxError`s" -> "errors caused by the stylesheets"

nex3:
Done reviewing. I apologize for the long wait.

eriwen:
@nex3 I hope I've addressed all of your feedback with these recent commits. Please let me know if there are any other improvements to be made.

Issue #642, opened 2013-01-25
Add a map data type to Sass


ByScripts:
_Edited by @nex3 to change the title._

I don't find tickets about collections.

I read some closed tickets abouts namespaces and scopes, but they don't seems to describe what I'm thinking about.

So, instead of a description, here is an example :

``` scss
// variables.scss
$categories: ($name: 'first', $c: #000, $bg: #111, $mt: 10px, $h: 128px, $w: 100px),
             ($name: 'second', $c: #222, $bg: #333, $mt: 14px, $h: 124px, $w: 90px),
             ($name: 'third', $c: #444, $bg: #555, $mt: 12px, $h: 126px, $w: 104px);

// menu.scss
@mixin build-menu($name, $c, $w) {
    body.#{$name} {
        nav {
            width: $w;
            a {
                color: $c;
                &.#{$name} { color: darken($c, 15%); }
            }
        }
    }
}

// info.scss
@mixin build-info-block($name, $c, $bg, $mt) {
    body.#{$name} {
        div.info {
            margin-top: $mt;
            background-color: $bg;
            h1 { color: $c; }
        }
    }
}

// global.scss
@import "variables";
@import "menu";
@import "info";

@each $category in $categories {
    @include build-menu($category...);
    @include build-info-block($category...);
}
```

This would output :

``` css
body.first nav { width: 100px; }
body.first nav a { color: #000; }
body.first nav a.first { color: #000; }
body.first div.info { margin-top: 10px; background-color: #111; }
body.first div.info h1 { color: #000; }

body.second nav { width: 90px; }
body.second nav a { color: #222; }
body.second nav a.second { color: #1a1a1a; }
body.second div.info { margin-top: 14px; background-color: #333; }
body.second div.info h1 { color: #222; }

body.third nav { width: 104px; }
body.third nav a { color: #444; }
body.third nav a.third { color: #3a3a3a; }
body.third div.info { margin-top: 12px; background-color: #555; }
body.third div.info h1 { color: #444; }
```

More than collections, for this to work, we should be able to pass more arguments than mixins wants, as long as they are named.

In short, this should work without throwing error (sass should just ignore named arguments which are not in its signature):

``` scss
@mixin mymixin($foo, $bar) {}
@include mymixin($whatever: 10px, $foo: #fff, $other: bold, $bar: 2em);
```

Sorry for being so long.

nex3:
[...]

Issue #832, opened 2013-07-02
`index` should return `null` for a non-existent element rather than `false`


nex3:
The current API for `index` was written before SassScript had a `null` type, but now that we do we should move it to return `null`.

We could go through a deprecation process for this, which would involve detecting whenever the return value was used somewhere where the distinction between `null` and `false` matters (e.g. `index(...) == false`). It's unlikely that anyone's doing that, though, so we may be able to get away with no deprecation process.

chriseppstein:
So we'd need to make a DeprecatedFalse class or something to that effect?

KittyGiraudel:
Why not returning `false`? It makes sense.

chriseppstein:
I think we should also return `null` instead of `0` for `str-index()`.

nex3:
> @chriseppstein wrote:
> So we'd need to make a DeprecatedFalse class or something to that effect?

If we wanted to do the deprecation thing, yes.

> I think we should also return `null` instead of `0` for `str-index()`.

Agreed.

> @HugoGiraudel wrote:
> Why not returning `false`? It makes sense.

Because `null` is the standard value for indicating a lack of a result.

pdaoust:
@chriseppstein I agree with returning `null` too -- I was just toying around with `str-index()` for the first time today, expecting it to work like `index()`, and discovered this myself. (By the way, thank you for having made `index()` return a falsy value rather than `-1` like JavaScript does.)

nex3:
`0` isn't falsy.

pdaoust:
@nex3 I was referring to the return value of `index()` rather than `str-index()`.

PR #864, opened 2013-07-26
Maps

nex3:
Implements #642


doc-src/SASS_CHANGELOG.md
```diff
@@ -29,6 +29,85 @@ Thanks to Alexander Pavlov for implementing this.
...
+    }
+
+    h2 {
+      font-size: 1.5em;
+    }
+
+    h3 {
+      font-size: 1.2em;
+    }
+
```
chriseppstein, line 110: This documentation should explain that maps are ordered by original insertion order.

chriseppstein, line 110: This documentation should explain that `nth` returns a pair from a map. And that length($map) returns the number of pairs.

Correction: I see a mention of nth now. But it was pretty subtle, can we call it with some bullet points, and maybe show how maps are iterated in `@each` if only one variable is provided.

doc-src/SASS_CHANGELOG.md
```diff
@@ -29,6 +29,85 @@ Thanks to Alexander Pavlov for implementing this.
...
+values (often strings, but potentially any value) to other SassScript values.
+They look like this:
+
+    $map: (key1: value1, key2: value2, key3: value3);
+
+Unlike lists, maps must always be surrounded by parentheses. `()` is now an
+empty map in addition to an empty list.
+
+Maps will allow users to collect values into named groups and access those
+groups dynamically, without needing variable interpolation. For example, you
```
chriseppstein: I think mentioning variable interpolation is unnecessary. It's an undefined term and I think it makes this more confusing. But if you feel strongly about it I think you should define it here. E.g. "Some authors expect to be able to use interpolation with variables (E.g. $theme-#{$theme-name}), but this is not a feature in Sass. Instead..."

doc-src/SASS_CHANGELOG.md
```diff
@@ -29,6 +29,85 @@ Thanks to Alexander Pavlov for implementing this.
...
+      // ...
+    );
+
+    @mixin themed-header($theme-name) {
+      h1 {
+        color: map-get(map-get($themes, $theme-name), header);
+      }
+    }
+
+There are a variety of functions for manpulating maps:
```
chriseppstein: manipulating => working with

doc-src/SASS_CHANGELOG.md
```diff
@@ -72,6 +151,13 @@ Thanks to Alexander Pavlov for implementing this.
 * You can call a function by name by passing the function name to the
   call function. For example, `call(nth, a b c, 2)` returns `b`.
 
+* `@each` now supports "destructuring", which makes it easier to iterate over
+  lists of lists. If you write `@each $var1, $var2 in a b, c d, d e`, the
+  elements of the sub-lists will be assigned individually to the variables.
+  `$var1` and `$var2` will be `a` and `b`, then `c` and `d`, and then `e` and
+  `f`. For more information, see {file:SASS_REFERENCE.md#each-destructuring the
+  `@each` reference}.
+
```
chriseppstein, line 192: I think this should show sub lists of three elements, all the other examples for maps show two, So this will explain the generality of it better. 

Update: I see you did this below, I think it's good to have the same here in the changelog.

doc-src/SASS_CHANGELOG.md
```diff
@@ -29,6 +29,85 @@ Thanks to Alexander Pavlov for implementing this.
...
+  the keys in a map as a comma-separated list.
+
+* The {Sass::Script::Functions#map_values `map-values($map)` function} returns
+  all the values in a map as a comma-separated list.
+
+* The {Sass::Script::Functions#map_has_key `map-has-key($map, $key)` function}
+  returns whether or not a map contains a pair with the given key.
+
+All the existing list functions also work on maps, treating them as lists of
+pairs. For example, `nth((foo: 1, bar: 2), 1)` returns `foo, 1`. Maps can also
```
chriseppstein: Because of the convention that maps can be a comma delimited list of space delimited pairs, I think `nth` should return a space separated list here.

doc-src/SASS_REFERENCE.md
```diff
@@ -1831,7 +1860,7 @@ is compiled to:
 
 ### `@each` {#each-directive}
 
-The `@each` rule has the form `@each $var in <list>`.
+The `@each` directive usually has the form `@each $var in <list>`.
```
chriseppstein: `<list>` => `<list or map>`

doc-src/SASS_REFERENCE.md
```diff
@@ -722,14 +722,43 @@ However, they aren't the same when they're Sass:
...
+for each key/value pair in a map.
+
+Maps can also be used anywhere lists can. When used by a list function, a map is
+treated as a list of pairs. For example, `(key1: value1, key2: value2)` would be
+treated as the nested list `key1 value1, key2 value2` by list functions. Lists
+cannot be treated as maps, though, with the exception of the empty list. `()`
+represents both a map with no key/value pairs and a list with no elements.
+
+Maps cannot be converted to plain CSS. Using one as the value of a variable or
+an argument to a CSS function will cause an error.
```
chriseppstein, line 762: The `Data Types` section of this documentation should mention maps.

doc-src/SASS_REFERENCE.md
```diff
@@ -1856,6 +1885,52 @@ is compiled to:
     .salamander-icon {
       background-image: url('/images/salamander.png'); }
 
+#### Destructuring {#each-destructuring}
```
chriseppstein: Not a big fan of using the term "Destructuring" in our documentation. It is programmer jargon that may be hard for our users to understand. I think "Multiple assignment" is more clear, but am open to other terms.

doc-src/SASS_REFERENCE.md
```diff
@@ -1856,6 +1885,52 @@ is compiled to:
...
+      background-image: url('/images/puma.png');
+      border: 2px solid black; }
+    .sea-slug-icon {
+      background-image: url('/images/sea-slug.png');
+      border: 2px solid blue; }
+    .egret-icon {
+      background-image: url('/images/egret.png');
+      border: 2px solid white; }
+
+Since [maps](#maps) are treated as lists of pairs, destructing works with them
```
chriseppstein: Typo: "Destructing" => "Destructuring"

test/sass/functions_test.rb
```diff
@@ -1224,6 +1258,41 @@ def test_unique_id
...
+    assert_equal "null", perform("map-get((), foo)").to_sass
+  end
+
+  def test_map_merge
+    assert_equal("(foo: 1, bar: 2, baz: 3)",
+      perform("map-merge((foo: 1, bar: 2), (baz: 3))").to_sass)
+    assert_equal("(foo: 1, bar: 2)",
+      perform("map-merge((), (foo: 1, bar: 2))").to_sass)
+    assert_equal("(foo: 1, bar: 2)",
+      perform("map-merge((foo: 1, bar: 2), ())").to_sass)
```
chriseppstein, line 1282: Map-merge should work for lists of pairs.

```
map-merge((), (a 1, b 2, c 3, b 4)) => (a: 1, b: 4, c: 3)
```

test/sass/functions_test.rb
```diff
@@ -1224,6 +1258,41 @@ def test_unique_id
...
+  end
+
+  def test_map_keys
+    assert_equal("foo, bar",
+      perform("map-keys((foo: 1, bar: 2))").to_sass)
+    assert_equal("()", perform("map-keys(())").to_sass)
```
chriseppstein, line 1306: Map keys should work for lists of pairs:

```
map-keys((a 1, b 2, c 3)) => a, b, c
```

test/sass/functions_test.rb
```diff
@@ -1224,6 +1258,41 @@ def test_unique_id
     end
   end
 
+  def test_map_get
+    assert_equal "1", evaluate("map-get((foo: 1, bar: 2), foo)")
+    assert_equal "2", evaluate("map-get((foo: 1, bar: 2), bar)")
+    assert_equal "null", perform("map-get((foo: 1, bar: 2), baz)").to_sass
+    assert_equal "null", perform("map-get((), foo)").to_sass
```
chriseppstein, line 1265: map-get should work for lists of pairs:

```
map-get((a 1, b 2, c 3), b) => 2
```

test/sass/functions_test.rb
```diff
@@ -1224,6 +1258,41 @@ def test_unique_id
...
+    assert_equal("foo, bar",
+      perform("map-keys((foo: 1, bar: 2))").to_sass)
+    assert_equal("()", perform("map-keys(())").to_sass)
+  end
+
+  def test_map_values
+    assert_equal("1, 2", perform("map-values((foo: 1, bar: 2))").to_sass)
+    assert_equal("1, 2, 2",
+      perform("map-values((foo: 1, bar: 2, baz: 2))").to_sass)
+    assert_equal("()", perform("map-values(())").to_sass)
```
chriseppstein, line 1322: map-values should work for lists of pairs:

```
map-values((a 1, b 2, c 3)) => (1, 2, 3)
```

test/sass/functions_test.rb
```diff
@@ -1224,6 +1258,41 @@ def test_unique_id
...
+  end
+
+  def test_map_has_key
+    assert_equal "true", evaluate("map-has-key((foo: 1, bar: 1), foo)")
+    assert_equal "false", evaluate("map-has-key((foo: 1, bar: 1), baz)")
+    assert_equal "false", evaluate("map-has-key((), foo)")
```
chriseppstein, line 1336: map-has-key should work for list of pairs.

We had agreed for backwards compatibility that all map functions would accept old-style maps-as-lists. If we want, this support can be deprecated in this release or a future one. But I think it's very important for transitioning to the new data type.

chriseppstein:
I don't see any tests for sass-convert.

lib/sass/script/functions.rb
```diff
@@ -1740,14 +1760,94 @@ def index(list, value)
...
+    #   map-merge(("foo": 1, "bar": 2), ("bar": 3)) => ("foo": 1, "bar": 3)
+    # @overload map_merge($map1, $map2)
+    # @param $map1 [Sass::Script::Value::Map]
+    # @param $map2 [Sass::Script::Value::Map]
+    # @return [Sass::Script::Value::Map]
+    # @raise [ArgumentError] if either parameter is not a map
+    def map_merge(map1, map2)
+      assert_type map1, :Map
+      assert_type map2, :Map
+      Sass::Script::Value::Map.new(map1.to_h.merge(map2.to_h))
```
chriseppstein: This will need to be updated when we have merged the value helpers pull request.

lib/sass/script/functions.rb
```diff
@@ -1740,14 +1760,94 @@ def index(list, value)
...
+    def map_get(map, key)
+      assert_type map, :Map
+      map.to_h[key] || Sass::Script::Value::Null.new
+    end
+    declare :map_get, [:map, :key]
+
+    # Merges two maps together into a new map. Keys in `$map2` will take
+    # precedence over keys in `$map1`.
+    #
+    # This is the best way to add new values to a map.
```
chriseppstein, line 1795: This should specify that the ordering of keys is that new keys will be at the end, but existing keys will maintain the index from `$map1`

lib/sass/script/functions.rb
```diff
@@ -1740,14 +1760,94 @@ def index(list, value)
...
+    #   map-get(("foo": 1, "bar": 2), "baz") => null
+    # @overload map_get($map, $key)
+    # @param $map [Sass::Script::Value::Map]
+    # @param $key [Sass::Script::Value::Base]
+    # @return [Sass::Script::Value::Base] The value indexed by `$key`, or `null`
+    #   if the map doesn't contain the given key
+    # @raise [ArgumentError] if `$map` is not a map
+    def map_get(map, key)
+      assert_type map, :Map
+      map.to_h[key] || Sass::Script::Value::Null.new
```
chriseppstein: An idea here would be to set the default for all ruby hashes backing a map to `Sass::Script::Value::Null.new`

Thoughts?

doc-src/SASS_REFERENCE.md
```diff
@@ -722,14 +722,43 @@ However, they aren't the same when they're Sass:
...
+you try to do e.g. `font-family: ()`, Sass will raise an error. If a list
+contains empty lists or null values, as in `1px 2px () 3px` or `1px 2px null
+3px`, the empty lists and null values will be removed before the containing list
+is turned into CSS.
+
+#### Maps
+
+Maps represent an association between keys and values, where keys are used to
+look up values. They make it easy to collect values into named groups and access
+those groups dynamically, without needing variable interpolation. They have no
```
chriseppstein: My comments about variable interpolation apply here as well.

doc-src/SASS_REFERENCE.md
```diff
@@ -722,14 +722,43 @@ However, they aren't the same when they're Sass:
...
+functions](Sass/Script/Functions.html#map-functions). The
+{Sass::Script::Functions#map-get `map-get` function} looks up values in a map
+and the {Sass::Script::Functions#map-merge `map-merge` function} adds values to
+a map. The [`@each` directive](#each-destructuring) can be used to add styles
+for each key/value pair in a map.
+
+Maps can also be used anywhere lists can. When used by a list function, a map is
+treated as a list of pairs. For example, `(key1: value1, key2: value2)` would be
+treated as the nested list `key1 value1, key2 value2` by list functions. Lists
+cannot be treated as maps, though, with the exception of the empty list. `()`
```
chriseppstein, line 758: I thought we agreed that all map functions would accept lists of pairs for backwards compatibility.

lib/sass/script/parser.rb
```diff
@@ -253,11 +253,35 @@ def range(start_pos, end_pos=source_position)
...
+        map = node(Sass::Script::Tree::MapLiteral.new([[key, value]]), start_pos)
+        while tok = try_tok(:comma)
+          key, value = assert_expr(:map_pair)
+          map.pairs << [key, value]
+        end
+        map
+      end
+
+      def map_pair(key=nil)
+        start_pos = source_position
```
chriseppstein: `start_pos` is not used.

chriseppstein:
Will the `keywords` function be implemented in a subsequent pull request?

lib/sass/script/value/list.rb
```diff
@@ -36,6 +36,10 @@ def eq(other)
         self.separator == other.separator)
     end
 
+    def hash
+      [value, separator].hash
```
chriseppstein: I think our immutable objects should memoize the computed hash value. Especially since this creates an object in the process which increases the need for GC.

lib/sass/script/functions.rb
```diff
@@ -380,7 +397,13 @@ def initialize(environment)
       # @param name [String, Symbol, nil] The name of the argument.
       # @raise [ArgumentError] if value is not of the correct type.
       def assert_type(value, type, name = nil)
-        return if value.is_a?(Sass::Script::Value.const_get(type))
+        klass = Sass::Script::Value.const_get(type)
+        if value.is_a?(klass) ||
+            (klass == Sass::Script::Value::Map &&
+             value.is_a?(Sass::Script::Value::List) &&
+             value.value.empty?)
```
chriseppstein: It seems a lot of things I requested would work fine if this considered whether a list was a list of pairs (for efficiency I think only the first few elements should be checked for the convention, and full checking would happen during processing the list as a map)

lib/sass/script/functions.rb
```diff
@@ -175,6 +175,23 @@ module Sass::Script
...
+  #
+  # \{#map_keys map-keys($map)}
+  # : Returns a list of all keys in a map.
+  #
+  # \{#map_values map-values($map)}
+  # : Returns a list of all values in a map.
+  #
+  # \{#map_has_key map-has-key($key)}
+  # : Returns whether a map has a value associated with a given key.
+  #
```
chriseppstein, line 196: The list functions should be updated to mention they accept maps.

lib/sass/script/functions.rb
```diff
@@ -1627,12 +1650,10 @@ def join(list1, list2, separator = Sass::Script::Value::String.new("auto"))
       unless %w[auto space comma].include?(separator.value)
         raise ArgumentError.new("Separator name must be space, comma, or auto")
       end
-      sep1 = list1.separator if list1.is_a?(Sass::Script::Value::List) && !list1.value.empty?
-      sep2 = list2.separator if list2.is_a?(Sass::Script::Value::List) && !list2.value.empty?
       Sass::Script::Value::List.new(
         list1.to_a + list2.to_a,
         if separator.value == 'auto'
-          sep1 || sep2 || :space
+          list1.separator || list2.separator || :space
```
chriseppstein, line 1659: I was confused by this until I saw that separator is moved to the base class. Which, given that single value objects act like lists of one element, makes sense.

lib/sass/tree/each_node.rb
```diff
@@ -5,18 +5,18 @@ module Sass::Tree
...
+    # The names of the loop variables.
+    # @return [Array<String>]
+    attr_reader :vars
 
     # The parse tree for the list.
     # @return [Script::Tree::Node]
     attr_accessor :list
 
-    # @param var [String] The name of the loop variable
+    # @param vars [String] The names of the loop variables
```
chriseppstein: Array&lt;String&gt;

lib/sass/tree/visitors/convert.rb
```diff
@@ -104,7 +104,8 @@ def visit_directive(node)
   end
 
   def visit_each(node)
-    "#{tab_str}@each $#{dasherize(node.var)} in #{node.list.to_sass(@options)}#{yield}"
+    vars = node.vars.map {|var| "$#{dasherize(var)}"}.join(", ")
```
chriseppstein, line 107: I'm pretty sure this transforms the variable to underscores in sass-conversions instead of preserving the authored format.

lib/sass/util.rb
```diff
@@ -30,6 +30,36 @@ def scope(file)
...
 
+    # Converts a hash or a list of pairs into an order-preserving hash.
+    #
+    # On Ruby 1.8.7, this uses the orderedhash gem to simulate an
+    # order-preserving hash. On Ruby 1.9 and up, it just uses the native Hash
+    # class, since that preserves the order itself.
+    #
+    # @overload ordered_hash(hash)
+    #   @param hash [Hash] a normal hash to convert to an ordered hash
+    #   @return [Hash]
```
chriseppstein, line 41: I don't understand the purpose of the hash-based constructor. In 1.8.7, this would result in an indeterminate order.

lib/sass/util.rb
```diff
@@ -30,6 +30,36 @@ def scope(file)
...
+
+      if pairs_or_hash.length == 1 && pairs_or_hash.first.is_a?(Hash)
+        hash = pairs_or_hash.first
+        return hash unless ruby1_8?
+        return OrderedHash.new.merge hash
+      end
+
+      return Hash[pairs_or_hash] unless ruby1_8?
+      return OrderedHash[*flatten(pairs_or_hash, 1)]
+    end
```
chriseppstein, line 61: I think this would read cleaner and be faster if we just had two functions that are defined differently at load time instead of switching so much at runtime.

test/sass/functions_test.rb
```diff
@@ -975,6 +975,10 @@ def test_type_of
     assert_equal("color", evaluate("type-of(#fff)"))
     assert_equal("color", evaluate("type-of($value: #fff)"))
     assert_equal("null", evaluate("type-of(null)"))
+    assert_equal("list", evaluate("type-of(1 2 3)"))
+    assert_equal("list", evaluate("type-of((1, 2, 3))"))
+    assert_equal("list", evaluate("type-of(())"))
+    assert_equal("map", evaluate("type-of((foo: bar))"))
```
chriseppstein, line 981: Given the fluidity of the same value that can act as a map, list or single value depending on context, I'm having second thoughts about this function. I think it should be `is-type-of($obj, $type-name)` with this api we can return `true` for both `is-type-of((), map)` and `is-type-of((), list)`. Basically the contract here is "Can I use this object like a <type>"?

I'm not sure `type-of` must be deprecated, but I think the above function is going to be saner to work with.

test/sass/functions_test.rb
```diff
@@ -1131,6 +1160,9 @@ def test_index
     assert_equal("false", evaluate("index(1px solid blue, 1em)"))
     assert_equal("false", evaluate("index(1px solid blue, notfound)"))
     assert_equal("false", evaluate("index(1px, #00f)"))
+
+    assert_equal("1", evaluate("index((foo: bar, bar: baz), (foo, bar))"))
+    assert_equal("false", evaluate("index((foo: bar, bar: baz), (foo: bar))"))
```
chriseppstein, line 1165: Relevant: #832.

test/sass/functions_test.rb
```diff
@@ -1141,6 +1173,8 @@ def test_list_separator
     assert_equal("space", evaluate("list-separator(#f00)"))
     assert_equal("space", evaluate("list-separator(())"))
     assert_equal("space", evaluate("list-separator(1 2 () 3)"))
+
+    assert_equal("comma", evaluate("list-separator((foo: bar, bar: baz))"))
```
chriseppstein, line 1177: The list separator function is new in 3.2, should it just be `separator` and work for any value now that separator is defined on the base class?

chriseppstein:
I'm done reviewing.

doc-src/SASS_CHANGELOG.md
nex3: The goal was to guide people searching for "Sass variable interpolation" or whatever to the right place, but you're probably right that using that specific term is unlikely. Removed.

doc-src/SASS_CHANGELOG.md
nex3: Done.

doc-src/SASS_CHANGELOG.md
nex3: > the convention that maps can be a comma delimited list of space delimited pairs

I'm not sure what you mean by this.

I made the pairs comma-separated because I think that reads better in general, but if you feel strongly I can make them space-separated instead.

doc-src/SASS_CHANGELOG.md
nex3, line 110: > This documentation should explain that maps are ordered by original insertion order.

"Insertion order" isn't meaningful for immutable maps, but I've added a note about ordering.

> This documentation should explain that `nth` returns a pair from a map. And that length($map) returns the number of pairs.
> 
> Correction: I see a mention of nth now. But it was pretty subtle, can we call it with some bullet points, and maybe show how maps are iterated in `@each` if only one variable is provided.

I don't think people are going to want to use most list operations other than `length()` and `@each` with maps very often. `length()` doesn't require an understanding of pairs, and `@each` will work fine if they just always use two variables. I don't think it's worth it to call out the specific structure of listified maps any more than it already is.

doc-src/SASS_CHANGELOG.md
nex3, line 192: Done.

doc-src/SASS_REFERENCE.md
nex3: Done.

doc-src/SASS_REFERENCE.md
nex3, line 758: I thought we agreed the opposite :p. You're probably right, but now I disagree with my past self.

That feels like the kind of feature that is going to be weird and annoying forever in exchange for being mildly convenient in the near term. I think the compatibility of the list functions will make it easy enough to swap out lists of pairs for real maps that users won't have a strong desire to keep using lists of pairs.

doc-src/SASS_REFERENCE.md
nex3, line 762: Done.

doc-src/SASS_REFERENCE.md
nex3: Done.

doc-src/SASS_REFERENCE.md
nex3: Changed to "multiple assignment."

doc-src/SASS_REFERENCE.md
nex3: Changed to "multiple assignment" anyway.

lib/sass/script/functions.rb
nex3, line 196: I've added a note in the top-level listing of all functions, and to the documentation of each individual function for which it seemed likely that people would want to use it on maps.

lib/sass/script/functions.rb
nex3: I assume this is in reference to lists-as-maps? See my comment above.

For this in particular, I'd want to check up front to be sure that all elements are pairs. We'd be iterating over all the elements soon enough anyway to generate the Ruby hash.

lib/sass/script/functions.rb
nex3: That seems likely to screw with anyone who writes a third-party function that deals with maps, or us if we try to implement a custom `to_h` in the future and forget to set the default. It seems safer to make them behave exactly like a normal Ruby hash.

lib/sass/script/functions.rb
nex3, line 1795: Done.

lib/sass/script/parser.rb
nex3: Done.

lib/sass/script/value/list.rb
nex3: Done.

lib/sass/tree/each_node.rb
nex3: Done.

lib/sass/tree/visitors/convert.rb
nex3, line 107: Empirical testing indicates that this is not the case.

test/sass/functions_test.rb
nex3, line 981: Since everything can be used as a list, I don't think `is-type-of(..., list)` will be especially useful. I think the only case where `type-of` really falls down right now is checking whether `()` is a map. If you think that's a strong enough case for `is-type-of`, open an issue for it.

test/sass/functions_test.rb
nex3, line 1165: Yep.

test/sass/functions_test.rb
nex3, line 1177: I don't think so. The notion of a separator is still only useful in the context of lists. Just think of every object in SassScript being some sort of list.

test/sass/functions_test.rb
nex3, line 1336: See above comment.

nex3:
Finished responding, uploaded new changes.

lib/sass/util.rb
nex3, line 41: It supports hashes coming in from third-party functions. We want those functions to be able to write `Sass::Script::Value::Hash.new({key => value})`.

We also occasionally pass in hashes here internally, but always OrderedHashes on 1.8.

lib/sass/util.rb
nex3, line 61: No other utils functions are defined that way. I'd rather stick with code consistency here. We can refactor all the version-specific functions later, or just wait until we drop support for 1.8.

doc-src/SASS_CHANGELOG.md
chriseppstein: I do think it should be space separated for consistency.

If I had `nth($map, 1), nth($map, 2)` I'd want this to return `a 1, b 2` instead of `(a, 1), (b, 2)` which outputs to CSS as a flattened list instead of a list of lists.

doc-src/SASS_REFERENCE.md
chriseppstein, line 758: Here's the issue: There are libraries that have a list of pairs as part of their public API. If we don't add support for migrating this API, each library author will have to do it. I think it would be nicer if we did it. Their code would be immediately simpler. If we have the list support immediately deprecated for some of the functions, I think that would be fine, the libraries' end users would simply need to convert their list of pairs to a map to silence the deprecations.

lib/sass/script/functions.rb
chriseppstein: The only benefit in my mind is that in ruby extensions, `map[some_key]` always returns a value object. It's not a big deal. I'm ok with keeping the existing behavior.

chriseppstein:
Done responding to your updates and comments.

We need to come to an agreement on how lists of pairs will be handled by the map functions. Also: I didn't get a reply about whether `keywords($arglist)` is supposed to be in this patch or not.

test/sass/functions_test.rb
chriseppstein, line 981: I'm not sure how urgent it is, but I created #885.

doc-src/SASS_CHANGELOG.md
nex3: Done.

doc-src/SASS_REFERENCE.md
nex3, line 758: Libraries that expose lists of pairs get free backwards compatibility if they start returning maps instead. Libraries that accept lists of pairs will have to use old pair-based methods as long as they want backwards-compatibility, but that doesn't affect users of those libraries, who can start using maps immediately.

Given that this won't introduce any backwards incompatibilities in existing code, I don't want to saddle it with a deprecation warning as soon as it starts existing.

nex3:
I've changed the map pairs to be space-separated.

> We need to come to an agreement on how lists of pairs will be handled by the map functions.

I'm still pretty convinced that the short-term gain isn't worth the long-term pain.

> Also: I didn't get a reply about whether `keywords($arglist)` is supposed to be in this patch or not.

I missed the question. I'll add that after this PR gets in.

doc-src/SASS_REFERENCE.md
```diff
@@ -1856,6 +1886,52 @@ is compiled to:
...
+#### Multiple Assignment {#each-multi-assign}
+
+The `@each` directive can also use multiple variables, as in `@each $var1,
+$var2, ... in <list>`. If `<list>` is a list of lists, each element of the
+sub-lists is assigned to the respective variable. For example:
+
+    @each $animal, $color, $cursor in (puma, black, default),
+                                      (sea-slug, blue, pointer),
+                                      (egret, white, move) {
+      .#{animal}-icon {
```
eevee: should be `$animal`

doc-src/SASS_REFERENCE.md
```diff
@@ -1856,6 +1886,52 @@ is compiled to:
...
+        border: 2px solid $color;
+        cursor: $cursor;
+      }
+    }
+
+is compiled to:
+
+    .puma-icon {
+      background-image: url('/images/puma.png');
+      border: 2px solid black; }
```
eevee: where'd `cursor` go?  :)

doc-src/SASS_REFERENCE.md
nex3: Done.

doc-src/SASS_REFERENCE.md
nex3: Done.

nex3:
I've added support for passing lists of pairs to the map functions. Please take a look.

lib/sass/script/functions.rb
```diff
@@ -380,7 +398,12 @@ def initialize(environment)
       # @param name [String, Symbol, nil] The name of the argument.
       # @raise [ArgumentError] if value is not of the correct type.
       def assert_type(value, type, name = nil)
-        return if value.is_a?(Sass::Script::Value.const_get(type))
+        klass = Sass::Script::Value.const_get(type)
+        return if value.is_a?(klass)
+        if value.is_a?(Sass::Script::Value::List) && type == :Map &&
+            value.to_a.all? {|e| e.is_a?(Sass::Script::Value::List) && e.to_a.length == 2}
```
chriseppstein: This is an enormous performance penalty to pay at each read on a list. We should move this check to the value object and memoize the result. I've seen some lists as maps with hundreds of entries.

lib/sass/script/functions.rb
```diff
@@ -1847,5 +1961,23 @@ def _adjust(color, amount, attr, range, op, units = "")
...
+        pair
+      end)
+      return hash if hash.empty?
+
+      fn_name = Sass::Util.caller_info.last.gsub('_', '-')
+      Sass::Util.sass_warn <<WARNING + environment.stack.to_s.gsub(/^/, '        ')
+DEPRECATION WARNING: Passing lists of pairs to #{fn_name} is deprecated and will
+be removed in future versions of Sass. Use Sass maps instead. For details, see
+http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#maps.
+WARNING
```
chriseppstein, line 1970: This causes a deprecation warning to be produced each time a list is used by a map function. To test, I made a simple sass file:

``` scss
$lmap: a 1, b 2, c 3, d 4;
a {
  @each $key in map-keys($lmap) {
    #{$key}: map-get($lmap, $key);
  }
}
```

This produced 5 deprecation warnings:

```
DEPRECATION WARNING: Passing lists of pairs to map-keys is deprecated and will
be removed in future versions of Sass. Use Sass maps instead. For details, see
http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#maps.
        on line 5 of list_as_map.scss
DEPRECATION WARNING: Passing lists of pairs to map-get is deprecated and will
be removed in future versions of Sass. Use Sass maps instead. For details, see
http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#maps.
        on line 6 of list_as_map.scss
DEPRECATION WARNING: Passing lists of pairs to map-get is deprecated and will
be removed in future versions of Sass. Use Sass maps instead. For details, see
http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#maps.
        on line 6 of list_as_map.scss
DEPRECATION WARNING: Passing lists of pairs to map-get is deprecated and will
be removed in future versions of Sass. Use Sass maps instead. For details, see
http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#maps.
        on line 6 of list_as_map.scss
DEPRECATION WARNING: Passing lists of pairs to map-get is deprecated and will
be removed in future versions of Sass. Use Sass maps instead. For details, see
http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#maps.
        on line 6 of list_as_map.scss
```

I think a deprecation warning should only be produced the first time a list is used as a map. It's a LOT of noise otherwise.

chriseppstein:
Three remaining issues:
1. performance of `assert_type :Map, a_list_of_pairs`
2. too many deprecation warnings
3. tests for sass-convert.

Once those are addressed, :+1:

lib/sass/script/functions.rb
nex3: I think you may be overestimating the cost of iterating through the list, especially since it will be unavoidably iterated when it's converted to a map, but it's easy to memoize so I've done so.

lib/sass/script/functions.rb
nex3, line 1970: Done.

lib/sass/script/functions.rb
chriseppstein: So is a new map created with each access? or is that result memoized? I may have missed this.

PR #865, opened 2013-07-28
Site Update

HamptonMakes:
Sorry for having a "mega commit", but I don't think much here is questionable and I think we need some new stuff up on the site... nowish!

Included:
- Use the site title that was agreed upon in a previous PR. "Sass: CSS with Superpowers" I like it, ship it.
- Replaced the "Blog" link (2 years old!) and put in a link to the Twitter account that Chris has been keeping up
- Replaced the "/try.html" page... which was looking terrrrible and was almost unusable... with a link to "http://trysass.com" which is an independent open source project that I'm keeping up. No compass support yet, but neither does the "try.html" page. Much simpler, much more modern, and type as you update, etc. The new site will have an embedded version, but I think this is a good idea for now.
- Small update to the <meta> description to have a good explanation for our Google results.

### DIFF
--- a/serve.rb
+++ b/serve.rb
@@ -1,24 +1,9 @@
-require '.haml/lib/haml'
-require '.sass/lib/sass'
 require 'rubygems'
 require 'sinatra'
 require 'sinatra/dynamicmatic'
 require 'timeout'
 require 'date'
 
-get('/try.html') {haml :try}
-
-post('/try.html') do
-  begin
-    Timeout.timeout(5) do
-      syntax = (params[:syntax] == 'sass') ? :sass : :scss
-      @result = Sass::Engine.new(params[:input], :syntax => syntax).render
-    end
-  rescue Sass::SyntaxError => e
-    @result = "Sass Error: " + e
-  rescue Timeout::Error
-    @result = "Timed out!"
-  end
-
-  haml :try
-end
+get('/try.html') {
+  redirect "http://trysass.com"
+}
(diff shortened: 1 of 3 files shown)


HamptonMakes:
Oh, and make sure to ping @chriseppstein

Snugug:
Considering that sassmeister allows for "try" plus has Compass and a bunch of Compass extension support, maybe point to that?

HamptonMakes:
@Snugug - that is a larger conversation to have. At the moment, Sass-meister is a completely independent project from Sass, and its long term management, direction, and financial management of its donation system haven't been discussed directly. 

As it stands today, officially pointing to Sassmeister is a bit like http://git-scm.com/ pointing to github.com. It's not impossible, but the official git project would want to have to talk a bit before that happened.

I _really want to_ have those discussions, but I believe that for today, linking to the more official 'drop-in replacement' of '/try.html' is the simplest first step to figuring out how an official "Try Sass" feature might work into the seemingly more startupy / codepen.io-style roadmap of Sassmeister. 

Also, note that the Sass site doesn't link to Compass either... only in Chris' bio.

roryokane:
[Try Sass](http://trysass.com/) is currently not a complete replacement for the [Try Online](http://sass-lang.com/try.html) page on the Sass website. Try Online lets you choose between SCSS and the [indented syntax](http://sass-lang.com/docs/yardoc/file.INDENTED_SYNTAX.html), while Try Sass only supports SCSS. I've created [an issue for this](https://github.com/hcatlin/trysass/issues/2) in Try Sass's issues.

I actually used the indented syntax option of Try Online just yesterday, when I had an idea for a mixin and wanted to try writing it right away. So the indented syntax option _is_ useful.

Of course, if it turns out that adding support for the indented syntax would take more time than you have, it might still be best to link to Try Sass anyway, for its other advantages. You can determine the tradeoffs, if such a decision becomes necessary.

nex3:
I'm closing this. If it's still relevant, file is again against sass/sass.github.com.

PR #879, opened 2013-08-01
Provide a better error message for the -E option in ruby 1.8.

### LINKED ISSUE #875: More of a Question -E Parameter
As per some issues on Prepros https://github.com/sbspk/Prepros/issues/190

Just wondering when the `-E utf-8` parameter was implemented, seems to be related to the Ruby Version?

chriseppstein:
Closes #875.

### DIFF
--- a/lib/sass/exec.rb
+++ b/lib/sass/exec.rb
@@ -293,8 +293,16 @@ END
           @options[:for_engine][:cache] = false
         end
 
-        unless ::Sass::Util.ruby1_8?
-          opts.on('-E encoding', 'Specify the default encoding for Sass files.') do |encoding|
+        encoding_desc = if ::Sass::Util.ruby1_8?
+          'Does not work in ruby 1.8.'
+        else
+          'Specify the default encoding for Sass files.'
+        end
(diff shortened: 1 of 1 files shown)


lib/sass/exec.rb
```diff
@@ -293,8 +293,16 @@ def set_opts(opts)
...
-        unless ::Sass::Util.ruby1_8?
-          opts.on('-E encoding', 'Specify the default encoding for Sass files.') do |encoding|
+        encoding_desc = if ::Sass::Util.ruby1_8?
+          'Does not work in ruby 1.8.'
+        else
+          'Specify the default encoding for Sass files.'
+        end
+        opts.on('-E encoding', encoding_desc) do |encoding|
+          if ::Sass::Util.ruby1_8?
+            warn "Specifying the encoding is not supported in ruby 1.8."
```
nex3, line 303: Elsewhere in this file we use `$stderr.puts` rather than `warn`. I don't have a strong preference between the two, but we should be consistent.

nex3:
One comment, otherwise :thumbsup:

chriseppstein:
Rebased for merge onto stable.

PR #891, opened 2013-08-17
Script & and @at-root

nex3:


lib/sass/environment.rb
```diff
@@ -43,6 +44,21 @@ def content
       @content || (@parent && @parent.content)
     end
 
+    # The selector for the current CSS rule, or nil if there is no
+    # current CSS rule.
+    #
+    # @return [Selector::CommaSequence?] The current selector, with any
+    #   nesting fully resolved.
+    def selector
+      parent_selector = caller ? caller.selector : (@parent && @parent.selector)
```
chriseppstein, line 54: I know it's not new, but I'm uncomfortable with the fact that we override `Kernel#caller`

lib/sass/engine.rb
```diff
@@ -839,6 +840,25 @@ def parse_directive(parent, line, root)
           Sass::Source::Position.new(@line, to_parser_offset(line.offset) + line.text.length),
           @options[:filename], @options[:importer])
         node
+      when 'at-root'
```
chriseppstein: This case statement is getting out of control. Especially this one. This code should be in a `parse_at_root` method. Ideally each of these clauses would be in a `parse_xxx_directive` method and we would simply use method dispatch instead of a giant case statement here.

lib/sass/environment.rb
nex3, line 54: I'm not too worried about it. It's very unlikely that any code in `Environment` will access `Kernel#caller`, especially since `Sass::Util` already wraps it.

lib/sass/engine.rb
nex3: Done.

Yedfree (approved):

PR #897, opened 2013-08-26
Add warning that recursion is forbidden to the Sass Reference

roryokane:
This documents the limitation I stumbled upon when writing a mixin.

I also fix a small grammar error in the following paragraph.

### DIFF
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -1974,7 +1974,13 @@ For example:
     @mixin highlighted-background { background-color: #fc0; }
     @mixin header-text { font-size: 20px; }
 
-Mixins that only define descendent selectors, can be safely mixed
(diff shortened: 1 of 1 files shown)


doc-src/SASS_REFERENCE.md
```diff
@@ -1974,7 +1974,13 @@ For example:
     @mixin highlighted-background { background-color: #fc0; }
     @mixin header-text { font-size: 20px; }
 
-Mixins that only define descendent selectors, can be safely mixed
+A mixin may not include itself, directly or indirectly. That is,
+mixin recursion is forbidden. Attempting to recurse will produce
+the following compilation error:
+
+    An @include loop has been found: xxxxx includes itself
```
nex3, line 1981: I don't think it's necessary to include the error message here. The rest of the documentation never does so.

PR #903, opened 2013-08-28
Added random number generator.

BeesBoxler:
Stolen from https://gist.github.com/chriseppstein/1561650


nex3:
I don't think this is something that's widely-applicable enough to be in the core set of functions.

PR #904, opened 2013-08-28
Added random number generator

BeesBoxler:
Added random number generator, taken from https://gist.github.com/chriseppstein/1561650, all credit to him.


twalpole:
Couple of things -- your test isnt actually being run since you have it as a private method.   Also I beliee If you had actually run the test it would fail since rand produces a number greater than or equal to 0 and less than max, which means the comment describing random is wrong too

BeesBoxler:
Apologies, I am more used to rspec for testing. Will fix asap. And thanks for pointing that out. I don't know where my head was when I wrote that!

BeesBoxler:
Was unsure how to test for a random number. Settled instead for generating a number and multiplying that by 0. If a string is generated, that will return a blank string, and therefore failing the test, looking for a zero.

Sorry, I'm clutching at straws with this, I could just not think of any other suitable method to test.

twalpole:
@benjaminboxler    random(1) should return always 0 so you could test that. Also, since its based on rand which is a pseudo random number generator you can seed it via a call to srand and then rand will produce the same "random" sequence each time - so you could verify using that -  although that really ties the testing to the implementation rather than just the functionality.

BeesBoxler:
Yes, that seems simple enough, thank you. It has been updated and passes tests now.

nex3:
As I mentioned on #903, I don't think this is widely-useful enough to warrant including in the core function library.

BeesBoxler:
Okay, no worries. I only opened this after seeing a few pens using the function after its inclusion on CodePen.

PR #907, opened 2013-08-31
Use system temporary directory as default cache_location

Moosemorals:
This is actually to fix a bug in the Netbeans plug-in, where @import doesn't work because sass tries to create a directory in \Program Files\Netbeans, and it can't because the user doesn't have Administrator access.

The patch uses the builtin Dir::tmpdir to find the system temporary directory, and then uses File.join to build the cache path. This replaces the default behaviour which creates a directory wherever sass happens to be running.


nex3:
I don't think we want to put the cache in a temporary directory where it can get blown away unless the user explicitly asks for it.

Why can't the Netbeans plugin configure this itself?

PR #919, opened 2013-09-13
Variable argument maps

nex3:
Add support for passing maps as variable arguments and using the `keywords()` function to access them.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -167,6 +167,26 @@ Produces:
       font-size: 1.2em;
     }
 
+#### Variable Keyword Arguments
+
+Maps can be passed as variable arguments, just like lists. For example, if
+`$map` is `(lightness: 10%, blue: -30%)`, you can write `scale-color($color,
+$map...)` and it will do the same thing as `scale-color($color, $lightness: 10%,
+$blue: 30%)`. To pass a variable argument list and map at the same time, just do
+the list first, then the map, as in `fn($list..., $map...)`.
+
+You can also access the keywords passed to a function that accepts a variable
+argument list using the new {Sass::Script::Functions#keywords `keywords($args)`
+function}. For example:
+
+    @function create-map($args...) {
+      @return keywords($args);
+    }
+
+    create-map($foo: 10, $bar: 11); // returns (foo: 10, bar: 11)
+
+#### Lists of Pairs as Maps
+
 The new map functions work on lists of pairs as well, for the time being. This
 feature exists to help libraries that previously used lists of pairs to simulate
 maps. These libraries can now use map functions internally without introducing
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -170,8 +170,8 @@ Produces:
 #### Variable Keyword Arguments
 
 Maps can be passed as variable arguments, just like lists. For example, if
-`$map` is `(lightness: 10%, blue: -30%)`, you can write `scale-color($color,
-$map...)` and it will do the same thing as `scale-color($color, $lightness: 10%,
+`$map` is `(alpha: -10%, "blue": 30%)`, you can write `scale-color($color,
+$map...)` and it will do the same thing as `scale-color($color, $alpha: -10%,
 $blue: 30%)`. To pass a variable argument list and map at the same time, just do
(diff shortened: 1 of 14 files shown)


lib/sass/script/tree/funcall.rb
```diff
@@ -23,20 +23,33 @@ class Funcall < Node
     # @return [{String => Node}]
     attr_reader :keywords
 
-    # The splat argument for this function, if one exists.
+    # The first splat argument for this function, if one exists.
+    #
+    # This could be a lsit of positional arguments, a map of keyword
```
chriseppstein: `lsit` => `list`

test/sass/conversion_test.rb
```diff
@@ -1684,6 +1722,24 @@ def test_at_root_with_selector
 SCSS
   end
 
+  def test_function_var_kwargs
```
chriseppstein: this method name was already used above.

lib/sass/tree/visitors/perform.rb
```diff
@@ -1,89 +1,135 @@
...
+            callable.args.map {|var| var.first.underscored_name})
+          if callable.splat && unknown_args.include?(callable.splat.underscored_name)
+            raise Sass::SyntaxError.new("Argument $#{callable.splat.name} of #{downcase_desc} cannot be used as a named argument.")
+          elsif unknown_args.any?
+            description = unknown_args.length > 1 ? 'the following arguments:' : 'an argument named'
+            raise Sass::SyntaxError.new("#{desc} doesn't have #{description} #{unknown_args.map {|name| "$#{name}"}.join ', '}.")
+          end
         end
+      rescue Sass::SyntaxError => keyword_exception
       end
```
chriseppstein, line 35: Why are we just discarding this exception? Looks like a bug, if not, it needs a comment explaining what's going on here.

lib/sass/tree/visitors/perform.rb
```diff
@@ -1,89 +1,135 @@
...
+        env.set_local_var(var.name, value)
+      end
 
-    keywords = keywords.dup
-    env = Sass::Environment.new(callable.environment)
-    callable.args.zip(args[0...callable.args.length]) do |(var, default), value|
-      if value && keywords.include?(var.underscored_name)
-        raise Sass::SyntaxError.new("#{desc} was passed argument $#{var.name} both by position and by name.")
+      if callable.splat
+        rest = args[callable.args.length..-1]
```
chriseppstein: This should be `args[callable.args.length..-1] || []` because it's `nil` when the the number of args is the same.

I pushed a test and fix for this to the branch.

chriseppstein:
``` scss
$adjust-by: (alpha: -10%, "blue": 30%);
.color {c: scale-color($color: purple, $red: -20%, $adjust-by...)};
```

Raises the error:

```
Syntax error: Positional arguments must come before keyword arguments.
```

But the arglist contains only keywords. I think we should allow arglist passing of the form:

``` scss
something(arg1, arg2, $kwarg1: v1, $kwarg2: v2, $positional_args..., $kwargs...)
// or
something(arg1, arg2, $positional_args..., $kwarg1: v1, $kwarg2: v2, $kwargs...)
```

doc-src/SASS_CHANGELOG.md
```diff
@@ -167,6 +167,26 @@ Produces:
       font-size: 1.2em;
     }
 
+#### Variable Keyword Arguments
+
+Maps can be passed as variable arguments, just like lists. For example, if
+`$map` is `(lightness: 10%, blue: -30%)`, you can write `scale-color($color,
+$map...)` and it will do the same thing as `scale-color($color, $lightness: 10%,
+$blue: 30%)`. To pass a variable argument list and map at the same time, just do
```
chriseppstein, line 175: this example doesn't actually work. I updated it.

lib/sass/script/parser.rb
```diff
@@ -517,6 +524,7 @@ def literal
         :default => "expression (e.g. 1px, bold)",
         :mixin_arglist => "mixin argument",
         :fn_arglist => "function argument",
+        :splat => "...",
```
chriseppstein, line 527: There's a utf-8 character `…` I wonder if we should allow it as splat as well ^_^

chriseppstein:
Landed as a4e9233341aeaf94bd00b4c85b95e004c94df2f7

Issue #847, opened 2013-07-11
Feature: Add YUI-style optimisations for compressed CSS output


ghost:
Just to name a few optimisations that the YUI compressor handles but SASS doesn't:
- Removing the leading zero from floating-point numbers (`0.5em` -> `.5em`)
- Removing the units from 0 (`0px` -> `0`)
- Removing spaces in the middle of font lists (`"Font 1", "Font 2"` -> `"Font 1","Font 2"`)

Anahkiasen:
...does that really provide an enormous speed boost to the loading time ?

kevinSuttle:
I'd be interested in this. Every little bit helps, though a linter seems a bit out of scope. http://csslint.net/

ghost:
In hindsight, having a linter does seem a bit out of scope, although things like removing units from values can reduce file size by quite a bit if you have a lot of definitions. Stuff like condensing border definitions into one rule is kind of out of the scope of SASS.

kevinSuttle:
Don't get me wrong, I still _very_ much want that kind of compression and minification, along with linting. In fact, I have been trying to handle this with Grunt or Rake tasks lately, so I'm glad this discussion is happening now (yes, I use and am aware of the compressed style of output). What we're talking about is pushing Sass beyond compression into linting. Maybe it could fit into the goals of Sass, I dunno. I think the discussion would have to be with a wider audience.

Maybe it could be an option in Compass? + @chriseppstein

kevinSuttle:
Linting could also be something that could be a part of a GUI app like Codekit. Thoughts @bdkjones?

bdkjones:
CodeKit already offers the compressed option for Sass output. 

Is there a good Sass lint tool available?

kevinSuttle:
Ooh now that's a good question.

kevinSuttle:
I really thought there'd be more traction and opinions on this. 

A quick search didn't yield much other than the following: 
https://github.com/causes/scss-lint

mhulse:
+1 on remove leading zero.

The other optimizations look good as well. :)

LeeMallabone:
@Undeterminant It looks like the font list compression is already present in master. If I've misunderstood, please add another example and I'll take a look.

ghost:
This bug was reported 2 months ago, so, it may have silently been fixed with a code update. Either way, if it's in master, you can just accept it as implemented.

kevinSuttle:
Any updates re: Sass/SCSS linting?

[...]

PR #922, opened 2013-09-16
Add more compression for numbers (implementing issue #847)

LeeMallabone:
As requested in issue #847, this change implements the following when outputting css in `:compressed` mode:
- Removing the leading zero from floating-point numbers (`0.5em` -> `.5em`)
- Removing the units from 0 (`0px` -> `0`)

These are implemented in the YUI compressor but previously missing from sass.


chriseppstein:
It's a good feature, but it needs an iteration to take code review into account.

nex3:
Closing due to unresponsiveness.

Issue #691, opened 2013-03-20
@for can't decrement


ghost:
I was a bit surprised to see this isn't possible:

```
@for $i from 7 through 1 {
    @debug $i;
}
```

The expected output being:

```
./test.scss:34 DEBUG: 7
./test.scss:34 DEBUG: 6
./test.scss:34 DEBUG: 5
./test.scss:34 DEBUG: 4
./test.scss:34 DEBUG: 3
./test.scss:34 DEBUG: 2
./test.scss:34 DEBUG: 1
```

The output, however was nothing.
Was this by design? I can't imagine it would have been.

robwierzbowski:
`@for` only counts up. Whether this is a bug or not, you can get around it with negative numbers and the `abs()` (absolute value) function:

``` scss
@for $i from -7 through -1 {
    @debug abs($i);
}
```

Result:

```
/style.scss:3 DEBUG: 7
/style.scss:3 DEBUG: 6
/style.scss:3 DEBUG: 5
/style.scss:3 DEBUG: 4
/style.scss:3 DEBUG: 3
/style.scss:3 DEBUG: 2
/style.scss:3 DEBUG: 1
```

ghost:
Thanks @robwierzbowski: Yea, there are various workarounds, they're just less intuitive. I don't know how high of a priority a native implementation would be, but it'd still be nice to see.

chriseppstein:
I think adding a `by` clause to `@for` would be very nice. Sure it can always be done via math, but it's a nice sugar.

Examples:

``` scss
@for $v from 0 through 10 by 2 { ... }
@for $v from 10 to 1 by -1 { ... }
```

Also, Sass can fairly easily change from increment to decrement when the orientation of the values is flipped. This avoids infinite loops. I'm :thumbsup: on this one.

nex3:
I'm a little ambivalent about supporting `@for ... by`. I'm not sure it's widely useful enough to warrant expanding the `@for` syntax, especially given how easy it is to do by hand.

I do think `@for $i from 10 to 1` should work, though.

dustinhorton:
Just ran into this, so if actual use cases help: I was using a for loop to generate a series of max-width media queries, issue being it was generating larger & larger max-width media queries which would simply override all those prior.  Fixed using `abs()` to get a decrementing `@for` loop.

Can post code if that'd help, but +1 for `@for $i from 10 to 1`, or the alternate syntax Chris proposed.

ghost:
[...]

PR #923, opened 2013-09-17
Add support for decrementing for-loops

robinroestenburg:
PR to implement decrementing for-loops as discussed in issue #691 (sorry, couldn't get Hub to attach the pull-request to the issue directly).

From the CHANGELOG:

```
It is now possible to create decrementing loops using the default @for-directive. For example:

    .foo {
      @for $var from 5 through 1 { a: $var }
    }

Produces:

    .foo {
      a: 5;
      a: 4;
      a: 3;
      a: 2;
      a: 1; }
```

I did not add the `by` part to the `@for` loop as suggested by @chriseppstein in issue #691. I'd be happy to implement this in another PR. 

I am a bit unsure about adding the constants to the `ForNode`, they seem a bit out of place there. At first I was going to calculate the direction of the loop in the `ForNode` as well, but you can't do that because the environment can change the values of `from` and `to` later on, right?

Let me know what you think!

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -5,6 +5,27 @@
 
 ## 3.3.0 (Unreleased)
 
+### Decrementing for-loops
+
+It is now possible to create decrementing loops using the default `@for`
+directive by having the `from` number be larger than the `to` number. For
+example:
+
+    .foo {
+      @for $var from 5 through 1 { a: $var }
+    }
+
+Produces:
+
+    .foo {
+      a: 5;
+      a: 4;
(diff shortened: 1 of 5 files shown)


chriseppstein:
This breaks in ruby 1.8.7

chriseppstein:
Thanks for the patch, it needs to be reworked before we can accept it. Please address the comments from the code review.

robinroestenburg:
I have addressed the various comments from your code review, thanks for that.
Just let me know if you have any other comments.

chriseppstein:
Looks good to me. All merges are currently blocked pending #850.

PR #934, opened 2013-09-24
If with delayed evaluation

chriseppstein:
This is another take on an implementation for making the `if` function behave correctly (only the branch of code corresponding to the condition is evaluated.

I like this one better than #836.

The approach here is to allow the `if` sass function to declare some of its arguments as delayed. For those arguments the unevaluated syntax tree node is passed to the function instead of the value.

This approach makes functions with delayed arguments behave identically to normal functions with respect to their calling syntax.

The evaluation context now provides a `perform` method which can be used to perform a tree node in the correct environment. If we decide other Sass builtins require this sort of delayed evaluation it will now be trivial to build them. For instance we could trivially implement something like:

```
for-each(1 2 3 4 5, i, $i * 2) => (2 4 6 8 10)
```

This ability is limited to the if function by forbidding other functions during declaration.

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -298,10 +298,12 @@ module Sass::Script
 
     # A class representing a Sass function signature.
     #
-    # @attr args [Array<Symbol>] The names of the arguments to the function.
+    # @attr args [Array<String>] The names of the arguments to the function.
+    # @attr delayed_args [Array<String>] The names of the arguments whose evaluation should be
+    #   delayed.
     # @attr var_args [Boolean] Whether the function takes a variable number of arguments.
     # @attr var_kwargs [Boolean] Whether the function takes an arbitrary set of keyword arguments.
-    Signature = Struct.new(:args, :var_args, :var_kwargs)
+    Signature = Struct.new(:args, :delayed_args, :var_args, :var_kwargs)
 
     # Declare a Sass signature for a Ruby-defined function.
     # This includes the names of the arguments,
@@ -338,9 +340,23 @@ module Sass::Script
     #   In addition, if this is true and `:var_args` is not,
     #   Sass will ensure that the last argument passed is a hash.
     def self.declare(method_name, args, options = {})
+      delayed_args = []
+      args = args.map do |a|
+        a = a.to_s
+        if a[0] == ?&
+          a = a[1..-1]
+          delayed_args << a
+        end
+        a
+      end
+      # We don't expose this functionality except to certain builtin methods.
+      if delayed_args.any? && method_name != :if
+        raise ArgumentError.new("Delayed arguments are not allowed for method #{method_name}")
+      end
       @signatures[method_name] ||= []
       @signatures[method_name] << Signature.new(
-        args.map {|s| s.to_s},
+        args,
+        delayed_args,
         options[:var_args],
         options[:var_kwargs])
     end
@@ -476,6 +492,24 @@ module Sass::Script
           raise ArgumentError.new("Expected $#{name} to be an integer but got #{number}")
         else
           raise ArgumentError.new("Expected #{number} to be an integer")
+        end
+      end
+
+      # Performs a node that has been delayed for execution.
+      #
+      # @private
+      # @param node [Sass::Script::Tree::Node,
+      #   Sass::Script::Value::Base] When this is a tree node, it's
+      #   performed in the caller's environment. When it's a value
+      #   (which can happen when the value had to be performed already
+      #   -- like for a splat), it's returned as-is.
+      # @param env [Sass::Environment] The environment within which to perform the node.
+      #   Defaults to the (read-only) environment of the caller.
+      def perform(node, env = environment.caller)
+        if node.is_a?(Sass::Script::Value::Base)
+          node
+        else
+          node.perform(env)
         end
       end
     end
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -1997,17 +1997,17 @@ module Sass::Script
     # @overload if($condition, $if-true, $if-false)
     # @param $condition [Sass::Script::Value::Base] Whether the `$if-true` or
     #   `$if-false` will be returned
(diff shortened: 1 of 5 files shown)


cjcheshire:
Forgive my stupidity but would this delayed approach now allow us to perform code within a mixin content block making it react to conditions set within the mixin?

chriseppstein:
@cjcheshire nope.

lib/sass/script/functions.rb
```diff
@@ -380,14 +394,23 @@ class EvaluationContext
       # @return [Environment]
       attr_reader :environment
 
+      # The environment of the caller.
+      #
+      # Only set when the function has delayed argument evaluation.
+      #
+      # @return [Environment]
+      attr_reader :environment_of_caller
```
chriseppstein: This needs to use `environment.caller` instead after #821 lands.

nex3:
I continue to strongly dislike this approach. In an abstract sense, I understand the motivation to find a more generic solution for this than just a one-off node subclass. Practically speaking, though, I don't _want_ to add more functions that delay the evaluation of their arguments. Macros are confusing; their very existence erodes the confidence users have in their ability to reason about what a function might be doing.

`if()` is a case where it's been shown very clearly that the function itself is useful and users expect it to behave differently than other functions. I think it's very unlikely that we'll find any other use cases for this infrastructure that even approach the intuitive clarity of `if()`.

Given this, I'd like to switch back to the previous approach of adding a separate script node for `if()`.

chriseppstein:
My motivation is not re-use. My motivation is consistency of implementation and maintainability. This solution is a more elegant way to implement this feature given the requirements you defined for exact function call semantics. My old code still exists in the other PR. Feel free to dust it off and finish that branch if you prefer. I want this feature in 3.3 and I will not accept holding it out because of an implementation detail for which the only argument against is a slippery slope.

lib/sass/script/functions.rb
```diff
@@ -284,7 +284,7 @@ module Functions
     # @attr args [Array<Symbol>] The names of the arguments to the function.
     # @attr var_args [Boolean] Whether the function takes a variable number of arguments.
     # @attr var_kwargs [Boolean] Whether the function takes an arbitrary set of keyword arguments.
-    Signature = Struct.new(:args, :var_args, :var_kwargs)
+    Signature = Struct.new(:args, :delayed_args, :var_args, :var_kwargs)
```
nex3, line 306: Add an `@attr` doc comment for `delayed_args`.

lib/sass/script/functions.rb
```diff
@@ -321,9 +321,23 @@ module Functions
...
+        a = a.to_s
+        if a[0] == ?&
+          a = a[1..-1]
+          delayed_args << a
+        end
+        a
+      end
+      # We don't expose this functionality except to certain builtin methods.
+      if delayed_args.any? && method_name != :if
+        raise ArgumentError.new("Delayed arguments not allowed for method #{method_name}")
```
nex3: "arguments not allowed" -> "arguments are not allowed"

lib/sass/script/functions.rb
nex3: Mark this `@private`, if it sticks around.

lib/sass/script/functions.rb
```diff
@@ -454,6 +477,21 @@ def assert_integer(number, name = nil)
...
       end
+
+      # Performs a node that has been delayed for execution
+      #
+      # @param node [Sass::Script::Tree::Node, Sass::Script::Value::Base] When a tree node, the
+      #   node is performed in the environment of the caller. When a value (happens when the value
+      #   had to be performed already -- like for a splat), it is simply returned.
+      # @param node [Sass::Environment] The environment within which to perform the node.
+      #   Defaults to the environment of the caller.
+      def perform(node, env = environment_of_caller)
```
nex3: Mark this `@private` as well.

lib/sass/script/functions.rb
```diff
@@ -454,6 +477,21 @@ def assert_integer(number, name = nil)
           raise ArgumentError.new("Expected #{number} to be an integer")
         end
       end
+
+      # Performs a node that has been delayed for execution
```
nex3: Nit: period.

lib/sass/script/functions.rb
```diff
@@ -454,6 +477,21 @@ def assert_integer(number, name = nil)
...
+      #
+      # @param node [Sass::Script::Tree::Node, Sass::Script::Value::Base] When a tree node, the
+      #   node is performed in the environment of the caller. When a value (happens when the value
+      #   had to be performed already -- like for a splat), it is simply returned.
```
nex3: This comment is complex enough already that it would probably be easier to read with complete sentences.

nex3:
I'm not finished reviewing yet, but you've convinced me that this is a better way to go. I'll finish the review on Monday.

lib/sass/script/tree/funcall.rb
```diff
@@ -114,9 +114,26 @@ def deep_copy
     # @return [Sass::Script::Value] The SassScript object that is the value of the function call
     # @raise [Sass::SyntaxError] if the function call raises an ArgumentError
     def _perform(environment)
-      args = @args.map {|a| a.perform(environment)}
+      has_delayed_arguments = false
+      signature = Sass::Script::Functions.signature(name.to_sym, @args.size, @keywords.size)
+      args = []
+      @args.each_with_index do |a, i|
```
nex3: `args = enum_with_index(@args).map do |a, i|`

test/sass/functions_test.rb
```diff
@@ -1461,4 +1481,9 @@ def assert_error_message(message, value)
     assert_equal(message, e.message)
   end
 
+  def render(sass, options = {})
+    munge_filename options
+    Sass::Engine.new(sass, options).render
+  end
+
```
nex3: Nit: extra newline

nex3:
Done reviewing. A few small changes, otherwise :+1:.

nex3:
Made the changes myself, at Chris's request.

PR #939, opened 2013-09-26
Make access to the dependency cache threadsafe.

chriseppstein:
This wraps all accesses to the dependency cache in a mutex so that the global thread-unsafe hash object won't trigger hangs in JRuby.


nex3:
One nit, otherwise :+1:.

chriseppstein:
squashed and merged to stable.

Issue #930, opened 2013-09-20
new scriptable `&` can't see parents defined in mixins


pdaoust:
**Edit**: I've modified this issue because I now believe it all has to do with scriptable `&` rather than `@at-root`, because if you run the code below, the comment at the beginning of the `context` mixin shows that `&` has already lost knowledge of its parent, before `@at-root` is even used.

I've got a snippet of code in the new Sass 3.3 which is giving me a bit of trouble:

``` scss
@mixin question-mark-symbol($selector) {
    #{$selector} {
        /* In the question-mark-symbol mixin, the current context is #{&} */
        @include generated-content-context {
            &::before {
                content: '?';
            }
        }
    }
}

@mixin generated-content-context {
    /* In the context mixin, the current context is #{&} */
    $current-selector: #{&};
    @if $current-selector == "" { $current-selector: "&"; }
    @at-root .generated-content {
        #{$current-selector} {
            @content;
        }
    }
}

/* This (example 1)... */
@include question-mark-symbol('.i-have-a-question-mark');

/* I expected to be the same as this (example 2)... */
.i-have-a-question-mark {
    @include question-mark-symbol('&');
}
```

Output:

``` css
/* This (example 1)... */
.i-have-a-question-mark {
  /* In the question-mark-symbol mixin, the current context is .i-have-a-question-mark */
  /* In the context mixin, the current context is  */ }
  .generated-content::before {
    content: '?'; }

/* I expected to be the same as this (example 2)... */
.i-have-a-question-mark {
  /* In the question-mark-symbol mixin, the current context is .i-have-a-question-mark */
  /* In the context mixin, the current context is .i-have-a-question-mark */ }
  .generated-content .i-have-a-question-mark::before {
    content: '?'; }
```

Apologies for the convoluted code; this is a simplification of a real-life situation.
1. I invoke the `question-mark-symbol` mixin.
2. `question-mark-symbol` creates a rule, using the passed argument as a selector.
3. Inside that rule, it uses a simplification of my `context` mixin (which handles classes on the root element, but in real life also handles `@media` queries and other junk).
4. The `context` mixin outputs the style in a `.generated-content` class (which is a class that some feature-detecting JS puts on the `<html>` element, if you're curious).

The funny thing is that, whenever a rule is created in a mixin and then that mixin calls another mixin, the inner mixin can't see the selector from the previous mixin. I don't know if this is a design choice or a bug, but it was very unexpected to me, because I assumed that a rule was always added to the parent.

[...]

PR #940, opened 2013-09-27
added test for script & when the parent is created in a mixin

pdaoust:
On Sass 3.3 alpha 255, Sass seems to lose certain parts of the parent selector when it's interpolated in Sass script: that is, in a nested mixin, it seems to lose any rule nodes that were created in containing mixins. More info and sample code in #930. Sorry I can't help any further -- I did clone a copy of the repo onto my computer and tried to fish through the code, but I think that fixing this issue would demand a lot more knowledge of the code than I have!

Oh, and sorry for submitting a pull request that intentionally breaks the CI build :-)

### DIFF
--- a/test/sass/scss/scss_test.rb
+++ b/test/sass/scss/scss_test.rb
@@ -2025,6 +2025,25 @@ CSS
 SCSS
   end
 
+  def test_selector_script_through_mixin_with_parent_created_in_mixin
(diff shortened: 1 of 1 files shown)


pdaoust:
Now that commit cc1787d includes a very similar test, I'm going to close this pull request.

PR #958, opened 2013-10-09
Refactor @at-root to ensure that it works properly with a plain &.

nex3:
This changes the structure of the intermediate trees a little. Instead of having the RuleNode nesting resolved entirely in the Cssize visitor, the selectors are now resolved in Perform, while the nesting itself is removed in Cssize. This is necessary to make the resolved selectors consistently available to script &.


chriseppstein:
:thumbsup:

Issue #955, opened 2013-10-08
Upgrade to listen 2.0


nex3:
[guard/listen](/guard/listen) has released version 2.0.0. We should make sure it plays nicely with Sass, then upgrade to it.

nex3:
This isn't going to make it into 3.3; listen 2.0 is incompatible with Ruby 1.8.7, so we can't use it until we move our earliest supported version to something in the 1.9.\* range. Unfortunately, many people are still using versions of OS X that ship with 1.8.7, so this may not be able to happen very soon.

spicycode:
@nex3 Given mavericks shipped with Ruby 2.0, is there any movement on supporting newer versions?  We're evaluating whether we want to maintain a fork in the interim vs waiting for a fix in master.

nex3:
We'll need to wait until Ruby 1.8.7 is well out of circulation before we drop support for it. We can't just tell all our users to upgrade their OS if they want to use Sass; we need to support the OSes that our users are actually using.

That said, #982 adds support for listen 2.0 with a fallback to listen 1.1, so that should work if you want 2.0.

spicycode:
Excellent, we'll give that a try in 2.0 and 1.9.X.

metaskills:
Supporting 1.8.7 seems like it is over-heroic and penelizes those using current software right? Like I have Guard bundled in my Rails app and I need to use the latest versions for my MiniTest support. Now I cant bundle to the latest Sass? So someone is going to pay a price, I vote it be the people using Ruby 1.8.7.

nex3:
Ruby 1.8.7 was bundled with Mac OS X up until version 10.9, which was released in October 2013. Anyone with an older version of OS X who isn't themself a Ruby user will be using 1.8.7. A large proportion of Sass users are on OS X, and many of those users don't use Ruby. These users don't know or want to learn how to upgrade their Ruby version, and we don't want to force them to do so.

That said, we've re-bundled listen with Sass, so listen 2.0.0 now works properly if it's installed.

metaskills:
Thanks @nex3 awesome work!

PR #959, opened 2013-10-10
Upgrade to listen 2.0

nex3:
This preserves the existing `sass --watch` functionality. There are still some performance issues and small behavior issues stemming from guard/listen#111, but those are unlikely to be solved in time for 3.3.

Addresses #955.


nex3:
Oops, this PR was supposed to be against master.

PR #960, opened 2013-10-10
Upgrade to listen 2.0

nex3:
This preserves the existing `sass --watch` functionality. There are still some performance issues and small behavior issues stemming from guard/listen#111, but those are unlikely to be solved in time for 3.3.

Addresses #955.

### DIFF
--- a/lib/sass/plugin/compiler.rb
+++ b/lib/sass/plugin/compiler.rb
@@ -284,6 +284,7 @@ module Sass::Plugin
 
       listener.start
       listener.thread.join
+      listener.stop # Partially work around guard/listen#146
     end
 
     # Non-destructively modifies \{#options} so that default values are properly set,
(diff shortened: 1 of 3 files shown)


chriseppstein:
Listen 2.0 depends on Celluloid which depends on ruby 1.9.2. 

Test failures: https://travis-ci.org/nex3/sass/jobs/12377471

A good portion of our users are using standard ruby installs on macs that came with 1.8.7. This is a non-starter.

:-1:

chriseppstein:
If we ship a stand-alone installer, we could do this. But as things stand right now, we can't ask our users to upgrade their ruby versions.

nex3:
I didn't realize that Listen 2.0 didn't run on Ruby 1.8.7; it's definitely not time to drop support for that yet. I'm going to close this PR but keep the branch around so we can use it later.

PR #961, opened 2013-10-10
Be more liberal in the allowed ordering of var args.

### LINKED ISSUE #926: Nonsensical error passing keywords as a splat after explicit keywords.
``` scss
$adjust-by: (alpha: -10%, "blue": 30%);
.color {c: scale-color($color: purple, $red: -20%, $adjust-by...)};
```

Raises the error:

```
Syntax error: Positional arguments must come before keyword arguments.
```

At the very least, we should have a better error in this case. But the arglist contains only keywords. I think we should allow arglist passing of the form:

``` scss
something(arg1, arg2, $kwarg1: v1, $kwarg2: v2, $positional_args..., $kwargs...)
// or
something(arg1, arg2, $positional_args..., $kwarg1: v1, $kwarg2: v2, $kwargs...)
```

nex3:
Keyword arguments may occur after the first "...".

Closes #926

### DIFF
--- a/lib/sass/script/parser.rb
+++ b/lib/sass/script/parser.rb
@@ -437,6 +437,7 @@ RUBY
 
         return [args, keywords] unless e
 
+        splat = nil
         loop do
           if @lexer.peek && @lexer.peek.type == :colon
             name = e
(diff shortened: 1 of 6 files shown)


chriseppstein:
All tests are failing.

PR #962, opened 2013-10-11
@at-root w/directives

nex3:


nex3:
supposed to be against master

PR #968, opened 2013-10-15
Random function

chriseppstein:
Lots of people at SassConf want this. so here it is.

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -2249,5 +2249,17 @@ WARNING
     def generator
       Sass::Script::Functions.random_number_generator
     end
+
+    unless defined?(::Random)
+      # Shim for ruby 1.8.7 support.
+      class Random
+        def initialize(seed = nil)
+          srand(seed) if seed
+        end
+        def rand(*args)
+          Kernel.rand(*args)
+        end
+      end
+    end
   end
 end
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -401,6 +401,25 @@ module Sass::Script
       @signatures[method_name].first
     end
 
+    # Sets the random seed used by Sass's internal random number generator.
+    #
+    # This can be used to ensure consistent random number sequences which
+    # allows for consistent results when testing, etc.
+    #
+    # @param seed [Integer]
+    # @return [Integer] The same seed.
+    def self.random_seed=(seed)
+      @random_number_generator = Random.new(seed)
+      seed
+    end
+
+    # Get Sass's internal random number generator.
+    #
+    # @return [Random]
+    def self.random_number_generator
+      @random_number_generator ||= Random.new
+    end
+
     # The context in which methods in {Script::Functions} are evaluated.
     # That means that all instance methods of {EvaluationContext}
     # are available to use in functions.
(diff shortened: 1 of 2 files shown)


chriseppstein:
@nex3 Please review. You promised this would get into 3.3. Though you may have been too drunk to remember. But I have a witness: @snugug ;)

Snugug:
I am a witness!

lib/sass/script/functions.rb
```diff
@@ -401,6 +401,25 @@ def self.signature(method_name, arg_arity, kwarg_arity)
...
+    # Sets the random seed used by Sass's internal random number generator.
+    #
+    # This can be used to ensure consistent random number sequences which
+    # allows for consistent results when testing, etc.
+    #
+    # @param seed [Integer]
+    # @return [Integer] The same seed.
+    def self.random_seed=(seed)
+      @random_number_generator = Random.new(seed)
+      seed
```
nex3, line 413: This is unnecessary; Ruby automatically returns the right-hand side of an assignment, even if the assignment method itself doesn't return its argument. Similarly, the doc comment describing this return value is unnecessary.

lib/sass/script/functions.rb
```diff
@@ -2168,6 +2186,25 @@ def inspect(value)
     end
     declare :inspect, [:value]
 
+    # @overload random()
+    #   Return a decimal between 0 and 1.
+    #   @param $limit The maximum of the random integer to be returned.
```
nex3, line 2191: Copy/paste error?

lib/sass/script/functions.rb
```diff
@@ -2205,5 +2242,25 @@ def to_h(obj)
...
+      class Random
+        def initialize(seed = nil)
+          srand(seed) if seed
+        end
+
+        def rand(*args)
+          Kernel.rand(*args)
+        end
+      end
+    end
```
nex3, line 2264: Let's move this shim into `Sass::Util`.

lib/sass/script/functions.rb
```diff
@@ -2168,6 +2186,25 @@ def inspect(value)
     end
     declare :inspect, [:value]
 
+    # @overload random()
+    #   Return a decimal between 0 and 1.
```
nex3, line 2190: Inclusive or exclusive?

test/sass/functions_test.rb
```diff
@@ -1574,6 +1574,20 @@ def test_inspect
...
+    assert_equal "13", evaluate("random(100)")
+  end
+
+  def test_random_works_without_a_seed
+    if Sass::Script::Functions.instance_variable_defined?("@random_number_generator")
+      Sass::Script::Functions.send(:remove_instance_variable, "@random_number_generator")
+    end
+    assert_nothing_raised do
+      evaluate("random()")
+    end
```
nex3, line 1589: Let's at least assert that this is within the expected range.

lib/sass/script/functions.rb
```diff
@@ -2168,6 +2186,25 @@ def inspect(value)
...
+    #   @param $limit The maximum of the random integer to be returned.
+    #   @return [Sass::Script::Number] A decimal value.
+    # @overload random($limit)
+    #   Return an integer between 1 and $limit inclusive.
+    #   @param $limit The maximum of the random integer to be returned.
+    #   @return [Sass::Script::Number] An integer.
+    def random(limit = nil)
+      if limit
+        assert_integer limit, "limit"
```
nex3, line 2199: We should check that `limit` is greater than zero, and test that check below.

nex3:
Done reviewing (and I remember agreeing to this, don't worry).

PR #971, opened 2013-10-17
Improve performance when processing files with few newlines

shepmaster:


seejee:
@shepmaster This is awesome. Nice work! I'm copying your commit message here, so that your fantastic explanation is front-and-center:

```
Using `rindex` works well when we can guarantee that a newline will
occur reasonably close to the end of the string; this does not hold
true for source files that have already been minimized.

When processing source that has few newlines, we have to walk a very
long way through the entire string. This is compounded by the fact
that we may have to walk the same string many times, once for each
subsequent directive we are processing. This leads to behavior that
looks O(n^2).

Instead, we lazily pay a one-time cost to figure out where all the
newlines are, then quickly scan through those locations when
calculating a specific column.

This will be most significant on JRuby, which currently (1.7.5) has a
less-than-efficient implementation of `rindex` when non-ASCII
characters are involved. This fix speeds up our real-life asset
precompilation from 4:30 to 2:00!

Some synthetic benchmarks that worst expose the problem:

MRI (1.9.3-p448) - 250KiB of CSS
Before: 71.5
After:  24.8 (34.6%)

JRuby (1.7.5) - 30KiB of CSS
Before: 188.4
After:   27.3 (14.5%)
```

robwierzbowski:
Performance gains :+1: :smile:.

shepmaster:
Is there anything else I can do to help get this commit merged? Love to help! :heart:

shepmaster:
@nex3 Doing my monthly check-in to see if there is anything more I can do to help push this along. Thanks!

:v: :deciduous_tree: :bear:

nex3:
No, you're fine; this is just waiting on me to have enough time to do a review.

lib/sass/util/source_location_finder.rb
```diff
@@ -0,0 +1,42 @@
...
+    # column number given the file (as a string) and an offset into
+    # the string.
+    class SourceLocationFinder
+      attr_reader :source_string
+
+      def initialize(source_string)
+        @source_string = source_string
+      end
+
+      def column(index)
```
nex3, line 16: Document this method and the `#source_string` attribute.

lib/sass/util/source_location_finder.rb
```diff
@@ -0,0 +1,42 @@
...
+        preceding_newlines = newline_positions.take_while {|pos| pos < index}
```
nex3, line 17: This is technically still `O(source_string.length)`, although it will certainly have a much smaller constant than iterating through the whole string. It's worth investigating making this `O(1)` by precomputing an array that maps source positions to the position of the preceding newline. Then you could just return `newline_positions[index]` here and be done with it.

nex3:
Finished reviewing. This is a very nice-looking patch.

chriseppstein:
This patch is against stable. I've been trying to get it working on master.

I've also implemented @nex3's suggestion to memoize all the offsets and it also tracks line numbers now too. I should be able to get a bow on this tomorrow then we should benchmark it again.

@shepmaster do you have the files you were using for benchmarking your changes anywhere?

chriseppstein:
@nex3 A general problem I've noticed with stringscanner is that scanner.pos tracks the byte offset and not the character offset. I'm pretty sure we're misreporting column numbers when multibyte characters are involved. I'm not going to touch that aspect here and it's not clear whether sourcemaps are character based or byte-based. If they are byte-based, we need to track bytes for sourcemaps and characters for error reporting. Sigh.

nex3:
@chriseppstein `Sass::Util::MultibyteStringScanner` should handle multibyte characters correctly. Until we hear otherwise, I think it's safe to assume that source maps use characters rather than bytes.

chriseppstein:
Meh, I couldn't get the tests back to green before I gave up on this because I had to move on to higher priority tasks. Hopefully @shepmaster will keep going with it.

nex3:
@shepmaster I'd like to get this into the upcoming 3.3.0-rc.3, but I can't do so unless we iterate on this code review.

nex3:
I'm removing this from the 3.3 milestone due to lack of responsiveness.

nex3:
Closing due to lack of responsiveness.

PR #982, opened 2013-10-22
Support use of   listen >=1.1.0  < 2.2

twalpole:
This takes the work from pr#960 and extends it so that listen >=1.1.0 and < 2.2 will work - thereby allowing those on ruby 1.9.3+ to move to newer releases of listen while still supporting 1.1.x for older ruby versions

### DIFF
--- a/lib/sass/plugin/compiler.rb
+++ b/lib/sass/plugin/compiler.rb
@@ -316,12 +316,11 @@ module Sass::Plugin
         listener.thread.join
         listener.stop # Partially work around guard/listen#146
       else
-        # rubocop:disable RescueException
         begin
           listener.start!
         rescue Interrupt
+          # Squelch Interrupt for clean exit from Listen::Listener
         end
-        # rubocop:enable RescueException
       end
     end
 
--- a/lib/sass/plugin/compiler.rb
+++ b/lib/sass/plugin/compiler.rb
@@ -283,6 +283,7 @@ module Sass::Plugin
 
       listener.start
       listener.thread.join
+      listener.stop # Partially work around guard/listen#146
     end
 
     # Non-destructively modifies \{#options} so that default values are properly set,
--- a/lib/sass/plugin/compiler.rb
+++ b/lib/sass/plugin/compiler.rb
@@ -281,13 +281,8 @@ module Sass::Plugin
       # https://github.com/nex3/sass/commit/a3031856b22bc834a5417dedecb038b7be9b9e3e
       listener.force_polling(true) if @options[:poll] || Sass::Util.windows?
 
-      # rubocop:disable RescueException
-      begin
-        listener.start!
-      rescue Exception => e
-        raise e unless e.is_a?(Interrupt)
-      end
-      # rubocop:enable RescueException
+      listener.start
+      listener.thread.join
     end
 
     # Non-destructively modifies \{#options} so that default values are properly set,
@@ -310,7 +305,7 @@ module Sass::Plugin
 
     def create_listener(*args, &block)
       require 'listen'
-      Listen::Listener.new(*args, &block)
+      Listen.to(*args, &block)
     end
 
     def remove_redundant_directories(directories)
--- a/lib/sass/plugin/compiler.rb
+++ b/lib/sass/plugin/compiler.rb
@@ -281,9 +281,7 @@ module Sass::Plugin
       # https://github.com/nex3/sass/commit/a3031856b22bc834a5417dedecb038b7be9b9e3e
       listener.force_polling(true) if @options[:poll] || Sass::Util.windows?
 
-      listener.start
-      listener.thread.join
-      listener.stop # Partially work around guard/listen#146
+      listen_to(listener)
     end
 
     # Non-destructively modifies \{#options} so that default values are properly set,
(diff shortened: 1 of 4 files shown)


nex3:
I like this approach, but I have a couple comments/suggestions.

twalpole:
@nex3  Comments taken and code updated

nex3:
A couple more comments.

Also, are you confident that the `RUBY_VERSION` trick in the gemspec does what it's intended to? Does the client's local rubygems actually run the gemspec locally, or does it deal with some sort of pre-parsed representation? Shouldn't it be fine if we have a broader dependency on listen, since listen's gemspec itself specifies the Ruby versions it's compatible with? Is rubygems smart enough to do the right thing in that case?

Ideally it would be nice to actually set up a dummy package with the dependencies we intend to publish and make sure it does the right thing on Ruby 1.8.7 and Ruby 1.9+.

twalpole:
The RUBY_VERSION trick is only used for development dependencies and I'm 99% sure it is read locally there and not a pre-parsed representation.  Its required for travis because otherwise travis tried to load celluloid when testing with 1.8.7, which means its loading listen 2+
Users of sass staying on < 1.9.3 may have to specify listen ~> 1.1.0 in their Gemfiles to have the right thing happen, which would require a readme update

twalpole:
@nex3  ok - changes made and support extended to the current released version of Listen - From my tests a note will need to go with the 3.3 release that if users are on ruby < 1.9.3 they will need to also specify listen < 2.0 in their gemfile for gem to be able to work out the correct versions

nex3:
Ugh, I'm pretty unhappy with the user experience this will present to Mac users stuck on 1.8.7. As far as I know, there's no way to hook into the gem error message that gets printed, so what they'll see is something like this:

```
> sudo gem install sass
...
Fetching: celluloid-0.15.2.gem (100%)
ERROR:  Error installing listen:
        celluloid requires Ruby version >= 1.9.2.
```

Imagine being someone who's not using Bundler and isn't familiar with Ruby (e.g. a Sass user who just wants the CLI). Where do you go from there? You need to know both that you need to install an earlier version of `listen` (_not_ `celluloid`), **and** that the specific version you need to install is `<2.0`. Then you need to try installing Sass again.

The only ways you have to figure this out are to either come from a tutorial written that knows about this restriction (of which there are currently zero), or to Google the issue and end up finding someone else complaining about it (which is very risky to rely on). I don't think we can put all our potential non-Mavericks OS X users in that position.

chriseppstein:
Sass really needs a 1-click installer that would bundle ruby and sandbox the entire gem installation.

  Chris Eppstein  
 Sent via Intro for iPhone

nex3:
> Sass really needs a 1-click installer that would bundle ruby and sandbox the entire gem installation.

Yes; if we had a reliable and well-maintained installer, I'd feel much more comfortable making this change.

twalpole:
@nex3  What about  removing the listen dependency from the Gem, and only enabling the --watch functionality if listen is separately included into the users project?  That way there could be a message explaining the issue if the user tries to use watch without having listen available, but wouldn't limit 1.9.3+ users to an old listen

twalpole:
@nex3  or I could implement the method from http://en.wikibooks.org/wiki/Ruby_Programming/RubyGems#How_to_install_different_versions_of_gems_depending_on_which_version_of_ruby_the_installee_is_using. To possibly solve the issue?

chriseppstein:
I'm wary of how that would work with bundler. 

  Chris Eppstein  
 Sent via Intro for iPhone

twalpole:
@chriseppstein Ok -- what about the option of removing the dependency and having a message if someone tries to use the watch functionality telling them they need to also include listen?

chriseppstein:
What is the core issue we're trying to solve here?

  Chris Eppstein  
 Sent via Intro for iPhone

twalpole:
Not being able to use listen 2.0+ (or any gems that require it)  on systems running ruby 1.9.3+ if also
using sass

chriseppstein:
Sure, but why do you need to use listen 2.0? Why does this conflict exist?

twalpole:
My specific reason is to use Guard 2.0+  which depends on listen 2.1+

chriseppstein:
And is the issue just that you have them all in the same gemfile? Or you use guard to compile sass instead of our watcher?

  Chris Eppstein  
 Sent via Intro for iPhone

twalpole:
Yes they are all in the same gemfile, we use guard for a number of things in development - running specs, etc, the sass compilation is handled via sass-rails and the rails asset pipeline

chriseppstein:
So to handle this scenario, instead of futzing with Sass's dependency requirements, what I'd rather do is just split sass into a sass-engine gem that is just the core of the language meant to be embedded in other frameworks and the CLI and plugin code can then stop creating dependencies that are unnecessary at best, and conflicting at worst. 

  Chris Eppstein  
 Sent via Intro for iPhone

imathis:
:+1: 

I'm currently working on writing a Sass + Jekyll plugin but Jekyll relies on `listen ~> 1.3` so I'm unable to use Sass 3.3.

julik:
I concur. I am using Sass without it's watchers and perfectly happy with it (I have full on-demand recompilation). I've upgraded my test installation to v3 for source maps upport only to find out that I have to remove Guard just to satisfy some watcher functionality I am not even using.

I'd much rather have something that does the compilation, source map and assembly parts of the equation but does not mandate an outdated dependency.

julik:
Actually it's not very clear to me why SASS has to depend on `listen` at all in 3.3 while it didn't have to in 3.2 leaving the whole FS watching business to the user. To be clear on this: currently I cannot have SASS and Guard in the same Ruby project, which is just absurd.

nex3:
Sass 3.2 embedded its own copy of listen. This caused a considerable amount of friction, since it made it difficult for people to use Sass with any version of listen other than the one that was embedded.

julik:
@nex3 I've used 3.2 with multiple projects using Guard and haven't encountered any issues, primarily since I explicitly avoid any SASS-related watches. I just run my stuff through the compiler. Obviously what @chriseppstein is proposing is the sane alternative, but how long would that take? I would not mind to just get an error message if the version of 'listen' I depend on cannot be used for one reason or another, or to be "downgraded" to on-demand compilation. I mean, how would a setup like it is now work out for someone with Guard on a Rails project for example? There clearly are two use cases here: one where SASS is run as a library, and the other where it's run standalone and they are getting mixed up.

julik:
Is there a possibility to roll up a pre-release gem for this?

PR #991, opened 2013-11-01
Source map 'version' property should be a number.

harthur:
As per the source map [spec](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit), the 'version' property of a source map should be a positive integer. Currently Sass makes it a string.

### DIFF
--- a/lib/sass/source/map.rb
+++ b/lib/sass/source/map.rb
@@ -104,7 +104,7 @@ module Sass::Source
       css_uri ||= css_path.relative_path_from(sourcemap_path.dirname).to_s
 
       result = "{\n"
-      write_json_field(result, "version", "3", true)
+      write_json_field(result, "version", 3, true)
 
       source_uri_to_id = {}
       id_to_source_uri = {}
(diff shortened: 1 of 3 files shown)


fitzgen:
+1, we had to special case mozilla/source-map to handle Sass source maps since they don't match the spec (which admittedly could be more explicit about number vs string).

Also, the `names` property should always exist on the source map, even if it is only an empty array.

chriseppstein:
@nex3 let's get this in 3.3?

nex3:
@fitzgen is there a more polished spec than [this one](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k)? There are a lot of things it doesn't make clear.

fitzgen:
@nex3 Nope, that is it. If you have specifics you'd like to see cleared up or have questions, this is the place: https://lists.mozilla.org/listinfo/dev-js-sourcemap

PR #998, opened 2013-11-04
Add --bell and --bwatch command line options

collink:
Add `--bell` and `--bwatch` (shortcut for `--bell --watch`) command line options.  --bell rings a system bell when a file is compiled.  For my use case, this is really nice because I typically have Visual Studio in one window and my browser in another, and I don't want to give up any screen real estate to the cmd.exe to keep an eye on my `sass --watch`.  With the custom framework we use, it can take up to 5 seconds sometimes to fully compile everything, and having the bell has really made my life easier in that regard.

### DIFF
--- a/lib/sass/exec.rb
+++ b/lib/sass/exec.rb
@@ -239,8 +239,8 @@ END
         end
         opts.on('--bwatch', 'Watches file or directories for changes and outputs a bell after compilation is finished.',
                             'The location of the generated CSS can be set using a colon:',
-                            "  #{@default_syntax} --watch input.#{@default_syntax}:output.css",
-                            "  #{@default_syntax} --watch input-dir:output-dir") do
+                            "  #{@default_syntax} --bwatch input.#{@default_syntax}:output.css",
+                            "  #{@default_syntax} --bwatch input-dir:output-dir") do
           @options[:watch] = true
           @options[:bell] = true
 		end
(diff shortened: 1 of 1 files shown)


chriseppstein:
:-1: I dislike that this bells after each file is compiled. Most files take a very short time, this would be quite annoying. I also dislike the extra option for watching.

As specified in http://sass-lang.com/community#Contribute it's a good idea to propose ideas to the issue tracker and get feedback on them before coding them up. if we were to accept this, it would need docs, tests, etc as specified in the contribution guide. But I would wait on doing any further work until @nex3 has weighed in.

collink:
There should be no harm in it being an option that's not on by default though.  And while I agree, yes, most SCSS files are compiled quickly, not everyone's projects are small.  It's not like I'm attacking SASS's performance or anything either, we have a ton of includes and whatnot for different packages in our framework.  I've been using this all morning today though and it's already proven itself SO much more convenient.  The way I look at it is, if you don't like it, by all means, don't enable the flag on the command line, but if it's useful to you, it's there.

collink:
I can concede on the extra option for watching.  It's definitely redundant, just a convenience thing.  I've removed that.

nex3:
I agree with @chriseppstein. This isn't something that's widely useful enough to warrant inclusion. Sass also exposes a pretty reasonable Ruby API for watching files, so it's easy to make something like this for yourself without needing it to be in the Sass core.

PR #1006, opened 2013-11-07
3 3 perf tuning

chriseppstein:
This was some low hanging fruit that gave a big help to the performance of Sass 3.3. I don't yet have metrics for wether this faster or slower than 3.2 as it involved compass which has huge changes under the hood and makes it not an apples-to-apples comparison.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -21,7 +21,7 @@ end
 
 # ----- Code Style Enforcement -----
 
-if RUBY_VERSION !~ /^(1\.8|2\.1)/ && (ENV.has_key?("RUBOCOP") && ENV["RUBOCOP"] == "true" || !ENV.has_key?("RUBOCOP"))
+if RUBY_VERSION !~ /^(1\.8|2\.1)/ && (ENV.has_key?("RUBOCOP") && ENV["RUBOCOP"] == "true" || !(ENV.has_key?("RUBOCOP") || ENV.has_key?("TEST")))
   require 'rubocop/rake_task'
   Rubocop::RakeTask.new do |t|
     t.patterns = FileList["lib/**/*"]
--- a/lib/sass/engine.rb
+++ b/lib/sass/engine.rb
@@ -145,7 +145,7 @@ module Sass
     # Designates block as mixin definition rather than CSS rules to output
     MIXIN_DEFINITION_CHAR = ?=
 
-    # uncludes named mixin declared using MIXIN_DEFINITION_CHAR
+    # Includes named mixin declared using MIXIN_DEFINITION_CHAR
     MIXIN_INCLUDE_CHAR    = ?+
 
     # The regex that matches and extracts data from
--- a/lib/sass/engine.rb
+++ b/lib/sass/engine.rb
@@ -88,8 +88,6 @@ module Sass
   #     output = sass_engine.render
   #     puts output
   class Engine
-    include Sass::Util
-
     # A line of Sass code.
     #
     # `text`: `String`
@@ -147,7 +145,7 @@ module Sass
     # Designates block as mixin definition rather than CSS rules to output
     MIXIN_DEFINITION_CHAR = ?=
 
-    # Includes named mixin declared using MIXIN_DEFINITION_CHAR
+    # uncludes named mixin declared using MIXIN_DEFINITION_CHAR
     MIXIN_INCLUDE_CHAR    = ?+
 
     # The regex that matches and extracts data from
@@ -375,7 +373,7 @@ ERR
     end
 
     def encode_and_set_charset(rendered)
-      return rendered if ruby1_8?
+      return rendered if Sass::Util.ruby1_8?
       begin
         # Try to convert the result to the original encoding,
         # but if that doesn't work fall back on UTF-8
@@ -431,7 +429,7 @@ ERR
     def check_encoding!
       return if @checked_encoding
       @checked_encoding = true
-      @template, @original_encoding = check_sass_encoding(@template) do |msg, line|
+      @template, @original_encoding = Sass::Util.check_sass_encoding(@template) do |msg, line|
         raise Sass::SyntaxError.new(msg, :line => line)
       end
     end
@@ -766,7 +764,7 @@ WARNING
           value = self.class.parse_interp(
             line.text, line.index, to_parser_offset(line.offset), :filename => @filename)
         end
-        value = with_extracted_values(value) do |str|
+        value = Sass::Util.with_extracted_values(value) do |str|
           str = str.gsub(/^#{line.comment_tab_str}/m, '')[2..-1] # get rid of // or /*
           format_comment_text(str, silent)
         end
--- a/lib/sass/script/lexer.rb
+++ b/lib/sass/script/lexer.rb
@@ -374,7 +374,7 @@ MESSAGE
         return unless str
         c = str.count("\n")
         @line += c
-        @offset = (c == 0 ? @offset + str.size : str[/\n([^\n]*)/, 1].size + 1)
+        @offset = (c == 0 ? @offset + str.size : (str.size - str.rindex("\n")) + 1)
         str
       end
 
--- a/lib/sass/script/lexer.rb
+++ b/lib/sass/script/lexer.rb
@@ -117,18 +117,28 @@ module Sass
       # while the boolean represents whether or not the string
       # is following an interpolated segment.
       STRING_REGULAR_EXPRESSIONS = {
-        [:double, false] => string_re('"', '"'),
-        [:single, false] => string_re("'", "'"),
-        [:double, true] => string_re('', '"'),
-        [:single, true] => string_re('', "'"),
-        [:uri, false] => /url\(#{W}(#{URLCHAR}*?)(#{W}\)|#\{)/,
-        [:uri, true] => /(#{URLCHAR}*?)(#{W}\)|#\{)/,
+        :double => {
+          false => string_re('"', '"'),
+          true => string_re('', '"')
+        },
+        :single => {
+          false => string_re("'", "'"),
+          true => string_re('', "'")
+        },
+        :uri => {
+          false => /url\(#{W}(#{URLCHAR}*?)(#{W}\)|#\{)/,
+          true => /(#{URLCHAR}*?)(#{W}\)|#\{)/
+        },
         # Defined in https://developer.mozilla.org/en/CSS/@-moz-document as a
         # non-standard version of http://www.w3.org/TR/css3-conditional/
-        [:url_prefix, false] => /url-prefix\(#{W}(#{URLCHAR}*?)(#{W}\)|#\{)/,
-        [:url_prefix, true] => /(#{URLCHAR}*?)(#{W}\)|#\{)/,
-        [:domain, false] => /domain\(#{W}(#{URLCHAR}*?)(#{W}\)|#\{)/,
-        [:domain, true] => /(#{URLCHAR}*?)(#{W}\)|#\{)/,
+        :url_prefix => {
+          false => /url-prefix\(#{W}(#{URLCHAR}*?)(#{W}\)|#\{)/,
+          true => /(#{URLCHAR}*?)(#{W}\)|#\{)/
+        },
+        :domain => {
+          false => /domain\(#{W}(#{URLCHAR}*?)(#{W}\)|#\{)/,
+          true => /(#{URLCHAR}*?)(#{W}\)|#\{)/
+        }
       }
 
       # @param str [String, StringScanner] The source text to lex
@@ -265,7 +275,7 @@ module Sass
       end
 
       def string(re, open)
-        return unless scan(STRING_REGULAR_EXPRESSIONS[[re, open]])
+        return unless scan(STRING_REGULAR_EXPRESSIONS[re][open])
         if @scanner[2] == '#{' # '
           @scanner.pos -= 2 # Don't actually consume the #{
           @offset -= 2
--- a/lib/sass/script/lexer.rb
+++ b/lib/sass/script/lexer.rb
@@ -374,7 +374,7 @@ MESSAGE
         return unless str
         c = str.count("\n")
         @line += c
-        @offset = (c == 0 ? @offset + str.size : (str.size - str.rindex("\n")) + 1)
+        @offset = (c == 0 ? @offset + str.size : str.size - str.rindex("\n") + 1)
         str
       end
 
--- a/lib/sass/script/parser.rb
+++ b/lib/sass/script/parser.rb
@@ -219,7 +219,7 @@ module Sass
               interp = try_ops_after_interp(#{ops.inspect}, #{name.inspect})
               return interp if interp
               return unless e = #{sub}
-              while tok = try_several_tokens(#{ops.map {|o| o.inspect}.join(', ')})
+              while tok = try_toks(#{ops.map {|o| o.inspect}.join(', ')})
                 if interp = try_op_before_interp(tok, e)
                   other_interp = try_ops_after_interp(#{ops.inspect}, #{name.inspect}, interp)
                   return interp unless other_interp
@@ -321,7 +321,7 @@ RUBY
 
       def try_ops_after_interp(ops, name, prev = nil)
         return unless @lexer.after_interpolation?
-        op = try_several_tokens(*ops)
+        op = try_toks(*ops)
         return unless op
         interp = try_op_before_interp(op, prev)
         return interp if interp
@@ -543,7 +543,7 @@ RUBY
       end
 
       def literal
-        t = try_several_tokens(:color, :bool, :null)
+        t = try_toks(:color, :bool, :null)
         return literal_node(t.value, t.source_range) if t
       end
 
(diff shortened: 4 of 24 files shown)


lib/sass/util.rb
```diff
@@ -512,28 +524,44 @@ def av_template_class(name)
     #
     # @return [Boolean]
     def windows?
-      RbConfig::CONFIG['host_os'] =~ /mswin|windows|mingw/i
+      if defined?(@windows)
```
chriseppstein: the `||=` idiom doesn't work for these ivars because they're mostly false so we have to do a `defined?` check.

lib/sass/util/normalized_map.rb
```diff
@@ -7,14 +7,11 @@ module Util
     # to the original keys that were stored. If several different values normalize
     # to the same value, whichever is stored last wins.
     require 'sass/util/ordered_hash' if ruby1_8?
-    class NormalizedMap < DelegateClass(ruby1_8? ? OrderedHash : Hash)
+    class NormalizedMap
```
chriseppstein, line 10: This was the biggest offender. Let no one ever user `DelegateClass` in any frequently called code ever again. I don't even know what ruby is doing to make this so slow.

lib/sass/tree/node.rb
```diff
@@ -30,6 +30,11 @@ module Tree
     class Node
       include Enumerable
 
+      class << self
+        # store the node name here for caching by the visitors.
+        attr_accessor :node_name
```
nex3: Can we put the logic for determining and caching the node name in here as well? In addition to keeping the logic in one place, `node.class.node_name` is a lot nicer than `Sass::Tree::Visitors::Base.node_name(node.class)`.

lib/sass/tree/visitors/check_nesting.rb
```diff
@@ -19,6 +19,16 @@ def visit(node)
     raise e
   end
 
+  # @return [String] The method name that determines if the parent is invalid.
+  def invalid_child_method_name(parent)
+    "invalid_#{Sass::Tree::Visitors::Base.node_name parent.class}_child?"
```
nex3: Should these be interpolated symbols rather than strings?

lib/sass/tree/visitors/check_nesting.rb
```diff
@@ -19,6 +19,16 @@ def visit(node)
     raise e
   end
 
+  # @return [String] The method name that determines if the parent is invalid.
+  def invalid_child_method_name(parent)
```
nex3: Would it make a difference to cache these on the node class as well? E.g. `parent.class.invalid_child_method_name`. You could even monkey-patch the caching code in in this class to keep the concerns a little more separate.

lib/sass/util.rb
```diff
@@ -1075,16 +1128,27 @@ def lcs_table(x, y)
...
+      if block_given?
+        return lcs_backtrace(c, x, y, i, j - 1) {|xx, yy| yield(xx, yy)} if c[i][j - 1] > c[i - 1][j]
+        lcs_backtrace(c, x, y, i - 1, j) {|xx, yy| yield(xx, yy)}
+      else
+        return lcs_backtrace(c, x, y, i, j - 1) if c[i][j - 1] > c[i - 1][j]
+        lcs_backtrace(c, x, y, i - 1, j)
+      end
     end
+
+    require 'sass/util/ordered_hash' if ruby1_8?
```
nex3: Move this near the definition of `#ordered_hash`.

lib/sass/util.rb
```diff
@@ -101,7 +99,11 @@ def map_keys(hash)
     # @see #map_keys
     # @see #map_hash
     def map_vals(hash)
-      to_hash(hash.map {|k, v| [k, yield(v)]})
+      rv = hash.dup
```
nex3: Add a comment here and below explaining that we're copying and modifying rather than using `#to_hash` for performance reasons.

Also, why does `#map_keys` use `#map_hash` but `#map_vals` expands out the iteration manually? Is one just called much less frequently than the other? If so, make a note of that (although it may still be worth expanding out `#map_keys` anyway in case it gets more use later).

Finally, what's the purpose of duplicating the hash rather than creating a new empty hash?

lib/sass/util.rb
```diff
@@ -244,11 +251,16 @@ def paths(arrs)
...
       y = [nil, *y]
-      block ||= proc {|a, b| a == b && a}
-      lcs_backtrace(lcs_table(x, y, &block), x, y, x.size - 1, y.size - 1, &block)
+      if block_given?
+        lcs_backtrace(lcs_table(x, y) {|xx, yy| yield(xx, yy)},
+                      x, y, x.size - 1, y.size - 1) {|xx, yy| yield(xx, yy)}
+      else
+        lcs_backtrace(lcs_table(x, y) {|a, b| a == b && a},
+                      x, y, x.size - 1, y.size - 1) {|a, b| a == b && a}
+      end
```
nex3: Add a comment explaining that we're not using `&block` for performance reasons.

lib/sass/util.rb
nex3: I prefer the following structure to cut down on the number of lines:

``` ruby
return @windows if defined?(@windows)
@windows = (RbConfig::CONFIG['host_os'] =~ /mswin|windows|mingw/i)
```

lib/sass/util.rb
```diff
@@ -512,28 +524,44 @@ def av_template_class(name)
     #
     # @return [Boolean]
     def windows?
-      RbConfig::CONFIG['host_os'] =~ /mswin|windows|mingw/i
+      if defined?(@windows)
+        @windows
+      else
+        @windows = (RbConfig::CONFIG['host_os'] =~ /mswin|windows|mingw/i)
+      end
```
nex3: Now that we're caching a bunch of stuff here, we should get rid of the last few places that still include `Sass::Util` rather than treating it as a namespace full of static method.

Also it might be nice to add a single comment before the first cached method that explains why they're cached (that is, that they're called a lot and doing string equality is relatively expensive).

lib/sass/util/normalized_map.rb
```diff
@@ -53,12 +51,42 @@ def as_stored
...
+        @map.map {|h, k| yield(h, k)}
+      end
+
+      def dup
+        d = super
+        d.send(:instance_variable_set, "@map", @map.dup)
+        d
       end
     end
   end
```
nex3, line 113: I'm worried about the following scenario:
- We write some code that takes a map and uses a method that's not defined here.
- That code is never passed a NormalizedMap in our tests, but could theoretically be passed one.
- A user stumbles onto a code path that passes a NormalizedMap to that code and they get horribly broken.

It's a little cumbersome, but I think we should explicitly forward _all_ the public methods on `Hash`, and add a test that checks the public instance methods of `NormalizedMap` against those of `Hash` so that when more are added we remember to add them as well.

lib/sass/util/normalized_map.rb
nex3, line 10: What about `MultibyteStringScanner` on Rubinius? Is DelegateClass more efficient there, or should we manually forward?

nex3:
This is great. Thanks for looking into it. I've got a few suggestions, but overall I'm really happy with this.

It does look like Travis is broken, though.

lib/sass/util.rb
chriseppstein: I would prefer that too, but when it's written on a single line the defined check always evaluates to true. 

Hunt & pecked on my iPhone... Sorry if it's brief!

chriseppstein: Ah, I see -- ya that works.

lib/sass/util/normalized_map.rb
chriseppstein, line 10: We should profile it and see if it's a performance issue for that platform. It may explain why our builds on rbx are so slow. Unfortunately, optimizing rbx is not a priority for me right now given the mountain of work I have right now. I filed #1008 to track the issue, maybe a rubinius user will submit a patch.

lib/sass/tree/node.rb
chriseppstein: I moved it and deprecated the old method.

lib/sass/tree/visitors/check_nesting.rb
chriseppstein: Yes, they should be symbols, also they should not be interpolated as that is basically the same as interning a string. Now, these are precalculated at boot up.

lib/sass/tree/visitors/check_nesting.rb
chriseppstein: Yep. It's now added to the node class when it inherits from the base node.

lib/sass/util.rb
chriseppstein: I had to move the method because `ruby1_8?` is not defined until later.

lib/sass/util.rb
chriseppstein: > what's the purpose of duplicating the hash rather than creating a new empty hash?

This allows the method to work with arbitrary instances of map-like classes without making any assumptions about what class it was passed or how instances of that class are constructed. Another option is to do `hash.class.new` and assume the class has a zero argument constructor.

> Also, why does #map_keys use #map_hash but #map_vals expands out the iteration manually?

`map_keys`, like `map_hash` allows keys to be removed and replaced with new keys, this requires a delete on the key and a re-insertion to maintain the hash-order.

Since `map_vals` preserves the key, we can be more optimal and avoid the key removal and re-insertion.

chriseppstein: I have added code comments clarifying the same as I mentioned above.

lib/sass/util.rb
chriseppstein: All the includes of `Sass::Util` have been removed and all the methods are now declared with `module_function` which makes them private and forces them to be called on the module itself.

I've added the comment you requested.

lib/sass/util.rb
chriseppstein: comment added.

lib/sass/util.rb
nex3: > This allows the method to work with arbitrary instances of map-like classes without making any assumptions about what class it was passed or how instances of that class are constructed. Another option is to do `hash.class.new` and assume the class has a zero argument constructor.

I think I'd prefer `hash.class.new`, especially since you were saying that `#dup` can be pretty expensive for large objects.

> `map_keys`, like `map_hash` allows keys to be removed and replaced with new keys, this requires a delete on the key and a re-insertion to maintain the hash-order.
> 
> Since `map_vals` preserves the key, we can be more optimal and avoid the key removal and re-insertion.

If you start with an empty hash, this won't be an issue, right?

Rakefile
```diff
@@ -21,7 +21,7 @@ end
 
 # ----- Code Style Enforcement -----
 
-if RUBY_VERSION !~ /^(1\.8|2\.1)/ && (ENV.has_key?("RUBOCOP") && ENV["RUBOCOP"] == "true" || !ENV.has_key?("RUBOCOP"))
+if RUBY_VERSION !~ /^(1\.8|2\.1)/ && (ENV.has_key?("RUBOCOP") && ENV["RUBOCOP"] == "true" || !(ENV.has_key?("RUBOCOP") || ENV.has_key?("TEST")))
```
nex3, line 24: What's this `TEST` thing?

lib/sass/engine.rb
```diff
@@ -147,7 +145,7 @@ def comment?
     # Designates block as mixin definition rather than CSS rules to output
     MIXIN_DEFINITION_CHAR = ?=
 
-    # Includes named mixin declared using MIXIN_DEFINITION_CHAR
+    # uncludes named mixin declared using MIXIN_DEFINITION_CHAR
```
nex3: I think "Includes" is correct :wink:.

lib/sass/script/lexer.rb
```diff
@@ -364,7 +374,7 @@ def scan(re)
         return unless str
         c = str.count("\n")
         @line += c
-        @offset = (c == 0 ? @offset + str.size : str[/\n([^\n]*)/, 1].size + 1)
+        @offset = (c == 0 ? @offset + str.size : (str.size - str.rindex("\n")) + 1)
```
nex3: Nit: the parentheses around `str.size - str.rindex("\n")` are unnecessary.

lib/sass/script/parser.rb
```diff
@@ -564,13 +564,26 @@ def assert_expr(name, expected = nil)
...
-      def assert_tok(*names)
-        t = try_tok(*names)
+      def assert_tok(name)
+        # Avoids an array allocation caused by argument globbing in assert_several_tokens.
+        t = try_tok(name)
+        return t if t
+        @lexer.expected!(Lexer::TOKEN_NAMES[name] || name.to_s)
+      end
+
+      def assert_several_tokens(*names)
```
nex3: I'd probably call this `assert_toks` (and `try_toks` below) for consistency with the singular method names.

lib/sass/script/value/color.rb
```diff
@@ -166,14 +176,14 @@ class << self; include Sass::Util; end
...
-        rgba
-      end
+     }, &method(:int_to_rgba))
 
     # A hash from `[red, green, blue, alpha]` value arrays to color names.
-    COLOR_NAMES_REVERSE = map_hash(hash_to_a(COLOR_NAMES)) {|k, v| [v, k]}
+    COLOR_NAMES_REVERSE = COLOR_NAMES.invert.freeze
+
+    # We add the alternate color names after inverting because
+    # different ruby implementations and versions vary on the result of invert.
```
nex3: "on the result of invert" -> "on the ordering of the result of invert"

lib/sass/script/value/color.rb
```diff
@@ -14,13 +14,31 @@ module Sass::Script::Value
   # If only the alpha channel is modified using \{#with},
   # the cached RGB and HSL values are retained.
   class Color < Base
-    class << self; include Sass::Util; end
+    # Convert a ruby integer to a rgba components
+    # @param color [Fixnum]
+    # @return [Array<Fixnum>] Array of 4 numbers representing r,g,b and alpha
+    def self.int_to_rgba(color)
```
nex3, line 22: Mark this `@private`.

lib/sass/tree/node.rb
```diff
@@ -30,6 +30,29 @@ module Tree
...
       include Enumerable
 
+      def self.inherited(base)
+        node_name = base.name.gsub(/.*::(.*?)Node$/, '\\1').downcase
+        base.instance_eval <<-METHODS
+          # @return [Symbol] The name that is used for this node when visiting.
+          def node_name
+            :#{node_name}
+          end
+          # @return [Symbol] The method that is used on the visitor to visit nodes of this type.
```
nex3, line 41: Nit: Newlines between each of these methods.

lib/sass/tree/node.rb
```diff
@@ -30,6 +30,29 @@ module Tree
     class Node
       include Enumerable
 
+      def self.inherited(base)
```
nex3, line 33: This is a nice way to do this.

lib/sass/tree/visitors/base.rb
```diff
@@ -53,15 +52,13 @@ def visit_children(parent)
...
-    NODE_NAME_RE = /.*::(.*?)Node$/
-
     # Returns the name of a node as used in the `visit_*` method.
     #
     # @param [Tree::Node] node The node.
     # @return [String] The name.
     def self.node_name(node)
-      @node_names ||= {}
-      @node_names[node.class.name] ||= node.class.name.gsub(NODE_NAME_RE, '\\1').downcase
+      Sass::Util.deprecated(self, "Call node.class.node_name instead.")
```
nex3, line 60: I don't really consider this method to be part of the public API; I'd be fine with just removing it without a deprecation process.

lib/sass/tree/visitors/perform.rb
```diff
@@ -113,8 +113,8 @@ def perform_splat(splat, performed_keywords, kwarg_splat, environment)
           args = splat.to_a
         end
       end
-
-      kwargs = kwargs.merge(performed_keywords)
+      kwargs ||= Sass::Util.ordered_hash
```
nex3, line 116: Replacing the initial assignment with this seems like an unnecessary optimization. It can't be that expensive to just create a hash that will only occasionally be unused.

lib/sass/util.rb
```diff
@@ -415,6 +398,17 @@ def abstract(obj)
...
     end
 
+    # Prints a deprecation warning for the caller method.
+    #
+    # @param obj [Object] `self`
+    # @param message [String] A message describing what to do instead.
+    def deprecated(obj, message = nil)
+      obj_class = obj.is_a?(Class) ? "#{obj}." : "#{obj.class}#"
+      Sass::Util.sass_warn <<WARNING[0...-1]
+DEPRECATION WARNING: #{obj_class}#{caller_info[2]} will be removed.#{("\n" + message) if message}
```
nex3: "will be removed" -> "will be removed in a future version of Sass."

I'm not sure how useful this will be as a utility method, since mostly what we deprecate are language features that don't fit this pattern.

nex3:
Done reviewing. Did you miss my final comment last time on `normalized_map.rb`?

Rakefile
chriseppstein, line 24: `TEST` is an environment variable that you can pass to rake to run just a single test. When a single test is being run, we shouldn't run rubocop.

lib/sass/engine.rb
chriseppstein: oops.

lib/sass/script/lexer.rb
chriseppstein: removed.

lib/sass/script/parser.rb
chriseppstein: changed.

lib/sass/script/value/color.rb
chriseppstein, line 22: marked.

lib/sass/script/value/color.rb
chriseppstein: reworded.

lib/sass/tree/node.rb
chriseppstein, line 33: :) it wasn't the first thing I tried.

lib/sass/tree/node.rb
chriseppstein, line 41: added

lib/sass/tree/visitors/base.rb
chriseppstein, line 60: It's easy enough.

lib/sass/tree/visitors/perform.rb
chriseppstein, line 116: Death by a thousand cuts.

lib/sass/util.rb
chriseppstein: changed.

Issue #1033, opened 2013-12-06
Rails.root is nil


chriseppstein:
Filed against compass initially: https://github.com/chriseppstein/compass/issues/1462

ahmeij:
I am not sure the issue is with sass, as including sass 3.3.0.rc.2 in clean new rails 3.2.x works like expected, only when adding compass 1.0.0.alpha.15 the error occurs in the rails boot process. 

As stated in the referenced issue, I'm willing to help solve this issue but I need a pointer or two as to where to start looking.

glebm:
`require "sass/plugin"` by compass causes Sass to `require "sass/plugin/rails"` and access `Rails.root` before Rails has initialized. We can move Rails-specific code to the railtie.

ahmeij:
Nice, it looks like #1039 fixes the problem indeed, thanks @glebm !

ServusJon:
it works! please do the pull request @nex3 
thanks @glebm

billsaysthis:
Can a new RC be released that includes this fix please?

chrisgriffin:
I'm having this problem, is there a workaround I can use until you cut a new RC?

billsaysthis:
Add the following to your Gemfile, before Sass-Rails:

```
gem 'sass', github: 'nex3/sass'
```

PR #1039, opened 2013-12-10
Fix for #1033

glebm:
Fixes latest Compass + Rails loading problem described in #1033.

### DIFF
--- a/lib/sass/plugin.rb
+++ b/lib/sass/plugin.rb
@@ -124,7 +124,10 @@ module Sass
   end
 end
 
-if defined?(Merb::Plugins)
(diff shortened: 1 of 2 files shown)


ServusJon:
It works beautiful!

nex3:
How far back in terms of Rails versions does this work? I believe currently Sass works back to 2.3, and I'd like to preserve that compatibility if at all possible.

glebm:
I think Railties are only available on Rails 3.0+. Rails 2.3 had reached end of life on June 25, 2013.

glebm:
Updated with Rails 2 support. Please note that I have not tested Rails 2 compatibility, hopefully the test suite covers it.

Issue #833, opened 2013-07-02
comparable()'s arguments should be named $number1 and $number2


nex3:
Every other function with numbered arguments uses the `$name1`, `$name2` format. Only `comparable()` uses `$name-1`, `$name-2`. It should be brought in line with the other functions.

This will require a deprecation process for the old names.

nex3:
Relatedly, `percentage`, `round`, `ceil`, `floor`, and `abs` should have their argument names changed from `$value` to `$number` to match the pattern of naming arguments after their types.

nex3:
`mix()` also uses `$color-1` and `$color-2`.

PR #1044, opened 2013-12-14
Fix for  #833 

JonathanTR:
Brings argument names for `comparable`, `percentage`, `round`, `ceil`, `floor`, `abs`, and `mix` in line with other functions as requested in issue #833.


nex3:
Please make this pull request against master instead of stable.

JonathanTR:
Got it - all moved over.

PR #1045, opened 2013-12-15
Fix for #Issue 833

JonathanTR:
Brings argument names for `comparable`, `percentage`, `round`, `ceil`, `floor`, `abs`, and `mix` in line with other functions as requested in issue #833.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -5,6 +5,12 @@
 
 ## 3.3.0 (Unreleased)
 
+  * mix()'s arguments now use $name1, $name2 format, in keeping with other functions
+
+  * arguments for percentage(), round(), ceil(), floor(), and abs() now use 'number' instead of 'value' for arguments
+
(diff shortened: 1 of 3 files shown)


nex3:
This is good, but unfortunately we can't just change the names and break backwards-compatibility. We'll need a deprecation period where the old names still work but produce warnings indicating that the new names should be used instead.

Right now we don't have any infrastructure for this. The best way to add it is probably to add a `:deprecated` option to `Functions.declare` and a matching field to `Functions::Signature`. Then in `Tree::Funcall#_perform`, if a deprecated signature is being used, issue a helpful warning.

JonathanTR:
Thanks for the feedback. I'm pretty new with this stuff, but I love Sass and want to contribute :+1:  I'll work on that and try to get a new pull request together. Thanks for reviewing!

nex3:
Feel free to push new commits to this pull request instead, if you want. If you do so, make sure to add a comment to GitHub notifies me of the additions.

PR #1061, opened 2014-01-04
Add Ruby 2.1.0 to .travis.yml

salimane:


felixbuenemann:
That's a good idea. Currently two specs are failing for me on 2.1.0, because of different order of css rules.

chriseppstein:
I added this in bf4c1a0861217c2fe301315ac55a912a81d9eb6f and b499fa1d0f3984fbf0e3c4ea735ad2af82b94388.

PR #1062, opened 2014-01-04
Random function with Fixes

HamptonMakes:
I read through @nex3's code comments on the random function pull request and implemented the requested changes. I also added some extra tests that weren't requested.

See original pull request here: https://github.com/nex3/sass/pull/968

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -2249,5 +2249,17 @@ WARNING
     def generator
       Sass::Script::Functions.random_number_generator
     end
+
+    unless defined?(::Random)
+      # Shim for ruby 1.8.7 support.
+      class Random
+        def initialize(seed = nil)
+          srand(seed) if seed
+        end
+        def rand(*args)
+          Kernel.rand(*args)
+        end
+      end
+    end
   end
 end
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -2189,10 +2189,10 @@ module Sass::Script
     #   Return a decimal between 0 and 1, inclusive of 0, but not 1.
     #   @return [Sass::Script::Number] A decimal value.
     # @overload random($limit)
-    #   Return an integer between 1 and $limit inclusive.
-    #   @param $limit The maximum of the random integer to be returned, a positive integer.
+    #   Return an integer between 1 and `$limit` inclusive.
+    #   @param $limit [Sass::Script::Value::Number] The maximum of the random integer to be returned, a positive integer.
     #   @return [Sass::Script::Number] An integer.
-    #   @raise [ArgumentError] if the limit is not 1 or greater
+    #   @raise [ArgumentError] if the `$limit` is not 1 or greater
     def random(limit = nil)
       if limit
         assert_integer limit, "limit"
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -401,6 +401,25 @@ module Sass::Script
       @signatures[method_name].first
     end
 
+    # Sets the random seed used by Sass's internal random number generator.
+    #
+    # This can be used to ensure consistent random number sequences which
+    # allows for consistent results when testing, etc.
+    #
+    # @param seed [Integer]
+    # @return [Integer] The same seed.
+    def self.random_seed=(seed)
+      @random_number_generator = Random.new(seed)
+      seed
+    end
+
+    # Get Sass's internal random number generator.
+    #
+    # @return [Random]
+    def self.random_number_generator
+      @random_number_generator ||= Random.new
+    end
+
     # The context in which methods in {Script::Functions} are evaluated.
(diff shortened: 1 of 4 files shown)


lib/sass/script/functions.rb
```diff
@@ -2168,6 +2185,28 @@ def inspect(value)
     end
     declare :inspect, [:value]
 
+    # @overload random()
+    #   Return a decimal between 0 and 1, inclusive of 0, but not 1.
+    #   @return [Sass::Script::Number] A decimal value.
+    # @overload random($limit)
+    #   Return an integer between 1 and $limit inclusive.
```
nex3: I think this is exclusive of `$limit` as well.

Also, `$limit` should be code-quoted.

HamptonMakes: Damnit. I misread it. Here is the documentation in Ruby. Good catch, Nathan. Sorry about that. I feel like some other documentation had confused me. Changes coming. http://www.ruby-doc.org/core-2.1.0/Random.html#method-c-rand

HamptonMakes: Oh wait. Sorry. So, the way that @chriseppstein implemented this, its actually "Random.rand($limit) + 1" So, while the ruby implementation is not inclusive, by adding one, we are making ours inclusive of the limit.

HamptonMakes:
So, code-quoted the `$limit` param (plus added its type to the rdoc).

Also, the Random function _is_ inclusive of $limit in this implementation, since its defined in Ruby as "Random.rand(limit) + 1".

lib/sass/script/functions.rb
```diff
@@ -2168,6 +2185,28 @@ def inspect(value)
...
+    # @overload random($limit)
+    #   Return an integer between 1 and `$limit` inclusive.
+    #   @param $limit [Sass::Script::Value::Number] The maximum of the random integer to be returned, a positive integer.
+    #   @return [Sass::Script::Number] An integer.
+    #   @raise [ArgumentError] if the `$limit` is not 1 or greater
+    def random(limit = nil)
+      if limit
+        assert_integer limit, "limit"
+        if limit.value < 1
+          raise ArgumentError.new("Expected the limit to be 1 or greater")
```
nex3, line 2200: For consistency, this message should be `"$limit #{limit} must be greater than or equal to 1"`.

lib/sass/script/functions.rb
```diff
@@ -2168,6 +2185,28 @@ def inspect(value)
...
+        end
+        number(1 + generator.rand(limit.value))
```
nex3, line 2202: You're right that this is inclusive of `limit`; for consistency with `random()`, it should probably be exclusive instead.

lib/sass/util/random.rb
```diff
@@ -0,0 +1,15 @@
+# In Ruby 1.9+, there is a built-in Random class, which is globally defined.
+# In case we are running on Ruby 1.8.7 (detected, by having no Random class defined),
+# this method defines a mock Random class.
+unless defined?(::Random)
+  # Shim for ruby 1.8.7 support.
+  class Random
```
nex3, line 6: Name this CrossPlatformRandom or something and put it within the Sass::Util namespace. We don't want to cause conflicts if someone else defines a top-level Random class on 1.8.

test/sass/functions_test.rb
```diff
@@ -1574,6 +1574,55 @@ def test_inspect
...
+      assert result.value >= 0, "Random number was below 0"
+      assert result.value <= 1, "Random number was above 1"
+    end
+  end
+
+  def test_random_works_with_limit
+    if Sass::Script::Functions.instance_variable_defined?("@random_number_generator")
+      Sass::Script::Functions.send(:remove_instance_variable, "@random_number_generator")
+    end
+    assert_nothing_raised do
```
nex3, line 1599: This is unnecessary, since `assert` will produce an appropriate error.

test/sass/functions_test.rb
```diff
@@ -1574,6 +1574,55 @@ def test_inspect
...
+      # Passing 1 as the limit should always return 1, since limit calls return
+      # integers from 1 to the argument, so when the argument is 1, its a predicatble
+      # outcome
+      assert "1", evaluate("random(1)")
+    end
+
+    # Limit must be 1 or greater
```
nex3, line 1606: I'd prefer these be separate tests with descriptive names.

test/sass/functions_test.rb
```diff
@@ -1574,6 +1574,55 @@ def test_inspect
...
+    assert_raises Sass::SyntaxError do
+      evaluate("random(0)")
+    end
+
+    # Limit cannot be negative
+    assert_raises Sass::SyntaxError do
+      evaluate("random(-1)")
+    end
```
nex3, line 1614: There should also be a test for a non-integer limit.

nex3:
Finished reviewing. I'd like Chris's okay on this as well before it goes in.

PR #1063, opened 2014-01-05
Fix exception on bad directive '@'

felixbuenemann:
When sass encounters a sass-file that has a single @ character, it trips up the parser:

```
undefined method `gsub' for nil:NilClass
sass-3.2.13/lib/sass/engine.rb:716:in `parse_directive'
sass-3.2.13/lib/sass/engine.rb:566:in `parse_line'
sass-3.2.13/lib/sass/engine.rb:471:in `build_tree'
sass-3.2.13/lib/sass/engine.rb:490:in `block in append_children'
sass-3.2.13/lib/sass/engine.rb:489:in `each'
sass-3.2.13/lib/sass/engine.rb:489:in `append_children'
sass-3.2.13/lib/sass/engine.rb:345:in `_to_tree'
sass-3.2.13/lib/sass/engine.rb:274:in `to_tree'
```

In this case `directive` is `nil` so it triggers the else branch, which tries to call gsub on it.

### DIFF
--- a/lib/sass/engine.rb
+++ b/lib/sass/engine.rb
@@ -712,6 +712,8 @@ WARNING
       when 'media'
         parser = Sass::SCSS::Parser.new(value, @options[:filename], @line)
         Tree::MediaNode.new(parser.parse_media_query_list.to_a)
+      when nil
+        raise SyntaxError.new("Invalid directive: '@'.")
       else
         unprefixed_directive = directive.gsub(/^-[a-z0-9]+-/i, '')
         if unprefixed_directive == 'supports'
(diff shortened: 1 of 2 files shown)


felixbuenemann:
This problem was ignored on 3.2.3, but caused above exception after upgrading to 3.2.13, so I added the test in the block for regressions. I hope that's fine.

chriseppstein:
Nice catch. Most people don't type this, so I guess it makes sense that it hasn't been hit before. I guess this would mostly be an issue with live compilers like sassmeister.

felixbuenemann:
Or bad copy'n'paste ;-)

Btw. should I rebase this on master? Just noticed it was made against stable, because that the default branch.

chriseppstein:
@felixbuenemann yes please. Master is a god place for it.

felixbuenemann:
~~That rebase looks wrong, because github didn't pick up that it's now based on master, I'll revert to previous state.~~

OK, reverted, I'll send a separate PR with the changes for master.

nex3:
I'm going to merge this into stable, since it's non-invasive and fixes a bug.

PR #1072, opened 2014-01-07
Fix specs on Ruby 2.1.0

### LINKED ISSUE #1064: Specs fail on MRI Ruby 2.1.0
```
[ 188/1376] ExtendTest#test_multiple_extends_with_multiple_extenders_and_single_target = 0.01 s
  4) Failure:
ExtendTest#test_multiple_extends_with_multiple_extenders_and_single_target [/Users/felix/dev/gems/sass/test/sass/extend_test.rb:111]:
<".foo.bar, .bar.baz, .baz.bang, .foo.bang {\n  a: b; }\n"> expected but was
<".foo.bar, .bar.baz, .bang.baz, .foo.bang {\n  a: b; }\n">.

[ 229/1376] ExtendTest#test_parent_and_sibling_extend = 0.01 s
  5) Failure:
ExtendTest#test_parent_and_sibling_extend [/Users/felix/dev/gems/sass/test/sass/extend_test.rb:1250]:
<".parent1 .parent2 .child1.child2, .parent2 .parent1 .child1.child2 {\n  c: d; }\n"> expected but was
<".parent1 .parent2 .child2.child1, .parent2 .parent1 .child2.child1 {\n  c: d; }\n">.
```

For some reason the order of the class names is not alphabetical.

felixbuenemann:
Fixes #1064. See issue for more info. TL;DR: Ruby 2.1.0 has different order for uniq, which makes some tests fail.

28a9911 is the simplest possible fix I could come up with, but which I found ugly to read. So 47b9249 implements an `assert_permutation` method to match ignoring class order.

### DIFF
--- a/test/sass/extend_test.rb
+++ b/test/sass/extend_test.rb
@@ -108,10 +108,8 @@ CSS
 .bang {@extend .bar}
 SCSS
 
-    assert_equal <<CSS, render(<<SCSS)
-.foo.bar, .bar.baz, .baz.bang, .foo.bang {
-  a: b; }
-CSS
+    assert_match /\.foo\.bar, \.bar\.baz, \.(baz\.bang|bang\.baz), \.foo\.bang {
+  a: b; }/, render(<<SCSS)
 .foo.bar {a: b}
 .baz {@extend .foo}
 .bang {@extend .bar}
@@ -1190,10 +1188,8 @@ SCSS
   end
 
   def test_parent_and_sibling_extend
-    assert_equal <<CSS, render(<<SCSS)
-.parent1 .parent2 .child1.child2, .parent2 .parent1 .child1.child2 {
-  c: d; }
-CSS
+    assert_match /\.parent1 \.parent2 \.(child1\.child2|child2\.child1), \.parent2 \.parent1 \.(child1\.child2|child2\.child1) {
+  c: d; }/, render(<<SCSS)
 %foo %bar%baz {c: d}
 
 .parent1 {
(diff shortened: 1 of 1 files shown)


felixbuenemann:
@chriseppstein Choose whichever solution you prefer and I'll squash.

The `assert_permutation` method should be generic enough to use it in other tests as well. However it is complex enough to warrant its own unit test.

chriseppstein:
I toyed around with what it would take to make the Sass AST have a semantic equality check and it got hairy pretty fast.

None of these options make me particularly happy. Ultimately, I think we will find that and AST-based comparison is the most robust way of testing CSS equality -- especially as a way to test the output from other Sass implementations where minor output differences are irrelevant to be considered "compatible". However, for the time being, I think I prefer 28a99115b9b6925e22189d3d789ca019effba191 because it's a small fix even though it's not general.

@nex3 I leave the final decision to you.

HamptonMakes:
I've been playing around with this with sass-spec branches. Hard. Hard.
Hard.

felixbuenemann:
I believe @nex3 already commented, that he would prefer to wait on the fix to ruby 2.1.0, which should land in the next minor release. I believe in that case ruby 2.1.0 should be marked as "Allowed Failures" on Travis CI for the time being.

nex3:
Yeah, I just want to wait for the Ruby 2.1.0 fix. 05fc8e2d835296c5abde788c2af7057b06212f32 marks the tests as failing on 2.1.0 for the time being.

PR #1075, opened 2014-01-07
Fix exception on bad directive '@'

felixbuenemann:
Without this fix the following exception was raised:
NoMethodError: undefined method `gsub' for nil:NilClass

This is adapted to master form PR #1063 which was made against stable.


nex3:
Duplicate of #1063.

felixbuenemann:
@nex3 Why did you close this? The fix had to be adapted to master, because the underlying code had changed significantly compared to stable.

nex3:
I plan to just manually merge it next time I do a stable-to-master merge.

Issue #839, opened 2013-07-03
Sourcemap and different prefix issue


ArmorDarks:
Whenever I'm trying to build scss file with --sourcemap, I'm always receiving following error:

```
C:/Ruby200-x64/lib/ruby/2.0.0/pathname.rb:500:in `relative_path_from': different prefix: "C:/" and "C:\\Users/Username/Desktop/sass" (ArgumentError)
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/source/map.rb:125:in `block in to_json'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/source/map.rb:120:in `each'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/source/map.rb:120:in `to_json'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/plugin/compiler.rb:318:in `update_stylesheet'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/plugin/compiler.rb:182:in `block in update_stylesheets'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/plugin/compiler.rb:179:in `each'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/plugin/compiler.rb:179:in `update_stylesheets'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/plugin.rb:81:in `update_stylesheets'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/exec.rb:462:in `watch_or_update'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/exec.rb:324:in `process_result'
    from C:/Ruby200-x64/lib/ruby/gems/2.0.0/gems/sass-3.3.0.alpha.198/lib/sass/exec.rb:41:in `parse'
[...]

PR #1084, opened 2014-01-13
Move the sourcemap filename warning.

chriseppstein:
The location of the sourcemap warning was tightly coupled to the calling
code. This made it impossible for an importer to have source files for
which there is no public_url without a warning from Sass.

This patch updates Sass so that the importer can disable the warning by
simply returning nil from public_url without calling
`warn_about_public_url(uri)` which is now provided by the
`Importer::Base` class.


lib/sass/engine.rb
```diff
@@ -346,16 +346,24 @@ def _dependencies(seen, engines)
...
-        !(@options[:importer].public_url(@options[:filename]) ||
-          @options[:importer].is_a?(Sass::Importers::Filesystem))
+      elsif importer.nil?
         raise Sass::SyntaxError.new(<<ERR)
-Error generating source map: couldn't determine public URL for "#{@options[:filename]}".
+Error generating source map: couldn't determine public URL for "#{filename}".
+  Without a public URL, there's nothing for the source map to link to.
+  An importer was not set for this file.
+ERR
+      elsif Sass::Util.silence_warnings {!importer.public_url(filename, sourcemap_filename)}
```
nex3: Style nit: I prefer `.nil?` to `!` here.

lib/sass/engine.rb
```diff
@@ -346,16 +346,24 @@ def _dependencies(seen, engines)
     private
 
     def _render_with_sourcemap(sourcemap_uri)
-      if @options[:filename].nil?
+      filename = @options[:filename]
+      importer = @options[:importer]
+      sourcemap_filename = @options[:sourcemap_filename] &&
```
nex3: Since this variable stores the directory containing the sourcemap, call it `sourcemap_dir`.

lib/sass/importers/base.rb
```diff
@@ -129,8 +129,12 @@ def key(uri, options)
       # from this importer.
       #
       # @param uri [String] A URI known to be valid for this importer.
+      # @param relative_to_directory [String, NilClass] A directory (on disk) the url
+      #   should be made relative to, if possible. This directory may not be
+      #   provided.
```
nex3: This documentation is confusing, since in general it doesn't make sense for a server URL to be relative to a directory on disk. I suggest calling it `sourcemap_directory` and documenting it as such:

> The directory on disk where the sourcemap will be saved. If `uri` refers to a file on disk that's accessible relative to `sourcemap_directory`, this may return a relative URL.

lib/sass/importers/base.rb
```diff
@@ -129,8 +129,12 @@ def key(uri, options)
...
       # @return [String?] The publicly-visible URL for this file.
```
nex3: ", or `nil` indicating that no publicly-visible URL exists."

lib/sass/importers/base.rb
```diff
@@ -129,8 +129,12 @@ def key(uri, options)
...
-      def public_url(uri)
+      def public_url(uri, relative_to_directory = nil)
+        warn_about_public_url(uri)
```
nex3: I'm not a big fan of using a protected method where inheritance will do. I'd rather see the current contents of `warn_about_public_url` become the default contents of `public_url`, and have subclasses call `super` if they want to emit a warning.

lib/sass/importers/base.rb
```diff
@@ -165,6 +169,25 @@ def directories_to_watch
...
       end
+
+      protected
+
+      # Issues a warning about being unable to determine a public url.
+      #
+      # @param uri [String] A URI known to be valid for this importer.
+      # @return [NilClass] nil
+      def warn_about_public_url(uri)
+        @warnings_issued ||= Set.new
```
nex3: What's the benefit of issuing a separate warning for each URI as opposed to a warning for each importer? The issue that needs to be addressed is that the importer doesn't define its own `#public_url`.

lib/sass/importers/filesystem.rb
```diff
@@ -66,6 +66,18 @@ def watched_file?(filename)
           filename.start_with?(root + File::SEPARATOR)
       end
 
+      def public_url(name, relative_to_directory = nil)
+        if relative_to_directory.nil?
+          warn_about_public_url(name)
```
nex3, line 70: I don't think the filesystem importer should warn here. I know it did prior to this patch, but it doesn't make sense to issue a warning about how a built-in importer doesn't implement a method that it actually does implement.

It should probably return an absolute `file:` URI here.

lib/sass/importers/filesystem.rb
```diff
@@ -66,6 +66,18 @@ def watched_file?(filename)
...
+        else
+          file_path = Pathname.new(@root).join(Pathname.new(remove_root(name.to_s)))
+          file_path.relative_path_from(Pathname.new(relative_to_directory.to_s)).to_s
+        end
+      rescue ArgumentError
```
nex3: This wraps way more code than it needs to. It should wrap _only_ the call to `#relative_path_from`. All other computation should happen outside the `begin`/`rescue` block.

Also, add a comment explaining what causes an ArgumentError.

lib/sass/importers/filesystem.rb
```diff
@@ -66,6 +66,18 @@ def watched_file?(filename)
...
+        warn_about_public_url(name)
```
nex3: As above, this also shouldn't warn.

lib/sass/importers/filesystem.rb
```diff
@@ -66,6 +66,18 @@ def watched_file?(filename)
           filename.start_with?(root + File::SEPARATOR)
       end
 
+      def public_url(name, relative_to_directory = nil)
+        if relative_to_directory.nil?
+          warn_about_public_url(name)
+        else
+          file_path = Pathname.new(@root).join(Pathname.new(remove_root(name.to_s)))
+          file_path.relative_path_from(Pathname.new(relative_to_directory.to_s)).to_s
```
nex3: `Pathname#relative_path_from` will break if one path is absolute and the other is relative. Are we guaranteeing that `relative_to_directory` is always absolute? If so, document it. If not, make it absolute here.

Also, calling `to_s` on `relative_to_directory` here and `name` above is unnecessary.

lib/sass/importers/filesystem.rb
```diff
@@ -66,6 +66,18 @@ def watched_file?(filename)
           filename.start_with?(root + File::SEPARATOR)
       end
 
+      def public_url(name, relative_to_directory = nil)
+        if relative_to_directory.nil?
+          warn_about_public_url(name)
+        else
+          file_path = Pathname.new(@root).join(Pathname.new(remove_root(name.to_s)))
```
nex3: Just call `File.absolute_path(name)`.

test/sass/importer_test.rb
```diff
@@ -213,10 +225,58 @@ def test_source_map_with_only_css_uri_supports_public_url_imports
 JSON
   end
 
+  def test_source_map_with_only_css_uri_can_have_no_public_url_without_warning
+    $TEST = true
```
nex3: Left over from debugging?

test/sass/scss/scss_test.rb
```diff
@@ -1650,42 +1650,6 @@ def test_function_triple_keyword_splat_merge
...
-  @return $str;
-}
-
-.foo {
-  $map: (a-1: 1, b_2: 2, c-3: 3, d_4: 4);
-  val: foo($map...);
-}
-SCSS
-  end
-
```
nex3, line 1838: Why is this deleted?

test/test_helper.rb
```diff
@@ -37,6 +38,15 @@ def filename_for_test(syntax = :sass)
     "#{test_name}_inline.#{syntax}"
   end
 
+  def sourcemap_filename_for_test(syntax = :sass)
+    test_name = caller.
+      map {|c| Sass::Util.caller_info(c)[2]}.
+      compact.
+      map {|c| c.sub(/^(block|rescue) in /, '')}.
+      find {|c| c =~ /^test_/}
```
nex3: Abstract out this `test_name` computation.

test/test_helper.rb
```diff
@@ -37,6 +38,15 @@ def filename_for_test(syntax = :sass)
...
+    "#{test_name}.css.map"
```
nex3: Add "_inline" before ".css".

nex3:
Done reviewing.

test/sass/scss/scss_test.rb
chriseppstein, line 1838: These same tests were in the code twice.

lib/sass/importers/filesystem.rb
nex3: One more thing: we can fix #839 by calling `Pathname#cleanpath` on both paths here before calling `#relative_path_from`.

lib/sass/engine.rb
chriseppstein: Changed.

lib/sass/engine.rb
chriseppstein: Changed.

lib/sass/importers/base.rb
chriseppstein: There are two issues at play. The first is that the importer may not have defined what a public_url means. The second issue is that even though the semantics of a public_url for filesystems is fairly well defined, it's possible for there to be no way to construct the public url for the filesystem paths in use. E.g. If the sourcemap is being generated into a different disk (on windows) than the stylesheet. In this case, the per-uri warning makes sense to me. I've separated this into two warnings. There's the default warning about implementing the method, but there's also a warning in the filesystem importer.

lib/sass/importers/base.rb
chriseppstein: `warn_about_public_url` is moved to the filesystem importer now.

lib/sass/importers/filesystem.rb
chriseppstein, line 70: >  It should probably return an absolute `file:` URI here.

This should be done in a different pull request. It's unrelated to this change.

lib/sass/importers/base.rb
chriseppstein: Changed.

lib/sass/importers/base.rb
chriseppstein: Changed.

lib/sass/importers/filesystem.rb
chriseppstein: I don't understand how that's useful in this context.

lib/sass/importers/filesystem.rb
chriseppstein: Yes. It's always absolute. I've documented it and refactored this method and added cleanpath calls.

lib/sass/importers/filesystem.rb
chriseppstein: Moved. Comment added.

lib/sass/importers/filesystem.rb
chriseppstein: It should. As noted above.

test/sass/importer_test.rb
chriseppstein: Yep. Removed.

test/test_helper.rb
chriseppstein: added.

test/test_helper.rb
chriseppstein: refactored.

lib/sass/importers/filesystem.rb
nex3: `Pathname.new(@root).join(Pathname.new(remove_root(name.to_s)))` is effectively making `name` absolute, but it's doing so in a very roundabout way that runs the risk of failing if `name`'s prefix isn't textually identical to `@root`'s.

lib/sass/importers/base.rb
```diff
@@ -125,8 +125,18 @@ def key(uri, options)
       # from this importer.
       #
       # @param uri [String] A URI known to be valid for this importer.
-      # @return [String?] The publicly-visible URL for this file.
-      def public_url(uri)
+      # @param sourcemap_directory [String, NilClass] The absolute path to a
+      #   directory on disk where the sourcemap will be saved. If uri refers to
+      #   a file on disk that's accessible relative to sourcemap_directory, this
+      #   may return a relative URL.
```
nex3: Add "This may be `nil` if the sourcemap's eventual location is unknown."

lib/sass/importers/base.rb
```diff
@@ -125,8 +125,18 @@ def key(uri, options)
...
+      # @return [String?] The publicly-visible URL for this file, or `nil`
+      #   indicating that no publicly-visible URL exists.
+      def public_url(uri, sourcemap_directory = nil)
+        return if @public_url_warning_issued
+        @public_url_warning_issued = true
+        Sass::Util.sass_warn <<WARNING
```
nex3, line 138: It looks like there are no tests that check for this warning. We should add one.

lib/sass/importers/base.rb
nex3: Sounds good to me.

nex3:
Done reviewing. There are also a couple replies in the inline comments from the last iteration.

lib/sass/importers/base.rb
chriseppstein, line 138: That would cause an error instead of a warning.

lib/sass/importers/base.rb
chriseppstein: Added.

chriseppstein:
Squashed and merged to master as f9a615c1e3ab0758976b438b3244487b3249c4d7.

PR #1146, opened 2014-03-05
Attempt to Fix all warnings from Sass.

vipulnsward:
Hi all, This is an attempt to fix different kinds of warnings form 
- Circular dependency loading
- Un- initialized variable.
- Use of File.exists? instead of File.exist?
- Parenthesis around ambiguous arguments, etc.

Not sure what the policy is about this. Just an attempt to clear out some warnings I see on Sinatra builds. (https://travis-ci.org/sinatra/sinatra/jobs/19503469)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -19,6 +19,7 @@ Rake::TestTask.new do |t|
   test_files.exclude(scope('test/plugins/*'))
   t.test_files = test_files
   t.verbose = true
+  t.warning = true
 end
 
 # ----- Code Style Enforcement -----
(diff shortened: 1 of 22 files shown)


nex3:
We're not interested in avoiding warnings caused by uninitialized instance variables. We consider instance variables defaulting to `nil` to be a feature of Ruby, and use it accordingly.

If you remove the instance-variable changes, I'll merge this.

vipulnsward:
Actually it just returns nil value, doesn't default to setting/initializing it.
https://github.com/ruby/ruby/commit/b996367eb50402cfb343e671a2ea2cf1450a8ec4

Do you wan't to me to remove the initializations to nil, with such a behaviour from Ruby?
Let me know, I'l update the PR.

nex3:
Yes, please do remove the `nil` initializations.

vipulnsward:
I see that all changes are already in master.

PR #1161, opened 2014-03-10
Updated install description to Debian-like

andrebian:


nex3:
I'm uncomfortable including installation instructions for a channel that's not one we publish directly to, nor one that necessarily exposes the exact code we publish.

andrebian:
Sorry, I sent 'cause have problems to install on Ubuntu 13.10 in my work place. Tomorrow I'll put a error snapshot that suggest the installation via apt-get. In my home PC I didn't have any problem and was installed typing sudo gem install sass with no errors.

PR #1171, opened 2014-03-14
Fix marshaling of mixin def node with variable arguments

spohlenz:
After seeing this issue for a long time and originally thinking it was an issue in Sprockets (sstephenson/sprockets#536), I've finally narrowed this down to an issue within Sass.

The problem occurs when defining a mixin with a splat argument, and when the engine options are unmarshalable (which is the case with Sprockets). e.g.

``` sass
=kaboom($args...)
  color: red
```

which gives an error like:

```
Warning. Error encountered while saving cache 5eeafa3aa0f7f3a490743b762a56c00651a72d7e/test.css.sassc: can't dump anonymous class #<Class:0x007fcacb3ef150>
```

Curiously this does not occur when using SCSS syntax. I'm not sure why.

### DIFF
--- a/lib/sass/tree/visitors/set_options.rb
+++ b/lib/sass/tree/visitors/set_options.rb
@@ -63,6 +63,7 @@ class Sass::Tree::Visitors::SetOptions < Sass::Tree::Visitors::Base
       k.options = @options
       v.options = @options if v
     end
+    node.splat.options = @options if node.splat
     yield
   end
 
--- a/test/sass/cache_test.rb
+++ b/test/sass/cache_test.rb
@@ -73,10 +73,22 @@ class CacheTest < Test::Unit::TestCase
   end
 
   def test_cache_node_with_unmarshalable_option
-    engine = Sass::Engine.new("foo {a: b + c}",
-      :syntax => :scss, :object => Unmarshalable.new, :filename => 'file.scss',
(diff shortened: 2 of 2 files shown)


smudge:
Hey, sorry to jump on this thread, and I appreciate the work done on this issue so far.

I upgraded to 3.3.3 and I'm still seeing: `Warning. Error encountered while saving cache /path/to/file.css.scssc: can't dump anonymous class #<Class:0xasdf>`

Here's everything in my gemfile that might be relevant:

```
gem 'rails', '~> 3.2.17'
gem 'sass-rails'
gem 'sass', '~> 3.3.3'
gem 'compass-rails'
gem 'compass', '~> 1.0.0.alpha.19 '
gem 'susy', "~> 2.1.1" # the reason for using sass 3.3.3 and compass 1.0.0 alpha
```

Is there an obvious workaround I'm missing? If not, how can I further isolate the issue I'm seeing?

nex3:
There are still issues within Sprockets that are causing marshaling errors, although I believe they will be cleared up soon.

smudge:
@nex3 Cool, and thanks. (I assume you refer to sstephenson/sprockets#541? Any other changes I should expect to wait for?)

nex3:
Just that one, I believe.

PR #1178, opened 2014-03-18
Allow custom script functions to be passed to an Engine

josh:
Currently, the suggested and only api to extend Sass script functions are to monkey patch the `Sass::Script::Functions` module.

This can often lead to some plugin functions conflicting with others. Multiple `Sass::Engine`s maybe running in the same Ruby environments with different purposes. Its common to use Sass just as a css minifer since it does a good job. In this use case, you might not want any other functions leaking into the environment.

The idea is that you could pass in a module thats only extended on the `Engine`'s environment rather than globally.

``` ruby
module MyFunctions
  def my_url(url)
  end
end

Engine.new(css, { :functions => MyFunctions })
```

I'm not attached to the api or option name. Just setting up a straw man. 

cc @nex3 @rafaelfranca

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -458,6 +458,7 @@ module Sass::Script
       def initialize(environment)
         @environment = environment
         @options = environment.options
+        extend @options[:functions] if @options[:functions]
       end
 
       # Asserts that the type of a given SassScript value
--- a/test/sass/engine_test.rb
+++ b/test/sass/engine_test.rb
@@ -27,6 +27,12 @@ module Sass::Script::Functions::UserFunctions
   end
 end
 
+module Sass::Script::Functions::MyFunctions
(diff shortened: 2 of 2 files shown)


nex3:
I've got this working on [my own branch](/nex3/sass/tree/engine-custom-functions), but it's missing an important piece. Custom functions that are defined directly on `Sass::Script::Functions` have access to the `declare` method, which is important for making named arguments work. We need to find an easy way to expose this to modules passed in via `:functions` before this can be merged in.

josh:
> Custom functions that are defined directly on Sass::Script::Functions have access to the declare method

Good point.

I guess the custom functions module would have to `extend Sass::Script::Something` to get a `declare` and `signature` helper.

nex3:
`signature` isn't necessary; that's called by the Sass internals, not the functions themselves.

PR #1179, opened 2014-03-19
Don't freeze nil.

BMorearty:
In Ruby, the nil singleton object is not frozen. Freezing it breaks some other gems. When `Sass::Script::Value::Null` calls `new(nil)`, the Base class ends up freezing nil.

```
~$ irb
2.1.1 :001 > nil.frozen?
 => false
```

Conveniently, the `unless` syntax leaves `@value` as good ol' nil, so there's no need for a ternary. :honeybee: 

Tests pass.


BMorearty:
P.S. Please also pull this up to master.

SamSaffron:
+1

PR #1198, opened 2014-04-01
corrected a small misprint in opacify/transparentize description

everyonesdesign:
With this commit I corrected a small typo in the sass reference: "rgba(255, 0, 0, 0.5)" opacified by .3 equals "rgba(255, 0, 0, 0.8)", not "rgba(255, 0, 0, 0.9)".


nex3:
Thanks for the fix!

PR #1204, opened 2014-04-05
Fix incorrect documentation of map-has-key

alyssais:
`map-has-key` was previously documented as taking a single parameter, `$key`. Since this is obviously not the case, I updated the documentation to show `map-has-key` taking two parameters, `$map` and `$key`.


nex3:
Thanks!

Issue #1094, opened 2014-01-30
Add a module system


JulienCabanes:
- [x] [Proposal](https://github.com/sass/language/blob/master/accepted/module-system.md)
- [x] [Specs](https://github.com/sass/sass-spec/tree/feature.use)
- [x] [Documentation](https://github.com/sass/sass-site/issues/263)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/554)
- [ ] [LibSass](https://github.com/sass/libsass/issues/2807)

See also [the module system project](https://github.com/orgs/sass/projects/3).

---

This is a compilation of issues and wishes about @import
### Import CSS files #556

``` scss
@import "normalize";
```

Would import sass, scss or **css** file

> There's a [plugin](https://github.com/chriseppstein/sass-css-importer) for that
### Import Once #139

``` scss
@import-once "foo";
@import "foo" !once;
@depend "foo";
```

Multiple syntax suggestions for one time import

> There is a [plugin](https://github.com/chriseppstein/compass/tree/master/import-once) for that
### Import Directory #690

`@import "foo";`

If "foo" is a folder than it tries to import "foo/_module.scss"

> reasonable
### Namespaces and Aliases #353

``` scss
@import "foo" as "bar";
```

This would put every "foo" components (mixins, vars, placeholders...) in a "bar" package.

What would be the separator ? `.` or `/` or `\` or `::` ?
Where would go the `$` and `%` ? Before or after the namespace ?
What it would look like :

``` scss
// Mixins
@include bar.baz();
or
@include bar/baz();

// Variables
color: $bar.baz;
or
color: bar.$baz;
or
color: $bar/baz;
or
color: bar/$baz;

// Placeholders
div { @extend %bar.baz; }
or
div { @extend bar.%baz; }
or
div { @extend %bar/baz; }
or
div { @extend bar/%baz; }

// Functions
color: bar.baz();
or
color: bar/baz();
```
### Non-transitive imports [#353](https://github.com/nex3/sass/issues/353#issuecomment-5146513)

> if A imports B and B imports C, A shouldn't necessarily see everything from C.
> A way of getting around non-transitive imports.
### Renaming classes [#353](https://github.com/nex3/sass/issues/353#issuecomment-5146513)

> This isn't something I'm 100% sold on, but it has been requested. Maybe the correct way to do this is some combination of namespacing and @extend.
### Disabling CSS output [#353](https://github.com/nex3/sass/issues/353#issuecomment-5146513)

[...]

Issue #451, opened 2012-07-18
Allow @import within @if


AriaMinaei:
In one of the recent versions, the ability to have `@import` within `@if` has been deprecated. I looked through the changelog. Couldn't find anything mentioning that change,

Anyway, `@import` within `@if` gives the dev a lot of flexibility.

**Example:**

```
// File: latin.scss
$lang: 'latin';
@imoprt "./master";

// File: persian.scss
#lang: 'persian';
@import "./master";

// File: _master.scss
@if $lang == 'persian' {
   @impot "fonts/arabic";
}
```

**Explanation:**
Simply, it imports some arabic fonts only in persian.css.
I actually use this feature in almost all of my designs, since they are all multi-language, multi-directional designs. Deprecating it simply breaks all of those styles.

I will be downgrading to an earlier version for now.

AriaMinaei:
3.1.7 works fine, change came in 3.1.8. It's not documented in the [changelog](http://sass-lang.com/docs/yardoc/file.SASS_CHANGELOG.html#318).

chriseppstein:
Given this file:

``` scss
@if true {
  @import "foo";
}
```

When I compile it using Sass 3.1.7 I get the following error:

```
$ sass _3.1.7_ --scss nested_import.scss
Syntax error: Import directives may not be used within control directives or mixins.
        on line 2 of nested_import.scss
```

There was a bug where an `@import` that was nested conditionally in a file that was itself imported didn't get checked properly for this condition. I believe that bug was fixed in `3.1.8`.

It was never intended that `@import` would work in a conditional context, this makes it impossible for us to build a dependency tree for recompilation without fully executing the file -- which would be simply terrible for performance.

AriaMinaei:
Maybe have a flag in the config file to enable/disable this restriction?
It's too big of a feature to not have.

chriseppstein:
It's not a feature that ever existed. Users who do this would encounter many strange bugs. If you like, please open a new issue and we can discuss adding this feature in a future release.

AriaMinaei:
Well I was using this for at least 9 months and didn't encounter anything
strange.

Thanks, I will open a ticket.

AriaMinaei:
I have to add:

Just checked the compiled css files. Turns out all the fonts had been included for all languages. Such a huge bug in my code :))

jslegers:
See also :
- https://github.com/sass/sass/issues/115
- https://github.com/sass/sass/issues/739
[...]

Issue #1218, opened 2014-04-17
Intelligent SASS imports


drewwells:
Modern web development is much more than translating design to code.  We must consider how different pages inherit code, when to override styles vs fork code, managing a/b tests and managing file size and build time considerations.

In doing so, we have hit a major hurdle with SASS.  Something I'm finding easily supported by [less](http://lesscss.org/features/#variables-feature-import-statements).

In order to continue using SASS, our developers find themselves copying large portions to dance around the inflexibility in the import/build philosophy of SASS.  Here's an example.

Home Page is made of many sass partials.

Home
- Base
  - sprite
  - font
  - layout
- Modules
  - Section
  - Widget
- Header
  - Search bar
  - Header links

Let's say I run a test that changes the styling of the Search bar.  In an ideal world my Header partial would look something like this:

```
@import $search_bar; //Default values when variable is undefined would be even better
```

In the test, I would set $search_bar to `"_new_search_bar"`.  Due to the static importing this is not possible.  I must fork `Home` make a copy of `Header` just to change the one line `@import "_new_search_bar"`.

The process of copying and pasting SASS is proving unsustainable as an organization.  We require more flexibility. #1094 Paths is a nice step forward, but inevitably we need more fine grained control of what partials are loaded in a target.

Earlier today, we brainstormed the ideal solution for this.  It hinges on dynamic imports.

control.scss

``` sass
@import "paths";
@import "header"
```

_paths.scss

``` sass
$search_bar: "_search_bar";
```

_header.scss

``` sass
@import $search_bar;
```

test.scss

``` sass
@import "paths";
@import "test_overrides";
@import "base";
@import "header";
```

_test_overrides.scss

``` sass
$search_bar: "_new_search_bar";
```

By making every import a variable, the developer can pick and choose what files to replace and what styles to override.  This is something we are deeply interested in and would like to know how you plan to facilitate this use case.

nex3:
Duplicate of #739.

At least for now, I suggest you put the contents of the files that you want to dynamically import into mixins, and dynamically include the mixins instead.

Issue #1155, opened 2014-03-09
Project configuration for Sass


chriseppstein:
I'd like to make a simple configuration system for running sass with the same options for a given project.

My proposal is to simply allow command line options to be specified in a hidden file called `.sass` in the current working directory.

```
GLOBAL: [options for all core commands]
--update,--watch: [options for update or watch] 
--update: [options for update] 
--watch: [options for watch]
```

So for example:

```
GLOBAL: --compass -I ../my_library_of_sass -r ./some_custom_stuff
--update,--watch: from_dir:to_dir
--update: -t compressed
--watch: -t expanded  --sourcemap
```

Then simple user commands can result in more complex, repeatable commands:

| Typed | Command ran |
| --- | --- |
| `sass --watch` | `sass --compass -I ../my_library_of_sass -r ./some_custom_stuff --watch from_dir:to_dir -t expanded  --sourcemap` |
| `sass --update` | `sass --compass -I ../my_library_of_sass -r ./some_custom_stuff --update from_dir:to_dir -t compressed` |
| `sass -i` | `sass -i --compass -I ../my_library_of_sass -r ./some_custom_stuff` |

Any additional options typed by the user when issuing the command would simply go at the end. We could also create an option to ignore the `.sass` file.

Anyways, this is a strawman. I'd like to see a way to establish reproducible sass compiles to enable team collaboration even if they don't use compass. If there are other ideas for how to do this, let's discuss them. However, I'd like to avoid making this a ruby file.

mirisuzanne:
I love this idea, but think it could get even more clear. If you are building a config file, why limit yourself to command-line style syntax? The Compass config file handles similar settings, and is much more readable. Names like "sass_dir" and "preferred_syntax" tell me more than "-r" and "-t".

Snugug:
Having recently started to need to build config files myself, I'm with @ericam here. Visible > Hidden, Human Readable > Machine Readable. I've recently found that users really like `.yml` files; they're easy to read, easy to extend, and super human readable. We could have something like a `sass.yml` file available at the file root like `config.rb` or Compass, and then other systems (including Compass) could extend `sass.yml` to add their options. Something like the following:

``` yml
# General Sass options
sass:
  watch: true
  output style: expanded
  sourcemap: true
  line numbers: true

# Input/Output Directories
directories:
  sass: sass
  stylesheets: css
```

[...]

PR #1220, opened 2014-04-17
Added @path annontation to perform path mapping.

emphaticsunshine:
See: #1094, #451
- Made changes in perform.rb to visit path directives
  to map files.
- Made changes in import.rb to replace the mapped file.

### DIFF
--- a/lib/sass/tree/import_node.rb
+++ b/lib/sass/tree/import_node.rb
@@ -47,10 +47,10 @@ module Sass
 
       def import
         paths = @options[:load_paths]
-         
+        immutable_path = self.immutable_path
         if @options[:importer]
-          if(self.immutable_path[imported_filename])
-            @imported_filename = self.immutable_path[imported_filename]
+          if  immutable_path[imported_filename]
+            @imported_filename = immutable_path[imported_filename]
           end
           f = @options[:importer].find_relative(
             @imported_filename, @options[:filename], options_for_importer)
--- a/lib/sass/tree/import_node.rb
+++ b/lib/sass/tree/import_node.rb
@@ -12,6 +12,8 @@ module Sass
       # Sets the imported file.
       attr_writer :imported_file
 
+      attr_accessor :immutable_path
+
       # @param imported_filename [String] The name of the imported file
       def initialize(imported_filename)
         @imported_filename = imported_filename
@@ -31,7 +33,9 @@ module Sass
       # Returns whether or not this import should emit a CSS @import declaration
       #
       # @return [Boolean] Whether or not this is a simple CSS @import declaration.
-      def css_import?
+      def css_import?(immutable_path = {})
+        self.immutable_path = immutable_path
+        self.immutable_path = {} if self.immutable_path.nil?
         if @imported_filename =~ /\.css$/
           @imported_filename
         elsif imported_file.is_a?(String) && imported_file =~ /\.css$/
@@ -43,8 +47,11 @@ module Sass
 
       def import
         paths = @options[:load_paths]
-
+         
         if @options[:importer]
+          if(self.immutable_path[imported_filename])
+            @imported_filename = self.immutable_path[imported_filename]
+          end
           f = @options[:importer].find_relative(
             @imported_filename, @options[:filename], options_for_importer)
           return f if f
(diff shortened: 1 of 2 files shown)


emphaticsunshine:
See the following gist for implementation.
https://gist.github.com/emphaticsunshine/10959128

emphaticsunshine:
@nex3 or @chriseppstein can you please give it a look? This is very important feature for the company I work for. This feature enables us to do some A/B testing in sensible manner.

nex3:
In general when adding a new feature, it's a good idea to communicate us to make sure the feature is one we want and that we approve of its design. Otherwise you can end up spending effort writing code that will go unused.

In this case, it's difficult to tell exactly what the feature is doing, since there's no documentation in the patch, but I'm pretty sure it's not something we're interested in supporting. It has weird global effects, it subverts the importer system, and in general adding load paths from within a stylesheet isn't a good practice.

drewwells:
@nex3 rather than abruptly closing this PR, please describe "weird global effects" and other concerns you have.  The paths feature is explicitly requested in import_v2 wishlist and something that the community finds value in, but you seem to disagree here.  Paths is a basic feature and critical to how other static file tools use ie. RequireJS.

If you can describe your thoughts on this matter, we can find a solution to implementing this feature.

#1094 

### Paths

A way to setup shortcut paths (like Require.js)

// No syntax intended, only the purpose

```
@paths (
    "bootstrap": "../components/bootstrap-sass/vendor/assets/stylesheets/bootstrap"
);

@import "bootstrap";
```

nex3:
> If you can describe your thoughts on this matter, we can find a solution to implementing this feature.

I dislike including configuration in source files in general for a number of reasons. It violates separation of concerns by mixing configuration with styling, it hurts reusability by encoding the author's preferences indelibly in the unit of reuse, it adds a considerable amount of complexity to the language that would otherwise be isolated in the external configuration, and it requires that global options be specified locally to a single stylesheet.

I dislike encoding load paths in source files in particular because paths are usually system-specific, so it adds a surface where it's easy to make a stylesheet brittle and nonportable. In addition, the behavior isn't well-defined in terms of Sass's `Importer` abstraction; every load path is encoded as a separate `Importer` object, and these objects may not necessarily correspond to the filesystem. There's not a simple mapping between strings and `Importer`s, so it's not clear how this would work with non-filesystem importers.

chriseppstein:
I don't understand why Sass's existing importer abstraction and load path mechanism is not sufficient for the use case.

RE abruptly closing this PR, the [contribution guidelines](http://sass-lang.com/community#Contribute) are very clear:

> New feature requests should be made on the issue tracker and discussed with the core team before implementation begins to avoid wasted effort.

Once a feature has been vetted and agreed to as a proposal, then we're happy to entertain patches that implement that feature.

nex3:
> I don't understand why Sass's existing importer abstraction and load path mechanism is not sufficient for the use case.

Currently the language itself (as opposed to the command-line interface or the `Sass::Engine` API) is entirely importer-agnostic. I think that's a good property that's worth preserving, but it means that we can't just always instantiate a FilesystemImporter when someone says "add this string as a load path". I think the place for that sort of thing is external to the language, likely in the configuration file proposed in #1155.

chriseppstein:
@nex3 My comment was directed to @drewwells and @emphaticsunshine. I don't understand why the features in Sass 3.3 are not sufficient for them.

drewwells:
@chriseppstein, what feature?

> Imports may contain #{} interpolation, but only with certain restrictions. It’s not possible to dynamically import a Sass file based on a variable; interpolation is only for CSS imports. As such, it only works with url() imports. For example:

Issues have been filed #1218 , but marked as duplicate #739, a ticket with no obvious objective.

chriseppstein:
@drewwells In that ticket, @nex3 explained that you can define a mixin that implements the branching logic that you need.

#739 is a generic proposal to allow interpolation within and conditionals around @import statements.

drewwells:
You're saying we should put our entire codebase in mixins?  If I told my
team that is the solution offered by SASS, they would throw me in the mad
house.

chriseppstein:
Only the parts that have to vary for your test. Other solutions also exist. For instance, you could define the features using placeholder selectors and only extend them with concrete classes when you know you're going to need them.

drewwells:
I agree those are good one off solutions but we do a lot of development within tests.  We also expect to write code that can easily be folded into the main once complete.  Mixins and placeholders aren't practical for entire page redesigns.

drewwells:
If we could put sass files inside control directives that would be much easier.  Although paths would offer better options for the organization of our code.

Issue #1230, opened 2014-04-25
map-remove should allow removing several keys at once


chriseppstein:
for parity with `map-merge` which allows setting several keys and for efficiency.

nex3:
Agreed.

PR #1275, opened 2014-06-08
Give map-remove ability to remove multiple keys

myabc:
#1230

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -1978,7 +1978,7 @@ module Sass::Script
       hash.delete_if {|key, _| keys.include?(key)}
       map(hash)
     end
-    declare :map_remove, [:map], :var_args => true
+    declare :map_remove, [:map, :key], :var_args => true
 
     # Returns a list of all keys in a map.
     #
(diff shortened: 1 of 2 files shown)


nex3:
Thanks for the patch!

For backwards-compatibility, using `$key` as a keyword argument should continue to work, although it shouldn't be documented. You should be able to do this by just declaring an additional argument without modifying the definition. Please add an additional test for this as well.

myabc:
@nex3 done. I hope I've handled it correctly. I'm presuming we don't need to handle `$keys` as a keyword argument.

lib/sass/script/functions.rb
```diff
@@ -1967,18 +1967,18 @@ def map_merge(map1, map2)
...
+    #   @param $map  [Sass::Script::Value::Map]
+    #   @param $keys [[Sass::Script::Value::Base]]
     # @return [Sass::Script::Value::Map]
     # @raise [ArgumentError] if `$map` is not a map
-    def map_remove(map, key)
+    def map_remove(map, *keys)
       assert_type map, :Map, :map
       hash = map.to_h.dup
-      hash.delete key
+      hash.delete_if {|key, _| keys.include?(key)}
```
myabc, line 1978: I find `delete_if` more elegant, but this alternative implementation is a tad faster:

``` ruby
keys.each do |key|
  hash.delete(key)
end
```

https://gist.github.com/myabc/bc58d5cb49c679fd561c

PR #1341, opened 2014-07-28
Support compass watching

chriseppstein:
These are the changes I needed to make in order to remove Compass's compiler/watcher and replace it with a simple adapter to the Sass compiler/watcher.

None of the watching code in Sass is tested at all and these changes are relatively simple, but if you think some of it needs testing, I can add tests.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -3,6 +3,23 @@
 * Table of contents
 {:toc}
 
+## 3.3.11 (UNRELEASED)
+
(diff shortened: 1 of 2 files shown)


nex3:
Done reviewing. Please also add changelog entries for the user-visible changes here. Tests would be nice, but don't feel obligated to block the PR on them.

chriseppstein:
I added changelog entries. I'd like to add a `--clean` option to sass but I will do that as a separate PR.

chriseppstein:
@nex3 :+1: ?

nex3:
:thumbsup:

PR #1379, opened 2014-08-19
Docs typo: “now only only outputs UTF-8”, not “not only”

alanhogan:
Assuming I understood that correctly!


nex3:
Thanks!

alanhogan:
NP, I only wish I had the time and drive to be a more active contributor ;) thanks for all the hard work. I’m enjoying Sass 3.4, Natalie

PR #1438, opened 2014-09-15
Use flat_map instead of map.flatten

joevandyk:
Should be a bit faster, less allocations.  Not sure if I should be doing this on stable or master..

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -98,7 +98,7 @@ end
 
 desc "Install Sass as a gem. Use SUDO=1 to install with sudo."
 task :install => [:package] do
-  gem  = RUBY_PLATFORM =~ /java/  ? 'jgem' : 'gem' 
+  gem  = RUBY_PLATFORM =~ /java/  ? 'jgem' : 'gem'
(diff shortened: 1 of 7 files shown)


nex3:
This is busted on Ruby 1.8.7 (as the Travis build indicates). I suppose we could add a `Sass::Util` function to switch between `.map.flatten` and `.flat_map`, but it's very cumbersome to use those functions and I don't want to switch to it without compelling evidence that it produces a noticeable performance improvement.

joevandyk:
Hm, thought 1.8.7 was EOL by now. At what point will sass stop supporting it?

nex3:
Unfortunately, 1.8 was bundled with OS X through Mountain Lion, and is _de facto_ the only Ruby version available for users of those systems who aren't themselves Ruby developers. OS X Mavericks was the first version to include a more recent Ruby, and it was released less than a year ago; since quite a few of our users are on OS X, we want to wait until we're confident we'll break very few of them to drop support for 1.8.

It's worth considering metrics to determine when exactly to remove it, though. I've filed #1444 to track that.

PR #1471, opened 2014-10-14
Fix exception on Sass::Importers::Filesystem#eql?

nitoyon:
`Sass::Importers::Filesystem#eql?` raises an exception if `other` is `nil` or `other` doesn't respond to `root` method.

I found this bug when I executed `Compass.configuration.sass_load_paths.uniq` (`NoMethodError` is occasionally raised):

``` bash
$ ruby -e 'require "compass"; p Compass.configuration.sass_load_paths'
[/home/nitoyon/sass, /path/to/gems/compass-core-1.0.1/stylesheets, Compass::SpriteImporter]

$ ruby -e 'require "compass"; Compass.configuration.sass_load_paths.uniq'
/path/to/gems/sass-3.4.3/lib/sass/importers/filesystem.rb:54:in `eql?': undefined method `root' for
Compass::SpriteImporter:Compass::SpriteImporter (NoMethodError)
        from -e:1:in `uniq'
        from -e:1:in `<main>'
```

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -3,6 +3,10 @@
 * Table of contents
 {:toc}
 
+## 3.4.6 (Unreleased)
+
+* Fix exception on `Sass::Importers::Filesystem#eql?`.
+
 ## 3.4.5 (19 September 2014)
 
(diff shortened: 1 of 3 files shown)


nitoyon:
thanks!

PR #1481, opened 2014-10-21
Fixed an error caused by compiling :not selectors.

### LINKED ISSUE #1476: Regression with 3.4.6: undefined method `specificity'
Hi, I just upgraded from sass 3.4.5 to 3.4.6 and am now getting the error when running `compass compile`:

```
    error client/sass/styles.scss (/Library/Ruby/Gems/2.0.0/gems/sass-3.4.6/lib/sass/selector/abstract_sequence.rb:96:in `block in _specificity': undefined method `specificity' for [:not(:last-child)]:Array)
```

I can confirm it's a regression with 3.4.6 because when I downgraded to 3.4.5 it worked fine with the same sass.

Here are my gem versions:

```
brad: emailer-app (master) gem list

*** LOCAL GEMS ***

animation (0.1.alpha.3)
bigdecimal (1.2.0)
CFPropertyList (2.2.0)
chunky_png (1.3.1)
compass (1.0.1)
compass-core (1.0.1)
compass-flexbox (1.1.3)
compass-import-once (1.0.5)
ffi (1.9.5)
fssm (0.2.10)
io-console (0.4.2)
json (1.7.7)
libxml-ruby (2.6.0)
minitest (4.3.2)
multi_json (1.10.1)
nokogiri (1.5.6)
psych (2.0.0)
rake (0.9.6)
rb-fsevent (0.9.4)
rb-inotify (0.9.5)
rdoc (4.0.0)
sass (3.4.6)
sqlite3 (1.3.7)
test-unit (2.0.0.0)
```

andrewjtait:
Fixes #1476 

From what I can see it is [this commit](https://github.com/sass/sass/commit/c717726c4307899bec87482673bfb7fb6fe1f509) which has introduced the bug. The fact that it now [returns an Array](https://github.com/sass/sass/blob/c717726c4307899bec87482673bfb7fb6fe1f509/lib/sass/selector/pseudo.rb#L93-L97) is not being handled in [this method](https://github.com/sass/sass/blob/stable/lib/sass/selector/abstract_sequence.rb#L96) - it is expecting an instance of the `Sass::Selector::Pseudo` class, but is instead now given an `Array` containing the instance.

This PR reverts the changes made in the `with_selector` method, and adds the introduced `:not` logic into a new method which is called in the relevant place.

Let me know if this is any good and if I need to update anything, including the changelog.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -3,6 +3,11 @@
 * Table of contents
 {:toc}
 
+## 3.4.7 (tbc)
+
+* Fixed a bug introduced in 3.4.6 where `:not` selectors were causing a
+compilation error.
+
 ## 3.4.6 (16 October 2014)
 
 * Parent selectors now work in selector pseudoclasses (for example, `:not(&)`).
--- a/lib/sass/selector/pseudo.rb
+++ b/lib/sass/selector/pseudo.rb
@@ -51,9 +51,9 @@ module Sass
       # Returns a copy of this with \{#selector} set to \{#new\_selector}.
       #
       # @param new_selector [CommaSequence]
-      # @return [Array<Simple>]
+      # @return [CommaSequence]
       def with_selector(new_selector)
-        result = Pseudo.new(syntactic_type, name, arg,
+        Pseudo.new(syntactic_type, name, arg,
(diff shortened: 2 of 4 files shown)


andrewjtait:
I've updated the PR to include a new test, which failed with the introduced bug.

nvanheuverzwijn:
Great job Andrew !

PR #1492, opened 2014-10-29
Add set-nth to the list functions group in the function documentation

xzyfer:
This PR updates the functions reference documentation by adding `set-nth` to the list functions table.

screen shot 2014-10-29 at 3 28 02 pm

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -170,6 +170,9 @@ module Sass::Script
   # \{#nth nth($list, $n)}
   # : Returns a specific item in a list.
   #
+  # \{#length set-nth($list, $n, $value)}
+  # : Replaces the nth item in a list.
+  #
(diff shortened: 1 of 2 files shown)


lib/sass/script/functions.rb
```diff
@@ -170,6 +170,9 @@ module Sass::Script
   # \{#nth nth($list, $n)}
   # : Returns a specific item in a list.
   #
+  # \{#length set-nth($list, $n, $value)}
```
xzyfer, line 173: Sorry @nex3 I just noticed I failed at copy paste. This should be `# \{#set-nth set-nth($list, $n, $value)}`

nex3, line 173: Can you send another PR to fix this?

xzyfer, line 173: Yep, done. https://github.com/sass/sass/pull/1501

Issue #1088, opened 2014-01-22
Multiline declaration of maps in SASS


idflood:
Looking at the documentation and commits I see many examples of maps but always in scss syntax. The multiline declaration is really scalable and clear so I would like to use somthing similar in sass.

For instance this scss syntax:

```
$layout: (
  columns: 12,
  gutters: 20px
);
```

In sass I have only found a "one line" syntax. Eventually with a lot of properties this become unreadable.

```
$layout: (columns: 12, gutters: 20px, ...)
```

Is there an existing syntax for multiline map declaration in .sass?

lolmaus:
There's a number of issues with the indented syntax, and Sass maintainers aren't going to fix them. :( They say, the .sass parser is weird and hard to refactor.

I find Sass syntax to be quicker to type and easier to read. It is deprived of the visual noise:

![indented_vs_bracketed](http://d.pr/i/7sbD+)

It's also much easier to do copy-pasting.

So Sass maintainers, PLEASE don't let the indented syntax fall behind!

gisu:
I love the indented syntax, but i switch more and more too the 'ugly' Syntax, because the Syntax have more and more Problems. 

New Member in the Team or Freelancers that never have worked with SASS, have problem with "WTF why i get compile errors!?!?!%§$!" 
SassScript Maps are hard to read in the indented syntax. Goto Definition in Sublimetext will not work with indented syntax (because the Parser have problems when he didn't find curly braces).

In Libsass itself you have no indented Syntax, so i work with braces and hatefull semicolons.

nex3:
Duplicate of #216.

This is definitely something I'd like to add, but it would take a considerable amount of effort due to the refactoring of the indented syntax that would be required. Right now that's just lower priority than adding features that benefit everyone.

ezekg:
@nex3 how much refactoring are we talking about here? Would you mind pointing me to the relevant files so that I could take a look? I was going to make the switch to the indented syntax, but this was kind of a deal breaker since we use maps a lot. We're using Coffee and MtHaml, so wanted to keep the indented syntax consistent throughout our development stack.

nex3:
@ezekg Basically the entire indented-syntax parser would have to be rewritten from the ground up to handle this correctly. It's a major task, but if you want to take a look, the current parser is in `lib/sass/engine.rb`.

ghost:
[...]

Issue #216, opened 2011-11-23
Multiline Expressions in Indented Syntax


talsafran:
_Edited by @nex3 to change the title._

[cross-posted from http://stackoverflow.com/questions/8248976/sass-indented-syntax-on-multiple-lines]

I love indented syntax, I think it's much cleaner.

There's one issue I have with it. If I have a really long mixin declaration, there's no way to split it into multiple lines (obeying the 80 character limit, for example)

Take this example, first written in SCSS.

``` scss
@mixin col($cols, $mleft: 0, $mright: 0, $include-margin: false, $border: 0,
           $pleft: 0, $pright: 0, $include-padding: true, $extra: 0, 
           $clear: false, $lead: true, $container: false) {
    color: red;
    display: block;
}
```

I'm able to split up one long declaration in to multiple lines. With the indented syntax, I don't think there's a way. I have to put the declaration on one line, which is way less readable.

``` sass
@mixin col($cols, $mleft: 0, $mright: 0, $include-margin: false, $border: 0, $pleft: 0, $pright: 0, $include-padding: true, $extra: 0, $clear: false, $lead: true, $container: false)
    color: red
    display: block
```

Is there some way I don't know of? And if not, can there be one? :)

miracle2k:
I also think sass should support this, and not just in mixins. It's a serious deficiency in the indented syntax.

pengwynn:
+1 Haml allows multi-line statements if lines end with a comma.

talsafran:
@pengwynn I like that approach. It's nicer than the way I was thinking, which was adding a \ to the end of line like in Python.

cybersquid:
Just ran into this myself.  It's a real problem.  box-shadow statements can be really long.

Anahkiasen:
+1

while0pass:
+1
another example of currently valid, but not very readable src property in @font-face rule

```
@font-face
    font-family: 'SPEdessa'
    src: url('fonts/spedessa-webfont.eot')
    src: url('fonts/spedessa-webfont.eot?#iefix') format('embedded-opentype'), url('fonts/spedessa-webfont.woff') format('woff'), url('fonts/spedessa-webfont.ttf') format('truetype')
    font-weight: normal
    font-style: normal
```

and currently invalid for SASS, but much more readable:

```
@font-face
    font-family: 'SPEdessa'
    src: url('fonts/spedessa-webfont.eot')
[...]

PR #1504, opened 2014-11-02
[wip] Multi-line map and list support for indented "Sass" syntax

myabc:
This PR contains minimally-invasive, albeit not particularly elegant, patches to allow for multi-line maps and lists as variable values.

Based on the discussion/feature requests here: #216, #1088

**NOT COVERED** is support for other multi-line expressions: mix-in declarations (https://github.com/sass/sass/issues/216#issue-2334772), multi-line properties (https://github.com/sass/sass/issues/216#issuecomment-9028170), etc. From cursory glance, it does look like this would need a rewrite of indented syntax parser.

Would love to get feedback on this…

### DIFF
--- a/lib/sass/engine.rb
+++ b/lib/sass/engine.rb
@@ -726,10 +726,12 @@ WARNING
 
     def parse_variable(line)
       name, value, flags = line.text.scan(Script::MATCH)[0]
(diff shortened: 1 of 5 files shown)


nex3:
As I mentioned [in the issue thread](https://github.com/sass/sass/issues/216#issuecomment-61728214), this is good work but I don't think it's the direction I want to go with this parsing.

PR #1527, opened 2014-11-16
Fix compatibility issues with listen 2.7.12

e2:
- Fixes https://github.com/guard/listen/issues/275
- reworked how different versions of Listen are handled

### DIFF
--- a/lib/sass/plugin/compiler.rb
+++ b/lib/sass/plugin/compiler.rb
@@ -5,6 +5,7 @@ require 'sass'
 require 'sass/callbacks'
 require 'sass/plugin/configuration'
 require 'sass/plugin/staleness_checker'
+require 'listen/compat/wrapper'
 
 module Sass::Plugin
   # The Compiler class handles compilation of multiple files and/or directories,
@@ -298,14 +299,12 @@ module Sass::Plugin
         @watched_files << Sass::Util.realpath(source).to_s
         directories << File.dirname(source)
       end
+
       directories = remove_redundant_directories(directories)
 
-      # A Listen version prior to 2.0 will write a test file to a directory to
-      # see if a watcher supports watching that directory. That breaks horribly
-      # on read-only directories, so we filter those out.
-      unless Sass::Util.listen_geq_2?
-        directories = directories.select {|d| File.directory?(d) && File.writable?(d)}
-      end
+      # The native windows listener is much slower than the polling option, according to
+      # https://github.com/nex3/sass/commit/a3031856b22bc834a5417dedecb038b7be9b9e3e
+      poll = @options[:poll] || Sass::Util.windows?
 
       # TODO: Keep better track of what depends on what
       # so we don't have to run a global update every time anything changes.
@@ -313,29 +312,13 @@ module Sass::Plugin
       # a deprecated feature is removed. It may be removed without warning.
       listener_args = directories +
                       Array(options[:additional_watch_paths]) +
-                      [{:relative_paths => false}]
+                      [{:relative_paths => false, :force_polling => poll}]
 
-      # The native windows listener is much slower than the polling option, according to
-      # https://github.com/nex3/sass/commit/a3031856b22bc834a5417dedecb038b7be9b9e3e
-      poll = @options[:poll] || Sass::Util.windows?
-      if poll && Sass::Util.listen_geq_2?
-        # In Listen 2.0.0 and on, :force_polling is an option. In earlier
-        # versions, it's a method on the listener (called below).
-        listener_args.last[:force_polling] = true
-      end
-
-      listener = create_listener(*listener_args) do |modified, added, removed|
+      listener = Listen::Compat::Wrapper.create
+      listener.listen(*listener_args) do |modified, added, removed|
         on_file_changed(individual_files, modified, added, removed)
         yield(modified, added, removed) if block_given?
       end
-
-      if poll && !Sass::Util.listen_geq_2?
-        # In Listen 2.0.0 and on, :force_polling is an option (set above). In
-        # earlier versions, it's a method on the listener.
-        listener.force_polling(true)
-      end
-
-      listen_to(listener)
     end
 
     # Non-destructively modifies \{#options} so that default values are properly set,
@@ -383,30 +366,6 @@ module Sass::Plugin
     end
 
     private
-
-    def create_listener(*args, &block)
-      Sass::Util.load_listen!
-      if Sass::Util.listen_geq_2?
-        # Work around guard/listen#243.
-        options = args.pop if args.last.is_a?(Hash)
-        args.map do |dir|
-          Listen.to(dir, options, &block)
-        end
-      else
-        Listen::Listener.new(*args, &block)
-      end
-    end
-
-    def listen_to(listener)
-      if Sass::Util.listen_geq_2?
-        listener.map {|l| l.start}
-        sleep
-      else
-        listener.start!
-      end
-    rescue Interrupt
-      # Squelch Interrupt for clean exit from Listen::Listener
-    end
 
     def remove_redundant_directories(directories)
       dedupped = []
--- a/test/sass/compiler_test.rb
+++ b/test/sass/compiler_test.rb
@@ -4,75 +4,13 @@ require File.dirname(__FILE__) + '/../test_helper'
 require 'sass/plugin'
 require 'sass/plugin/compiler'
 
+require "listen/compat/test/session"
+
 class CompilerTest < MiniTest::Test
-  class FakeListener
-    attr_accessor :options
-    attr_accessor :directories
-    attr_reader :start_called
-    attr_reader :thread
-
-    def initialize(*args, &on_filesystem_event)
-      self.options = args.last.is_a?(Hash) ? args.pop : {}
-      self.directories = args
-      @on_filesystem_event = on_filesystem_event
-      @start_called = false
-      reset_events!
-    end
-
-    def fire_events!(*args)
-      @on_filesystem_event.call(@modified, @added, @removed)
-      reset_events!
-    end
-
-    def changed(filename)
-      @modified << File.expand_path(filename)
-    end
-
-    def added(filename)
-      @added << File.expand_path(filename)
-    end
-
-    def removed(filename)
-      @removed << File.expand_path(filename)
-    end
-
-    def on_start!(&run_during_start)
-      @run_during_start = run_during_start
-    end
-
-    # used for Listen < 2.0
-    def start!
-      @run_during_start.call(self) if @run_during_start
-    end
-
-    # used for Listen >= 2.0
-    def start
-      parent = Thread.current
-      @thread = Thread.new do
-        @run_during_start.call(self) if @run_during_start
-        parent.raise Interrupt
-      end
-    end
-
-    def stop
-    end
-
-    def reset_events!
-      @modified = []
-      @added = []
-      @removed = []
-    end
-  end
-
   module MockWatcher
-    attr_accessor :run_during_start
     attr_accessor :update_stylesheets_times
     attr_accessor :update_stylesheets_called_with
     attr_accessor :deleted_css_files
-
-    def fake_listener
-      @fake_listener
-    end
 
     def update_stylesheets(individual_files)
       @update_stylesheets_times ||= 0
@@ -83,44 +21,10 @@ class CompilerTest < MiniTest::Test
     def try_delete_css(css_filename)
       (@deleted_css_files ||= []) << css_filename
     end
-
-    private
-    def create_listener(*args, &on_filesystem_event)
-      if Sass::Util.listen_geq_2?
-        options = args.pop if args.last.is_a?(Hash)
-        args.map do |dir|
-          @fake_listener = FakeListener.new(*args, &on_filesystem_event)
-          @fake_listener.on_start!(&run_during_start)
-          @fake_listener
-        end
-      else
-        @fake_listener = FakeListener.new(*args, &on_filesystem_event)
-        @fake_listener.on_start!(&run_during_start)
-        @fake_listener
-      end
-    end
-  end
-
-  def test_initialize
-    watcher
-  end
-
-  def test_watch_starts_the_listener
-    start_called = false
-    c = watcher do |listener|
-      start_called = true
-    end
-    c.watch
-    assert start_called, "start! was not called"
   end
 
   def test_sass_callbacks_fire_from_listener_events
-    c = watcher do |listener|
-      listener.changed "changed.scss"
-      listener.added "added.scss"
-      listener.removed "removed.scss"
-      listener.fire_events!
-    end
+    c = watcher
 
     modified_fired = false
     c.on_template_modified do |sass_file|
@@ -140,21 +44,35 @@ class CompilerTest < MiniTest::Test
       assert_equal "removed.scss", sass_file
     end
 
-    c.watch
+    files = %w(changed.scss added.scss removed.scss)
+    events = [%w(changed.scss), %w(added.scss), %w(removed.scss)]
 
-    assert_equal 2, c.update_stylesheets_times
+    session = _create_session {c.watch(files)}
+    session.simulate_events(*events)
+    session.interrupt
+
     assert modified_fired
     assert added_fired
     assert removed_fired
+    assert_equal 2, c.update_stylesheets_times
   end
 
   def test_removing_a_sass_file_removes_corresponding_css_file
-    c = watcher do |listener|
-      listener.removed "remove_me.scss"
-      listener.fire_events!
-    end
+    c = watcher
 
-    c.watch
+    session = _create_session {c.watch([["foo"]])}
+    session.simulate_events([], [], ["remove_me.scss"])
+    session.interrupt
+
+    assert_equal "./remove_me.css", c.deleted_css_files.first
+  end
+
+  def test_removing_a_individually_watched_sass_file_removes_corresponding_css_file
+    c = watcher
+
+    session = _create_session {c.watch([["foo"]])}
+    session.simulate_events([], [], ["remove_me.scss"])
+    session.interrupt
 
     assert_equal "./remove_me.css", c.deleted_css_files.first
   end
@@ -166,10 +84,7 @@ class CompilerTest < MiniTest::Test
         filename =~ /\.png$/
       end
     end
-    c = watcher(:load_paths => [image_importer]) do |listener|
-      listener.changed "image.png"
-      listener.fire_events!
-    end
+    c = watcher(:load_paths => [image_importer])
 
     modified_fired = false
     c.on_template_modified do |f|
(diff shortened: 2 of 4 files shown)


e2:
Any feedback for improving more than welcome. (I'm spoiled by RSpec, so I don't know what is good practice with regards to Minitest, and I decide about using minitest/stub_const).

While the patch is big and squashed into a single commit, most of changes remove existing code and add new classes in new files.

Also merges cleanly into master (except for doc-src/SASS_CHANGELOG.md of course.).

Personally, I've been _very_ careful to keep Listen backward compatible and avoid bugs (to a fault) - so if you have any suggestions or questions for Listen specificly, don't hesiate (we're already duplicating some of the work).

e2:
#### TL; DR - I'd like to extract the above to a gem which would _guarantee_ a minimum API/functionality forever

A few thoughts:

1) Obviously, it doesn't make sense for Sass to be responsible for issues in Listen.

2) Also, Sass doesn't needy anything from Listen's API besides "run forever and when something changes do this, handle Ctrl-C, shutdown gracefully, complain meaningfully when something is wrong, etc.".

3) I'm astounded at the accumulated effort put in to supporting various versions of Listen - Sass shouldn't have to care

4) There's a ton of OS-specific stuff that Sass deals with that Listen could abstract out

So I was thinking of creating a "mini-listen" gem that would guarantee a minimum working API ... forever.

Meaning, it would be this gem's responsibility to keep the API working, regardless of changes in Listen (which needs lots of internal rework to fix all the known issues).

And so, I was wondering:
- what would reduce the burden of cases that Sass has to deal with regarding watching, paths, etc.?
- would removing recursion be ok? (since it gives more control over ignoring/excluding)
- is there any behavior that could help users avoid issues or diagnose them better?
- would glob-type expressions be better than regexps? Or keep both?
- can I drop the distinction between changed/added/removed (since that can't be 100% reliable anyway and can cause misleading errors)
- would it be better for users to explicitly say which files they are interested in? (as opposed to watching directories and never ignoring "noise")

I'm sure dealing with issues in Listen takes everyone's precious time, and it doesn't make sense for others to spend hours on something I could do in minutes (and vice-versa) - so any feedback would be kindly appreciated.

lib/sass/plugin/compiler.rb
```diff
@@ -384,27 +365,9 @@ def clean(individual_files = [])
...
-      if Sass::Util.listen_geq_2?
-        listener.map {|l| l.start}.each {|thread| thread.join}
-      else
-        listener.start!
-      end
-    rescue Interrupt
-      # Squelch Interrupt for clean exit from Listen::Listener
+    # Method to allow tests to mock/stub Listen
+    def listener
+      @listener || Sass::Util.listener
```
nex3: `@listener` is never assigned in this module, and `Sass::Util.listener` already caches it.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
+module Sass
+  module Util
+    module ListenWrapper
+      extend ::Sass::Util
```
nex3: I try to avoid doing this these days in favor of explicitly calling methods on `Util`, to make it easier to tell where a given function is coming from.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+        @module ||= const_get("::Listen")
+      end
+
+      def self.listen_module=(mod)
+        @module = mod
+      end
+
+      # History of bugs/workarounds:
+      #
+      # Ancient (< 2.0.0) - very old version
```
nex3: It's important to note that this is the version we bundle with Sass, because it works on Ruby 1.8.7 and the more recent versions don't.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
+module Sass
+  module Util
+    module ListenWrapper
+      extend ::Sass::Util
+
+      def self.listen_module
+        @module ||= const_get("::Listen")
```
nex3: Why use `const_get` here over just `::Listen` directly?

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+        end
+
+        def _stop
+          _listen_module.stop
+        end
+      end
+
+      # Workarounds for pre 2.0 versions of Listen
+      class Ancient < Common
+        LAST_VERSION = Gem::Version.new("1.99.99") # fictional version
```
nex3: Rather than using a fictional version here, why not make it `NEXT_VERSION` and say it's the version after the workarounds in the class?

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+        end
+
+        protected
+
+        def _listen_module
+          ListenWrapper.listen_module
+        end
+
+        def _listen_class
+          _listen_module.const_get("Listener")
```
nex3: I'm pretty sure you can just write `_listen_module::Listener` here.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+      # Stale (>= 2.7.7) - broke mutliple dir handling on OSX (#243)
+      #   - devious threads hack in Sass works by accident (!)
+      #   = 2.7.11 - last version where start still returns a thread
+      #
+      # Current (>= 2.7.12)- start() no longer returns a thread
+      #   - fixed multiple dir handling (#243)
+      #   = 2.8.0 - current version
+
+      # "Expected" functionality from any Listen version
+      class Common
```
nex3: Please document these functions.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+
+        # A Listen version prior to 2.0 will write a test file to a directory to
+        # see if a watcher supports watching that directory. That breaks horribly
+        # on read-only directories, so we filter those out.
+        def watchable_directories(directories)
+          directories.select {|d| File.directory?(d) && File.writable?(d)}
+        end
+
+        def listen(*args, &block)
+          # In Listen 2.0.0 and on, :force_polling is an option (set above). In
```
nex3: I don't see it "set above".

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+
+      # >= 2.0.0, <= 2.7.6
+      class Old < Common
+        LAST_VERSION = Gem::Version.new("2.7.6")
+      end
+
+      # >= 2.7.7, <= 2.7.11
+      class Stale < Common
+        LAST_VERSION = Gem::Version.new("2.7.11")
+        # Work around guard/listen#243 (>= v2.7.9, < v2.8.0)
```
nex3: Nit: add a newline before this.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+      class Current < Common
+        LAST_VERSION = Gem::Version.new("2.99.99")
+      end
+
+      class Future < Current
+        LAST_VERSION = Gem::Version.new("99.99.99")
+      end
+
+      # Test helper
+      class Fake < Common
```
nex3: Let's move this to a file in `test/sass`.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+
+      private
+
+      def self._detect_listen_version
+        _load_listen!
+        require 'listen/version'
+        ::Listen::VERSION
+      end
+
+      # This method is crying "please halp!"
```
nex3: Please write a more descriptive comment.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+            begin
+              require 'listen'
+            rescue LoadError => e
+              if version_geq(RUBY_VERSION, "1.9.3")
+                version_constraint = "~> 2.7"
+              else
+                version_constraint = "~> 1.1"
+              end
+              e.message << "\n" <<
+              "Run \"gem install listen --version '#{version_constraint}'\" to get it."
```
nex3: Nit: indent +2.

lib/sass/util/listen_wrapper.rb
```diff
@@ -0,0 +1,233 @@
...
+          end
+        else
+          begin
+            require 'listen'
+          rescue LoadError => e
+            dir = scope("vendor/listen/lib")
+            if $LOAD_PATH.include?(dir)
+              raise e unless File.exist?(scope(".git"))
+              e.message << "\n" <<
+              'Run "git submodule update --init" to get the bundled version.'
```
nex3: Nit: indent +2.

test/sass/util/listen_wrapper_test.rb
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env ruby
+require File.dirname(__FILE__) + '/../../test_helper'
+require 'sass/util'
+require 'fileutils'
+
+require 'sass/util/listen_wrapper'
+
+# TODO: Personally I'd use RSpec or a a mocking library, but ...
```
nex3: Feel free to add a mocking library if you'd like.

nex3:
I'd love to see this encapsulated in a gem, and I'd love even more not to be responsible for maintaining it.

> And so, I was wondering:
> - what would reduce the burden of cases that Sass has to deal with regarding watching, paths, etc.?

I've been fine with each individual iteration of Listen's API over the years. There are two main issues we've run into: dealing with the frequent change of the API, and dealing with the fact that post-2.0.0 it didn't support Ruby 1.8.7.

The former would be helped substantially if listen followed semantic versioning. There have been a number of breaking changes in patch releases, which is difficult for us to track. If we knew that nothing would break until 3.0.0, that would help a lot.

The latter ship has already sailed, but it's the reason we have to support all of these different versions in the first place. If we could just ditch support for the "ancient" Listen versions, we could use a proper RubyGems dependency and gradually drop support for older Listen versions.

> - would removing recursion be ok? (since it gives more control over ignoring/excluding)

No, we need recursion for `--watch` mode.

> - is there any behavior that could help users avoid issues or diagnose them better?

User complaints that stem from Listen are almost always of the form "I didn't touch anything and Sass started giving me this ugly stack trace." The reason is that a Listen API has changed an Sass is suddenly using the wrong calling convention. So this is another thing that would be helped with less frequent breaking changes.

> - would glob-type expressions be better than regexps? Or keep both?

We don't use any type of filtering, since our importers can theoretically import any type of file.

> - can I drop the distinction between changed/added/removed (since that can't be 100% reliable anyway and can cause misleading errors)

Nothing's ever 100% when you're dealing with the filesystem, but this distinction is useful for us. If it went away we'd probably just end up rebuilding it by hand.

> - would it be better for users to explicitly say which files they are interested in? (as opposed to watching directories and never ignoring "noise")

Watching full directories is important for our command-line API.

nex3:
@e2 Before you address the code review comments, we should decide whether we want to pull this out into a separate gem. 4c003d3c67237dfebb781dec96050e833579ebed will get Sass working for now, so the added complexity from this PR may be better spent towards a separate reusable gem.

e2:
@nex3 - yes I considered that too, I'm just overloaded with bugfixes right now - so the overhead of a new gem will be greater (considering I need to rethink some points), while I'm sure a lot of people would be happy to get this fixed right now. 

I'll get the review points addressed soon, so that's no problem on my side - no matter how happy I'd be to skip merging this and have a new `listen-compat` gem ready. Unfortunately, I strongly believe bugfixes have a higher priority than inter-gem refactorings.

e2:
I think it's up to you to decide - the time spent making this work will save me time later when extracting it to a gem (I can commit to doing it, so no problem there). If you prefer to stick to the patch, that's fine too, because it squashes the bug.

e2:
I've made every change except for a mocking library.

I've tested this on various listen/ruby versions (watch scss files+ ctrl-c):

Ruby 1.8.7:

vendored 1.3.1 # works
'1.3.0' '1.3.1' '2.0.0' '2.0.1' '2.0.2' '2.1.0' # ok
'2.1.1' '2.1.2' '2.2.0' # doesn't work (no response)
'2.3.1' '2.4.0' '2.4.1' '2.5.0' '2.6.0' # ctrl-c doesn't work
'2.6.1' # celluloid no longer works with 1.8.7

Ruby 1.9.3:
'2.6.2' # 1.9.3 - ctrl-c doesn't work
'2.7.0' # 1.9.3 - works again
'2.7.4' '2.7.5' '2.7.6' '2.7.7' '2.7.8' '2.7.9' '2.7.11' '2.7.12' '2.8.0' '2.8.1' # ok

> I'd love even more not to be responsible for maintaining it.

That's the point - though as a backup plan, I'd gladly give you the option of reverting a broken release (a good policy in this case).

> dealing with the frequent change of the API

I did my best since I took over - though I the multiple dirs on OSX is my fault (I broke it, then I broke-fixed-broke it), and my excuse is I didn't have OSX to test it out properly (and the acceptance tests were as much a nightmare as they were brittle). Though using the thread incidentally returned by start() really confused me at first.

> post-2.0.0 it didn't support Ruby 1.8.7

Celluloid requires 1.9, so I can't help here much. Are there users who still desperately need 1.8.7? Would 1.8 support be still worthwhile? (I'm guessing not).

> If we could just ditch support for the "ancient" Listen versions, we could use a proper RubyGems dependency and gradually drop support for older Listen versions.
> 
> No, we need recursion for --watch mode.

Obviously, yes - I was just trying to see what complex parts of Listen I could "cut off". While removing ignores and excludes would make Listen quite a bit simpler, polling basically pull every other functionality as a dependency (even ignores and excludes). The added/modified/removed feature is relatively "cheap" compared to that.

Would a listen-specific config file be useful (kind of like Jekyll uses)? There'd be no need for a polling option in sass, while the only option needed from sass would be the directories.

I'll let you know once something is ready (probably weeks from now).

e2:
I've extracted the code out into a separate namespace - so I can later extract it to a gem with a smaller change of screwing things up. (Ideally, only files will be removed from Sass and a gem dependency will be added).

Let me know if you want any changes before I create a new gem and freeze this functionality.

e2:
One major issue, though:

While I'd gladly take dealing with changes in Listen off your hands and I can accept PRs for any missing Ruby 1.8 support, I don't see a point in supporting the "vendored" / 'git submodule' / "LOAD_PATH" stuff.

It would be up to the user to either deal with the LOAD_PATH/vendor/git themselves, or install at least Rubygems 1.8.0 (and use RUBYOPT to require it by default, if necessary).

Even Listen 0.1.0 is available through RubyGems so there's no point in dragging this further.

If someone has a super-important-expensive-project requiring Sass on Ruby 1.8.7 and without RubyGems, I'm sure it's important enough for them to fork Sass and Listen and fund active maintenance for whatever they need.

e2:
Ok, I created the `listen-compat` gem and updated this PR to use it.

It basically removes a lot of code from Sass and nothing else.

The problem remains with how to deal with no-rubygems support. I even thought about vendoring a Listen version in Listen-compat, but I don't think it makes sense. Instead, I think there should be a sass-watch with a dependency on _any_ version of Listen (so Sass can be used without Listen).

Let me know what you think.

test/sass/compiler_test.rb
```diff
@@ -177,45 +92,55 @@ def watched_file?(filename)
...
+    scss = _abs_path('foo.scss')
+    css = _abs_path('foo.css')
+
+    session = _create_session {c.watch([[scss, css, nil]])}
+    session.simulate_events([], [], [scss])
+    session.interrupt
+
+    # TODO: separate test?
+    # directories = listener_instances.first.directories
+    # assert directories.include?(_abs_path(".")), directories.inspect
```
nex3, line 114: What's the deal with this?

nex3:
This looks pretty good, although I think we need to retain the `load_listen` code—it's important to make sure we load the vendored version if necessary.

> It would be up to the user to either deal with the LOAD_PATH/vendor/git themselves, or install at least Rubygems 1.8.0 (and use RUBYOPT to require it by default, if necessary).
> 
> Even Listen 0.1.0 is available through RubyGems so there's no point in dragging this further.
> 
> If someone has a super-important-expensive-project requiring Sass on Ruby 1.8.7 and without RubyGems, I'm sure it's important enough for them to fork Sass and Listen and fund active maintenance for whatever they need.

The problem is that many Sass users are not Ruby users, and many of them are on old versions of Mac OS X, which bundles Ruby 1.8.7 and an old version of RubyGems. If we break these users, they won't have the technical know-how to update their Ruby or RubyGems version. I know because we've accidentally broken them in the past and been flooded with issue reports.

test/sass/compiler_test.rb
e2, line 114: It takes a different path in Sass's code (that's why I mentioned this should be a separate test). I didn't want to look into it, because it was unrelated to the issue I was dealing with.

e2:
> it's important to make sure we load the vendored version if necessary.

Well, then by definition I can't handle that from a separate gem.

Which means - e.g. Sass could capture the Gem NameError (?) from create(), set a load path, require the file and use an instance of the Ancient wrapper. (Ideally, there should at least be a Travis test for vendor loading).

> The problem is that many Sass users are not Ruby users, and many of them are on old versions of Mac OS X, which bundles Ruby 1.8.7 and an old version of RubyGems. If we break these users, they won't have the technical know-how to update their Ruby or RubyGems version. I know because we've accidentally broken them in the past and been flooded with issue reports.

This feels wrong on so many levels. I totally understand the "flood of issues", but it's creating a much bigger problem.

It's much easier to get a non-technical person to upgrade Rubygems than it is to "train" and "recruit" an open source developer (which are also many times fewer in number).

Just like it doesn't make sense to further burden the limited group of doctors just so that things are more convenient for patients.

It's counterproductive, because Sass isn't the _only_ useful gem out there - and all because Sass isn't currently getting flooded, it doesn't mean other gems aren't affected by the same problem. (And _not_ getting flooded may mean other people are perceiving other tools as completely disfunctional).

There isn't even a test for this on Travis (which is something to think about).

Maybe someone should create a user-friendly tool for upgrading Ruby on MacOS? I can't do this, because I don't even have a Mac to begin with (here, on Linux, I have rubygems >= 1.8.3). Maybe there's just a bug that needs fixing in Rubygems? I don't even know how to easily reproduce the scenario to even test it properly (even on a virtual machine).

Vendoring Listen itself makes as much sense as bundling Sass with the OS. What makes more sense is bundling Rubygems with Sass - though that seems silly too.

I'm not suggesting to "push forward and break everything" - on the contrary, there shouldn't be any "dependency management" code in Sass, just like you wouldn't expect Sass to be bundled with a specific version of Ruby.

If a "flood of issues" is a problem, it doesn't mean side-stepping a bigger problem is the correct solution. If upgrading Rubygems is a big issue, then we should drop everything and get that fixed first - to the point where Sass could simply perform the upgrade on the user's behalf (and so _everyone_ would benefit).

Just sharing my thoughts on this.

chriseppstein:
We have decided to stop vendoring the listen gem in Sass 3.5. See  http://blog.sass-lang.com/posts/560719

PR #1575, opened 2014-12-27
Remove extra parens; fixes #1565

### LINKED ISSUE #1565: Inspect function printing too many parenthesis
Test:

``` scss
$map: (
  'a': 1,
  'b': (
    'c': 'd'
  )
);

.test {
  content: inspect($map);
}
```

Expected:

``` css
.test {
  content: ("a": 1, "b": ("c": "d"));
}
```

Output:

``` css
.test {
  content: ("a": 1, "b": (("c": "d")));
}
```

jshawl:
Remove extra set of parentheses from inspect function when inspecting a map that contains children lists

### DIFF
--- a/lib/sass/script/value/map.rb
+++ b/lib/sass/script/value/map.rb
@@ -57,7 +57,7 @@ module Sass::Script::Value
 
       to_sass = lambda do |value|
         if value.is_a?(Map) || (value.is_a?(List) && value.separator == :comma)
-          "(#{value.to_sass(opts)})"
+          "#{value.to_sass(opts)}"
         else
           value.to_sass(opts)
         end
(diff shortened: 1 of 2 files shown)


lib/sass/script/value/map.rb
```diff
@@ -57,7 +57,7 @@ def to_sass(opts = {})
 
       to_sass = lambda do |value|
         if value.is_a?(Map) || (value.is_a?(List) && value.separator == :comma)
-          "(#{value.to_sass(opts)})"
+          "#{value.to_sass(opts)}"
```
nex3, line 60: The parens here are still necessary for lists. Also, as the code stands currently, there's no reason to have this branch different than the one below.

jshawl, line 60: That's weird. The output of inspecting a list contains no parentheses - http://sassmeister.com/gist/5051ea9588061673d807

Also, I wonder why the parens aren't being applied from list's own inspect method: https://github.com/sass/sass/blob/2bdc564dea068c4e19064daea25349d93f8d01b8/lib/sass/script/value/list.rb#L73

Lastly, this bugfix pr is to stable, whereas the other is a feature implementation, so it is submitted to master.

nex3, line 60: To match CSS, lists don't require parentheses unless it would be ambiguous otherwise, as here.

PR #1578, opened 2015-01-01
Fix warnings of  uninitialized variables

sivagollapalli:


nex3:
We consider the use of uninitialized variables to be a feature of Ruby, and we aren't interested in including useless lines of code to avoid it.

PR #1613, opened 2015-01-28
Add link to full list of SassScript functions.

xtian:
There are a couple links to the `Sass::Script::Functions` page in the reference doc, but not one under the main "Functions" heading, which was the first place I checked.

Thought it might be useful to add one!


nex3:
Thanks!

PR #1672, opened 2015-03-18
Add alpha to Color#hsla

paradox460:
`Color#hsla` was not returning an array with the proper arity, missing the alpha channel. This adds it back.

### DIFF
--- a/lib/sass/script/value/color.rb
+++ b/lib/sass/script/value/color.rb
@@ -382,7 +382,7 @@ module Sass::Script::Value
     # @return [Array<Fixnum>] A frozen four-element array of the hue,
     #   saturation, lightness, and alpha values (respectively) of the color
     def hsla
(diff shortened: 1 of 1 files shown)


Lordnibbler:
:+1: this should be tested too!

paradox460:
It _should_ already be tested, [here, to be precise](https://github.com/sass/sass/blob/stable/test/sass/functions_test.rb#L104-L109), but that passes regardless of the arity.

There don't seem to be any other tests for the sass color functions, unfortunately

nex3:
This getter isn't actually used in the Sass codebase itself.

paradox460:
Ah, that would explain why it hasn't been an issue.

In an application I worked on, we had to do some processing of colors on the fly, and I extended the sass color object into my own service object, and was surprised I had to code it myself.

Thanks for merging it!

PR #1696, opened 2015-04-14
Show filename in raised SyntaxError.

mib32:


nex3:
Any `Sass::SyntaxError`s that are thrown while evaluating a document are automatically given line and file-name information by the time they hit the top-level error handler.

PR #1697, opened 2015-04-14
Double space around rules

srawlins:
Today, sass-convert double spaces top-level nodes (like rules), but only those. This change would double-space non-top-level rules as well:

``` bash
$ echo "a { b:c; d: e; f { g: h; } i { j: k} }" |./bin/sass-convert -F scss -T scss
a {
  b: c;
  d: e;

  f {
    g: h;
  }

  i {
    j: k;
  }
}
```

I know that sass-convert does not guarantee (nor even really specify) the style of its output, but people do use it as a Sass formatter ([google search](https://www.google.com/#q=sass%20formatter)). I also know that the Sass team is not interested in maintaining sass-convert as a formatter, for maintainability reasons, such as adding features to sort properties or merge rules.

However, I think this one small change would be welcome, as it is so standard. All of the examples in Sass's documentation double space around rules, and top-level rules (CSS, Sass, and SCSS) have always been typically double-spaced.

Thanks much!


nex3:
I'm in favor of doing this, but I think the behavior should more closely match that of the top level, which double-spaces entities by default and only single-spaces specific things. In particular, declarations and variables should be single-spaced when nested, but directives should be double-spaced.

srawlins:
Ok cool. I think I can rewrite the PR to this behavior. I'm on vacation
until the end of June, so this won't move until then.

srawlins:
Hi @nex3, I've redone the implementation to match the behavior of top-level double-spacing. Looks better.

PR #1701, opened 2015-04-16
fix from_hex method.Now he leading hash is really optional

bernEsp:
##### summary

 The description of method say that the leading hash is optional, and it was not. Now it is optional and return a valid color instance

```
before:
  Color.from_hex('000000') => ##{hex_string}

now return the correct hex color
```


nex3:
Thanks!

PR #1705, opened 2015-04-20
There might be still a small chance that the file was deleted

phanle:
between checking it is readable and actually reading it.
We were compiling in parallel with 32 processors and encountered a race condition intermittently.


phanle:
Bump this PR.

PR #1742, opened 2015-06-08
Store line number in Sequence#line=

sds:
The [`scss-lint` project](https://github.com/brigade/scss-lint) performs a number of static checks against SCSS
code to find style offenses.

`Sequence` objects were returning `nil` when calling `#line`. This was
because `Sequence#line=` was not recording the actual line in an
instance variable like its superclass `AbstractSequence`. Fix this by ensuring the line
number is stored in `@line`.

This will allow the `scss-lint` project to implement
https://github.com/brigade/scss-lint/issues/456


nex3:
Thanks!

Issue #853, opened 2013-07-16
Strange behavior of the subtraction operator


piouPiouM:
I noticed the `-` operator needs to be wrapped by spaces in order to work as intented. Operation `1-1` for example, will return a list and not 0.

What is the origin of this unexpected behavior?

Thx.

Some examples:

``` shell
$ sass -v
Sass 3.3.0.alpha.184 (Bleeding Edge)
$ sass -i
>> 1 - 1 // returns a number
0
>> 1- 1 // returns a number
0
>> 1 -1  // returns a list and not a number
(1 -1)
>> 1-1  // returns a list and not a number
(1 -1)
>> a-b
"a-b"
>> b-a
"b-a"
>> a-2 // returns a string as expected
"a-2"
>> 1-b // returns a list and not a string
(1 "-b")
```

KittyGiraudel:
This is probably due to the fact `-` is used as both a hyphenation mark and a substraction operator. In order to differenciate the two, I believe the operator needs to be wrapped by spaces.

I'd love to hear a word of Nathan / Chris on this though. :)

nex3:
This is expected behavior, although unfortunately kind of unintuitive at the margins. In Sass, `-` can be part of an identifier, part of a negative number, a binary subtraction operator, or a unary negation operator. Being an identifier/number has the highest precedence; unary negation has the lowest. This gives rise to all the cases you listed. Let's go over them one by one:

In `1 - 1` it's unambiguously subtraction. Easy peasy. `1- 1` is the same; since `1-` isn't a valid number, the `-` must be the subtraction operator.

`1 -1` starts to get tricky. `-1` is parsed as a number before Sass ever starts to look for operators. This follows the [CSS grammar](http://www.w3.org/TR/WD-css3-syntax-20030813/#grammar0). `1-1` is parsed the same way.

`a-b` is a valid CSS identifier, so it's parsed as an unquoted string. Same for `b-a` and `a-2`.

`1-b` is similar to the `1 -1` case; `-b` is a valid identifier, so it gets parsed as that before any operators are parsed.

KittyGiraudel:
Thanks for detailed explanations Nathan.

piouPiouM:
Thx for taking the time to explain it so clearly Nathan!

Issue #1105, opened 2014-02-06
Regression in hyphens when used in property lists


xzyfer:
I understand there has been a lot of talk about unintuitive hyphen behaviour recently (#1030, #1023, #853) but I feel like this a regression since `inset -1.5em 0 1.5em -0.75em rgba(0, 0, 0, 0.25)` is a valid css value. Needing to throw in extra `()`s to please sass is very unintuitive. Requiring the extra `()`s makes however make sense if variables were involved.

``` scss
// source.scss
a {
  box-shadow: inset -1.5em 0 1.5em -0.75em rgba(0, 0, 0, 0.25);
}
```

Output from 3.3.0rc2

``` scss
a {
  box-shadow: inset -1.5em 0 1.5em -0.75em rgba(0, 0, 0, 0.25);
}
```

Output from master 3.3.0_980eece359

``` scss
a {
  box-shadow: inset-1.5em 0 0.75em rgba(0, 0, 0, 0.25);
}
```

The space between `inset` and `-1.5em` has been removed, and  `1.5em -0.75em` has been turned into `0.75em`.

Issue #1030, opened 2013-12-03
Expressions with nested parentheses


zikkoerik:
When working with retina-sized graphics and design PSD's, I often want to keep measurements in actual pixels and divide by two. This is neat, but in situations where I want an expression first, then do the division (such as subtracting padding from a height), I get a "nested" expression that doesn't seem to be supported. For example:

```
height: ((100px-22px)/2);
```

becomes:

```
height: 100px -22px/2;
```

where I would expect:

```
height: 39px;
```

There might be some syntax-related reason why this can't be implemented, I just haven't though of any yet. And I'm sure I could use variables, mix-ins or something else to achieve the same thing, it just seems to me that this should work as well.

Otherwise, this being my first issue post: Yay for Sass!

lolmaus:
A hyphen in Sass can be part of variable name, so... H2O: Just Add Water!

```
height: ((100px - 22px) / 2);
```

nex3:
This isn't actually caused by the hyphen being consumed as part of an identifier, nor does it have anything to do with nested parentheses. It's caused by the space operator plus the unary `-` operator having higher precedence than the binary minus.

I think this is misleading enough that it's worth fixing. My instinct is that it's unlikely that anyone is relying on this behavior; browsers don't seem to consider `10px-20px` to be a valid substitute for `10px -20px`. I'm fine with fixing this without a deprecation period. @chriseppstein, what do you think?

chriseppstein:
:thumbsup:

  Chris Eppstein  
 Sent via Intro for iPhone

zikkoerik:
Well, I was perfectly satisfied with the first answer, although I haven't had time to test it yet. Silly of me not to try adding the whitespace myself. I guess I'm just not used to hyphens being valid characters in variable names and such, but it does make sense in the CSS context.

If there is a slight risk of anyone or anything expecting 10px-20px to be the same as 10px -20px, maybe you could go the same way as with the division operator, to have it interpreted as such only if it is part of an expression or enclosed in parentheses? Just an idea - I'm sure you'll see and do what's best.

Anyway, I'm glad this resulted into something that you considered worth fixing!

PR #1761, opened 2015-06-30
Add a doc section about subtraction and hyphen

srawlins:
Subtraction and hyphens tripped me up for a while today before I realized that `margin: 2em-1em` will not perform subtraction. I thought that since [Division and /](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#division-and-slash) was so well documented, then it was the only exception to the operators. :stuck_out_tongue: not so!

I'm not an authority on the statements I made in this change, but I used @nex3's explanation in #853 as guidelines.

### DIFF
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -884,6 +884,43 @@ is compiled to:
     p {
       font: 12px/30px; }
 
+##### Subtraction and `-`
+{#subtraction-and-hyphen}
+
+CSS allows `-` to appear in identifiers. SassScript must support this syntax,
+and also allow `-` to be used for subtraction. To ensure that a `-` is
+interpreted as subtraction, keep it spaced out from the operands. For example:
+
+    p {
+      margin: 20px-10px;         // Plain CSS, no subtraction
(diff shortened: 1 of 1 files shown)


r4j4h:
+1 I could see this tripping someone else up too

doc-src/SASS_REFERENCE.md
```diff
@@ -884,6 +884,43 @@ is compiled to:
...
+
+    p {
+      margin: 20px-10px;         // Plain CSS, no subtraction
+      padding: 20px - 10px 40px; // hyphen is spaced out, does subtraction
+    }
+
+is compiled to:
+
+    p {
+      margin: 10px;
```
nex3: This is incorrect. Sass will produce `20px-10px` here, because `px-10px` is a valid identifier and thus a valid name for a unit.

nex3:
I'm on board for documenting this, but I don't think this documentation gets at the heart of the matter. I'll try to come up with my own. It's also work mentioning that the behavior has changed a lot since #853; see #1030 and #1105.

srawlins:
Thanks for the fix, @nex3 !

PR #1813, opened 2015-08-26
remove identical hash key / value

tenderlove:
This hash key and value are identical, so we can remove one of them from
this hash literal.


nex3:
Thanks!

PR #1814, opened 2015-08-26
fix cicular require warning

tenderlove:
`sass/util` required `sass/util/normalized_map` and
`sass/util/normalized_map` required `sass/util`

Circular requires are considered dangerous, so I pulled the `sass/util`
require out of `sass/util/normalized_map` and changed
`sass/script/tree/funcall` to require `sass/util` instead of directly
requiring the normalized map file


nex3:
Thanks!

PR #1818, opened 2015-08-28
Further avoid race conditions in the filesystem cache store.

akapshuk:
As many people still use 3.2.x because of dependencies please release this fix as 3.2.20


chriseppstein:
What dependencies? Sass 3.3 was released 18 months ago and 3.4 was released a year ago.

Happy to take this commit against stable (3.4.x), but you'll need to install a custom forked gem for 3.2

akapshuk:
https://rubygems.org/gems/compass/versions/0.12.7
as you can see compass 0.12.7 latest stable on 0.x dependent on sass 3.2, which has a bug fixed only in 3.4.x

PR #1819, opened 2015-08-28
Handle floating point errors more gracefully.

chriseppstein:
In SassScript, `(29 / 7 * 7) == 29` returns false. Which is bad because anyone who knows math (or in this case, @mknadler) can see this should be true instead.

This PR creates the concept of a floating point epsilon which is derived from the current sass script precision. Any two numbers will now be considered equal if they are within an `epsilon` of each other. If the stylesheet's precision is increased by 1, the epsilon gets divided by 10.

The default precision is `5` and so the default epsilon is `1 / (10 ** (5 + 1))` or `0.000001`.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -3,6 +3,12 @@
 * Table of contents
 {:toc}
 
+## 3.4.19 (UNRELEASED)
+
+* Sass numeric equality now better handles float-point errors. Any
+  numbers that are within 1 / (10 ^ (precision + 1)) of each other are
+  now considered equal.
+
 ## 3.4.18 (25 August 2015)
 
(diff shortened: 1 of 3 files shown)


chriseppstein:
Pending review from @nex3

mknadler:
```
29.0 / 7 * 7
=> 29.000000000000004
```

I'm not sure what math is, but I can definitely say with confidence that if you take 29 cats, and then split them into seven cat-packs (and have a stray cat with no pack that is also in all of the packs), and then herd all of the cats back into one big cat-congery, and there's a tiiiiny piece of cat-ear lying on the ground when you're all done, it is a _very bad day indeed_.

chriseppstein:
merged into stable as cc96f0c85d5ec09324e9158585c23ddbe922b2ab

PR #1821, opened 2015-08-31
Remove 1 8 6 workarounds

joerixaop:
There are still a few helper methods around that are only used for ruby 1.8.6 workarounds, but the gemspec has set a minimal version of ruby 1.8.7 for a long time (and for example arr.flatten(n) was already widely used instead of the Sass::Utils#flatten method, meaning sass wouldn't work on ruby 1.8.6 anyway). So it seems cleaner to clean up the code by removing these workarounds.


chriseppstein:
Thanks!

PR #1822, opened 2015-08-31
Avoid slicing twice when once suffices

joerixaop:
The old approach duplicated a comment twice when there was a newline
before it. This can be avoided by passing a parameter to rindex to
indicate where it needs to start searching.


chriseppstein:
Can you please add a test case for this fix?

joerixaop:
This doesn't change anything functional, the exact same output will be provided in both cases, just with one less copy of the string in certain situations. There seem to be a lot of tests for comments already, so I'm not quite sure what new thing there is to test.

chriseppstein:
Ok. I misunderstood "duplicated a comment twice" and thought it was duplicated in the output.

Issue #1812, opened 2015-08-25
REPL output of division of complex units is inconsistent


chriseppstein:
```
>> $div: 50em / 2px
50em/2px
>> $div
25em/px
```

In the repl, I think we should fully resolve the division before echoing the result.

nex3:
I'm moving this issue to [the new Ruby Sass repository](https://github.com/sass/ruby-sass) because it's specific to Ruby Sass's implementation.

PR #1824, opened 2015-09-02
Output the result of divison of complex units

ThaiWood:
For #1812, show the result of the division instead of just repeating it

### DIFF
--- a/lib/sass/repl.rb
+++ b/lib/sass/repl.rb
@@ -41,7 +41,7 @@ module Sass
           environment.set_var(name, val.perform(environment))
         end
 
-        p environment.var(name)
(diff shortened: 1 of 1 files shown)


nex3:
Thanks for the PR! Sorry it too so long to get to it.

I don't think this does quite what you want. You haven't modified the case where division is printed in its original form—the `Script::MATCH` expression matches variable declarations.

The underlying issue here is that we print division as literal slashes in order to make CSS properties that include slashes work. That's controlled by the [`Number.original`](http://sass-lang.com/documentation/Sass/Script/Value/Number.html#original-instance_method) property. If you set that property to `false`, it should print as you expect.

PR #1844, opened 2015-09-26
Adding a command line option --sourcemap-filename=custom.css.map

liquidmetal:
It is sometimes required that the sourcemap file be placed in a location that is different from the original scss/sass file. This command line does just that.

### DIFF
--- a/lib/sass/exec/sass_scss.rb
+++ b/lib/sass/exec/sass_scss.rb
@@ -172,8 +172,6 @@ MESSAGE
         @options[:sourcemap] = (type || :auto).to_sym
       end
 
-      # If the user gives --sourcemap-filename, we write out the sourcemap file there
-      # The default is to write in the same location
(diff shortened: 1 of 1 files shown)


liquidmetal:
Any suggestions / criticism here?

chriseppstein:
> It is sometimes required that the sourcemap file be placed in a location that is different from the original scss/sass file.

I'd like to understand this use case better. Is it simply that you need them in a different directory instead of being in the css output directory? If not, then I don't understand why our convention of adding `.map` to the filename doesn't work -- anything else just seems confusing.

Furthermore,  from my reading of the code here, it looks like this option will break when used with the `--update` and `--watch` options because it would write all the sourcemap files to the same file.

It seems like adding an option `--sourcemap-location` that points to the directory containing sourcemaps (e.g. right next to the css directory) would work with all modes and avoid confusion.

chriseppstein:
No response. Closing.

Issue #1754, opened 2015-06-23
Add support for @import ... supports()


nex3:
According to [this post](http://www.w3.org/blog/CSS/2015/06/19/cascade4-fpwd/), CSS is adding support for `supports()` annotations to CSS `@import` directives. Sass should support these in the same way it supports `media()` annotations.

chriseppstein:
Fixed by 25a73178f38bb939dac75132e03429f986ab57e6.

PR #1871, opened 2015-10-27
Import supports() clause.

chriseppstein:
http://www.w3.org/TR/css-cascade-4/#at-import

### DIFF
--- a/lib/sass/scss/parser.rb
+++ b/lib/sass/scss/parser.rb
@@ -55,6 +55,15 @@ module Sass
         interp_ident
       end
 
+      # Parses a supports clause for an @import directive
+      def parse_supports_clause
+        init_scanner!
+        ss
+        clause = supports_clause
+        ss
+        clause
+      end
+
       # Parses a media query list.
       #
       # @return [Sass::Media::QueryList] The parsed query list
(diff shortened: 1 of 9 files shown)


chriseppstein:
This addresses issue #1754.

lib/sass/scss/parser.rb
```diff
@@ -548,6 +561,15 @@ def supports_directive(name, start_pos)
         node(node, start_pos)
       end
 
+      def supports_clause
+        return unless tok(/supports\(/i)
+        ss
+        supports = supports_condition
+        ss
+        tok!(/\)/)
```
nex3: If you make this a plain string (`")"`), it'll provide a better error message.

nex3:
This needs SCSS and `sass-convert` tests. After that, it looks good.

lib/sass/scss/parser.rb
chriseppstein: @nex3 Using a string caused errors.

```
TypeError: wrong argument type String (expected Regexp)
    /Users/ceppstei/Projects/sass/lib/sass/scss/parser.rb:1193:in `scan'
    /Users/ceppstei/Projects/sass/lib/sass/scss/parser.rb:1193:in `tok'
    /Users/ceppstei/Projects/sass/lib/sass/scss/parser.rb:1103:in `tok!'
    /Users/ceppstei/Projects/sass/lib/sass/scss/parser.rb:569:in `supports_clause'
```

So I kept the regex

PR #1875, opened 2015-10-29
Add source ranges to CommentNodes; correct Interpolations in comments

### LINKED ISSUE #1868: Comment interpolations have wrong source_range
When visiting a Sass tree with a visitor (and expanding to visit things like comments, script interpolations and script funcalls), Sass::Script::Tree::Interpolation nodes (and nodes within) have wrong `#source_range`. It appears to not understand when comments are on their own lines, maybe? Here's a recreate script:

``` ruby
require 'sass'

class InterpolationVisitor < Sass::Tree::Visitors::Base
  def visit_comment(node)
    puts "Comment at:       #{node.source_range.inspect}"
    yield
  end

  def visit_script_funcall(node)
    puts "Funcall at:       #{node.source_range.inspect}"
    yield
  end

  def visit_script_interpolation(node)
    puts "Interpolation at: #{node.source_range.inspect}"
    yield
  end
end

module Sass::Tree
  class CommentNode
    def children
      super + value.select { |item| item.is_a?(::Sass::Script::Tree::Node) }
    end
  end
end

module Sass::Script
  class Tree::Funcall
    def self.visit_method; :visit_script_funcall; end
  end

  class Tree::Interpolation
    def self.visit_method; :visit_script_interpolation; end
  end

  class Tree::Variable
    def self.visit_method; :visit_script_variable; end
  end
end

sass = <<SASS
$color: purple;
/* \#{hue($color)} */

$color: purple;
/* \#{hue($color)} */
SASS

puts sass
tree = Sass::Engine.new(sass, {syntax: :scss}).to_tree
InterpolationVisitor.new().send(:visit, tree)
```

This produces:

```
Comment at:       nil
Interpolation at: (2:15 to 2:29)
Funcall at:       (2:17 to 2:28)
Comment at:       nil
Interpolation at: (5:53 to 5:67)
Funcall at:       (5:55 to 5:66)
```

which is definitely way wrong.

srawlins:
Fixes #1868 

Travis should show exactly 5 failures, all in `test/sass/source_map_test.rb`. To me, these failures look like just the result of comment nodes now being included in source maps. Can I just update the test expected outputs with the new actual outputs? Otherwise these test cases are a... mystery to me.

### DIFF
--- a/lib/sass/scss/parser.rb
+++ b/lib/sass/scss/parser.rb
@@ -159,20 +159,16 @@ module Sass
         silent = text =~ %r{\A//}
         loud = !silent && text =~ %r{\A/[/*]!}
         line = @line - text.count("\n")
+        comment_start = @scanner.pos - text.length
+        index_before_line = @scanner.string.rindex("\n", comment_start) || -1
+        offset = comment_start - index_before_line
 
         if silent
           value = [text.sub(%r{\A\s*//}, '/*').gsub(%r{^\s*//}, ' *') + ' */']
         else
-          value = Sass::Engine.parse_interp(
-            text, line, @scanner.pos - text.size, :filename => @filename)
-          newline_before_comment = @scanner.string.rindex("\n", @scanner.pos - text.length)
-          last_line_before_comment =
-            if newline_before_comment
-              @scanner.string[newline_before_comment + 1...@scanner.pos - text.length]
-            else
-              @scanner.string[0...@scanner.pos - text.length]
-            end
-          value.unshift(last_line_before_comment.gsub(/[^\s]/, ' '))
+          value = Sass::Engine.parse_interp(text, line, offset, :filename => @filename)
+          line_before_comment = @scanner.string[index_before_line + 1...comment_start]
+          value.unshift(line_before_comment.gsub(/[^\s]/, ' '))
         end
 
         type = if silent
@@ -182,8 +178,8 @@ module Sass
                else
                  :normal
                end
-        comment = Sass::Tree::CommentNode.new(value, type)
-        comment.line = line
+        start_pos = Sass::Source::Position.new(line, offset)
+        comment = node(Sass::Tree::CommentNode.new(value, type), start_pos)
         node << comment
       end
 
--- a/test/sass/source_map_test.rb
+++ b/test/sass/source_map_test.rb
@@ -28,7 +28,7 @@ a {
 CSS
 {
 "version": 3,
-"mappings": "AAAA,CAAE;EACA,GAAG,EAAE,GAAG;;EAER,SAAS,EAAE,IAAI",
+"mappings": "AAAA,CAAE;EACA,GAAG,EAAE,GAAG;EACV,kBAAkB;EAChB,SAAS,EAAE,IAAI",
 "sources": ["test_simple_mapping_scss_inline.scss"],
 "names": [],
 "file": "test.css"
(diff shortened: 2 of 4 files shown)


lib/sass/scss/parser.rb
```diff
@@ -150,20 +150,16 @@ def process_comment(text, node)
         silent = text =~ %r{\A//}
         loud = !silent && text =~ %r{\A/[/*]!}
         line = @line - text.count("\n")
+        comment_start = @scanner.pos - text.length
+        newline_before_comment = @scanner.string.rindex("\n", comment_start) || -1
```
nex3: If you're going to make this -1 if the comment is on the first line, it no longer represents a newline in all cases. Call it `index_before_line` or something instead.

lib/sass/scss/parser.rb
```diff
@@ -173,8 +169,8 @@ def process_comment(text, node)
                else
                  :normal
                end
-        comment = Sass::Tree::CommentNode.new(value, type)
-        comment.line = line
+        start_pos = Sass::Source::Position.new(line, comment_start - newline_before_comment)
```
nex3: Use `offset` here.

test/sass/source_map_test.rb
```diff
@@ -846,6 +846,118 @@ def test_sources_array_is_uri_escaped
 JSON
   end
 
+  def test_scss_comment_source_range
```
nex3, line 849: Use `assert_parses_with_mapping` for these tests.

nex3:
You're right that the existing tests just need to be updated.

lib/sass/scss/parser.rb
srawlins: Done.

lib/sass/scss/parser.rb
srawlins: Done.

test/sass/source_map_test.rb
srawlins, line 849: Done. Thanks for this tip!

srawlins:
Thanks @nex3 for the review. All comments addressed, and I've squashed into one commit.

lib/sass/scss/parser.rb
```diff
@@ -182,8 +178,8 @@ def process_comment(text, node)
                else
                  :normal
                end
-        comment = Sass::Tree::CommentNode.new(value, type)
-        comment.line = line
+        offset = Sass::Source::Position.new(line, comment_start - index_before_line)
```
nex3: I meant use the pre-existing `offset` variable:

``` ruby
start_pos = Sass::Source::Position.new(line, offset)
```

srawlins:
Oops, my bad @nex3. Addressed and squashed.

test/sass/source_map_test.rb
```diff
@@ -350,44 +350,44 @@ def test_selectors_properties_sourcemap_scss
...
-    {{11}}a .special:hover{{/11}} {
-      {{12}}foo{{/12}}: {{13}}bar{{/13}};
-      {{14}}cursor{{/14}}: {{15}}e-resize{{/15}};
-      {{16}}color{{/16}}: {{17}}rgba(255, 0, 0, 0.8){{/17}}; }
-    {{18}}a .special:after{{/18}} {
-      {{19}}content{{/19}}: {{20}}"I ate 15 pies 2px thick!"{{/20}}; }
-  {{21}}a:active{{/21}} {
-    {{22}}color{{/22}}: {{23}}#050709{{/23}};
-    {{24}}border{{/24}}: {{25}}2px solid black{{/25}}; }
+{{2}}  /* SOME COMMENT */{{/2}}
```
nex3: Is there any way we can make the source range not include the whitespace before the comment?

srawlins:
Thanks @nex3 , good catch. Addressed and squashed. Brought in a change to to_css.rb.

nex3:
Beautiful! Thanks so much for the patch!

Issue #1707, opened 2015-04-21
Support trailing comma in multi-line @include


henrik:
This is a syntax error because of the last trailing comma:

``` sass
@include foo(
  $one: 1,
  $two: 2,
);
```

I prefer trailing commas on every line since it reads consistently, diffs are cleaner and lines are easier to add and reorder without introducing "oops forgot the comma" errors, and I know I'm not alone in that.

So I propose supporting this syntax, if it's not hard to achieve. What do you think?

nex3:
Since we support this for maps and lists, we should support it for argument lists as well.

henrik:
@nex3 Excellent, thank you! Might look into contributing this on the next hack night if no one beats me to it (please do).

KittyGiraudel:
Makes sense. Seconded. :)

chriseppstein:
Agree. :+1:

PR #1878, opened 2015-11-01
Added support for trailing commas

royels:
Addresses issue #1707

### DIFF
--- a/lib/sass/script/parser.rb
+++ b/lib/sass/script/parser.rb
@@ -483,16 +483,11 @@ RUBY
             args << e if e
           end
 
-<<<<<<< HEAD
           return args, keywords, splat unless no_trailing_comma?(description)
-=======
-          return args, keywords, splat unless try_tok(:comma)
->>>>>>> parent of 3118711... Added a new method no_trailing_comma?
           e = assert_expr(subexpr, description)
         end
       end
 
-<<<<<<< HEAD
       def no_trailing_comma?(description)
         if description == 'mixin argument'
           val = try_tok(:comma)
@@ -503,8 +498,6 @@ RUBY
         end
       end
 
-=======
->>>>>>> parent of 3118711... Added a new method no_trailing_comma?
       def raw
         tok = try_tok(:raw)
         return special_fun unless tok
--- a/lib/sass/script/parser.rb
+++ b/lib/sass/script/parser.rb
@@ -483,9 +483,15 @@ RUBY
             args << e if e
           end
 
-          return args, keywords, splat unless try_tok(:comma)
+          return args, keywords, splat unless no_trailing_comma?
           e = assert_expr(subexpr, description)
         end
+      end
+
+      def no_trailing_comma?
+         val = try_tok(:comma)
+         peeked = @lexer.peek.type
+         !((val && peeked == :rparen) || (!val))
       end
 
       def raw
--- a/lib/sass/script/parser.rb
+++ b/lib/sass/script/parser.rb
@@ -483,17 +483,28 @@ RUBY
             args << e if e
           end
 
-          return args, keywords, splat unless no_trailing_comma?
+<<<<<<< HEAD
+          return args, keywords, splat unless no_trailing_comma?(description)
+=======
+          return args, keywords, splat unless try_tok(:comma)
+>>>>>>> parent of 3118711... Added a new method no_trailing_comma?
           e = assert_expr(subexpr, description)
         end
       end
 
-      def no_trailing_comma?
-         val = try_tok(:comma)
-         peeked = @lexer.peek.type
-         !((val && peeked == :rparen) || (!val))
+<<<<<<< HEAD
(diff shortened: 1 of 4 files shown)


lib/sass/script/parser.rb
```diff
@@ -483,11 +483,21 @@ def arglist(subexpr, description)
...
           end
 
-          return args, keywords, splat unless try_tok(:comma)
+          return args, keywords, splat unless no_trailing_comma?(description)
           e = assert_expr(subexpr, description)
         end
       end
 
+      def no_trailing_comma?(description)
+        if description == 'mixin argument'
```
nex3: Trailing commas should be supported in all argument lists, including definition arglists.

lib/sass/script/parser.rb
```diff
@@ -483,11 +483,21 @@ def arglist(subexpr, description)
             args << e if e
           end
 
-          return args, keywords, splat unless try_tok(:comma)
+          return args, keywords, splat unless no_trailing_comma?(description)
           e = assert_expr(subexpr, description)
         end
       end
 
+      def no_trailing_comma?(description)
```
nex3: A cleaner solution than modifying the comma-detection would be to refactor the loop itself so that it's more like

```
while there's a subexpression:
  parse the subexpression
  return if we don't parse a comma
```

test/sass/scss/scss_test.rb
```diff
@@ -3894,12 +3894,13 @@ def test_interpolation_with_bracket_on_next_line
 SCSS
   end
 
-  def test_extra_comma_in_mixin_arglist_error
-    assert_raise_message(Sass::SyntaxError, <<MESSAGE.rstrip) {render <<SCSS}
-Invalid CSS after "...clude foo(bar, ": expected mixin argument, was ");"
-MESSAGE
-@mixin foo($a1, $a2) {
-  baz: $a1 $a2;
+  def test_extra_comma_in_mixin_arglist
```
nex3, line 3897: We should still test that cases like `foo($bar,, $baz)` produce errors.

test/sass/scss/scss_test.rb
```diff
@@ -3908,6 +3909,44 @@ def test_extra_comma_in_mixin_arglist_error
 SCSS
   end
 
+  def test_extra_comma_in_mixin_arglist_multiline
```
nex3: I don't think there's a lot of value in testing this in the specifically multiline case. The script parser is pretty newline-agnostic in general.

nex3:
Done reviewing. Thanks for the PR, by the way!

chriseppstein:
@royels Do you plan to address the code review comments?

royels:
Whoops, I'm super sorry for falling off the face of the planet.
At any rate, yep, I addressed the code review comments (except for the first, I don't think I hit the spirit of "clean" refactor in my last commit. So still trying to figure that out.)

EDIT: I've removed the method, but more or less re-implemented its idea into the arglist handler. 
I'm genuinely unclear on what @nex3 meant by not modifying comma detection. If I understand correctly, isn't the point of the issue that we need to detect commas that are the last character before the right parentheses?

test/sass/scss/scss_test.rb
royels, line 3897: Added a new test for that!

lib/sass/script/parser.rb
```diff
@@ -458,7 +458,7 @@ def arglist(subexpr, description)
         return [args, keywords] unless e
 
         splat = nil
-        loop do
+        until subexpr.nil?
```
nex3, line 461: `subexpr` should never be nil here.

lib/sass/script/parser.rb
```diff
@@ -479,11 +479,13 @@ def arglist(subexpr, description)
...
-
             args << e if e
           end
 
-          return args, keywords, splat unless try_tok(:comma)
+          comma_present = try_tok(:comma)
+          if !comma_present || (@lexer.peek.type == :rparen && comma_present)
+            return args, keywords, splat
+          end
           e = assert_expr(subexpr, description)
```
nex3, line 489: This should be the condition in the loop.

``` ruby
while (e = send(subexpr))
  # ...
  return args, keywords, splat unless try_tok(:comma)
end
```

That way you can avoid the initial `send()` call, and you won't have to check to see if the next token is `)`.

nex3:
Did another review iteration. I tried to be more clear about what I meant by refactoring the loop, but if it's still confusing let me know and I'll just make the change locally.

chriseppstein:
@nex3 I think we should just clean this up and land it do to the latency of updates.

royels:
I agree, after hammering at it for a while, I was still puzzled on how to address the issue by incorporating @nex3 's suggestions. Quite interested to see what it was supposed to look like.

nex3:
@chriseppstein Can you take charge of that?

chriseppstein:
Closing. This landed via https://github.com/sass/sass/pull/2046

Issue #1739, opened 2015-06-04
Support nested maps in sass:map functions


blackfalcon:
*Edited by @nex3*

* [x] [Proposal](https://github.com/sass/sass/pull/2836)
* [x] [Specs](https://github.com/sass/sass-spec/issues/1568)
* [x] [Documentation](https://github.com/sass/sass-site/issues/476)
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1088)
* [ ] [LibSass](https://github.com/sass/libsass/issues/3127)

---

@lunelson project **sass-list-maps** has support for nested list maps, using the `map-get-z function`. 
https://github.com/lunelson/sass-list-maps

The following example says it all:

```
$list-map-z: (
  alpha (
    beta (
      gamma 3
    )
  )
);

.demo {
  out: map-get-z($list-map-z, alpha); // -> ( beta ( gamma 3 ) )
  out: map-get-z($list-map-z, alpha, beta); // -> ( gamma 3 )
  out: map-get-z($list-map-z, alpha, beta, gamma); // -> 3
}
```

FACT: Up to a point, even when using maps extended naming conventions come into play. Using `map-get-z` we can actually use structure to categorize versus naming conventions. I vote for structure every time. 

```
$carousel: (
  color (
    header (
      border gray,
    )
  )
);

$side-nav: (
  bg gray,
  color (
    hover (
      search yellow,
      home red,
      filter blue
    )
  )
);

block {
  color: map-get-z($side-nav, color, hover, filter);
}

block {
  border-color: map-get-z($carousel, color, header, border);
}
```

I don't mind using the plug-in, but I think that this is a feature that Sass should support natively.

nex3:
I'm not opposed to allowing `map-get()` to take multiple keys to do nested retrieval.

Off-topic, but:

``` scss
$side-nav: (
  bg gray,
  color (
    hover (
      search yellow,
      home red,
      filter blue
    )
  )
);
```

This isn't a map, it's a (nested) list of pairs. It works with map functions for compatibility reasons, but (in addition to being harder to read than a real map) it's likely to be much less efficient.

lunelson:
[...]

PR #1895, opened 2015-11-13
Nested map functions

HamptonMakes:
I've implemented the suggested map-get() and map-merge() functions in #1739, and also additionally included a nestable version of map-has-key() following a similar style to map-merge().


HamptonMakes:
Sorry, I figured this should be merged into master, not stable. Will re-open.

PR #1896, opened 2015-11-13
Nested Map Functions

HamptonMakes:
Now, map-merge, map-get, and map-has-key all accept various forms of nested Map logic as per @nex3's specifications in #1739, with the addition of a nested compatible map-has-key.

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -196,11 +196,13 @@ module Sass::Script
   # Maps in Sass are immutable; all map functions return a new map rather than
   # updating the existing map in-place.
   #
-  # \{#map_get map-get($map, $key)}
-  # : Returns the value in a map associated with a given key.
+  # \{#map_get map-get($map, $keys)}
+  # : Returns the value in a map (or nested map) associated with a given series of keys.
   #
   # \{#map_merge map-merge($map1, $map2)}
   # : Merges two maps together into a new map.
+  # \{#map_merge map-merge($map, keys, value)
+  # : Merges in the specified nested keys into a new map
   #
   # \{#map_remove map-remove($map, $keys...)}
   # : Returns a new map with keys removed.
@@ -211,8 +213,8 @@ module Sass::Script
   # \{#map_values map-values($map)}
   # : Returns a list of all values in a map.
   #
-  # \{#map_has_key map-has-key($map, $key)}
-  # : Returns whether a map has a value associated with a given key.
+  # \{#map_has_key map-has-key($map, $keys)}
+  # : Returns whether a map has a value associated with a given key or keys.
   #
   # \{#keywords keywords($args)}
   # : Returns the keywords passed to a function that takes variable arguments.
@@ -2053,15 +2055,23 @@ MESSAGE
     #   map-get(("foo": 1, "bar": 2), "baz") => null
     # @overload map_get($map, $key)
     #   @param $map [Sass::Script::Value::Map]
-    #   @param $key [Sass::Script::Value::Base]
-    # @return [Sass::Script::Value::Base] The value indexed by `$key`, or `null`
+    #   @param $keys [[Sass::Script::Value::Base]]
+    # @return [Sass::Script::Value::Base] The value indexed by `$keys`, or `null`
     #   if the map doesn't contain the given key
     # @raise [ArgumentError] if `$map` is not a map
-    def map_get(map, key)
+    def map_get(map, *keys)
       assert_type map, :Map, :map
-      map.to_h[key] || null
+      result = map
+      keys.each do |key|
+        if result.is_a?(Sass::Script::Value::Map) || result.is_a?(Hash)
+          result = result.to_h[key]
+        else
+          return null
+        end
+      end
+      result || null
     end
-    declare :map_get, [:map, :key]
+    declare :map_get, [:map, :key], :var_args => true
 
     # Merges two maps together into a new map. Keys in `$map2` will take
     # precedence over keys in `$map1`.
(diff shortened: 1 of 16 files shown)


lib/sass/script/functions.rb
```diff
@@ -196,11 +196,13 @@ module Sass::Script
   # Maps in Sass are immutable; all map functions return a new map rather than
   # updating the existing map in-place.
   #
-  # \{#map_get map-get($map, $key)}
-  # : Returns the value in a map associated with a given key.
+  # \{#map_get map-get($map, $keys)}
```
nex3: `$keys...` here and elsewhere.

lib/sass/script/functions.rb
```diff
@@ -2053,15 +2055,23 @@ def list_separator(list)
...
+        if result.is_a?(Sass::Script::Value::Map) || result.is_a?(Hash)
+          result = result.to_h[key]
+        else
+          return null
+        end
+      end
+      result || null
     end
-    declare :map_get, [:map, :key]
+    declare :map_get, [:map, :key], :var_args => true
```
nex3: Get rid of the `:key` declaration here—the the variable argument name can't be used as a named parameter anyway (see `call()` for another example of the same thing).

You should also preserve the previous declaration, so that `map-get()` calls with named parameters continue to work.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2088,35 @@ def map_get(map, key)
...
     # @return [Sass::Script::Value::Map]
     # @raise [ArgumentError] if either parameter is not a map
-    def map_merge(map1, map2)
+    # @overload map_merge($map1, $keys, $value)
+    #   @param $map1 [Sass::Script::Value::Map]
+    #   @param $keys [[Sass::Script::Value::Base]]
+    #   @param $value [Sass::Script::Value::Base]
+    # @return [Sass::Script::Value::Map]
+    # @raise [ArgumentError] if the first parameter is not a map
+    def map_merge(map1, *args)
```
nex3, line 2108: [I've changed my mind on how this should work.](https://github.com/sass/sass/issues/1739#issuecomment-156571180)

lib/sass/script/functions.rb
```diff
@@ -2053,15 +2055,23 @@ def list_separator(list)
...
+    # @return [Sass::Script::Value::Base] The value indexed by `$keys`, or `null`
     #   if the map doesn't contain the given key
     # @raise [ArgumentError] if `$map` is not a map
-    def map_get(map, key)
+    def map_get(map, *keys)
       assert_type map, :Map, :map
-      map.to_h[key] || null
+      result = map
+      keys.each do |key|
+        if result.is_a?(Sass::Script::Value::Map) || result.is_a?(Hash)
```
nex3: Why would `result` ever be a Ruby hash here?

lib/sass/script/functions.rb
```diff
@@ -2147,16 +2175,19 @@ def map_values(map)
...
-    #   @param $key [Sass::Script::Value::Base]
+    #   @param $keys [[Sass::Script::Value::Base]]
     # @return [Sass::Script::Value::Bool]
     # @raise [ArgumentError] if `$map` is not a map
-    def map_has_key(map, key)
+    def map_has_key(map, *keys)
       assert_type map, :Map, :map
-      bool(map.to_h.has_key?(key))
+      bool(keys.all? do |key|
+        map = map.to_h[key] || false
```
nex3, line 2207: `nil` is falsey, so you don't need the `|| false` here.

Also, what if `map` is no longer a map?

lib/sass/script/value/map.rb
```diff
@@ -43,6 +43,17 @@ def eq(other)
       Bool.new(other.is_a?(Map) && value == other.value)
     end
 
+    def surgical_merge(my_key, *keys)
```
nex3: Please document any methods you add.

test/sass/functions_test.rb
```diff
@@ -1357,10 +1357,20 @@ def test_unique_id
...
+    # Nested maps should return values, if found and matching the keys
+    assert_equal "d", evaluate("map-get((a: (b: (c: d))), a, b, c)")
+
+    # map-get should return a 'null' if the key isn't found
     assert_equal "null", perform("map-get((foo: 1, bar: 2), baz)").to_sass
     assert_equal "null", perform("map-get((), foo)").to_sass
+
+    # map-get should return a 'null' if the nested map key isn't found
+    assert_equal "null", perform("map-get((a: b), a, b, c, m)").to_sass
+    assert_equal "null", perform("map-get((a: (b: (c: d))), a, b, c, m)").to_sass
```
nex3, line 1373: Test accessing a nested empty map.

test/sass/functions_test.rb
```diff
@@ -1419,8 +1439,10 @@ def test_map_values_checks_type
 
   def test_map_has_key
     assert_equal "true", evaluate("map-has-key((foo: 1, bar: 1), foo)")
+    assert_equal "true", evaluate("map-has-key((foo: (bar: 1)), foo, bar)")
     assert_equal "false", evaluate("map-has-key((foo: 1, bar: 1), baz)")
     assert_equal "false", evaluate("map-has-key((), foo)")
+    assert_equal "false", evaluate("map-has-key((foo: (bar: 1)), foo, bing)")
```
nex3, line 1456: Test nesting deeper than actually exists, nesting through values that aren't maps, and nesting through empty maps.

nex3:
Don't forget to make the tests pass!

HamptonMakes:
Thaaaaat's embarrassing. :/

—

HamptonMakes:
I've updated based on feedback, but seems like rebasing on master caused this request to somehow calculate in more changes?

nex3:
It looks like you didn't change some of the tests. Also Travis is still failing.

Also, where did that documentation change come from?

HamptonMakes:
That documentation change came off master.

nex3:
If it were the same commit as the one on master, it wouldn't be showing up in the PR.

HamptonMakes:
Opening a new PR.

Issue #1291, opened 2014-06-19
Sass Function to determine if a string has quotes


chriseppstein:
There's currently no way to check if a string has quotes. I suggest we add:

`has-quotes($string) => true/false`

chriseppstein:
Currently, making a new string with the same "quote status" as another string is very hard to do.

nex3:
I'd prefer `is-quoted`, since we consistently refer to these strings as "quoted" or "unquoted".

ghost:
`quote($string) == $string` seems easy enough, although when would you need to use this? Not saying that it shouldn't be implemented, but I don't know when determining if something has quotes as opposed to turning them on/off should be useful.

chriseppstein:
@Undeterminant that doesn't work. strings are equal despite their quotes.

lunelson:
+1. It's possible to force the quoting or unquoting of a string, but in principle it makes sense to have a function for it—similar to `list-separator()` for lists

yairEO:
@Undeterminant - let me give you an example:

```
$typography : (
    font-family    : "Roboto, Arial, sans-serif",
    base-font-size : 14px
);
```

`font-family` must have quotes because of the commas. the other properties of `$typography` could be anything. I use this functions which I wrote as helpers for layout variables:

```
// map-get helper
@function mg($map, $key) {
    @if map-has-key($map, $key) {
        @return map-get($map, $key);
    }

    @warn "Unknown `#{$key}` in #{$map}.";
    @return null;
}

@function typography($key:"") {
    @return unquote(mg($typography, $key)));
}
```

I get an error from SASS:

> DEPRECATION WARNING: Passing 14px, a non-string value, to unquote()
>    will be an error in future versions of Sass.

If I don't use `unquote` in my `typography` function, then the font family won't work in CSS (it must be unquoted)

cimmanon:
@yairEO  You don't need quotes here, you need parentheses.

``` scss
$typography : (
    font-family    : (Roboto, Arial, sans-serif),
    base-font-size : 14px
);
```

lunelson:
:+1:@cimmanon's answer is the correct approach, since one of the items in that list may also be quoted: e.g.

``` scss
$font: (
  family: ("Helvetica Neue", helvetica, sans-serif),
  size: 16px
)
```

HamptonMakes:
@nex3 is this "planned" to do `str-is-quoted($str)`?

davidkpiano:
[...]

PR #1899, opened 2015-11-14
Added `str-is-quoted` function

HamptonMakes:
As discussed in issue #1291.


HamptonMakes:
Ugh again... this stable/master issue.

PR #1900, opened 2015-11-14
Implemented `str-is-quoted`

HamptonMakes:
For your coding pleasure.

As specified in sass/sass#1291

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -8,6 +8,8 @@
 * Combine ids and `:root` when unifying selectors with `@extend` and selector
   functions.
 
+* Add a `str-is-quoted($string)` function to detect whether a not a string is quoted or not (e.g. is an identifier)
+
 * Add a `$weight` parameter to `invert()`.
 
 ## 3.4.20 (Unreleased)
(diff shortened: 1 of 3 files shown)


nex3:
I'm going to put this on ice for now until we have support for modules. New functions in 4.0 will probably only be available via modules, and I don't want to set anyone up for breakage by adding a function that we're planning on relocating before release.

HamptonMakes:
I thought that the plan was to implement "planned" features for a maybe 3.5 release.

Since all that code needs refactoring for 4.0 into modules, seems like no patches for "planned" things can happen until after 4.0 is done blocking then?

nex3:
Even if we do a 3.5 release—which is possible but unlikely—I'd like to avoid adding new functions there for the same reason I want to avoid adding them to 4.0 prereleases: I don't want to release a feature with the full knowledge that we intend to make a breaking change to it in the future.

I probably won't land patches for planned functions until we have a module system. That doesn't mean the patches aren't useful, though; it'll be much quicker to land patches that already exist than it would be to start working on them only after modules are already finished.

nex3:
I've merged this into the [`feature.new-functions`](https://github.com/sass/sass/tree/feature.new-functions) branch.

PR #1904, opened 2015-11-15
WIP: Nested Map Functions

HamptonMakes:
Now, map-merge, map-get, and map-has-key all accept various forms of nested Map logic as per @nex3's specifications in #1739, with the addition of a nested compatible map-has-key.

Removed the weird commit that snuck in to the previous request and hopefully bought us Ruby 1.8.7 compatibility again.

### DIFF
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -10,9 +10,14 @@
 
 * Add a `$weight` parameter to `invert()`.
 
-* Add support for nested maps with `map-get()` and `map-has-key()` by passing more than 2 arguments
+* Add support for nested maps with `map-get()` and `map-has-key()` by passing
+  more than 2 arguments.
 
-* Add support for setting changes deeply into a nested hash by providing a new syntax for set that is generally non-destructive with the signature `map-set($map, $keys..., $value)` 
+* Add support for setting changes deeply into a nested hash by providing a new
+  syntax for set that is generally non-destructive with the signature
+  `map-set($map, $keys..., $value)`.
+
+* Add support for additional keys passed to `map-merge($map1, $keys..., $map2)`, allowing for nested Map merges.
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -10,6 +10,10 @@
 
 * Add a `$weight` parameter to `invert()`.
 
+* Add support for nested maps with `map-get()` and `map-has-key()` by passing more than 2 arguments
+
+* Add support for merging changes deeply into a nested hash by providing a new syntax for merge that is generally non-destructive with the signature `map-merge($map, $keys, $value)` 
+
 ### Backwards Incompatibilities -- Must Read!
 
 Certain ways of using `#{}` without quotes in property and variable values,
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -17,7 +17,8 @@
   syntax for set that is generally non-destructive with the signature
   `map-set($map, $keys..., $value)`.
 
-* Add support for additional keys passed to `map-merge($map1, $keys..., $map2)`, allowing for nested Map merges.
+* Add support for additional keys passed to `map-merge($map1, $keys..., $map2)`,
+  allowing for nested Map merges.
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/doc-src/SASS_CHANGELOG.md
+++ b/doc-src/SASS_CHANGELOG.md
@@ -12,7 +12,7 @@
 
 * Add support for nested maps with `map-get()` and `map-has-key()` by passing more than 2 arguments
 
-* Add support for merging changes deeply into a nested hash by providing a new syntax for merge that is generally non-destructive with the signature `map-merge($map, $keys, $value)` 
+* Add support for setting changes deeply into a nested hash by providing a new syntax for set that is generally non-destructive with the signature `map-set($map, $keys..., $value)` 
 
 ### Backwards Incompatibilities -- Must Read!
 
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -196,13 +196,16 @@ module Sass::Script
   # Maps in Sass are immutable; all map functions return a new map rather than
   # updating the existing map in-place.
   #
-  # \{#map_get map-get($map, $keys)}
+  # \{#map_get map-get($map, $keys...)}
   # : Returns the value in a map (or nested map) associated with a given series of keys.
   #
   # \{#map_merge map-merge($map1, $map2)}
   # : Merges two maps together into a new map.
-  # \{#map_merge map-merge($map, keys, value)
-  # : Merges in the specified nested keys into a new map
+  # \{#map_merge map-merge($map1, $keys..., $map2)}
+  # : Merges a map into another map, nested with the specified keys
+  #
+  # \{#map_set map-set($map, $keys..., $value)
+  # : Sets in the specified keys into a new map, supporting one or more keys, creating nested maps as needed
   #
   # \{#map_remove map-remove($map, $keys...)}
   # : Returns a new map with keys removed.
@@ -213,7 +216,7 @@ module Sass::Script
   # \{#map_values map-values($map)}
   # : Returns a list of all values in a map.
   #
-  # \{#map_has_key map-has-key($map, $keys)}
+  # \{#map_has_key map-has-key($map, $keys...)}
   # : Returns whether a map has a value associated with a given key or keys.
   #
   # \{#keywords keywords($args)}
@@ -2050,7 +2053,7 @@ MESSAGE
       assert_type map, :Map, :map
       result = map
       keys.each do |key|
-        if result.is_a?(Sass::Script::Value::Map) || result.is_a?(Hash)
+        if result.is_a?(Sass::Script::Value::Map)
           result = result.to_h[key]
         else
           return null
@@ -2058,7 +2061,8 @@ MESSAGE
       end
       result || null
     end
-    declare :map_get, [:map, :key], :var_args => true
+    declare :map_get, [:map, :key]
+    declare :map_get, [:map], :var_args => true
 
     # Merges two maps together into a new map. Keys in `$map2` will take
     # precedence over keys in `$map1`.
@@ -2082,28 +2086,40 @@ MESSAGE
     #   @param $map2 [Sass::Script::Value::Map]
     # @return [Sass::Script::Value::Map]
     # @raise [ArgumentError] if either parameter is not a map
-    # @overload map_merge($map1, $keys, $value)
+    # @overload map_merge($map1, $keys..., $map2)
+    #   @param $map1 [Sass::Script::Value::Map]
+    #   @param $keys [[Sass::Script::Value]]
+    #   @param $map2 [Sass::Script::Value::Map]
+    # @return [Sass::Script::Value::Map]
+    # @raise [ArgumentError] if either parameter is not a map
+    def map_merge(map1, *args)
+      assert_type map1, :Map, :map1
+      *keys, map2 = args
+      if map1.is_a? Sass::Script::Value::List
+        map1 = map(map1.to_h)
+      end
+      assert_type map2, :Map, :map2
+      map1.recursive_merge(keys, map2)
+    end
+    declare :map_merge, [:map1, :map2]
+    declare :map_merge, [:map1], var_args: true
+
+    # @overload map_set($map, $keys, $value)
     #   @param $map1 [Sass::Script::Value::Map]
     #   @param $keys [[Sass::Script::Value::Base]]
     #   @param $value [Sass::Script::Value::Base]
     # @return [Sass::Script::Value::Map]
     # @raise [ArgumentError] if the first parameter is not a map
-    def map_merge(map1, *args)
-      assert_type map1, :Map, :map1
-      # If we only have one argument, it must be a map
-      if args.size == 1
-        map2 = args.first
-        assert_type map2, :Map, :map2
-        map(map1.to_h.merge(map2.to_h))
-      else # Using the keyed version
-        if map1.is_a? Sass::Script::Value::List
-          map1 = map(map1.to_h)
-        end
-        map1.surgical_merge(*args)
+    def map_set(map, *args)
+      assert_type map, :Map, :map
+      *keys, new_value = args
+      if map.is_a? Sass::Script::Value::List
+        map = map(map.to_h)
       end
+      map.recursive_set(keys, new_value)
     end
-    declare :map_merge, [:map1, :map2]
-    declare :map_merge, [:map1, :keys, :value], :var_args => true
+    declare :map_set, [:map, :keys, :value], :var_args => true
+
 
     # Returns a new map with keys removed.
     #
@@ -2117,6 +2133,9 @@ MESSAGE
     # @overload map_remove($map, $keys...)
     #   @param $map  [Sass::Script::Value::Map]
     #   @param $keys [[Sass::Script::Value::Base]]
+    # @overload map_remove($map, $key)
+    #   @param $map  [Sass::Script::Value::Map]
+    #   @param $keys [Sass::Script::Value::Base]
     # @return [Sass::Script::Value::Map]
     # @raise [ArgumentError] if `$map` is not a map
     def map_remove(map, *keys)
@@ -2125,7 +2144,8 @@ MESSAGE
       hash.delete_if {|key, _| keys.include?(key)}
       map(hash)
     end
-    declare :map_remove, [:map, :keys], :var_args => true
+    declare :map_remove, [:map], var_args: true
+    declare :map_remove, [:map, :key]
 
     # Returns a list of all keys in a map.
     #
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -2103,7 +2103,7 @@ MESSAGE
       end
     end
     declare :map_merge, [:map1, :map2]
-    declare :map_merge, [:map1, :keys, :value]
+    declare :map_merge, [:map1, :keys, :value], :var_args => true
 
     # Returns a new map with keys removed.
     #
@@ -2125,7 +2125,7 @@ MESSAGE
       hash.delete_if {|key, _| keys.include?(key)}
       map(hash)
     end
-    declare :map_remove, [:map, :key], :var_args => true
+    declare :map_remove, [:map, :keys], :var_args => true
 
     # Returns a list of all keys in a map.
     #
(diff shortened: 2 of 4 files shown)


doc-src/SASS_CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 * Add a `$weight` parameter to `invert()`.
 
+* Add support for nested maps with `map-get()` and `map-has-key()` by passing more than 2 arguments
```
nex3: Please limit Markdown lines to 80 characters and make sure they're full sentences with proper punctuation.

In general, please try to match the surrounding style in prose or code.

doc-src/SASS_CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
...
+
+* Add support for setting changes deeply into a nested hash by providing a new syntax for set that is generally non-destructive with the signature `map-set($map, $keys..., $value)` 
+
```
nex3, line 19: Add a changelog entry about `map-merge()`.

lib/sass/script/functions.rb
```diff
@@ -196,11 +196,16 @@ module Sass::Script
   # Maps in Sass are immutable; all map functions return a new map rather than
   # updating the existing map in-place.
   #
-  # \{#map_get map-get($map, $key)}
-  # : Returns the value in a map associated with a given key.
+  # \{#map_get map-get($map, $keys...)}
+  # : Returns the value in a map (or nested map) associated with a given series of keys.
```
nex3: "a map (or nested map)" -> "a (potentially nested) map"

"a given series of key" -> "the given key(s)"

lib/sass/script/functions.rb
```diff
@@ -196,11 +196,16 @@ module Sass::Script
...
   #
   # \{#map_merge map-merge($map1, $map2)}
   # : Merges two maps together into a new map.
+  # \{#map_merge map-merge($map1, $keys..., $map2)}
+  # : Merges a map into another map, nested with the specified keys
```
nex3: Here I would just not mention nesting. It's a narrow enough use-case that it can be described in the body documentation only.

lib/sass/script/functions.rb
```diff
@@ -196,11 +196,16 @@ module Sass::Script
...
   # updating the existing map in-place.
   #
-  # \{#map_get map-get($map, $key)}
-  # : Returns the value in a map associated with a given key.
+  # \{#map_get map-get($map, $keys...)}
+  # : Returns the value in a map (or nested map) associated with a given series of keys.
   #
   # \{#map_merge map-merge($map1, $map2)}
   # : Merges two maps together into a new map.
+  # \{#map_merge map-merge($map1, $keys..., $map2)}
```
nex3: It might be clearer to write this as `[$keys...]`, to emphasize that passing no keys is valid.

lib/sass/script/functions.rb
```diff
@@ -211,8 +216,8 @@ module Sass::Script
   # \{#map_values map-values($map)}
   # : Returns a list of all values in a map.
   #
-  # \{#map_has_key map-has-key($map, $key)}
-  # : Returns whether a map has a value associated with a given key.
+  # \{#map_has_key map-has-key($map, $keys...)}
+  # : Returns whether a map has a value associated with a given key or keys.
```
nex3: "key or keys" makes this sound like it checks multiple keys in the base map. Maybe "Returns whether a map has a (potentially nested) value associated with the given key(s)."

lib/sass/script/functions.rb
```diff
@@ -2053,15 +2058,24 @@ def list_separator(list)
...
-    def map_get(map, key)
+    def map_get(map, *keys)
       assert_type map, :Map, :map
-      map.to_h[key] || null
+      result = map
+      keys.each do |key|
+        if result.is_a?(Sass::Script::Value::Map)
+          result = result.to_h[key]
+        else
+          return null
```
nex3: Move this to the beginning of the loop and make it short-circuit (`return null unless ...`).

lib/sass/script/functions.rb
```diff
@@ -2053,15 +2058,24 @@ def list_separator(list)
...
     #   if the map doesn't contain the given key
     # @raise [ArgumentError] if `$map` is not a map
-    def map_get(map, key)
+    def map_get(map, *keys)
       assert_type map, :Map, :map
-      map.to_h[key] || null
+      result = map
+      keys.each do |key|
+        if result.is_a?(Sass::Script::Value::Map)
+          result = result.to_h[key]
```
nex3: Move `|| null` up here to preserve the invariant that `result` is always a Sass value.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2092,51 @@ def map_get(map, key)
     # @example
     #   map-merge(("foo": 1), ("bar": 2)) => ("foo": 1, "bar": 2)
     #   map-merge(("foo": 1, "bar": 2), ("bar": 3)) => ("foo": 1, "bar": 3)
+    #   map-merge(("foo": 1, "bar": 2), bar, 3) => ("foo": 1, "bar": 3)
+    #   map-merge((a: (b: c)), a, b, d) => (a: (b: d))
```
nex3: These examples are no longer correct.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2092,51 @@ def map_get(map, key)
...
     # @overload map_merge($map1, $map2)
     #   @param $map1 [Sass::Script::Value::Map]
     #   @param $map2 [Sass::Script::Value::Map]
     # @return [Sass::Script::Value::Map]
     # @raise [ArgumentError] if either parameter is not a map
-    def map_merge(map1, map2)
+    # @overload map_merge($map1, $keys..., $map2)
```
nex3, line 2081: This shouldn't be a separate overload. Same goes for other functions.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2092,51 @@ def map_get(map, key)
...
+    #   @param $map1 [Sass::Script::Value::Map]
+    #   @param $keys [[Sass::Script::Value]]
+    #   @param $map2 [Sass::Script::Value::Map]
+    # @return [Sass::Script::Value::Map]
+    # @raise [ArgumentError] if either parameter is not a map
+    def map_merge(map1, *keys)
       assert_type map1, :Map, :map1
+      map2 = keys.pop
+      if map1.is_a? Sass::Script::Value::List
```
nex3: Nit: use parens for `is_a?`.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2092,51 @@ def map_get(map, key)
...
+        map1 = map(map1.to_h)
+      end
```
nex3: If a function has to check for a list value explicitly to handle an empty map, either the function itself or the map API is wrong.

In this case, I don't think `recursive_set` or `recursive_merge` should be defined on `Map`. They should just be implemented (iteratively, not recursively) directly in the function bodies.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2092,50 @@ def map_get(map, key)
...
       assert_type map2, :Map, :map2
-      map(map1.to_h.merge(map2.to_h))
+      map1.recursive_merge(keys, map2)
     end
     declare :map_merge, [:map1, :map2]
+    declare :map_merge, [:map1], :var_args => true
+
+    # @example
+    #   map-set((a: 1), b, 2) => (a: 1, b: 2)
+    #   map-set((a: (a: 1)), a, b, 2) => (a: (a: 1, b: 2))
```
nex3: Using the same letter at multiple levels of nesting here is confusing.

lib/sass/script/functions.rb
HamptonMakes: That was the original code, but it was terribly hard to understand... and required multiple passes through the tree. Unless there is something obvious I'm missing on how to do that.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2092,50 @@ def map_get(map, key)
...
+    # @raise [ArgumentError] if the first parameter is not a map
+    def map_set(map, *keys)
+      assert_type map, :Map, :map
+      new_value = keys.pop
+      if map.is_a? Sass::Script::Value::List
+        map = map(map.to_h)
+      end
+      map.recursive_set(keys, new_value)
+    end
+    declare :map_set, [:map, :keys, :value], :var_args => true
```
nex3: This is the wrong `declare` format. This should have the same structure as `map-merge` and `map-get`.

lib/sass/script/functions.rb
HamptonMakes: I was only trying to avoid using "unless" inline, but fine changing it.

lib/sass/script/functions.rb
```diff
@@ -2110,7 +2160,8 @@ def map_remove(map, *keys)
       hash.delete_if {|key, _| keys.include?(key)}
       map(hash)
     end
-    declare :map_remove, [:map, :key], :var_args => true
+    declare :map_remove, [:map], :var_args => true
```
nex3: Why are you changing this?

lib/sass/script/functions.rb
```diff
@@ -2147,16 +2198,19 @@ def map_values(map)
...
-    def map_has_key(map, key)
+    def map_has_key(map, *keys)
       assert_type map, :Map, :map
-      bool(map.to_h.has_key?(key))
+      bool(keys.all? do |key|
+        map = map.to_h[key] || false
+      end)
     end
-    declare :map_has_key, [:map, :key]
+    declare :map_has_key, [:map, :keys], :var_args => true
```
nex3: This is also the wrong declaration. Users that used `$key` as a named parameter would be broken by this.

HamptonMakes: I've been struggling to understand how the `declare` syntax works.

lib/sass/script/functions.rb
```diff
@@ -2147,16 +2198,19 @@ def map_values(map)
...
-    #   @param $key [Sass::Script::Value::Base]
+    #   @param $keys [[Sass::Script::Value::Base]]
     # @return [Sass::Script::Value::Bool]
     # @raise [ArgumentError] if `$map` is not a map
-    def map_has_key(map, key)
+    def map_has_key(map, *keys)
       assert_type map, :Map, :map
-      bool(map.to_h.has_key?(key))
+      bool(keys.all? do |key|
+        map = map.to_h[key] || false
```
nex3: As I mentioned in the previous PR, `|| false` is redundant here.

lib/sass/script/value/map.rb
```diff
@@ -43,6 +43,50 @@ def eq(other)
       Bool.new(other.is_a?(Map) && value == other.value)
     end
 
+    # Walks the Map, directed by the list
```
nex3: These days, we wrap comments at 80 columns. Older functions don't do this, but new ones should.

test/sass/functions_test.rb
```diff
@@ -1419,8 +1450,10 @@ def test_map_values_checks_type
 
   def test_map_has_key
     assert_equal "true", evaluate("map-has-key((foo: 1, bar: 1), foo)")
+    assert_equal "true", evaluate("map-has-key((foo: (bar: 1)), foo, bar)")
     assert_equal "false", evaluate("map-has-key((foo: 1, bar: 1), baz)")
     assert_equal "false", evaluate("map-has-key((), foo)")
+    assert_equal "false", evaluate("map-has-key((foo: (bar: 1)), foo, bing)")
   end
```
nex3, line 1463: This _still_ doesn't have the additional tests I asked for.

lib/sass/script/functions.rb
nex3: Can you go into detail? What don't you understand about it?

nex3:
Please also get the tests passing!

lib/sass/script/functions.rb
HamptonMakes: So, coming back to this, the original implementation was done iteratively and only in the function itself. Suffice it to say that dealing with the frozen nature of maps and the multi-directional modifications that needed to happen, it was incredibly memory intensive. It was only when it dawned on me that I could attempt a recursive version, that anything approximating acceptably efficient code appeared.

Iterative algorithms are almost always more efficient than their recursive counterparts, but in this case, quickly "diving in" through the data structure, and only making the required changes, seems a better representation of what's happening here, than a traditional "recursive calculation". It's the fact that these are nested maps (like a graph) that make it reasonable to be recursive.

For now, I'm leaving it, but I cleaned up the API to simplify the above code. If you _really_ want me to attempt an iterative version again, I'll do it, but otherwise I'm leaving it.

lib/sass/script/functions.rb
HamptonMakes: First of all, I'd say I'm not 100% sure about named parameters. It's something I haven't really seen in the wild in Sass code, so I always forget that's a thing, so the idea of var_args declaring an "unnamed" variable seems uncomfortable to me.

lib/sass/script/functions.rb
nex3: Please do make the changes I suggested. It's always possible to convert a recursive algorithm to an iterative one by using an explicit stack, and doing so will eliminate any risk of stack overflows on large data structures. For example, here's an iterative version of `map_set`:

``` ruby
def map_set(map, *keys)
  assert_type map, :Map, :map
  root_map = map.to_h.dup

  current_map = root_map
  keys[1...-1].each do |key|
    value = current_map[key]
    next_map = value ? value.to_h.dup : {}

    current_map[key] = Map.new(next_map)
    current_map = next_map
  end
  current_map[keys[-2]] = keys[1]

  Map.new(root_map)
end
```

Also, you didn't address my other requests: not explicitly checking for a list and getting rid of the `Map.recursive_*` functions.

doc-src/SASS_CHANGELOG.md
```diff
@@ -10,6 +10,15 @@
...
 * Add a `$weight` parameter to `invert()`.
 
+* Add support for nested maps with `map-get()` and `map-has-key()` by passing
+  more than 2 arguments.
+
+* Add support for setting changes deeply into a nested hash by providing a new
+  syntax for set that is generally non-destructive with the signature
+  `map-set($map, $keys..., $value)`.
+
+* Add support for additional keys passed to `map-merge($map1, $keys..., $map2)`, allowing for nested Map merges.
```
nex3: Please limit Markdown lines to 80 characters.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2089,61 @@ def map_get(map, key)
...
+    def map_set(map, *keys)
+      assert_type map, :Map, :map
+      new_value = keys.pop
+      if map.is_a?(Sass::Script::Value::List)
+        map = map(map.to_h)
+      end
+      map.recursive_set(keys, new_value)
+    end
+    declare :map_set, [:map, :value]
+    declare :map_set, [:map, :keys], :var_args => true
```
nex3, line 2133: The names of var args should never appear in declarations. It's not valid to pass var args by name in Sass.

Once you finish with this PR, I'd like to talk to you about how to improve the documentation for `declare`, since it seems to be a major sticking point.

lib/sass/script/functions.rb
```diff
@@ -2078,17 +2089,61 @@ def map_get(map, key)
...
+    # @raise [ArgumentError] if the first parameter is not a map
+    def map_set(map, *keys)
+      assert_type map, :Map, :map
+      new_value = keys.pop
+      if map.is_a?(Sass::Script::Value::List)
+        map = map(map.to_h)
+      end
+      map.recursive_set(keys, new_value)
+    end
+    declare :map_set, [:map, :value]
```
nex3, line 2132: Get rid of this declaration. This isn't a valid way of calling `map-set()`; `map-set((a: b), c)` doesn't make any sense.

test/sass/functions_test.rb
```diff
@@ -1368,12 +1378,21 @@ def test_map_get_checks_type
...
     assert_equal("(foo: 1, bar: 2)",
       perform("map-merge((foo: 1, bar: 2), ())").to_sass)
+
+    # Here, we test when the second argument is an array of keys, plus a value
+    assert_equal("(a: b, c: d)",
+            perform("map-merge((a: b), (c: d))").to_sass)
+    assert_equal("(a: (b: c, d: e))",
+            perform("map-merge((a: (b: c)), a, (d: e))").to_sass)
+    assert_equal("(a: (d: e))",
+            perform("map-merge((a: (b: c)), (a: (d: e)))").to_sass)
```
nex3, line 1397: Test nesting deeper than actually exists, nesting through values that aren't maps, and nesting through empty maps. That goes for all of these functions that don't already test these cases.

nex3:
Done reviewing!

test/sass/functions_test.rb
```diff
@@ -1381,6 +1400,18 @@ def test_map_merge_checks_type
...
+    assert_equal("(foo: 1, bar: (bing: ping))",
+      perform("map-set((foo: 1, bar: (bing: bong)), bar, bing, ping)").to_sass)
+    assert_equal("(foo: 1, bar: (bing: bong, ping: pong))",
+      perform("map-set((foo: 1, bar: (bing: bong)), bar, ping, pong)").to_sass)
+    # Create a nested Map, if it didn't exist already
+    assert_equal("(a: (b: c))",
+      perform("map-set((), a, b, c)").to_sass)
+    assert_equal("(b: d, a: (b: c))",
+      perform("map-set((b: d), a, b, c)").to_sass)
+  end
```
nex3, line 1415: Test what happens if too few arguments are passed.

lib/sass/script/functions.rb
HamptonMakes: Diving back in to implement this, and have smacked back into the problem that I had the first go around with an iterative approach... which is the same reason that your above code doesn't work. The line where you modify the "current_map[keys[-2]]" doesn't work because you have created new frozen hashes all along the process. Which, we can counter for. But, the fact we are dup-ing like this, means that we throw away all branches of the tree that we don't visit. This walks in like a laser, and nukes everything _not_ in it's path. And, the frozen hashes make walking-and-duplicating the tree pretty tricky, since there isn't an easy way to copy the tree.

Hence, my going with recursive, which got around all those issues with pretty minimal code.

I'll keep seeing if I can figure out a way to duplicate-and-modify the frozen tree.

chriseppstein:
@nex3 Can we land this in `next`?

nex3:
This is currently assigned to @hcatlin and marked as a work in progress—I wasn't aware it was ready for another round of review, let alone a candidate for landing.

chriseppstein:
@nex3 I'm willing to help wrangle it. My question was whether this feature set could go out in the 3.5 release.

nex3:
Oh, yes, it could. If you want to take charge of merging it in, feel free to add it to the 3.5 milestone and assign yourself.

HamptonMakes:
Yeah, I took another couple swings at doing a non-recursive version and failed twice... So this fell from my radar.

nex3:
I don't think @chriseppstein ended up having the cycles to work on this, so I'm going to close it out. If someone else wants to adopt in in the future, I'll re-open.

PR #1911, opened 2015-11-23
Change assert_sourcemaps_equal to just assert on the ranges' #inspect output

### LINKED ISSUE #1910: Testing: assert_ranges_equal file assertion unhelpful
While working on @nex3's comments on #1875, switching to `assert_parses_with_mapping`, it was super hard to diagnose the following error:

```
  1) Failure:
SourcemapTest#test_selectors_properties_sourcemap_sass [/Users/srawlins/code/sass/test/sass/source_map_test.rb:397]:
Input.
--- expected
+++ actual
@@ -1 +1 @@
-"test_selectors_properties_sourcemap_sass_inline.sass"
+nil
```

This comes from [`assert_ranges_equal`](https://github.com/sass/sass/blob/stable/test/sass/source_map_test.rb#L922..L926).

srawlins:
Fixes #1910 

Now these test failures look like:

```
  1) Failure:
SourcemapTest#test_selectors_properties_sourcemap_sass [/Users/srawlins/code/sass/test/sass/source_map_test.rb:397]:
Mapping input.
--- expected
+++ actual
@@ -1 +1 @@
-"(16:3 to 16:21 in test_selectors_properties_sourcemap_sass_inline.sass)"
+"(16:3 to 16:21)"
```

The alternate route would be a sophisticated failure message like in `assert_positions_equal`, but IMHO, this messaging is perfectly readable and actionable. Let me know if you'd prefer that route.

### DIFF
--- a/test/sass/source_map_test.rb
+++ b/test/sass/source_map_test.rb
@@ -903,28 +903,6 @@ JSON
     assert_sourcemaps_equal source, css, mapping, sourcemap
   end
 
-  def assert_positions_equal(expected, actual, lines, message = nil)
-    prefix = message ? message + ": " : ""
-    expected_location = lines[expected.line - 1] + "\n" + ("-" * (expected.offset - 1)) + "^"
-    actual_location = lines[actual.line - 1] + "\n" + ("-" * (actual.offset - 1)) + "^"
-    assert_equal(expected.line, actual.line, prefix +
-      "Expected #{expected.inspect}\n" +
-      expected_location + "\n\n" +
-      "But was #{actual.inspect}\n" +
-      actual_location)
-    assert_equal(expected.offset, actual.offset, prefix +
-      "Expected #{expected.inspect}\n" +
-      expected_location + "\n\n" +
-      "But was #{actual.inspect}\n" +
-      actual_location)
-  end
-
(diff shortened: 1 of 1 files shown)


nex3:
This looks like a substantial downgrade in readability when the ranges are misaligned. Compare:

```
Output start position: Expected 2:5
  y: foobar; }
----^

But was 2:6
  y: foobar; }
-----^.
Expected: 5
  Actual: 6
```

to

```
Mapping output.
Expected: "(2:5 to 2:12)"
  Actual: "(2:6 to 2:12)"
```

It's much easier to visually understand the failure in the previous example.

I don't think error for a missing filename is that confusing—like most assertion failures, you just have to look at the listed line to see what's going on—but if you want to make it clearer, I think the way forward is to add a failure message to the `assert_equal` call rather than scrapping all the existing presentation logic for the line and column numbers.

srawlins:
OK, I can do that change instead.

The original message reported in #1910 gives no indication as to what node is missing a file. There is no way to solve it without writing new debug messaging in `#assert_ranges_equal` or similar.

nex3:
In that case I'd include the messaging in the failure message.

nex3:
Are you planning on modifying this to clarify the failure message, or should I close it out?

PR #1919, opened 2015-12-02
Include import origin in import error message

onli:
This aims to improve the error message when a file to be imported is not found.

In a project using sass via jekyll we ran into confusion when the sass -> css conversion did not work. After some debugging it was found that the origin was simply a hidden @import (in a file that was itself imported, so not initially visible) so of a file that did not exist, or rather did exist only locally on the designers machine. This was overseen because the file name to be imported was the same as a folder name, misleading the initial debugging effort.

Mentioning where the file is imported would've helped, and this is what this small change does. 

If further work is needed (I was not sure about a test?) please just say so.

### DIFF
--- a/lib/sass/tree/import_node.rb
+++ b/lib/sass/tree/import_node.rb
@@ -55,7 +55,7 @@ module Sass
           return f if f
         end
 
-        lines = ["File to import not found or unreadable: #{@imported_filename}."]
+        lines = ["File to import in #{options[:filename]} not found or unreadable: #{@imported_filename}."]
 
         if paths.size == 1
           lines << "Load path: #{paths.first}"
--- a/test/sass/engine_test.rb
+++ b/test/sass/engine_test.rb
@@ -680,7 +680,7 @@ SASS
 
   def test_nonexistent_import
     assert_raise_message(Sass::SyntaxError, <<ERR.rstrip) do
-File to import not found or unreadable: nonexistent.sass.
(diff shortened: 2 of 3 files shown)


davidkpiano:
Looks like you need to update these tests:
- https://github.com/sass/sass/blob/stable/test/sass/engine_test.rb#L681
- https://github.com/sass/sass/blob/stable/test/sass/engine_test.rb#L689
- https://github.com/sass/sass/blob/stable/test/sass/plugin_test.rb#L167

onli:
Thanks, and sorry for not doing that initially.

I tried to change the tests accordingly. That it still fails I can't really explain – the line too long error  in https://travis-ci.org/sass/sass/jobs/95212050 is a style check, and I should just break the line? But why would this then only happen in specific ruby versions?

nex3:
This exception should already include a stack trace that has the file context—that's what [this line](https://github.com/onli/sass/blob/667f7f1cbfe27805a79d7dd3a4f3155ac1a9ae78/lib/sass/tree/import_node.rb#L67) is doing. Have you run into a situation where that stack trace is incorrect or omitted?

onli:
> Have you run into a situation where that stack trace is incorrect or omitted?

Yes. When building a site using sass via jekyll, that stack trace was invisible to us.

nex3:
> Yes. When building a site using sass via jekyll, that stack trace was invisible to us.

This sounds like a bug in Jekyll's Sass integration. Sass provides this information, but it's up to the code that uses it to expose it.

PR #1926, opened 2015-12-07
try to run/test on recent JRuby

mprins:


nex3:
Thanks for the patch!

PR #1961, opened 2016-01-08
Work with ruby 2.3's --enable-frozen-string-literal

jeremyevans:
These changes are the minimal ones necessary to allow Roda's specs
to pass.  There may well be other changes that are required.

### DIFF
--- a/lib/sass/scss/parser.rb
+++ b/lib/sass/scss/parser.rb
@@ -1022,7 +1022,7 @@ WARNING
       end
 
       def str
-        @strs.push ""
+        @strs.push String.new("")
         yield
         @strs.last
       ensure
--- a/lib/sass/tree/visitors/to_css.rb
+++ b/lib/sass/tree/visitors/to_css.rb
@@ -12,7 +12,7 @@ class Sass::Tree::Visitors::ToCss < Sass::Tree::Visitors::Base
     @tabs = 0
     @line = 1
     @offset = 1
-    @result = ""
+    @result = String.new("")
     @source_mapping = build_source_mapping ? Sass::Source::Map.new : nil
     @lstrip = nil
     @in_directive = false


chriseppstein:
Our style checker barfs on bare `String.new`. I'm not sure how to configure it correctly, so I filed an issue on rubocop: https://github.com/bbatsov/rubocop/issues/2645

matthewrudy:
can I suggest

``` ruby
String.new("")
```

Because `String.new` returns a string with ASCII encoding.
While `String.new("")` returns a string with the encoding of your source file.

```
>> "".encoding
=> #<Encoding:UTF-8>
>> String.new.encoding
=> #<Encoding:ASCII-8BIT>
>> String.new("").encoding
=> #<Encoding:UTF-8>
```

note: I'm assuming you're already working with frozen literals,

if you need to support both, then

``` ruby
String.new("".freeze)
```

jeremyevans:
Note that the encoding of the empty string doesn't generally matter, as the string's encoding will change to UTF-8 if you append UTF-8 data to it:

``` ruby
a = String.new
a.encoding # => #<Encoding:ASCII-8BIT>
a << "\u2004"
a.encoding => #<Encoding:UTF-8>
```

Considering bare `String.new` is simpler and faster, unless there is a real world use case where it causes problems, I would recommend using it.

nex3:
@jeremyevans I'm happy to merge something like this in, but I'll need you to merge with the latest stable and figure out how to get Rubocop happy if it's still complaining.

jeremyevans:
I updated the commit to apply to the current stable branch, and switched to `String.new("")` so hopefully rubocop won't complain.

nex3:
Can you rebase one more time? A dependency issue was breaking Travis, which should be fixed now.

jeremyevans:
OK, I rebased again.

nex3:
Rad, thanks!

PR #1973, opened 2016-01-19
Don't pet that kitty. Make a donation.

chriseppstein:
The kitty petting thing is creepy. I've removed it and made a more socially responsible recommendation instead (to donate to a non-profit). I also updated my info while I was in there.


jina:
all looks good to me!

HamptonMakes:
How about make a donation to us. ;)

chriseppstein:
@hcatlin When we can take donations online, let's add that.

davidkpiano:
I'd make a note that Eyeglass is for Node-Sass. Might cause confusion, as this is the Ruby Sass repo.

nex3:
:+1:

PR #2016, opened 2016-03-12
Fixes most ruby warnings (master)

glebm:
Rake since v11+ defaults to showing ruby warnings in the test task,
generating lots of noise when the tests use the sass gems.

This fixes most of the warnings present on the compilation code path in
Sass.

### DIFF
--- a/lib/sass/callbacks.rb
+++ b/lib/sass/callbacks.rb
@@ -51,12 +51,12 @@ module Sass
     def define_callback(name)
       class_eval <<RUBY, __FILE__, __LINE__ + 1
 def on_#{name}(&block)
-  @_sass_callbacks ||= {}
+  @_sass_callbacks = {} unless defined? @_sass_callbacks
   (@_sass_callbacks[#{name.inspect}] ||= []) << block
 end
 
 def run_#{name}(*args)
-  return unless @_sass_callbacks
+  return unless defined? @_sass_callbacks
   return unless @_sass_callbacks[#{name.inspect}]
   @_sass_callbacks[#{name.inspect}].each {|c| c[*args]}
 end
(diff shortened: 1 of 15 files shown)


glebm:
@chriseppstein Any blockers to merging this?

chriseppstein:
@glebm Well, @nex3 has consistently rejected patches that fix this so far, arguing that ruby warnings are not in line with our preferred coding standard for this project. So I will let her weigh in. Personally, I am not opposed to Sass running without warnings when running with `-w` set.

chriseppstein:
I will note that I've had to limit Sass to running with rake < 10.5 because of this. For some reason I couldn't figure out how to disable the warnings within rake.

glebm:
@chriseppstein This is undocumented in rake, but to disable warnings, `verbose` also needs to be disabled [like here](https://github.com/twbs/bootstrap-rubygem/blob/23b55fc0e064a9ccc105320ba8f05c99b7fad61a/Rakefile#L8-L9). I don't know how to make it only hide the Sass warnings but not other warnings.

jodosha:
@nex3 @chriseppstein Please consider to merge this for the Community sanity.

Sass is a dependency for hundreds of gems and thousands of applications that will see these warnings (and they are a lot). Of course they can decide to opt-out for Ruby warnings, but that prevents OSS maintainers to keep them enabled.

:pray: 

---

**UPDATE:** Here's an example of what I'm saying: a Travis CI build that fails and it reports:

> This log is too long to be displayed. Please reduce the verbosity of your build or download the raw log.

screen shot 2016-03-22 at 19 14 22

Source: https://travis-ci.org/hanami/assets/jobs/117776354

chriseppstein:
I've landed the other PR to stable and merged to all other branches in active development. No more warnings are emitted.

jodosha:
@chriseppstein Thank you for that merge and for maintaining Sass :green_heart:

PR #2020, opened 2016-03-27
docs: add examples about behavior of null

xi:
I was wandering how I can find out whether a variable is `null`. I guess the right answer is simply `$var == null`, but coming from JavaScript I am careful about stuff like this (and I guess others are, too).

I found that the docs are not very clear about the behavior of `null`, so I added some examples. I did not find a good place for the `$var == null` example though. If you have an idea where this could fit in I would like to add this, too.

### DIFF
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -2023,6 +2023,9 @@ is compiled to:
     p {
       border: 1px solid; }
 
+You can explicitly test for `$var == false` or `$var == null` if you want to
+distinguish between these.
+
 The `@if` statement can be followed by several `@else if` statements
 and one `@else` statement.
 If the `@if` statement fails,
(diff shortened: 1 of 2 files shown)


chriseppstein:
In Sass there are only two "falsy" values: `null` and `false`. All other values are "truthy" including `0` and empty strings, lists, and maps. You can explicitly test for `$var == false` or `$var == null` if you want to distinguish between these. In most cases, the best solution is to just use the value as the condition and let Sass cast it to a boolean for you. E.g. `@if $var { ... }` or `if($var, $if-true, $if-null)`.

I think a good place to add this would be to the documentation for `@if` and to reference it for the documentation for the `if()` function. A small note could also be made in the `null` data type section.

xi:
The fact that `false` and `null` are the only falsy values is already well documented. I now added a note about distinguishing the two in the `@if` section. I guess it is not important enough to be repeated in the `if()` reference.

chriseppstein:
OK. Thanks!

PR #2040, opened 2016-04-28
Logger & Number threadsafety

chriseppstein:
These global settings make sass unreliable in a concurrent environment. Using Thread locals now to maintain the present API while avoiding concurrency issues.

### DIFF
--- a/lib/sass/script/value/number.rb
+++ b/lib/sass/script/value/number.rb
@@ -34,30 +34,36 @@ module Sass::Script::Value
     attr_accessor :original
 
     def self.precision
-      @precision ||= 5
+      Thread.current[:sass_numeric_precision] || Thread.main[:sass_numeric_precision] || 5
     end
 
     # Sets the number of digits of precision
     # For example, if this is `3`,
     # `3.1415926` will be printed as `3.142`.
+    # The numeric precision is stored as a thread local for thread safety reasons.
+    # To set for all threads, be sure to set the precision on the main thread.
     def self.precision=(digits)
-      @precision = digits.round
-      @precision_factor = 10.0**@precision
-      @epsilon = 1 / (@precision_factor * 10)
+      Thread.current[:sass_numeric_precision] = digits.round
+      Thread.current[:sass_numeric_precision_factor] = 10.0**precision
+      Thread.current[:sass_numeric_epsilon] = 1 / (precision_factor * 10)
+
+      Thread.current[:sass_numeric_precision]
     end
 
     # the precision factor used in numeric output
     # it is derived from the `precision` method.
     def self.precision_factor
-      @precision_factor ||= 10.0**precision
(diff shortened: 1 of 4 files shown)


lib/sass/script/value/number.rb
```diff
@@ -34,31 +34,37 @@ class Number < Base
...
+    # To set for all threads, be sure to set the precision on the main thread.
     def self.precision=(digits)
-      @precision = digits.round
-      @precision_factor = 10.0**@precision
-      @epsilon = 1 / (@precision_factor * 10)
+      Thread.current[:sass_numeric_precision] = digits.round
+      Thread.current[:sass_numeric_precision_factor] = 10.0**precision
+      Thread.current[:sass_numeric_epsilon] = 1 / (precision_factor * 10)
+
+      Thread.current[:sass_numeric_precision]
```
nex3, line 50: I don't think you need to explicitly return this. I'm pretty sure functions whose names end with `=` _always_ implicitly return their argument.

lib/sass/script/value/number.rb
```diff
@@ -34,31 +34,37 @@ class Number < Base
...
     # `3.1415926` will be printed as `3.142`.
+    # The numeric precision is stored as a thread local for thread safety reasons.
+    # To set for all threads, be sure to set the precision on the main thread.
     def self.precision=(digits)
-      @precision = digits.round
-      @precision_factor = 10.0**@precision
-      @epsilon = 1 / (@precision_factor * 10)
+      Thread.current[:sass_numeric_precision] = digits.round
+      Thread.current[:sass_numeric_precision_factor] = 10.0**precision
+      Thread.current[:sass_numeric_epsilon] = 1 / (precision_factor * 10)
```
nex3, line 48: Rather than duplicating these here, it might be better to just let them be lazily initialized in the getters.

lib/sass/script/value/number.rb
```diff
@@ -34,31 +34,37 @@ class Number < Base
...
 
     # Used in checking equality of floating point numbers. Any
     # numbers within an `epsilon` of each other are considered functionally equal.
     # The value for epsilon is one tenth of the current numeric precision.
     def self.epsilon
-      @epsilon ||= 1 / (precision_factor * 10)
+      Thread.current[:sass_numeric_epsilon] ||= 1 / (precision_factor * 10)
     end
 
+    self.precision = 5 # sets the default on the main thread (probably)
```
nex3, line 66: Is this really necessary? The precision getter already defaults to 5.

nex3:
Added a few comments!

lib/sass/script/value/number.rb
chriseppstein, line 50: 👍

chriseppstein:
@nex3 Thanks for reviewing. Commit 6e8a053f3f1b3c58d3c7db057d7df80e8fee3752 has the changes you suggested.

PR #2046, opened 2016-05-03
Trailing commas

chriseppstein:
This PR builds upon #1878 but addresses code review comments made in that PR as well as adds more tests and implements the corresponding behavior for definition arglists.

Sass spec tests are in https://github.com/sass/sass-spec/pull/860 and will land after this.


chriseppstein:
@royels FYI this is landing finally.

Issue #1582, opened 2015-01-05
Content functions.


chriseppstein:
I'd like to introduce a function `content()` that returns a Sass data structure that represents the css specified as mixin content that is normally placed by the `@content` directive. The exact form of this data is something that would need to be well specified, I will work on that if people are in agreement that this is a good idea.

I also think Sass should provide a mixin that effectively translates this data back into CSS. E.g. `@include expand-content(content());` would be identical to `@content;` when used inside a mixin. Sass doesn't currently expose any mixins automatically so we probably just want to make a special import location so that Sass mixins can be manually imported if they are needed. `@import 'stdlib'` or something?

Here's the rationale. I see a lot of users clamoring to use SassScript to manipulate their CSS. To do this, they end up defining their CSS as maps instead of as CSS. I'd like things that are stylesheets to be specified in CSS syntax, not in some it's a map and it's almost css syntax just so that it can be manipulated.

Another use case is that, https://github.com/oddbird/true needs a way to verify that the output of a mixin is correct without relying on string-based output comparison tests. By providing this capability, the css output of a mixin can be tested using SassScript.

In the past, I've proposed special directives to accomplish this, but I don't think that's necessary.

stef-rausch:
Indeed that would be great to have!

Regarding the testing of `Sass` / `mixin`s output, there is currently a viable way via [SassUnit](https://github.com/penman/SassUnit). It works really well.

jakob-e:
:+1:  +1

jakob-e:
This may be a bit off topic but when talking content manipulation – how about using @content as a kind of getter-setter (if possible working like "map-merge" to reduce CSS output). I know this example has a lot of unresolved – like nested selector output, shorthands and fallbacks (font:1em serif; font-size:1rem; font-size: 16px;). It was not my intention to spam :-) 

``` SCSS

@mixin a(){
  left: 0;
  right: 100px;
}

@mixin b(){
  left: 100px;
  top: 100px;
}

@mixin c(){
  // Passed @content will be present before any modifications 

  // Set @content (merge)
  // Note! 
  // ; is a separator 
  // will require content to be "@at-root" (not inside selectors)
  @content( 
    @include a();
    @include b();
  );

  // Looking up property
  @if content(top) {
    // Merge properties 
    @content( bottom: 100px; ); 
    // Remove properties
    @content( width: null; height: null; ); 
  }

  // Print content 
  @content;
}


.class {
  @include c(){
    position: absolute;
[...]

Issue #871, opened 2013-07-30
Passing arguments from a mixin to a content block


chriseppstein:
- [x] [Proposal](https://github.com/sass/language/blob/master/proposal/content-args.md)
- [x] [Specs](https://github.com/sass/sass-spec/pull/1299)
- [x] [Documentation](https://github.com/sass/sass-site/issues/250)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/498)
- [x] [LibSass](https://github.com/sass/libsass/issues/2723)

---

Mixin content is great and has done wonders for code implementing wrapper and context abstractions. However, there is often some setup done in the mixin and that setup needs to be available to the content when it is included. The current strategy is to use global variables, which works, but is obviously not ideal.

I'd like to introduce a way for mixins to include their content block and to pass arguments to the content.
### Calling the content with arguments

In order to pass arguments to content blocks an argument list can be passed to the content directive. This accepts positional and keyword arguments for passing to a content block. Variable argument semantics are allowed.

``` scss
@mixin accepts-content {
  @for $i from 1 to 5 {
    @content ($i, 360deg * $i / 5);
  }
}
```
### Receiving arguments within a content block

A content block receives arguments using a new directive called `@receive`. The directive allows an argument list declaration that is the same as for functions and mixins. Positional arguments, default values, and variable arguments are all allowed.

``` scss
@include accepts-content {
  @receive ($number, $hue);
  .color-#{$number} {
    background-color: hsl($hue, 75%, 90%);
  }
}
```

Snugug:
I like this idea. :+1:

cimmanon:
Not crazy about the syntax for `@receive`, but very interested in the concept.

chriseppstein:
Comments of general distaste are not very useful without explanations and, ideally, alternative suggestions.

gabrielnau:
It's really fine ! 
1. global variables are not ideal like you said 
2. it explicits what variables the nested block needs

HamptonMakes:
At first blush, I like it.

How about @bind instead of @receive... maybe that's too nerdy?

ghost:
Could always pull a JavaScript and make an $arguments variable. Or an arguments($n) function.

chriseppstein:
The problem with `$arguments` is that it only work for positional arguments. I'd like consistency with all other argument passing and definitions. With `arguments(...)` I really don't like the idea of a function that sets instead of receives arguments.

[...]

Issue #472, opened 2012-08-06
anonymous functions


chriseppstein:
This code is in a patch for compass that I'm about to approve and I really feel like this can be cleaned up dramatically by adding some basic functional programming abilities to Sass.

In particular the importance of lists in css and the need to manipulate them using side-effects is causing severe pain in many of the more complex sass projects that I've seen under development recently.

Consider this code written in Sass's current syntax:

``` scss
@mixin rem($property, $values, $use-px-fallback: $rem-with-px-fallback) {
    // Create a couple of empty lists as output buffers.
    $px-values: ();
    $rem-values: ();

    // Ensure $values is a list.
    @if type-of($values) != 'list' {
        $values: join((), $values);
    }

    // Loop through the $values list
    @each $value in $values {
        // For each property value, if it's in rem or px, derive both rem and
        // px values for it and add those to the end of the appropriate buffer.
        // Ensure all pixel values are rounded to the nearest pixel.
        @if type-of($value) == number and not unitless($value) and (unit($value) == px or unit($value) == rem) {
            @if unit($value) == px {
                $px-values: join($px-values, round($value));
                $rem-values: join($rem-values, convert-length($value, rem));
            }
            @else {
                $px-values: join($px-values, round(convert-length($value, px)));
                $rem-values: join($rem-values, $value);
            }
        }
        @else {
            $px-values: join($px-values, $value);
            $rem-values: join($rem-values, $value);
        }
    }

    // Use pixel fallback for browsers that don't understand rem units.
    @if $use-px-fallback {
        #{$property}: $px-values;
    }

    // Use rem values for everyone else (overrides pixel values).
    #{$property}: $rem-values;
}
```

I'm not sure what the syntax should be, but so that we have something to throw darts at I propose that if the first statement of a ruleset is `->(<argument list>);` then the block would become an anonymous function if passed to a function. This same syntax could also be used by anonymous content blocks to receive arguments from mixins to their `@content` blocks.

``` scss
.mapped {
  foo: map-to(1 2 3) {->($v);
    @return $v * 2;
  };
}
```

which would emit:

``` css
.mapped {
  foo: 2 4 6;
}
```

This would necessitate the need for a built-in function that can be used in `@function` to call the anonymous function. Again, maybe there's a better name and semantics, but let's start with `yield`:

``` scss
@function foo($arg) {
  @return yield($arg * 2);
}

.foo {
  prop: foo(30px) {->($a); @return $a * 2; }
}
```

would compile to:

``` css
[...]

PR #2057, opened 2016-05-11
Dynamic @include. Closes #626.

### LINKED ISSUE #626: Dynamic mixin invocation

---

### Inside Extension

``` scss

  @if $output == 'isolation' {
    @include isolation($input);
  }
  @else if $output == 'float' {
    @include float($input);
  }
}

@mixin isolation($input) {…}
@mixin float($input) {…}
```

``` scss

@mixin table($input) {…}
```
### Inside User's file

``` scss
@debug $output-styles;
```

``` bash
```

``` scss
  @if $output == 'isolation' {
    @include isolation($input);
  }
  @else if $output == 'float' {
    @include float($input);
  }
@else if $output == 'table' {
    @include table($input);
  }
}
```

### Inside Extension

``` scss

    @if $output == $style {
    }
  }
}

@mixin isolation($input) {…}
@mixin float($input) {…}
```

``` scss

@mixin table($input) {…}
```
### Inside User's file

chriseppstein:
Implements the most requested feature: #626.

The following tasks remain:
- [x] sass-spec tests.
- [x] CHANGELOG.
- [x] Language Reference docs

This feature adds the ability to include a mixin dynamically. It is the mixin equivalent to the `call()` function. A new "magic" mixin named `mixin` is exposed who's behavior is to take in the arguments and call another mixin, taking the first positional argument as the mixin name to call. The mixin name can be passed a normal positional argument or it can be passed as a "splat" argument. 

We are not deprecating any mixin named `mixin` at this time because 3.5 will only break backwards compat for CSS syntax support. Like all global functions defined by the Sass core, this mixin can be overridden by a user-defined mixin named `mixin`. Instead, if a user defined a mixin named `mixin` they will not be able to use dynamic includes, nor will any of the libraries that they include. In Sass 4.0, the special mixin named `mixin` can be `@use`d thus preventing any possible namespace collision.

Given that in practice, people probably don't have a mixin named `mixin` this seems like an acceptable risk and one that is easily worked around.

Code review summary:
- [x] Add backtrace info for dynamic calls
- [x] Simply initialization of mixin_name
- [ ] Remove Test::Unit tests in favor of sass-spec tests
- [x] Test backtrace line #'s
- [x] Test passing keyword arguments both explicitly and through a captured arglist

### DIFF
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -2534,6 +2534,53 @@ passed block are related to the other styles around where the block is defined. 
       }
     }
 
+### Including a Mixin Dynamically
+
+There are times when it is convenient to call a mixin based on the
+result of a SassScript expression.
+
+For example, a library might want to delegate some styling to the
+application by allowing the application to register a mixin that the
+library would include where appropriate. Sometimes a content block is
+sufficient for that purpose but in many cases it is not. So instead a
+library can provide a way to register a mixin, or a library could have a
+naming convention and check if a user has defined such a mixin (via
+`mixin-exists`) based on a name constructed from a script expression.
+
+To including a mixin dynamically, Sass provides a built-in mixin named
+`mixin` that has the signature `@mixin mixin($mixin-name, $args...)`
+which will include the mixin named `$mixin-name` by passing it all the
+argument remaining arguments. This is similar to the built-in `call`
+method.
+
+Example:
+
+    @mixin border-styles($theme-base-color) {
+      border: 1px dashed darken($theme-base-color, 10%);
+      border-radius: 10px;
+    }
+    @mixin background-styles($theme-base-color) {
+      background: $theme-base-color url(logo.png);
+    }
+    .my-fancy-element {
+      @each $type in (background, font, border, color) {
+        $mixin-name: "#{$type}-styles";
+        @if mixin-exists($mixin-name) {
+          @include mixin($mixin-name, red);
+        }
+      }
+    }
+
+Compiles to:
+
+    .my-fancy-element {
+      background: red url(logo.png);
+      border: 1px dashed #cc0000;
+      border-radius: 10px;
+    }
+
+Note: The mixin name when performing a dynamic include must be passed as
+the first positional argument; it cannot be passed keyword-style.
 
 ## Function Directives {#function_directives}
 
--- a/lib/sass/tree/visitors/perform.rb
+++ b/lib/sass/tree/visitors/perform.rb
@@ -346,29 +346,47 @@ WARNING
 
   # Runs a mixin.
   def visit_mixin(node)
+    mixin_name = nil
     @environment.stack.with_mixin(node.filename, node.line, node.name) do
-      mixin = @environment.mixin(node.name)
-      raise Sass::SyntaxError.new("Undefined mixin '#{node.name}'.") unless mixin
-
-      if node.children.any? && !mixin.has_content
-        raise Sass::SyntaxError.new(%(Mixin "#{node.name}" does not accept a content block.))
-      end
-
       args = node.args.map {|a| a.perform(@environment)}
       keywords = Sass::Util.map_vals(node.keywords) {|v| v.perform(@environment)}
       splat = self.class.perform_splat(node.splat, keywords, node.kwarg_splat, @environment)
+
+      mixin_name = node.name
+      # We let any existing mixin named `mixin` shadow dynamic includes
+      # for backwards compatability even though it will break the hell
+      # out of libraries if anyone does this.
+      mixin = @environment.mixin(mixin_name)
+
+      # The while loop is to allow any number of dynamic includes to be chained.
+      while mixin.nil? && mixin_name == "mixin"
+        if args.first.is_a?(Sass::Script::Value::String)
+          mixin_name = args.shift.value
+        elsif splat.value.first.is_a?(Sass::Script::Value::String)
+          n, splat = splat.shift
+          mixin_name = n.value
+        else
+          raise Sass::SyntaxError.new("First argument to a dynamic include must be a string.")
+        end
+        mixin = @environment.mixin(mixin_name)
+      end
+      raise Sass::SyntaxError.new("Undefined mixin '#{mixin_name}'.") unless mixin
+
+      if node.children.any? && !mixin.has_content
+        raise Sass::SyntaxError.new(%(Mixin "#{mixin_name}" does not accept a content block.))
+      end
 
       self.class.perform_arguments(mixin, args, splat, @environment) do |env|
         env.caller = Sass::Environment.new(@environment)
         env.content = [node.children, @environment] if node.has_children
 
-        trace_node = Sass::Tree::TraceNode.from_node(node.name, node)
+        trace_node = Sass::Tree::TraceNode.from_node(mixin_name, node)
         with_environment(env) {trace_node.children = mixin.tree.map {|c| visit(c)}.flatten}
         trace_node
       end
     end
   rescue Sass::SyntaxError => e
-    e.modify_backtrace(:mixin => node.name, :line => node.line)
+    e.modify_backtrace(:mixin => mixin_name || node.name, :line => node.line)
     e.add_backtrace(:line => node.line)
     raise e
   end
--- a/lib/sass/tree/visitors/perform.rb
+++ b/lib/sass/tree/visitors/perform.rb
@@ -347,6 +347,7 @@ WARNING
   # Runs a mixin.
   def visit_mixin(node)
     mixin_name = node.name
+    dynamic_calls = 0
     @environment.stack.with_mixin(node.filename, node.line, node.name) do
       args = node.args.map {|a| a.perform(@environment)}
       keywords = Sass::Util.map_vals(node.keywords) {|v| v.perform(@environment)}
@@ -368,6 +369,7 @@ WARNING
           raise Sass::SyntaxError.new("First argument to a dynamic include must be a string.")
         end
         mixin = @environment.mixin(mixin_name)
+        dynamic_calls += 1 if mixin
       end
       raise Sass::SyntaxError.new("Undefined mixin '#{mixin_name}'.") unless mixin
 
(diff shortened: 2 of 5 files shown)


doc-src/SASS_REFERENCE.md
```diff
@@ -2534,6 +2534,53 @@ passed block are related to the other styles around where the block is defined.
...
     }
 
+### Including a Mixin Dynamically
+
+There are times when it is convenient to call a mixin based on the
+result of a SassScript expression.
+
+For example, a library might want to delegate some styling to the
+application by allowing the application to register a mixin that the
+library would include where appropriate. Sometimes a content block is
```
nex3, line 2544: This use-case won't work well in a module world—the application's mixin won't be in scope in the library context where the dynamic include is invoked. To allow users to "pass in" mixins, we'd probably need a combination of #1582 and #871 so that users can do something like:

``` scss
@use "fancy";

@include fancy.register-style("background") using ($color) {
  background: $color url(logo.png);
}
```

If we think users are likely to use this functionality as a means to pass mixin definitions upstream to libraries, we should discuss whether it makes sense to land it right now. I'm very worried about adding features that will make it hard or impossible for users to upgrade to 4.0 without breaking functionality.

chriseppstein, line 2544: @nex3 We have this problem with `call` already. This is a common use case for that function. I think 4.0 will have to have a way of passing a function or mixin reference that isn't a string so the lookup can be done in one module context and used in another.

E.g. `function($function-name)` and `mixin($mixin-name)` would return an object of type `reference` which would include whatever internal binding information we need. The `call` function and `mixin` mixin would then be modified to accept references as well as strings. I would also note that this could enable #472 down the road.

lib/sass/tree/visitors/perform.rb
```diff
@@ -346,29 +346,47 @@ def visit_mixindef(node)
...
       splat = self.class.perform_splat(node.splat, keywords, node.kwarg_splat, @environment)
 
+      mixin_name = node.name
+      # We let any existing mixin named `mixin` shadow dynamic includes
+      # for backwards compatability even though it will break the hell
+      # out of libraries if anyone does this.
+      mixin = @environment.mixin(mixin_name)
+
+      # The while loop is to allow any number of dynamic includes to be chained.
+      while mixin.nil? && mixin_name == "mixin"
```
nex3, line 362: The fully general thing to do here would be to duplicate (or factor out) a bunch of the logic in `funcall.rb` to provide support for arbitrary Ruby-defined mixins with stack-trace rewriting support and everything. That may be overkill, though, since it doesn't seem like we're likely to add more built-in mixins any time soon. The simpler way to get stack trace information for the intermediate `mixin()` calls would be to track the number of nested calls here, and then in the rescue block below do:

``` ruby
e.modify_backtrace(mixin: mixin_name, line: node.line)
e.add_backtrace(line: node.line)
dynamic_calls.times do
  e.modify_backtrace(mixin: "mixin", line: node.line)
  e.add_backtrace(line: node.line)
end
```

(This code isn't tested, it may need some tweaks)

lib/sass/tree/visitors/perform.rb
```diff
@@ -346,29 +346,47 @@ def visit_mixindef(node)
 
   # Runs a mixin.
   def visit_mixin(node)
+    mixin_name = nil
```
nex3: If you initialize this to `node.name` here, I think you can avoid `mixin_name || node.name` below.

test/sass/scss/scss_test.rb
```diff
@@ -4221,4 +4221,70 @@ def test_crlf
 SCSS
   end
 
+  def test_dynamic_include
```
nex3, line 4224: When writing Ruby tests, I've been trying to create new files for new features. I'm sick of having just a few gargantuan test files that we have to scroll through to find anything.

That said, can we just add tests only to sass-spec? I don't think it's necessary to have Ruby tests for something that's tested there.

test/sass/scss/scss_test.rb
```diff
@@ -4221,4 +4221,70 @@ def test_crlf
...
+  dear-god-why: 2px; }
+CSS
+@mixin mixin($dear-god-why) {
+  .mixin { dear-god-why: $dear-god-why }
+}
+@include mixin(2px);
+SCSS
+  end
+
+  def test_dynamic_include_missing_name
```
nex3, line 4270: For error tests, it would be good to assert the line number as well, especially since weird indirection is going on.

test/sass/scss/scss_test.rb
```diff
@@ -4221,4 +4221,70 @@ def test_crlf
...
+    assert_equal(<<CSS, render(<<SCSS))
+.mixin {
+  dear-god-why: 2px; }
+CSS
+@mixin mixin($dear-god-why) {
+  .mixin { dear-god-why: $dear-god-why }
+}
+@include mixin(2px);
+SCSS
+  end
```
nex3, line 4268: Make sure you also test passing keyword arguments both explicitly and through a captured arglist.

nex3:
Done reviewing! Let's talk about the broader strategy here at our meeting this afternoon.

lib/sass/tree/visitors/perform.rb
chriseppstein, line 362: > provide support for arbitrary Ruby-defined mixins

I thought about this. It's an experimental feature that libsass has, actually. I don't love the idea of people writing mixins in ruby, tbh. I think functions are cleaner for calling out to the host language. Even if it was just an internal API, my sense is that we should hold off on doing that until we have a second or third use case to fully understand the problem space.

test/sass/scss/scss_test.rb
chriseppstein, line 4268: I did this in the sass-spec tests actually. https://github.com/sass/sass-spec/blob/750725c5b4f4ac5d3656f92f0c0e99dbf7e9aa8c/spec/sass_3_5/dynamic-includes/simple/input.scss

test/sass/scss/scss_test.rb
chriseppstein, line 4270: The line numbers are in the sass-spec error assertions and look 👍 . https://github.com/sass/sass-spec/blob/750725c5b4f4ac5d3656f92f0c0e99dbf7e9aa8c/spec/sass_3_5/dynamic-includes/errors/undefined-mixin/error

test/sass/scss/scss_test.rb
chriseppstein, line 4224: Sure. I'll remove these tests once the sass-spec bits land, but would like to keep them here until after the merge.

I'm not loving this workflow and I need to think of a way for sass-spec tests to be more seamlessly integrated into the new feature work we do. One idea is that we can have a directory where we add new sass-spec tests and those test can be ran by the sass-spec runner even though they are not in the sass-spec repo. Then after a feature lands in ruby-sass we can migrate the tests to sass-spec.

test/sass/scss/scss_test.rb
nex3, line 4268: This is so confusing 😩. We should definitely talk through the spec workflow this afternoon.

chriseppstein:
Code review summary:

- [ x ] Add backtrace info for dynamic calls
- [ x ] Simply initialization of mixin_name
- [ ] Remove Test::Unit tests in favor of sass-spec tests
- [ ] Test backtrace line #'s
- [ x ] Test passing keyword arguments both explicitly and through a captured arglist

lib/sass/tree/visitors/perform.rb
chriseppstein, line 362: Implemented in: https://github.com/sass/sass/pull/2057/commits/c3a38aaccaeaf432f64773f23be9057077b4c5b3
Tested in:
https://github.com/sass/sass-spec/pull/864/commits/c02c59fd14e1b233c2c68de8a4d3d6a35d0ce65a

lib/sass/tree/visitors/perform.rb
chriseppstein: Implemented in 9292e93335229fe710c980d34e39efae8507e0b2

chriseppstein:
This depends on https://github.com/sass/sass/pull/2061 before it can land.

lifeiscontent:
@chriseppstein any updates on this getting merged?

internetErik:
@chriseppstein I'm also interested on if there are updates for this pull request. It seems like #2061 was merged in.

dawidk92:
+1

nex3:
Removing from the 3.5 milestone, since I don't think Chris will have time to iterate on it and we want to move forward with the release.

jslegers:
Holy fuck.
Dynamic dependencies are finally coming to Sass?!?!

party

doc-src/SASS_REFERENCE.md
ArmorDarks: I'm the only person here, which feels dirty when reading need to create reference to function name through additional function (which makes some kinda wired recursion in design)? Why can't it be done in more natural way, like

``` Sass
@function test() { return 1 }
$testRef: test;

$output: $testRef()
```

Well, I understand that Sass has issue with determination of function due to string literals, but from other side if we will do this:

``` Sass
$testRef: test; // string literal

$output: $testRef()
```

It might behave same as Sass behaves now — since `test` it not a function, reference will not be received and we will get as output a plain `$testRef()` as string literal.

And if we want to get string literal, not reference, we can do this:

``` Sass
$testRef: test; // string literal

$output: $testRef
```

My only concern here that it won't be possible to properly pass `$testRef` reference to other vars, since we can't know does it hold function reference or variable.

coxrichuk:
+1

IAMtheIAM:
Any news?

mattpilott:
So it's been a year now since the last comment, where are we at with this landing?

nex3:
My limited dev time is currently focused on launching Dart Sass. If someone wants to take charge of landing this feature here and in Dart Sass, that's definitely a possibility, but I personally probably won't have time for another quarter or so at least.

averydev:
Is there an issue or similar in dart sass to track this I could take a look st? Having this feature would enable very powerful meta programming that is otherwise impossible. 

To get a similar effect I’m going to have to add Stylus as an addition pre processor to my project and I really don’t want to.

nex3:
@averydev There isn't currently. #626 is the issue to watch.

I'm closing this, since new features are no longer being added to Ruby Sass.

PR #2061, opened 2016-05-18
Introduce get-function() which returns a function reference for use with call()

chriseppstein:
In 4.0, the call function taking a string won't work because name resolution has to be done lexically. So in 3.5 we are deprecating the string form of `call` and introducing a new function `function-reference` which returns a value that can preserve the lexical scope (closure) at the time the reference is created. In 4.0, this allows that function to be invoked in other lexical contexts. In 3.5, this will silence the deprecation warning and provide a basic facility for overriding global functions and then delegating to them.

Specs are in https://github.com/sass/sass-spec/pull/867/commits/bc48e5ec2164ae88753beb074f00f5ad62d0e4d6  and part of https://github.com/sass/sass-spec/pull/867.

Pending Tasks:
- [x] CHANGELOG
- [x] Reference Documentation.
- [x] Code Review
- [x] Iteration based on Code Review
- [ ] Final Review
- [ ] Land Sass-spec & remove branch dependency (post merge)


doc-src/SASS_CHANGELOG.md
```diff
@@ -23,6 +23,13 @@
   function now accepts a `$bracketed` parameter that controls whether the
   returned list has brackets.
 
+* Passing a string to `call($function-name, $args...)` indicating which function to invoke is
+  now deprecated. Instead pass callable returned from `function-reference($function-name)`.
```
nex3: "pass a callable"

doc-src/SASS_REFERENCE.md
```diff
@@ -676,6 +676,7 @@ SassScript supports seven main data types:
 * nulls (e.g. `null`)
 * lists of values, separated by spaces or commas (e.g. `1.5em 1em 0 2em`, `Helvetica, Arial, sans-serif`)
 * maps from one value to another (e.g. `(key1: value1, key2: value2)`)
+* callable values that refer to functions
```
nex3: I'd just describe these as "first-class functions" or "function references", at least until we have first-class mixins as well. That's more consistent with the terminology from other languages.

A possible plan for terminology going forward:
- The type returned by `function-reference()` is called a "function reference", short form "function", which we describe in the docs as a first-class function.
- Once we add `mixin-reference()` we similarly call its return type "mixin reference".
- We consider these to be effectively different types, and only use the term "callable" in functions that take either one or the other.

This is probably worth talking over in person next week.

doc-src/SASS_CHANGELOG.md
nex3: I saw a little of what you said about this in Slack, but what's the motivation behind calling this `function-reference()` instead of `function()` or something else shorter? It seems like the longer it is, the more difficult it'll be to convince people that it's better than passing the raw name.

doc-src/SASS_REFERENCE.md
```diff
@@ -811,6 +812,13 @@ color interpolated into a selector becomes invalid syntax when
 compressed. To avoid this, always quote named colors if they are meant
 to be used in the construction of a selector.
 
+#### Callables
+
+A callable value is returned by `function-reference($function-name)`.
+The callable value can be passed to `call($callable, $args...)` and the
```
nex3: We definitely shouldn't position `call()` as able to call any callable—we won't allow mixins to be passed to it. I'd call the parameter `$function` or `$reference`.

lib/sass/script/functions.rb
```diff
@@ -525,18 +529,27 @@ def initialize(environment)
       #   assert_type value, :String
       #   assert_type value, :Number
       # @param value [Sass::Script::Value::Base] A SassScript value
-      # @param type [Symbol] The name of the type the value is expected to be
+      # @param type [Symbol, Array<Symbol>] The name(s) of the type the value is expected to be
       # @param name [String, Symbol, nil] The name of the argument.
       # @raise [ArgumentError] if value is not of the correct type.
       def assert_type(value, type, name = nil)
```
nex3, line 538: I'm not totally convinced it's worth adding extra complexity and overhead here versus hand-rolling the verification for functions that need it.

lib/sass/script/functions.rb
```diff
@@ -1670,6 +1687,42 @@ def feature_exists(feature)
     end
     declare :feature_exists, [:feature]
 
+    # Returns a reference to a function for later invocation with the `call` function.
+    #
+    # The function reference created may refer to a function defined in
+    # your stylesheet, built-in to the host environment, or a CSS native
+    # function. As such, this function never returns an error, but the
+    # you can use `function-exists()` to check if the function reference
+    # created points at a Sass function.
```
nex3, line 1699: It think I'd prefer this not support CSS functions. The fact that function calls can be interpreted as plain CSS already makes typos very painful to debug; I don't think we should recapitulate that where it's not necessary.

Maybe we can add an optional `$css` function to control whether it's a plain-css function, with the default being `false`.

lib/sass/script/functions.rb
```diff
@@ -1670,6 +1687,42 @@ def feature_exists(feature)
...
+    #
+    # @example
+    #   function-reference(rgb)
```
nex3: I think we should encourage users to use quotes here—both because I think strings should generally be quoted when they aren't specifically going to be emitted in CSS, and because we specifically have built-in functions with names like `red()` where unquoted names will break. Unquoted names also won't work in the module world with namespaced functions.

lib/sass/script/functions.rb
```diff
@@ -1670,6 +1687,42 @@ def feature_exists(feature)
...
+    #
+    #   @function myfunc { @return "something"; }
+    #   function-reference(myfunc)
+    #
+    # @overload function_reference($name)
+    #   @param name [Sass::Script::Value::String] The name of the function being referenced.
```
nex3, line 1708: Document that a function reference can also be passed and will be returned as-is.

lib/sass/script/functions.rb
```diff
@@ -1670,6 +1687,42 @@ def feature_exists(feature)
...
+    #
+    # @return [Sass::Script::Value::Callable] A function reference.
+    def function_reference(name)
+      assert_type name, [:String, :Callable], :name
+
+      return name if name.is_a?(Sass::Script::Value::Callable)
```
nex3: I think we should verify that the callable is a function and not a mixin. It doesn't matter yet, but we may well not remember to change it when we add mixin references.

lib/sass/engine.rb
```diff
@@ -79,7 +79,10 @@ module Sass
   #
   # `type`: `String`
   # : The user-friendly name of the type of the callable.
-  Callable = Struct.new(:name, :args, :splat, :environment, :tree, :has_content, :type)
+  #
+  # `origin`: `Symbol`
+  # : From whence comes the callable: `:stylesheet`, `:builtin`, `:undefined`
```
nex3: I'd call it `:css` rather than `:undefined`.

Also, it would be good to describe the semantics of each of these in more detail.

lib/sass/script/functions.rb
```diff
@@ -2393,13 +2456,17 @@ def global_variable_exists(name)
     #
     # @overload function_exists($name)
     #   @param name [Sass::Script::Value::String] The name of the function to
-    #     check.
+    #     check or a callable.
```
nex3: I'm not a big fan of allowing this to take a callable. I get that it makes more existing code work as-is, but once the transition is done it's a weird behavioral corner case that exists forevermore. Making sure every corner of the APIs are compatible across functions and references is something I don't want to have to worry about; I think we should just have people do `type-of()` checks manually if they want to check "this function name exists or this is a callable".

Honestly it might also be simpler to just make `function-reference()` take only strings as well.

lib/sass/script/value/callable.rb
```diff
@@ -0,0 +1,48 @@
+module Sass::Script::Value
+  # A SassScript object representing a null value.
+  class Callable < Base
+    # Constructs a Callable value for use in SassScript.
+    #
+    # @param callable [Sass::Callable] The callable to be used when the
+    # callable is called.
+    def initialize(callable)
+      super(callable.dup)
```
nex3: Why are we duplicating this? Do we modify callables after the fact? God, I hope we don't.

lib/sass/script/value/callable.rb
```diff
@@ -0,0 +1,48 @@
...
+      raise Sass::SyntaxError.new("Positive is not defined for values of type: Callable")
+    end
+
+    def unary_minus
+      raise Sass::SyntaxError.new("Negation is not defined for values of type: Callable")
+    end
+
+    def unary_div
+      raise Sass::SyntaxError.new("Division is not defined for values of type: Callable")
+    end
```
nex3, line 23: We should probably just let these drop back to the superclass implementation. That's what maps—another CSS-incompatible type—do.

test/sass/functions_test.rb
```diff
@@ -1520,18 +1546,45 @@ def test_call_uses_local_scope
   a: call(foo);
 }
 SCSS
+    end
+  end
+
+  def test_call_with_ref_uses_local_scope
```
nex3: I thought this was the sort of thing we were putting in sass-spec only?

lib/sass/script/value/callable.rb
```diff
@@ -0,0 +1,48 @@
...
+    def to_s(opts = {})
+      raise Sass::SyntaxError.new("#{to_sass} isn't a valid CSS value.")
+    end
+
+    def inspect
+      to_sass
+    end
+
+    def to_sass
+      "#{value.type}-reference(#{value.name})"
```
nex3: `value.name` here definitely should be quoted at least some of the time—what if it's a color name?

nex3:
Finished reviewing! The core implementation seems right, but there's a bunch of design stuff we should hash out.

lib/sass/script/functions.rb
chriseppstein: @nex3 The issue is that callables can refer to functions that don't exist in SassScript. Initially my code raised an error if they tried to do that, but as I implemented it, the code really felt like matching the semantics of `call` (which can be used with functions that don't exist -- they become css function calls) made the most sense and the cleanest implementation. Once I made that change, it followed that `function-exists()` needed to accept a callable.  Let's discuss this in our weekly meeting.

lib/sass/script/value/callable.rb
chriseppstein: I wish we'd freeze objects that we don't intend to mutate. It's very hard to be sure whether a reference should be dup'd or not.

lib/sass/script/value/callable.rb
chriseppstein, line 23: The superclass impl gave a really confusing error:

```
SyntaxError: function-reference(asdf) isn't a valid CSS value.
```

As a user, whether something is a css value doesn't explain to me why I can't do math on it. I can't "do math" on any css value. I guess the proper thing is to file a new issue and address the bad error message properly for all types that don't "math".

test/sass/functions_test.rb
chriseppstein: we did. after I wrote this 😛

doc-src/SASS_CHANGELOG.md
chriseppstein: As I was writing tests with `function(foo)` I just didn't like the way it looked. It felt confusingly close to a declaration of a function. I'm very open to alternate naming here.

lib/sass/script/functions.rb
chriseppstein, line 538: It's not a lot of complexity and it's very useful when it's needed. I'd like to keep it.

lib/sass/script/functions.rb
chriseppstein, line 1699: See my reply below.

test/sass/functions_test.rb
chriseppstein: This was added when I was working on silencing deprecation warnings, I'll move it.

doc-src/SASS_REFERENCE.md
chriseppstein: Yes, let's discuss.

chriseppstein: decision: different type-of names: `function`, `mixin`

doc-src/SASS_CHANGELOG.md
chriseppstein: Decision: `get-function()` and `get-mixin()`

lib/sass/script/functions.rb
chriseppstein, line 1699: Decision: `get-function($name, $css: true)` will always return a css reference even if the sass function exists. `get-function($name)` will always return a sass reference or an error.

lib/sass/script/functions.rb
chriseppstein: Decision: Will remove this.

lib/sass/script/functions.rb
nex3: :+1:

lib/sass/script/value/callable.rb
nex3, line 23: > I guess the proper thing is to file a new issue and address the bad error message properly for all types that don't "math".

Agreed.

chriseppstein, line 23: https://github.com/sass/sass/issues/2077

lib/sass/script/functions.rb
nex3, line 1699: Don't forget to update the documentation!

lib/sass/script/functions.rb
```diff
@@ -1670,6 +1687,60 @@ def feature_exists(feature)
...
+    #
+    #   @function myfunc { @return "something"; }
+    #   get-function("myfunc")
+    #
+    # @overload get_function($name)
+    #   @param name [Sass::Script::Value::String] The name of the function being referenced.
+    #     if `$name` is a Sass::Script::Value::Function it is simply returned.
+    #
+    # @return [Sass::Script::Value::Function] A function reference.
+    def get_function(name, kwargs = nil)
```
nex3: Why not make `kwargs` default to `{}`?

lib/sass/script/functions.rb
```diff
@@ -1670,6 +1687,60 @@ def feature_exists(feature)
...
+              v.value
+            else
+              false
+            end
+
+      if kwargs && kwargs.any?
+        raise ArgumentError.new("Illegal keyword argument '#{kwargs.keys.first}'")
+      end
+
+      return name if name.is_a?(Sass::Script::Value::Function)
```
nex3: I thought we decided not to do this.

lib/sass/script/value/callable.rb
```diff
@@ -0,0 +1,25 @@
+module Sass::Script::Value
```
nex3, line 1: This file is now unused, right?

nex3:
Done reviewing!

lib/sass/script/value/callable.rb
chriseppstein, line 1: No, it's going to be a common base class for mixins and functions.

lib/sass/script/functions.rb
chriseppstein: We didn't decide. I can take it out.

lib/sass/script/functions.rb
chriseppstein: I like conditionals 😜

PR #2064, opened 2016-05-19
content-exists() function

### LINKED ISSUE #1988: any way to test if @content has content?
Hey guys,

is there any way to test @content for any content? like length(@content) ?

chriseppstein:
Closes #1988

Specs are in https://github.com/sass/sass-spec/tree/content-exists/spec/sass_3_5/functions/content-exists

Tasks Remaining:
- [x] CHANGELOG
- [x] Source documentation
- [x] Reference Docs
- [ ] Stop using the sass-spec branch once merged.

### DIFF
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -2497,12 +2497,14 @@ The same mixins can be done in the `.sass` shorthand syntax:
       #logo
         background-image: url(/logo.gif)
 
-**Note:** when the `@content` directive is specified more than once or in a loop, the style block will be duplicated with each invocation.
+**Note:** when the `@content` directive is specified more than once or
+ in a loop, the style block will be duplicated with each invocation.
 
 Some mixins may require a passed content block or may have different
 behavior depending on whether a content block was passed. The
-`content-exists()` function will return true when a content block is
-passed to the current mixin and can be used to implement such behaviors.
+[`content-exists()` function](Sass/Script/Functions.html#content_exists-instance_method)
+will return true when a content block is passed to the current
+mixin and can be used to implement such behaviors.
 
 #### Variable Scope and Content Blocks
 
--- a/doc-src/SASS_REFERENCE.md
+++ b/doc-src/SASS_REFERENCE.md
@@ -2499,6 +2499,11 @@ The same mixins can be done in the `.sass` shorthand syntax:
 
 **Note:** when the `@content` directive is specified more than once or in a loop, the style block will be duplicated with each invocation.
 
+Some mixins may require a passed content block or may have different
+behavior depending on whether a content block was passed. The
+`content-exists()` function will return true when a content block is
+passed to the current mixin and can be used to implement such behaviors.
+
 #### Variable Scope and Content Blocks
 
 The block of content passed to a mixin are evaluated in the scope where the block is defined,
--- a/lib/sass/environment.rb
+++ b/lib/sass/environment.rb
@@ -189,13 +189,14 @@ module Sass
       @caller ||= env.is_a?(ReadOnlyEnvironment) ? env : ReadOnlyEnvironment.new(env, env.options)
     end
 
-    # The content passed to this environment but the content's
-    # environment is made readonly if it isn't already.
+    # The content passed to this environment. If the content's environment isn't already
+    # read-only, it's made read-only.
     #
     # @see BaseEnvironment#content
     #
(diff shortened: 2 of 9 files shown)


chriseppstein:
sass-spec PR: https://github.com/sass/sass-spec/pull/870

doc-src/SASS_REFERENCE.md
```diff
@@ -2499,6 +2499,11 @@ The same mixins can be done in the `.sass` shorthand syntax:
 
 **Note:** when the `@content` directive is specified more than once or in a loop, the style block will be duplicated with each invocation.
 
+Some mixins may require a passed content block or may have different
+behavior depending on whether a content block was passed. The
+`content-exists()` function will return true when a content block is
+passed to the current mixin and can be used to implement such behaviors.
```
nex3: Typically, we don't list functions in the reference unless they're crucial to the use of a broader feature (like `map-get()`). When we do list them, though, we should link to their documentation.

lib/sass/environment.rb
```diff
@@ -175,6 +175,10 @@ class Environment < BaseEnvironment
 
   # A read-only wrapper for a lexical environment for SassScript.
   class ReadOnlyEnvironment < BaseEnvironment
+    def initialize(parent = nil, options = nil)
```
nex3: This is a good use-case for named parameters!

lib/sass/environment.rb
```diff
@@ -190,9 +194,19 @@ def caller
...
+        tree, env = read_write_content
+        if env && !env.is_a?(ReadOnlyEnvironment)
+          env = ReadOnlyEnvironment.new(env, env.options)
+        end
+        @content_cached = true
+        @content = [tree, env]
+      else
+        @content_cached = true
+        @content = nil
+      end
```
nex3, line 216: It's not clear to me what's going on in this block. Can you add some comments?

lib/sass/stack.rb
```diff
@@ -98,6 +98,16 @@ def with_mixin(filename, line, name)
...
 
+    # Pushes a function frame onto the stack.
+    #
+    # @param filename [String] See \{Frame#filename}.
+    # @param line [String] See \{Frame#line}.
+    # @param name [String] See \{Frame#name}.
+    # @yield [] A block in which the new frame is on the stack.
+    def with_function(filename, line, name)
+      with_frame(filename, line, :function, name) {yield}
+    end
```
nex3, line 109: I think adding functions to the stack is the right thing to do—as far as I can tell the only reason they weren't before was that we didn't really need them. It doesn't look like it'll affect a huge amount of stuff, but it _will_ affect the output of `@warn` when it's used within a function, and we should add specs for that.

lib/sass/script/functions.rb
```diff
@@ -2421,6 +2426,29 @@ def mixin_exists(name)
...
+    #     @if not content-exists() {
+    #       @error "You must pass a content block!"
+    #     }
+    #     @content;
+    #   }
+    #
+    # @overload content_exists()
+    # @return [Sass::Script::Value::Bool] Whether a content block was passed to the mixin.
+    def content_exists
+      mixin_frame = environment.stack.frames[-2]
```
nex3, line 2446: Add a comment explaining why this is `-2`.

nex3:
Done reviewing!

doc-src/SASS_REFERENCE.md
chriseppstein: Normally, I don't document them, but in this case, it seemed appropriate, I'll add a link.

lib/sass/stack.rb
chriseppstein, line 109: I also needed to add a frame for @warn to get the trace to output the previous caller (the function).

chriseppstein:
@nex3 All code review comments have been addressed. 👍 ?

lib/sass/environment.rb
```diff
@@ -189,15 +189,21 @@ def caller
       @caller ||= env.is_a?(ReadOnlyEnvironment) ? env : ReadOnlyEnvironment.new(env, env.options)
     end
 
-    # The read-only content passed to this environment.
+    # The content passed to this environment but the content's
+    # environment is made readonly if it isn't already.
```
nex3: I had a lot of trouble parsing this sentence. Maybe "The content passed to this environment. If the content's environment isn't already read-only, it's made read-only."

nex3:
:+1:

PR #2065, opened 2016-05-20
Simple supports clause in import

### LINKED ISSUE #1967: supports() clause doesn't support declaration-syntax
`supports()` clause support, that was introduced in 3.4.20 is not fully compatible with W3C CSS spec.

According to [CSS Cascading and Inheritance Level 4](https://www.w3.org/TR/2015/WD-css-cascade-4-20150908/#typedef-supports-condition) it's allowed to place `supports_conditions` **OR** `declaration` inside `supports()` clause.

```
@import [ <url> | <string> ]
        [ supports( [ <supports-condition> | <declaration> ] ) ]?
        <media-query-list>? ;
```

So the following imports are both valid and  equivalent:

``` css
@import "mystyle.css" supports(display: flex);
@import "mystyle.css" supports((display: flex));
```

However Sass compiler doesn't work for the following code:

``` scss
@import "mystyle.css" supports(display: flex);
```

And produces the following error:

```
Error: Invalid CSS after "....css" supports(": expected ")", was "display: flex);"
        on line 1 of test.scss
```

chriseppstein:
Specs are in https://github.com/sass/sass-spec/tree/import-supports/spec/sass_3_4_x/simple_supports_clause_in_import

This change fixes #1967.

### DIFF
--- a/lib/sass/scss/parser.rb
+++ b/lib/sass/scss/parser.rb
@@ -560,7 +560,7 @@ module Sass
       def supports_clause
         return unless tok(/supports\(/i)
         ss
-        supports = supports_condition
+        supports = import_supports_condition
         ss
         tok!(/\)/)
         supports
@@ -568,6 +568,10 @@ module Sass
 
       def supports_condition
         supports_negation || supports_operator || supports_interpolation
+      end
+
+      def import_supports_condition
+        supports_condition || supports_declaration
       end
 
       def supports_negation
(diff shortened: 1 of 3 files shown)


lib/sass/scss/parser.rb
```diff
@@ -589,6 +593,14 @@ def supports_operator
         cond
       end
 
+      def supports_declaration
+          return if tok?(/\(/)
+          name = sass_script(:parse)
+          tok!(/:/); ss
+          value = sass_script(:parse)
+          Sass::Supports::Declaration.new(name, value)
```
nex3, line 600: Now that this is factored out, can you re-use it in `supports_condition_in_parens`?

nex3:
:+1: as long as Travis succeeds!

Issue #512, opened 2012-09-11
gray(x%) as a shortcut for rgb(80,80,80)


adius:
I think it would be cool if there was the possibility to use the rgb function with just one value as a shortcut.
What do you think?

chriseppstein:
gray(n%) would feel more natural to me.

Anahkiasen:
+1 for this. I've had a `gray()` (well actually `grey`) function in my files for months, I use it all the time. Very very useful.

nex3:
I'm definitely not willing to overload `rgb()` in a way that may end up being incompatible with the CSS spec. `gray()` may be doable, though.

timohofmeijer:
`gray()` became part of the CSS4 draft :smile: .
`gray() = gray( [<number> | <percentage>] [, <alpha-value>]? )` 
https://drafts.csswg.org/css-color/#grays

bkazi:
Hi,
I was hoping to get some guidance on this issue being a new contributor. I have come up with an implementation for `gray()` that is based on the way `rgb()` is done and it seems to work. I would like to know what would be the proper way to go about adding such a feature and what considerations should be taken.
Thanks

nex3:
@bkazi Just send a pull request and we'll walk you through the process :smiley:

nex3:
Closing this in favor of #2078.

PR #2071, opened 2016-06-03
Implement the gray function.

thiagorb:
Hello!

This is my first attempt of contributing here. I saw the issue in https://github.com/sass/sass/issues/512, and thought it would be a good starting point for me.

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -47,6 +47,9 @@ module Sass::Script
   # : Creates a {Sass::Script::Value::Color Color} from hue, saturation,
   #   lightness, and alpha values.
   #
+  # \{#gray gray($lightness, $alpha)}
+  # : Creates a {Sass::Script::Value::Color Color} lightness and alpha values.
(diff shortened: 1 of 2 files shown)


chriseppstein:
Hi @thiagorb Thanks for the patch, sorry for the delay in reviewing it.

My first and biggest concern is that this implementation of the gray() is going to give different results than [the algorithm that is specified by the gray() function in CSS](https://drafts.csswg.org/css-color/#gray-to-rgb).

The color spec level 4 has a bunch of new algorithms for color handling and this is our first foray into them. And you thought this was going to be easy 😛?

Fortunately, there's some [example JS code provided](https://drafts.csswg.org/css-color/#color-conversion-code) that we can port to ruby. I think for the gray function, there's probably an optimized path through that code, but eventually we'll need to implement the full spectrum of that color handling (great pun I know), so I don't think it's worth the effort to find it.

Beyond the algorithm issue, things look pretty good.

nex3:

PR #2079, opened 2016-06-11
Unit Replace

ghost:
I'm from Brazil

They could do something like this function? Since I believe that it is not good

**Example**

`unit_replace(157px, rem)`

output: 157rem

`unit_replace(157, px)`

output: 157px

`unit_replace(157em, "")`

output: 157

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -1687,6 +1687,13 @@ MESSAGE
       bool(number.unitless?)
     end
     declare :unitless, [:number]
+    
+    def unit_replace(number, unit_str)
+      assert_type number, :Number, :number
+      assert_type unit_str, :String, :unit_str
+      number(number.value, unit_str.value)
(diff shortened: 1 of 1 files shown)


chriseppstein:
The concept of dimensions makes replacing a unit nonsensical. If there's a value specified in one unit, a conversion ratio can be used. E.g. `$px-per-rem: 16px / 1rem;` And then math expressions will respect and cancel units appropriately. E.g. `10rem * $px-per-rem` would result in `160px`.

Similarly, removing the unit of a number is nonsensical unless you know some conversion factor between raw numbers and the dimension for your use case.

In general, please open an issue to discuss proposed features before sending a pull request.

justinskolnick:
@vinivsl Check out https://github.com/justinskolnick/unitize. My approach to SCSS has changed in the years since I published it, so I don't need it anymore. But it looks like you might have a use.

PR #2080, opened 2016-06-11
Calling a variable with the given name

ghost:
**I'm from Brazil**

I've never seen a Python code before, but added for me to function calling a variable with the given name.

The reason I have done this is to have something like this from Less: `@@var`

**Thank You!**

---

**Exemplo:**

```
$a-example-var: value;
$var1: a-example;
$var2: -var;

call_var(#{$var1}#{$var2})

call_var(a-example-var)
```

### DIFF
--- a/lib/sass/script/functions.rb
+++ b/lib/sass/script/functions.rb
@@ -2229,6 +2229,27 @@ MESSAGE
       perform(funcall)
     end
     declare :call, [:name], :var_args => true, :var_kwargs => true
+    
+    # Calling a variable with the given name in the current scope or
+    # in the global scope.
+    #
+    # @example
+    #   $a-example-var: value;
+    #   $var1: a-example;
+    #   $var2: -var;
+    #
+    #   call_var(#{$var1}#{$var2})
+    #
+    #   call_var(a-example-var)
+    #
+    # @overload call_var($name)
+    #   @param $name [Sass::Script::Value::String] The name of the variable to
+    #     call. The name should not include the `$`.
+    def call_var(name)
+      assert_type name, :String, :name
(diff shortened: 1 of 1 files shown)


KittyGiraudel:
> I've never seen a Python code before

That might be because it’s Ruby.

I find it super rad that you actually submitted a pull-request to add a feature that you deem useful, unfortunately it cannot be accepted as is. For starters, design decisions are being made by core authors @chriseppstein and @nex3, usually based on long time feedback from the community. Sass has a long history of not blindly implementing features for the sake of pleasing developers, and that’s a very fine behaviour.

Secondly, any feature has to be spec-ed extensively before being implemented. That is to make sure there won’t be an edge case we missed somewhere, putting the whole language to its knees. Remember that once it’s out there, there is no coming back.

Last but not least, a feature —no matter what it is— cannot be merged without a robust test suite. Sass as a language has proven to have very few bugs throughout the years, and this has everything to do with spec-ing and writing a hell lots of test.

Again, I find it great that you contribute to this project! Next time, think of asking contributors before if it’s even worth implementing in the first place. :)

ghost:
I knew it would not be accepted, my purpose was just to give an idea.

He believes that there is a possibility of adding similar functionality?

I migrated from LESS to SASS, but call variables through other is something I'm feeling really miss LESS
**Google Translate**

KittyGiraudel:
Dynamic variables have been solved with maps in Sass. There is no need for such a feature anymore.

chriseppstein:
We suggest that Sass maps be used for this. dynamic variable constructruction as a feature by itself like this, is a less comprehensive feature than maps because with maps you get the ability to iterate over keys and values, inspect the data structure, etc.

PR #2092, opened 2016-06-29
Skip mappings with nil importer in Source::Map#to_json

djudd:
This was already happening in the case where `options[:type] != :inline`, but would cause an error in the case of :inline

I'm not sure exactly under what circumstances this condition was triggered, or if it's a valid state, but it occurred in my production Rails app, and an equivalent monkeypatch got everything working.

### DIFF
--- a/lib/sass/source/map.rb
+++ b/lib/sass/source/map.rb
@@ -118,12 +118,14 @@ module Sass::Source
       @data.each do |m|
         file, importer = m.input.file, m.input.importer
 
+        next unless importer
+
         if options[:type] == :inline
           source_uri = file
         else
           sourcemap_dir = sourcemap_path && sourcemap_path.dirname.to_s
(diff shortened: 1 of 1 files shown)


chriseppstein:
It seems to me that the issue is with [this line](https://github.com/sass/sass/pull/2092/files#diff-2f15a542ab2d3add779eeae7eddee77cR142). Skipping when there's no importer shouldn't be necessary in the case of `:inline` as we have the contents of the file (E.g. this may be the case when reading from `stdin`).

@nex3 knows this part of the code better than me, so I'll defer to her.

nex3:
I think this is correct. Even though we should have the source text for the file _somewhere_, we don't have any means of accessing it in this code except through the importer, which may not exist. If we want source maps generated for importer-less templates, it may be necessary to refactor the code such that they always have an importer of some sort.

PR #2093, opened 2016-07-01
Remove extra lines after @include

### LINKED ISSUE #1843: sass-convert adds extra line breaks between @include directives
This: 

``` scss
@include component-one;
@include component-two;
@include component-three;
```

Turns into this:

``` scss
@include component-one;

@include component-two;

@include component-three;

```

This was introduced in sass version 3.4.16. I came across this because I use a Sublime Text 3 plugin called [SassBeautify](https://github.com/badsyntax/SassBeautify) which uses `sass-convert` to beautify the files. This behaviour cases the plugin to give ugly output. :disappointed:

Addisonbean:
This closes #1843, which would prevent `sass-convert` from adding newlines in between consecutive `@include` statements.


nex3:
Can you add a test that verifies the fix as well?

Addisonbean:
Sure, I just added one.

nex3:
Thanks for the patch!

PR #2094, opened 2016-07-03
Add multiline expressions, issue #216

maciej-ka:
Enables splitting a long definition into several lines:

``` sass
.shadow
  box-shadow: \
    5px 5px 10px black,
    inset 0 0 10px white

```

Symbol `\` is replaced with a concatenation of lines in a nested block.
This can be usefull for long css properties, variable maps or variable
lists. `\` has to be at the end of a line to work like this.

``` sass
$rgb-colors: \
  red
  green
  blue
```

Transformation is done before running a parser, during building of a
lines array. It works in a similar fashion to processing block comments.

This pull request is related to issue #216

### DIFF
--- a/lib/sass/engine.rb
+++ b/lib/sass/engine.rb
@@ -113,6 +113,8 @@ module Sass
     # `comment_tab_str`: `String?`
     # : The prefix indentation for this comment, if it is a comment.
     class Line < Struct.new(:text, :tabs, :index, :offset, :filename, :children, :comment_tab_str)
+      attr_accessor :multiline
+
       def comment?
         text[0] == COMMENT_CHAR && (text[1] == SASS_COMMENT_CHAR || text[1] == CSS_COMMENT_CHAR)
       end
@@ -120,13 +122,8 @@ module Sass
       def initialize(*args)
         super
         return unless text[-1] == MULTILINE_CHAR
-        # Remove multiline indicator and remember that this is a different type of a line.
         text.chop!
-        @is_multiline = true
-      end
-
-      def multiline?
-        defined? @is_multiline
+        @multiline = true
       end
     end
 
@@ -456,7 +453,6 @@ ERR
       comment_tab_str = nil
       first = true
       lines = []
-      @active_multiline_tabs = Set.new
 
       string.scan(/^[^\n]*?$/).each_with_index do |line, index|
         index += (@options[:line] || 1)
@@ -506,27 +502,18 @@ END
         next if try_multiline(line, lines.last, line_tabs)
 
         lines << Line.new(line.strip, line_tabs, index, line_tab_str.size, @options[:filename], [])
-
-        # fix tabs nested in multiline
-        lines.last.tabs -= @active_multiline_tabs.size
       end
+
       lines
     end
 
     def try_multiline(line, last, tabs)
-      # detect a new multiline block
-      @active_multiline_tabs << last.tabs if last.multiline?
-      # detect the end of a multiline block
-      @active_multiline_tabs.delete_if {|n| n >= tabs}
-
-      return false unless @active_multiline_tabs.include?(tabs - 1)
-
-      # detect a nested multiline block
+      return false unless last.multiline
       line = Line.new(line.strip)
-      @active_multiline_tabs << tabs if line.multiline?
-
-      # concatenate a multiline
-      last.text << ' ' << line.text unless line.comment?
+      unless line.comment?
+        last.text << ' ' << line.text
+        last.multiline = line.multiline
+      end
       true
     end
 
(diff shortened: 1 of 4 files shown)


lokson:
On Ruby 2.3.1 it passes, I will try to run test against 1.8.7.

chriseppstein:
Hi, this is interesting. It seems like a good generalized solution. I'd like to see some tests for it to make sure it is working for selectors and directives as well as tests to ensure that the multiline expressions can be nested. E.g. I would expect the following to also work:

``` sass
#sidebar \
    div
    a
      color: red
```

``` sass
@media all \
    and (min-device-width: 100px)
    and (max-device-width: 300px)
        #sidebar
            width: 100%
```

``` sass
.foo
    @each \
        $color in \
            red
            orange
            yellow
            green
            blue
                color: $color
```

maciej-ka:
@chriseppstein 
Thanks for a feedback. Update is ready and it covers all of your examples.

In a first version, a multiline operator worked on all nested lines 
no matter how deep they were nested.

For example

``` sass
div \
  a
    color: red
```

Was transformed into:

``` sass
div a color: red
```

And now it becomes:

``` sass
div a
  color: red
```

After this commit multiline operator is concatenating only directly nested lines, 
(nested one level deeper than a line with operator).

And to have a multiline block nested in an another multiline block
(which will all end as a one line) - two multiline operators should be used.

``` sass
.nested
  @each \
    $color in \
      red
      green
      blue
        color: $color
```

Is transformed into:

``` sass
.nested
  @each $color in red green blue
    color: $color
```

nex3:
This isn't quite how `\` works as a multiline operator in other languages (Bash, Python, etc). Typically it's required on the end of every line, with the semantics that `\` followed by a newline is interpreted as non-semantic whitespace. I'd prefer that we follow that behavior rather than leaning on indentation here.

chriseppstein:
@nex3 Let's talk about this today. I know that's how the other languages work, but I find this approach in keeping with other indented syntax features an also having much less visual clutter.

nex3:
Chris and I talked this over and decided to go with my proposal. To flesh it out in more detail, if a line ends in `\`, the backslash should be replaced with ` ` and the following line's text appended. The indentation of the following line would be irrelevant. To translate some of the examples above:

``` dart
.shadow
  box-shadow:           \
    5px 5px 10px black, \
    inset 0 0 10px white

#sidebar \
    div  \
    a
  color: red

@media all                        \
    and (min-device-width: 100px) \
    and (max-device-width: 300px)
  #sidebar
    width: 100%

.foo
  @each $color in \
      red         \
      orange      \
      yellow      \
      green       \
      blue
    color: $color
```

Hopefully this should be pretty straightforward to implement, but if you have any questions let us know.

maciej-ka:
@nex3 
Commit is ready. A traditional syntax is surely easier to implement.

This change makes a multiline operator `\` act like in most languages.
It is replaced by a space and joined with a next line. Indentation of joined
line doesn't matter.

Multiline operator can be chained across several lines and it ignores
comment lines. For example:

``` sass
.comment
  box-shadow: \
    // 5px 5px 10px black,
    inset 0 0 10px white, \
    inset -5px 0 5px blue
```

Is changed into:

``` sass
.comment
  box-shadow: inset 0 0 10px white, inset -5px 0 5px blue
```

xzyfer:
Could we also get these specs added to sass-spec. We'll need to update sass2scss for LibSass support /cc @mgreter

maciej-ka:
@xzyfer 
1. To which sass-spec folder add the tests? (I got lost there)
2. Should I do it before end of this pull request?

chriseppstein:
@nex3 do you want this to land on `next` or on `stable`? I was assuming `next` but it's currently a PR against `stable`.

nex3:
@lokson Specs should go in the [sass-spec repo](https://github.com/sass/sass-spec), under `spec/sass_3_5`, probably in a new `mutliline` directory. Feel free to do that after this lands, though.

@chriseppstein This should definitely land on `next`.

maciej-ka:
@xzyfer https://github.com/sass/sass-spec/pull/895

xzyfer:
Thanks @lokson

maciej-ka:
I still miss you, dear short syntax:

One idea: lets add second multi line operator '|' that would not need to be repeated

``` sass
.shadow
  box-shadow: |
    5px 5px 10px black,
    inset 0 0 10px white
```

It would mimic similar construct of yaml format.

``` yaml
Key: |
  this is my very very very 
  long string
```

@chriseppstein @nex3

nex3:
I hate do to this after so much iteration, but I'm going to put this on ice for now. Unless we can find a new maintainer, [Ruby Sass is on track to be deprecated](http://blog.sass-lang.com/posts/1022316-announcing-dart-sass), so we aren't accepting non-critical new features at this time.

Issue #2106, opened 2016-07-26
The method `invalid_content_parent` is never called


marianposaceanu:
So I was investigation why using libsassc I was getting:

```
SassC::SyntaxError: Error: Mixin "something" does not accept a content block.
```

which is completely valid since `something` doesn't use `@content`.

https://github.com/sass/libsass/blob/master/src/expand.cpp#L666

---

Yet on the Ruby side nothing is reported and whilst reading a bit the code:

This method should set `has_content` to true:

https://github.com/sass/sass/blob/stable/lib/sass/tree/visitors/check_nesting.rb#L82

but it doesn't seem to be called anywhere.

Hence this logic won't raise an exception if the mixin has no `@content`:

https://github.com/sass/sass/blob/stable/lib/sass/tree/visitors/perform.rb#L354

chriseppstein:
I need to see a Sass source code example for this.

marianposaceanu:
Ok the _bug_ is quite an edge case it seems libsassc is much more strict:

``` sass
@mixin breakpoint($args: null) {
  //  CASE IF NO DATA IS PASSED WE SHALL OUTPUT MEDIA QUERY ONLY FOR SCREEN THIS MEANS STYLES WILL BE GLOBALLY AND INHERITABLE BY UPPER BREAKPOINTS
  @if $args == null {
    @media screen {
      @content;
    }
  }
  //  CASE IF DATA IS PASSED WE GENERATE BREAKPOINT FOR PASSED BREAKPOINT
  @else {
    @media screen and (min-width: $args) {
      @content;
    }
  }
}

@mixin clearfix() {
  a::after {
    content: '';
    display: table;
    clear: both;
  }
}

@include breakpoint() {
  padding: 30px 0;

  @include clearfix() {}
}
```

## libsassc

```
Error: Mixin "clearfix" does not accept a content block.

       Backtrace:
        app/assets/stylesheets/home.scss:59, in mixin `@content`
        app/assets/stylesheets/home.scss:37, in mixin `breakpoint`
        app/assets/stylesheets/home.scss:56
        on line 59 of app/assets/stylesheets/home.scss
>>   @include clearfix() {}
   -----------^
```

## Ruby sass

it renders without issue

In any case this is something minor - I just found it bc. the front-end guy made that small mistake and thus the assets would not compile anymore was curios why that happened.

chriseppstein:
[...]

PR #2112, opened 2016-08-06
Fix unallowed block error.

maciej-ka:
Displays error message 'Mixin "foo" does not accept a content block.'
for this example:

``` scss
@mixin foo() {}
@include foo() {}
```

Before, example was rendered without error and only failed if block
inside `@include` had some statements.

This is fix to:
https://github.com/sass/sass/issues/2106
Tests: sass/sass-spec#897

### DIFF
--- a/test/sass/engine_test.rb
+++ b/test/sass/engine_test.rb
@@ -368,6 +368,12 @@ SASS
     end
   end
 
+  def test_empty_block_in_no_content_mixin
+    assert_raise_message Sass::SyntaxError, 'Mixin "foo" does not accept a content block.' do
+      render "@mixin foo() {}\n @include foo() {}", :syntax => :scss
+    end
+  end
+
   def test_selector_tracing
     actual_css = render(<<-SCSS, :syntax => :scss, :trace_selectors => true)
       @mixin mixed {
@@ -894,7 +900,7 @@ foo {
 CSS
 foo
   bar : baz
-  bizz	: bap
+  bizz : bap
 SASS
   end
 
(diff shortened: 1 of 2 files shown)


test/sass/engine_test.rb
```diff
@@ -894,7 +900,7 @@ def test_properties_with_space_after_colon
 CSS
 foo
   bar : baz
-  bizz	: bap
+  bizz : bap
```
maciej-ka, line 903: I don't know why diff marks this as a change, I haven't touched this line :/

chriseppstein:
This looks good. We'll also want to add a corresponding spec to sass-spec.

maciej-ka:
@chriseppstein sure. It will be in `spec/sass_3_5` untill end of this week.

lokson:
@chriseppstein, @xzyfer https://github.com/sass/sass-spec/pull/896

nex3:
@lokson Sorry for the long turnaround time on this! Can you rebase this onto the latest `master`? It has a change that'll make this run specs against your sass-spec pull request.

maciej-ka:
@nex3 rebase onto master done.

maciej-ka:
@nex3 https://github.com/sass/sass-spec/pull/897 rebased onto master

test/sass/engine_test.rb
tagliala: the original file has a tabulation after `bizz`, the edited file has a space. Maybe the text editor converted that tabulation automatically?

PR #2149, opened 2016-09-24
Remove duplicate private

maciej-ka:
`Sass::Selector::Sequence` called `private` twice. The second call was
removed.


nex3:
Thanks for the fix!

PR #2150, opened 2016-09-26
Change sourcemap JSON order to match spec

nschonni:
From [the spec](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit?pli=1#) the order should be:

* Line 1: The entire file is a single JSON object
* Line 2: File version (always the first entry in the object) and must be a positive integer.
* Line 3: An optional name of the generated code that this source map is associated with.
* Line 4: An optional source root, useful for relocating source files on a server or removing repeated values in the “sources” entry.  This value is prepended to the individual entries in the “source” field.
* Line 5: A list of original sources used by the “mappings” entry.
* Line 6: An optional list of source content, useful when the “source” can’t be hosted. The contents are listed in the same order as the sources in line 5. “null” may be used if some original sources should be retrieved by name.
* Line 7: A list of symbol names used by the “mappings” entry.
* Line 8: A string with the encoded mapping data.

### DIFF
--- a/lib/sass/source/map.rb
+++ b/lib/sass/source/map.rb
@@ -101,6 +101,8 @@ module Sass::Source
       result = "{\n"
       write_json_field(result, "version", 3, true)
 
+      write_json_field(result, "file", css_uri)
+
       source_uri_to_id = {}
       id_to_source_uri = {}
       id_to_contents = {} if options[:type] == :inline
@@ -178,7 +180,6 @@ module Sass::Source
         end
       end
       line_data.push(segment_data_for_line.join(","))
-      write_json_field(result, "mappings", line_data.join(";"))
 
       source_names = []
       (0...next_source_id).each {|id| source_names.push(id_to_source_uri[id].to_s)}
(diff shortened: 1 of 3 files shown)


nschonni:
Cross-linking to fix on the libsass side https://github.com/sass/libsass/pull/2193

nschonni:
The libsass version of this change has now landed

nex3:
I don't think the spec is intending to mandate this specific ordering—after all, JSON is inherently order-independent. I think it's just using those line numbers to point out where the sections it's referring to appear in the example source map.

That's the a concrete benefit to having this ordering?

nschonni:
Ensuring each one follows the spec will allow comparing of the sourcemaps at the file level in the spec side (although i don't think that is a goal currently). You're :100: right that a JSON object doesn't really care about the order normally, and can just be parsed and compared by property.

nex3:
I'd rather have sass-spec compare the parsed maps than have to worry about textual identity of sourcemaps between implementations.

Issue #490, opened 2012-08-16
Make the parent selector perform semantically-aware selector unification


nex3:
For example,

``` scss
div:before { &[id] { content: "#" attr(id); } }
```

should become

``` css
div[id]:before { content: "#" attr(id); }
```

since the unification should know that `:before` must come after other selectors.

Also,

``` scss
*, a { .foo { color: red; } }
```

should become

``` css
.foo, a.foo { color: red; }
```

since the unification would be able to remove the universal selector.

rafegoldberg:
:+1: this would be quite useful!

donatj:
#360 which was merged into this was my first GH issue ever. It's still not fixed ;)

Issue #360, opened 2012-04-20
*:psuedo { &[attr]{ ... } } appends [attr] to the end after the psuedo


donatj:
``` scss
div:before {
    &[id] {
        content: "#" attr(id);
    }
}
```

results in 

``` css
div:before[id] { content: "#" attr(id); }
```

which is invalid and does not work. It instead should result in 

``` css
div[id]:before { content: "#" attr(id); }
```

nex3:
This is tricky. Currently, `&` doesn't perform actual selector unification; it just effectively "concatenates" the selectors. It's smarter than string concatenation, but not a ton.

If we were to make this work, we'd have to be clever about cases like the following as well:

``` scss
.foo { &.foo { ... } }
#foo { &#bar { ... } }
```

Maybe that's the right way to go, though. Interesting food for thought.

donatj:
I don't know the codebase, or ruby for that matter (so excuse my psuedo-code), but I'd think you could do something along the lines of just 

``` js
selector = [
  elm: "",
  ids: [],
  classes: [],
  attrs: [],
  psuedos: []
]
```

whereas the & would explode it into multiple selectors and from there the individual pieces - and only piece it back together on to_string.  Of course this may not be something easy to implement as like I said I do not know the codebase - I think something like this would be necessary to fix this strain of bugs though.

chriseppstein:
@nex3 it seems like the work done for #286 would probably enable this. I give it a +1.

nex3:
#286 goes in the opposite direction; it allows the parent selector to be treated more like a plain string value. This requires treating it more like a semantic unification. The two aren't mutually exclusive, but #286 won't give us this for free.

nex3:
Merging into #490.

PR #2151, opened 2016-09-26
Improve parent resolution

maciej-ka:
Places pseudo class selectors at the end, removes duplicate selectors
and removes universal selector if possible during resolution of parent
selector (&)
## Pseudo at the end

``` sass
div:before { &[id] { color: red; }}
```

Before:

``` css
div:before[id] {
  color: red; }
```

After:

``` css
div[id]:before {
  color: red; }
```
## Remove duplicates

``` sass
.foo { &.foo { color: red; }}
```

Before:

``` css
.foo.foo {
  color: red; }
```

After:

``` css
.foo {
  color: red; }
```
## Remove universal

``` sass
* { &.foo { color: red; }}
```

Before:

``` css
*.foo {
  color: red; }
```

After:

``` css
.foo {
  color: red; }
```

Related to: https://github.com/sass/sass/issues/490
Tests: https://github.com/sass/sass-spec/pull/931

### DIFF
--- a/lib/sass/selector/simple_sequence.rb
+++ b/lib/sass/selector/simple_sequence.rb
@@ -132,8 +132,15 @@ module Sass
             end
           end
 
+          resolution = parent_sub + resolved_members[1..-1]
+          resolution.sort_by! {|e| e.is_a?(Pseudo) ? 1 : 0}
+          resolution.uniq!
+          if resolution.any? {|e| [Element, Class, Id].include? e.class}
+            resolution.delete_if {|e| e.is_a? Universal}
+          end
+
           Sequence.new(members[0...-1] +
-            [SimpleSequence.new(parent_sub + resolved_members[1..-1], subject?)] +
+            [SimpleSequence.new(resolution, subject?)] +
             [newline].compact)
           end)
       end


long-lazuli:
Beware of removing duplicate selector.
It should not be done automatically. Sass never controlled what you are trying to achieve with CSS.

There is a lot of cases, for specificity purpose, where you exactly expect this :

``` SCSS
.foo.foo {
  color: red; }
```

from this :

``` SCSS
.foo { &.foo { color: red; }}
```

It's not a Sass/SCSS problem.
People want to do that, they should be able to.

maciej-ka:
Reason for this implementation was @nex3 suggestion in https://github.com/sass/sass/issues/360. @long-lazuli Could you give any example where `.foo.foo` would be desirable?

long-lazuli:
Let's say we use a plugin which use the class `.foo` to describe the width property to 10em.
We want to surcharge this width, to make this 'plugin/widget/whatever' responsive, to 100%.
We can achieve that to make a specificity higher using twice the class. As in the selector chain `.foo.foo`.

This is not a good practice, but I deal with that kind of tricks enough. That pull request would make some old-fashioned Sass code obsolete.

maciej-ka:
@long-lazuli Right, it is possible to use `.foo.foo` to rise selector specifity. I'm perfectly fine to remove this part of pull request. @nex3, @chriseppstein, what's your opinion on this?

nex3:

PR #2163, opened 2016-10-07
Fix inconsistent evaluation of unary operators inside interpolants

### LINKED ISSUE #2160: Bug with handling / in interpolation
@nex3 confirmed this is a bug. There is currently a sass spec asserting this broken behaviour which is causing LibSass to fail CI.

``` sass
h1 {
  delayd: #{5% / 2};
  delayd: #{+5% / 2};
  evaled: #{-5% / 2};
}
```

Sass 3.4.21

``` css
h1 {
  delayd: 5%/2;
  delayd: 5%/2;
  evaled: -2.5%;
}
```

Expected

```
h1 {
  delayd: 5%/2;
  delayd: 5%/2;
  evaled: -5%/2;
}
```

xzyfer:
I had a quick chat to @nex3 which pointed me to the lexer/parse
as the root cause of the observed inconsistencies.

I think the general approach is sound. I'm sure the implementation
could be tidied up.

Fixes #2160
Spec sass/sass-spec#940

### DIFF
--- a/lib/sass/script/parser.rb
+++ b/lib/sass/script/parser.rb
@@ -36,6 +36,7 @@ module Sass
         @lexer = lexer_class.new(str, line, offset, options)
         @stop_at = nil
         @css_variable_warning = nil
+        @in_interpolant = false
       end
 
       # Parses a SassScript expression within an interpolated segment (`#{}`).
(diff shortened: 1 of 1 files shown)


xzyfer:
This needs more thought.

xzyfer:
This needs more thought.

xzyfer:
@nex3 @chriseppstein this is ready for review

lib/sass/script/parser.rb
```diff
@@ -312,7 +313,12 @@ def unary_#{op}
               start_pos = source_position
 
               @css_variable_warning.warn! if @css_variable_warning
-              node(Tree::UnaryOperation.new(assert_expr(:unary_#{op}), :#{op}), start_pos)
+              e = Tree::UnaryOperation.new(assert_expr(:unary_#{op}), :#{op})
```
nex3, line 316: I don't think triggering based on interpolation is the right approach here—`#{- 5% / 2}` shouldn't parse the same as `#{-5% / 2}`. The lexer should really be lexing `-5%` as a single token.

PR #2165, opened 2016-10-09
Convert README badge to SVG

rmm5t:
Because raster badges make eyes bleed :sunglasses:


nex3:
Thanks for the patch, and sorry it took so long to get reviewed!

Issue #1425, opened 2014-09-10
Support "selector&"


maranomynet:
``` scss
.bar {
    foo& { 
        color: baz;
    }
}
```

Causes an error. (See: http://codepen.io/anon/pen/GIufH)

This is surprising (since `.bar { foo & { ... } }` works like a charm), and this error prevents several useful selector prefixing patterns.

lolmaus:
Use interpolation, @maranomynet:

``` scss
.bar {
    foo#{&} { 
        color: baz;
    }
}
```

Demo: http://sassmeister.com/gist/da2394d0d28b0fe95d59

Works in Sass 3.4.

maranomynet:
Interesting hack. It suppresses the error but...

``` css
.bar foo.bar {  color: baz;  }
```

...is not the desired CSS.

It should be:

``` css
foo.bar {  color: baz;  }
```

lolmaus:
It's not a hack, it's the way meant to be done.

> ...is not the desired CSS.

In this case, use the `@at-root` directive:

``` scss
.bar {
    @at-root foo#{&} { 
        color: baz;
    }
}
```

Demo: http://sassmeister.com/gist/8bd827341243d66e61d5

maranomynet:
Still, the arbitrary discrepancy between this working smoothly:

``` scss
.bar {
    foo & {  color: baz;  }
}
```

and this throwing an error:

``` scss
.bar {
    foo& {  color: baz;  }
}
```

seems like a bug. (Breaking the principle of least astonishment.)

lolmaus:
Using the `&` parent selector without interpolation was a temporary feature in Sass 3.3, it is now deprecated.

Sass follows strict logic for how the parent selector works. It is indeed not very intuitive in more complex cases, but there's no sense in flowing against the current.

maranomynet:
Neither intuitive nor explained in http://sass-lang.com/documentation/file.SASS_REFERENCE.html#parent-selector

lolmaus:
I might be wrong about the deprecation status.

Unfortunately, @chriseppstein [did not](https://github.com/sass/sass/issues/1424#issuecomment-55002054) explain the [differences](http://sassmeister.com/gist/a9a9a2edf2b43623a72d) in behavior between the normal and interpolated parent selector. I believe it might be as follows:
1. If `&` appears in the selector, the previous selector in hierarchy is reconstructed.
2. If `&` does not appear in the selector, the current selector is appended to the previous selector.
[...]

PR #2181, opened 2016-11-02
Enable parent at any position

maciej-ka:
Removes rule that the parent operator (`&`) has to be first symbol in a
sequence. Goal behind this change is to enable parent at suffix
position, like here:
```scss
.foo {
  bar& {
    color: red;
  }
}
```

This version is unsafe as it enables:
- double use of parent operator `bar&&`
- concatenation of selectors: `foo { bar& { a: b; } }` becomes `barfoo {
  a: b; }`

Closes: https://github.com/sass/sass/issues/1425
Tests: https://github.com/sass/sass-spec/pull/960


nex3:

PR #2182, opened 2016-11-02
Exit code 13 when there is a data error on Windows

### LINKED ISSUE #2132: Use platform independent exit codes rather than 65 for data errors
[I raised the issue that libsass doesn't match the Ruby Sass exit code](https://github.com/sass/libsass/issues/2162), which is causing issues when trying to port the sass-spec for node-sass. @mgreter helpfully pointed out that [65 has a different meaning on window](https://github.com/sass/libsass/pull/2163#issuecomment-244283645)

Not sure, but it seems like `EINVAL` might work http://blog.honeybadger.io/understanding-rubys-strange-errno-exceptions/

/cc @mgreter @xzyfer

nschonni:
Closes #2132

This was eyeballed from the web editor, so I might be wrong here

### DIFF
--- a/lib/sass/exec/base.rb
+++ b/lib/sass/exec/base.rb
@@ -23,7 +23,7 @@ module Sass::Exec
         # at_exit is a bit of a hack, but it allows us to rethrow when --trace
         # is active and get both the built-in exception formatting and the
         # correct exit code.
-        at_exit {exit 65} if e.is_a?(Sass::SyntaxError)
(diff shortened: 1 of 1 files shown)


nex3:
Sorry for the long response time! Can you merge in the latest stable so we can get proper Travis results?

nschonni:
Rebased, looks like the build is happy now

nex3:
Thanks!

Issue #2247, opened 2017-02-20
Selector unification is asymmetrical for the universal selector


nex3:
Currently the unification of the universal selector, `*`, isn't symmetrical. `selector-unify("*", ".a")` returns `(("*.a"))`, but `selector-unify(".a", "*")` returns `((".a"))`. The latter is correct; a universal selector without a namespace should be eliminated any time there's another selector present.

PR #2255, opened 2017-03-11
Fix universal selector unification

ivantsepp:
Should address #2247

Hi! trying to get myself involved in the Sass community. I saw #2247 and saw the associated dart changes to fix the issue. Thought I could help by following suit and make the same changes in ruby Sass.

Should I also create a PR for sass-spec ?

### DIFF
--- a/lib/sass/selector/pseudo.rb
+++ b/lib/sass/selector/pseudo.rb
@@ -130,6 +130,9 @@ module Sass
       #
       # @see SimpleSequence#unify
       def unify(sels)
+        if sels.length == 1 && sels.first.is_a?(Universal)
+          return sels.first.unify([self])
+        end
         return if type == :element && sels.any? do |sel|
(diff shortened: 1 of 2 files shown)


nex3:
Sorry for the long wait time for a review! I've been on medical leave for the past couple months.

It looks like this change only partially solves the issue. `selector-unify("*.foo", ".a")` still produces `(("*.foo.a"))`. I think fixing this will require going deeper into the unification code, rather than adding a special case at the top.

ivantsepp:
Ah no problem for the wait - I'm really appreciate you taking your time to review my code! Looks like I need to dive in deeper into the code for a more complete fix.

I may not have too much time on my hands in the upcoming weeks but I'll see what I can do. That being said, I welcome anyone to jump in and take over!

nex3:
Fixed in #2316.

PR #2283, opened 2017-05-06
add chinese-support

n0rv1n:
add chinese-support


nex3:
Sorry for the long wait time for a review! I've been on medical leave for the past couple months.

It's not safe for us to just force an external encoding for the entire process—that's not being a good Ruby citizen. It's up to the user to properly configure their computer's default encoding, and/or to annotate their Sass files to declare which encoding they use.

PR #2294, opened 2017-05-31
DeepCopy bugfix

langalex:
Fixes a bug in Tree::Visitors::DeepCopy, which converted node.keywords to Hash but it should remain a Sass::Util::NormalizedMap.

This bug show up when trying to render a deep-copied tree to scss, where it gives an error "undefined method `as_stored` for Hash".

Sorry that I didn't supply a test - I couldn't figure out where to add it, as there seem to be no tests for DeepCopy so far at all. Any pointers appreciated.


nex3:
Thanks! I think a test probably isn't necessary here.

Issue #303, opened 2012-03-02
Deprecate reference combinators


nex3:
These should be on our radar, although we may want to wait until they're implemented somewhere before we support them. See [Selectors Level 4](http://www.w3.org/TR/selectors4/#idref-combinators).

Here's [a page tracking browser support](http://css4.rocks/selectors-level-4/reference-combinator.php).

Tasks:

- [x] Deprecate existing behavior in `stable`.
- [x] Remove behavior from `master`.

jleclanche:
I don't think this is here to stay; as the issue says it's a lot easier to implement as a functional pseudoclass. The syntax is really ugly, too.

sturobson:
quickly looking in Chrome (latest and Canary), Firefox (latest and Aurora) and Opera (latest(presto)). There seems to be varied support for level 4 selectors now. I've found this test (wrongly named as CSS4) here - 

http://css4-selectors.com/browser-selector-test/

Chrome -

  current -

Indeterminate-value pseudo-class (E:indeterminate)
Default option pseudo-class (E:default)
Validity pseudo-class (E:in-range)
Validity pseudo-class (E:out-of-range)
Optionality pseudo-class (E:required)
Optionality pseudo-class (E:optional)
Mutability pseudo-class (E:read-only)
Mutability pseudo-class (E:read-write)

  canary -

Indeterminate-value pseudo-class (E:indeterminate)
Default option pseudo-class (E:default)
Validity pseudo-class (E:in-range)
Validity pseudo-class (E:out-of-range)
Optionality pseudo-class (E:required)
Optionality pseudo-class (E:optional)
Mutability pseudo-class (E:read-only)
Mutability pseudo-class (E:read-write)

Firefox -

  current -

Indeterminate-value pseudo-class (E:indeterminate)
Default option pseudo-class (E:default)
Validity pseudo-class (E:in-range)
Validity pseudo-class (E:out-of-range)
Optionality pseudo-class (E:required)
Optionality pseudo-class (E:optional)

  aurora -

Indeterminate-value pseudo-class (E:indeterminate)
Default option pseudo-class (E:default)
Validity pseudo-class (E:in-range)
Validity pseudo-class (E:out-of-range)
Optionality pseudo-class (E:required)
Optionality pseudo-class (E:optional)

Opera -

  current - 

Indeterminate-value pseudo-class (E:indeterminate)
Default option pseudo-class (E:default)
Validity pseudo-class (E:in-range)
[...]

PR #2298, opened 2017-06-02
Deprecate the reference combinator.

nex3:
See #303
See sass/sass-spec#1133


nex3:
I'm holding off on merging this until we follow up on https://github.com/sass/sass-spec/pull/1133#issuecomment-305692597.

Issue #2293, opened 2017-05-22
Considering Travis Speed Up


junaruga:
To simplify this topic, I want to talk about `next` branch.
But other branches `stable` and `master` are also same.

Right now `sass`'s Travis tests takes long time until finishing.
The one of the reason is running many Virtual Machines(VMs).

Right now 15 VMs are running on latest next branch.
https://travis-ci.org/sass/sass
https://travis-ci.org/sass/sass/builds/233836704

By this PR: https://github.com/sass/sass/pull/2269 , the number of VM is little bit reduced.
But maybe we still can reduce the number of VMs.

I have 2 ideas to improve this.

### Unify VMs by different UNIX environment variable to 1 VM

Look at below `.travis.yml` of `test_declarative`.

- Before: https://github.com/svenfuchs/test_declarative/blob/d839c374d72f4c9c84d3e0babc03174255debacc/.travis.yml

- After: https://github.com/svenfuchs/test_declarative/blob/master/.travis.yml , https://github.com/svenfuchs/test_declarative/blob/master/test/run.rb

The idea is that I reduced the number of VM by unifying 4 VMs to 1 VM, running `./test/run.rb`.
4 Gemfile cases are running on 1 process `./test/run.rb`.
As the result, `test_declarative`'s Travis speed was much faster than before.

For example, we have below cases for `sass` in `.travis.yml`. And we can unify these cases to 1 VM.

```
- env: MATHN=true RUBOCOP=false
- env: MATHN=false RUBOCOP=false
- env: MATHN=true RUBOCOP=true
```

### A specified version Ruby for 3 VMs?

Ruby 2.2.3, 2.1.7, 2.0.0 are running using 3 VMs for each other.

We really need these 3 cases?

```
- env: MATHN=true RUBOCOP=false
- env: MATHN=false RUBOCOP=false
- env: MATHN=true RUBOCOP=true
```

For example, below 2 cases isn't good enough?

```
- env: MATHN=true RUBOCOP=true
- env: MATHN=false RUBOCOP=true
```

How do you think?

Thanks.

nex3:
I'm definitely on board with consolidating the environment variable matrix. We only need to test Rubocop once, so

```yaml
- env: MATHN=true RUBOCOP=true
- env: MATHN=false RUBOCOP=false
```

should be fine.

As for doing multiple test runs in a single VM, it's at least worth experimenting to see if it produces a net speed gain.

junaruga:
[...]

PR #2302, opened 2017-06-02
Speed up Travis tests

junaruga:
Applied the case 5 of #2293. This fixes the ticket.
* Reduced 3 VMs (the idea 2)
* Unified VMs by different UNIX environment variable to 1 VM (the idea 1)
* Run "rake test:spec" just once for each Rubies.

We can expect the Travis total running time ("Ran for") to change from about 10 minutes to 3:50 minutes by this PR. (75% reduced!)


nex3:
Can we make `extra/rake_with_matrix.rb` part of the Rakefile? That's where other tasks like it go.

junaruga:
Yes! That's a great idea.
Let me try it.

junaruga:
@nex3 I added 1 more commit to make the matrix part in `Rakefile`.

I deleted the logic to control `rubocop` task by `RUBOCOP` environment variable.
Because I wanted to avoid the rubocop rake task to be pre-defined by the environment variable. I want to control it from the matrix test task, referring this `Rakefile`: https://github.com/colszowka/simplecov/blob/master/Rakefile
But surprisingly, Rubocop works for every Ruby including Ruby < 2.2.0 now.
Rubocop supports Ruby 2.0.0 too.
https://github.com/bbatsov/rubocop/blob/master/.travis.yml
So, I changed the warning message too.

Also I unified the matrix test with `test`, because it looks simple.

I left 2 commits to compare easily.
If you like my modification, I can squash my 2 commits to 1 commit by myself.
Or you can squash it by yourself, when merging my commits.

nex3:
This is looking good, but I'm worried that it's difficult to visually distinguish the various tests that are being run. Can you have it print a big header before each section? Something like

```
##################################################
################### Running tests ################
##################################################
```

junaruga:
OK, I updated like this!

```
################################################################################
############### Running task: test:ruby, env: {"MATHN"=>"true"} ################
################################################################################
```

nex3:
Looks great, thanks for all your hard work!

Issue #2295, opened 2017-05-31
Help generating docs


lostapathy:
I'd like to submit a PR fixing some broken links in the docs, but I'm having a hard time building them to match what's getting published on sass-lang.com

I've done the usual
```
git clone
bundle install
rake doc
```
But I seem to get a different output than is currently on the website.

For example, http://sass-lang.com/documentation/file.SASS_REFERENCE.html#options takes you to the correction section, but hitting doc/file.SASS_REFERENCE.html#options in my working copy does not.

My local copy is generally formatted differently as well, for example I don't have the nice right-hand navigation that shows up on the sass-lang.com link above.

I've tried building with both a current version of yard and the 0.8.7.6 that the docs on sass-lang.com indicate they were built with and get the same results.

I've got a few corrections already identified, and if I can get setup to build them correctly I'll take a shot at fixing some of the doc warnings as well.

nex3:
You may be using the wrong version of Yard... I just compiled with 0.8.7.6 (the version in the lockfile) and I got the sidebar and working inner links. Note that to get it to look exactly like the website, you'll need to compile from within the [sass-site](https://github.com/sass/sass-site) repo, since it adds some custom CSS.

lostapathy:
Thanks @nex3 - pointing me to the sass-site project seems to fill in a lot of pieces.  FWIW, it doesn't quite build out of the box in an otherwise fresh environment, but I'll submit a PR over there for that.

When I build that, it builds with yard 0.9.9 which I believe is a result of [this line of the gemspec](https://github.com/sass/sass/blob/stable/sass.gemspec#L22) - can you point me at the lockfile that's locked to 0.8.7.6?

lostapathy:
I found the root of this. [This commit](https://github.com/sass/sass/commit/2980ac65ae6adfee04243b9f16469277f6e2bc6d) changed from maruku to redcarpet, which explains what I'm seeing.  Mainly, that anchors in the generated html with redcarpet are different than in the older docs (that I assume were generated with maruku).

For example, maruku generates lowercase anchors, such as:
http://sass-lang.com/documentation/file.SASS_REFERENCE.html#options

But redcarpet generates them uppercase, it would generate:
http://sass-lang.com/documentation/file.SASS_REFERENCE.html#Options

Is it safe to assume that the correct path forward is to update any broken links to use the redcarpet-style anchors?

[...]

PR #2305, opened 2017-06-03
Updating broken doc links

lostapathy:
Updates several broken doc links as discussed on #2295

Downgraded dependency on yardoc to 0.8.7.6 to make output match current docs on sass-lang.org

There are still some broken links related to the switch from maruku to redcarpet.  Starting around line 156 of [SASS_REFERENCE.md](https://github.com/sass/sass/blob/stable/doc-src/SASS_REFERENCE.md) there is some markdown that maruku was turning into HTML dl/dt/dd that both redcarpet and github's markdown dialect don't support.  I left that alone since there wasn't an obvious path forward, but can follow up on that later if given guidance.

I did the work on stable branch - if future work should start on master instead, let me know.

### DIFF
--- a/lib/sass.rb
+++ b/lib/sass.rb
@@ -47,7 +47,7 @@ module Sass
   #
   # @param contents [String] The contents of the Sass file.
   # @param options [{Symbol => Object}] An options hash;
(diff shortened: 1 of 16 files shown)


nex3:
Thanks!

PR #2306, opened 2017-06-07
Chore: Test out Travis Stages

nschonni:
Cleans up the RVM since most weren't being run anymore and moves to the Travis build stages https://docs.travis-ci.com/user/build-stages#Examples

This could probably be changed later like node-sass where we have on lint (or Rubocop) build that runs before all of the other platforms

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: ruby
 cache: bundler
-install:
+before_install:
   # If we're running for a pull request, check out the revision of sass-spec
   # referenced by that pull request.
   - if-pr() { if [ ! -z "$TRAVIS_PULL_REQUEST" -a "$TRAVIS_PULL_REQUEST" != false ]; then "$@"; fi }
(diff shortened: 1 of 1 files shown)


nschonni:
Think I was mistaken with this one, looks like the next/master have been trimmed and this should probably stay as-is

junaruga:
@nex3 

Seeing this document
https://github.com/sass/sass/blob/stable/CONTRIBUTING.md
> The stable branch is the default—it's what GitHub shows if you go to sass/sass, and it's the default place for pull requests to go. This branch is where we work on the next patch release. Bug fixes and documentation improvements belong here, but not new features.

In my understanding, "moves to the Travis build stages" is new feature.
And the stage feature is used for the use case of current parallel tasks and "additional sequential task" such as deploy. So, the stage feature may be not suitable for current `sass` workflow.
Ref: https://docs.travis-ci.com/user/build-stages#Build-Stages-and-Build-Matrix-Expansion

However I think that "most weren't being run anymore" is a kind of bug. what we have to fix on stable branch.

Yes, next/master branches have been trimmed the outdated RVM list.
https://github.com/sass/sass/blob/next/.travis.yml
https://github.com/sass/sass/blob/master/.travis.yml

We have a choice.
* Apply the modification (https://github.com/sass/sass/commit/7ad03a953442619aa37dcc48283ceb16ca31e92f) on next/master branch  to stable branch too.
* Only remove outdated RVMs on stable branch.
  * On stable branch, for example, can we update "2.3.0" to "2.3.4"?

How do you think?

nex3:
Adjusting Travis config isn't relevant to the feature/bug dichotomy, because it's in no way visible to downstream users of the package. So we should probably cherry-pick 7ad03a9 onto stable.

junaruga:
@nex3 ok I understood the policy.

PR #2308, opened 2017-06-07
Back port: Speed up Travis tests

junaruga:
This is a cherry-pick and back port to stable branch from next branch: 7ad03a953442619aa37dcc48283ceb16ca31e92f .

This PR is from below discussion.
https://github.com/sass/sass/pull/2306#issuecomment-306887738

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,22 +14,13 @@ install:
   - bundle update sass-spec
 rvm:
   - 1.8.7
-  - 1.9.2
   - 1.9.3
   - 2.0.0
-  - 2.1.7
-  - 2.2.3
-  - 2.3.0
-  - jruby-18mode
-  - jruby-19mode
-  - jruby-21mode
-  - jruby-9.0.0.0
-  - rbx
-  - ree
+  - 2.1.10
+  - 2.2.7
+  - 2.3.4
+  - 2.4.1
 gemfile: Gemfile
-env:
-  - MATHN=true RUBOCOP=false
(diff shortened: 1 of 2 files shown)


nschonni:
I'm not sure if the 1.8.x and 1.9.x ones should stay since they are passing on the current stable.
:100: on removing the JRuby and `rbx` since they've been failing forever

junaruga:
@nschonni 

> I'm not sure if the 1.8.x and 1.9.x ones should stay since they are passing on the current stable.

Yeah, that is my concern.
In the case of `next` branch, those were removed by commit: https://github.com/sass/sass/commit/bceb3e61b6e070669b47fbe935ebe90f85e286ff .

Will we remove those Rubies? or Remove only `1.9.2`?

```
- 1.8.7
- 1.9.2
- 1.9.3
```

As a information, Ruby < 2.2 has not been maintained officially.
https://www.ruby-lang.org/en/downloads/

nex3:
We still need to support 1.8 and 1.9 for stable, so we should continue testing them.

junaruga:
Ok but why does Ruby 1.9 have 2 cases both 1.9.2 and 1.9.3?
Keeping 1.9.3 is good enough isn't it?

nex3:
Yeah, just 1.9.3 should be fine.

junaruga:
OK! I rebased with keeping Ruby 1.8 and 1.9.

nex3:
Thanks!

PR #2322, opened 2017-06-16
Suppress warning from minitest.

junaruga:
I fixed to suppress the warning "DEPRECATED: Use assert_nil if expecting nil" by minitest.

As a information, other kind of warnings are from other gem packages.

* [DEPRECATION] `last_comment` is deprecated.  Please use `last_description` instead.
  => From rubocop 0.33.0 (latest version is 0.49.0)
* ... celluloid-0.16.0/lib/celluloid/logging/incident_logger.rb:96: warning: parentheses after method name    
* ... celluloid-0.16.0/lib/celluloid.rb:513: warning: global variable `$CELLULOID_TEST' not initialized
* ... celluloid-0.16.0/lib/celluloid.rb:138: warning: instance variable @shutdown_registered not initialized
  => From celluloid 0.16.0 (latest version is 0.17.3)
* "fatal: Not a git repository (or any of the parent directories): .git"
  => From ?
Those can be suppressed when we update those gems to latest version in the future.

### DIFF
--- a/test/sass/engine_test.rb
+++ b/test/sass/engine_test.rb
@@ -3394,7 +3394,13 @@ SASS
   private
 
   def assert_hash_has(hash, expected)
-    expected.each {|k, v| assert_equal(v, hash[k])}
+    expected.each do |k, v|
+      if v.nil?
+        assert_nil(hash[k])
+      else
+        assert_equal(v, hash[k])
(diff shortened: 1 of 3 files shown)


nex3:
Excellent, thank you!

PR #2342, opened 2017-07-12
Remove uneeded shebang and executable bit

junaruga:
This PR is to `stable` branch.

Some test files are including shebang [1] (`#!/usr/bin/env ruby`), and have executable bit.

However if we do not run the file like this `./file`, we do not need it.
So, I removed those unneeded shebang and executable bit for the test files.

I executed the modification like this.

Check target test files for the shebang.

```
$ grep -rn '/usr/bin/env ruby' test/
```

Remove shebang line for the target test files.

```
$ grep -rl '/usr/bin/env ruby' test/ | xargs sed -i '/\/usr\/bin\/env ruby/d'
```

Check test files with executable bit.

```
$ find test/ -type f -perm -u=x
```

Remove the executable bit for the test files.

```
$ find test/ -type f -perm -u=x | xargs chmod 664
```

[1] https://en.wikipedia.org/wiki/Shebang_(Unix)


nex3:
Thanks!

PR #2353, opened 2017-07-22
Update sass-spec logic in .travis.yml for readability.

junaruga:
Just small modification.
The reason of PR is because I believe that readable `.travis.yml` promotes pull-requests.

YAML block style "|" helps for the readability. :)
I also used "$(STATEMENT) instead of legacy `STATEMENT`".
Ref: https://github.com/koalaman/shellcheck/wiki/Sc2006


nex3:
Thanks!

Issue #1799, opened 2015-08-19
Parsed ListLiterals have wrong end_pos


srawlins:
I found that ListLiteral nodes (particularly those parsed from an `@each` list), have the end_pos of their _first_ element, rather than the last. Using this patch [1], we can see:

```
 bundle exec ./bin/sass --scss <<EOF
> @each \$a, \$b in (foo, bar), (baz, quux) { }
> EOF
LIST `(("foo", "bar"), ("baz", "quux"))` HAS RANGE: (1:16 to 1:27)
```

For clarity, here's the positions of the string I gave it:

```
@each $a, $b in (foo, bar), (baz, quux) { }
   0   0   1   1   2   2   2   3   3   4
   4   8   2   6   0   4   8   2   6   0
               ____________
16 to 27 are >> (foo, bar),<<
               ------------
```

[1] patch file:

``` diff
diff --git i/lib/sass/scss/parser.rb w/lib/sass/scss/parser.rb
index a066e5d..4d0df48 100644
--- i/lib/sass/scss/parser.rb
+++ w/lib/sass/scss/parser.rb
@@ -299,6 +299,7 @@ module Sass

         tok!(/in/)
         list = sass_script(:parse)
+        puts "LIST `#{list.inspect}` HAS RANGE: #{list.source_range.inspect}"
         ss

         block(node(Sass::Tree::EachNode.new(vars, list), start_pos), :directive)
```

PR #2394, opened 2017-10-26
Update map source range as new elements are parsed

tagliala:
MapLiteral nodes have the end_pos of their first element, rather than
the last. This PR fixes the wrong behavior.

Ref: #1799


nex3:
Thanks!

PR #2402, opened 2017-11-14
Update Travis to 2.2.8, 2.3.5, 2.4.2, add ruby-head

MSP-Greg:
Somewhat in response to [Ruby Core Issue 14106](https://bugs.ruby-lang.org/issues/14106).

Updated versions are available and 'head' will be released soon.

May be an issue with Pathname...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,10 @@ install:
 rvm:
   - 2.0.0
   - 2.1.10
-  - 2.2.7
-  - 2.3.4
-  - 2.4.1
+  - 2.2.8
+  - 2.3.5
+  - 2.4.2
+  - ruby-head
 gemfile: Gemfile
 branches:
   only:
@@ -28,5 +29,8 @@ branches:
     - stable
     - stable_3_2
     - next
+matrix:
+  allow_failures:
+    - rvm: ruby-head
 notifications:
   irc: {channels: "irc.freenode.org#sass"}


nex3:
I'd like to avoid any `allow_failures` entries; they just slow down the build without providing much meaningful data.

MSP-Greg:
> I'd like to avoid any allow_failures entries

Up to you.  I use (and build) trunk, and I've found more than a few issues that were not shown in the ruby test suites.  I would think that finding out now is better than when 2.5 is released?

I'm more familiar with Appveyor builds, and `allow_failures` is not uncommon with ruby-head.  I've also seen several Travis builds set that way.  Two 'ruby only' examples are ruby/spec and rubygems/rubygems, which do so on both Travis & Appveyor.

Don't know how much you follow trunk / ruby-head, but it now includes bundler and rake (as before), and the current Travis build has been somewhat odd re bundled gems.  Compare builds available on [Appveyor](https://ci.appveyor.com/project/MSP-Greg/appveyor-ruby) vs [Travis](https://travis-ci.org/MSP-Greg/travis-ruby) (the trunk info is listed at the bottom).

If you'd like, I could try Appveyor ruby-head on my fork and see if it passes?

MSP-Greg:
Local `rake test` on `ruby 2.5.0dev (2017-11-16 trunk 60813) [x64-mingw32]`.  All three look OS related.
```
2115 runs, 12670 assertions, 2 failures, 1 errors, 0 skips
```

nex3:
In my experience, `allow_failures` builds either get ignored or they add a bunch of overhead to each build in figuring out whether the allowed failures are something we can actually do something about, and if so how. Look at [the head build](https://travis-ci.org/sass/sass/jobs/302211019) for this pull request for example: it fails with a block of assembly code that's totally opaque to me. What am I supposed to do with that? I don't want to spend a couple hours narrowing it down to a reproduction case, so probably I'd just ignore it. But do I check `allow_failures` for every build from here on to see if that goes away and gets replaced by some more actionable error? I just don't have time for that kind of thing.

nicolasleger:
You can use `2.x`notation to auto use patched versions as soon as available on Travis. Also add newest `2.5`

Issue #690, opened 2013-03-20
Allow 'index' to be a possible file to import


anthonyshort:
It would be great to allow an `index.scss` file to be imported if a directory is found. Example

``` scss
@import "./fruits"
```

If `./fruits` is found to be a directory, it would look for `./fruits/index.scss` or `./fruits/index.sass`. 

The reason behind this is to make it easier to consume packages containing Sass files. For example.

![Screen Shot 2013-03-20 at 11 07 41 PM](https://f.cloud.github.com/assets/36125/280431/cdef5206-9156-11e2-8c61-572863860a9b.png)

We can then download packages into a directory, add that directory as a single load path and then in our main file we can easily do:

``` scss
@import "nib-styles-common";
@import "nib-styles-dialog";
```

Which is a little neater than any alternative.

anthonyshort:
I attempted to do a pull request for this, but alas, my skills are weak :cry:

chriseppstein:
I love this idea.

nex3:
This seems reasonable to me.

robwierzbowski:
Isn't this basically [Sass globbing](https://github.com/chriseppstein/sass-globbing)?

anthonyshort:
Not really. It more like an alias. It could theoretically live anywhere. If Sass packages all do this then there is a guaranteed, single entry point for every library.

chriseppstein:
I think we should make this part of 4.0 when we add our module system. it
may be better to use a different file name than "index" that is in keeping
with whatever we come up with for our modules. E.g. `_module.s[ac]ss`.

robwierzbowski:
I think I see your point. So the difference is:
- Sass Globbing: imports all files in a directory or directory tree
- Index Importing imports only the index (or whatever naming convention is used) in a directory, reducing functionality but enforcing a uniform rule

If this rule would be for all stylesheets, including site/application stylesheets, `_style` might be better — `index` is an odd conflation with html. Chris's suggestion `_module` works too.

nex3:
`index` is nice because it's consistent with conventions used by other languages, but we would want it to be a partial. `_index` is less similar, so `_module` might be a better way to go for that reason.

anthonyshort:
I only suggested index because that is what I'm used to with node and Component. The name doesn't particular matter too much. It would be nice to get this before 4.0 as it seems like it could be trivial to implement and would make using Sass with tools like Bower and Component easier.

anthonyshort:
[...]

PR #2456, opened 2018-01-23
Implement Index a default file

john681611:
Attempting to fix https://github.com/sass/sass/issues/690

Simple implementation that if no file in current level is found it looks if it's a directory and searches for a valid index file inside.

Though I have the tests working I am unsure of how to run the code locally in a manual test.
 
Working with:
@liamshaunsmith
@mariann013 

See sass/sass-spec#1206

### DIFF
--- a/lib/sass/importers/filesystem.rb
+++ b/lib/sass/importers/filesystem.rb
@@ -155,7 +155,7 @@ module Sass
           end
         end.flatten(1)
         if found.empty? && split(name)[2].nil? && File.directory?("#{dir}/#{name}")
-            return find_real_file("#{dir}/#{name}", "index", options)
+          return find_real_file("#{dir}/#{name}", "index", options)
         end
 
         if found.size > 1 && !@same_name_warnings.include?(found.first.first)
--- a/lib/sass/importers/filesystem.rb
+++ b/lib/sass/importers/filesystem.rb
@@ -154,12 +154,8 @@ module Sass
             [Sass::Util.cleanpath(full_path).to_s, s]
           end
         end.flatten(1)
-        if found.empty?
-          if File.directory?("#{dir}/#{name}")
+        if found.empty? && _extension?(name) && File.directory?("#{dir}/#{name}")
             return find_real_file("#{dir}/#{name}", "index", options)
-          end
-
-          return
         end
 
         if found.size > 1 && !@same_name_warnings.include?(found.first.first)
@@ -219,6 +215,14 @@ WARNING
         options[:filename] = full_filename
         options[:importer] = self
         Sass::Engine.new(File.read(full_filename), options)
+      end
+
+      # Given a name checks for any extentions
+      #
+      # @param name [String] The filename.
+      # @return [Boolean] Wheather a exetntion exits or not.
+      def _extension?(name)
+        split(name)[2].nil?
       end
     end
   end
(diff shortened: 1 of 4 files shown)


nex3:
Thanks for the contribution!

Because we're in the process of [deprecating Ruby Sass](http://sass.logdown.com/posts/1022316-announcing-dart-sass), we aren't currently adding new features to Ruby only. If you're willing to add this behavior to [Dart Sass](https://github.com/sass/dart-sass) as well and add cross-platform tests to [sass-spec](https://github.com/sass/sass-spec), then we can land it in both places at once.

john681611:
Ok I've got zero experience in dart so I will have a go

nex3:
Give it a shot! Dart is pretty easy to learn.

nex3:
Same comments as on the Dart PR: if the import ends in `.scss` or `.sass` it shouldn't look for the index, and the tests aren't necessary here.

lib/sass/importers/filesystem.rb
```diff
@@ -154,7 +154,9 @@ def find_real_file(dir, name, options)
             [Sass::Util.cleanpath(full_path).to_s, s]
           end
         end.flatten(1)
-        return if found.empty?
+        if found.empty? && _extension?(name) && File.directory?("#{dir}/#{name}")
```
nex3: I don't understand what `_extension?(name)` is doing here... are we only checking for an index file if `name` has an extension? That seems wrong.

john681611: We figured it would be more readable than `split(name)[2].nil?` I'll switch back

nex3: I meant more "why do we need to check for an extension at all" than "why is it factored out into a private method"

john681611: Because of your above comment that if it has an extension in the name we shouldn't look at all.

nex3: I just meant a `.sass` or `.scss` extension. If we disallow *any* extension, [this spec](https://github.com/john681611/sass-spec/tree/bd1e7759b7641cc88b4f1a49b7058e08e5a48b42/spec/directives/import/index/dir_dot_foo) will fail (and in fact I'm not sure why it isn't failing already...)

nex3: Oh, sorry, I was being dumb here... I didn't realize that this was using the `split()` method defined below that already checks for the correct file extensions. Your original code was fine; I'll revert b196d9c.

AndrewBogdanovTSS:
Any info on when this feature might be released?

john681611:
I believe its released in https://sass-lang.com/dart-sass, ruby-sass is being deprecated so I'd look into moving to that.

nex3:
This was released as part of Ruby Sass 3.6.0.

john681611:
> This was released as part of Ruby Sass 3.6.0.

Something I have been wondering is did this make it into lib-sass or is that another separate piece of work I need to do?

Basically, we need a web pack sass-loader that supports dart-sass and does de-duping.

nex3:
> Something I have been wondering is did this make it into lib-sass or is that another separate piece of work I need to do?

It looks like this landed in https://github.com/sass/libsass/pull/2614.

> Basically, we need a web pack sass-loader that supports dart-sass and does de-duping.

The Webpack Sass loader support Dart Sass now. You just need to pass `implementation: require("sass")` to it.

PR #2547, opened 2018-07-30
Doc: @extends complex selector deprecation

butterflyhug:
Update the Sass reference documentation to account for the decision
to remove support for `@extends` with complex selectors.

Per https://github.com/sass/sass/issues/1599#issuecomment-371979590,
https://github.com/sass/libsass/issues/2418, and related issues.


nex3:
Thanks for this fix! Those docs have languished a bit in the last few years.

nex3:
Thanks for this fix! Those docs have languished a bit in the last few years. This repo's contents are no longer active, though; please submit this to [sass/ruby-sass](https://github.com/sass/ruby-sass) instead.

PR #2561, opened 2018-09-13
Remove link to Compass in intro

GregoryAveryWeir:
[Compass is no longer supported](https://github.com/Compass/compass/blob/stable/README.markdown), so it should probably not be endorsed as prominently as it is here.


nex3:
Thanks for the contribution! This repository is currently frozen; please file pull requests against [sass/ruby-sass](/sass/ruby-sass) instead.

Issue #2670, opened 2019-04-08
Add continuous integration


nex3:
We should add continuous integration testing that can catch (at least) the following errors:

- [x] Markdown links missing a referent.
- [x] Markdown links to headings that don't exist.
- [x] Tables of contents missing heading links.

TeresaPartidaS:
Hi! I'd like to help, can you tell me more about this?

nex3:
We need to set up a [Travis CI](https://travis-ci.org/) config that will run a series of tests on the Markdown files in this repository to ensure links aren't broken and tables of contents include all headings in the document. I'm not sure what tools exist for this; ideally I don't want to implement something from scratch, but we'll probably need to write a plugin for table-of-contents validation. A cursory search indicates that [remark](https://www.npmjs.com/package/remark-cli)'s lint mode might be a good option, if it's extensible enough.

nschonni:
@TeresaPartidaS I've used https://github.com/tcort/markdown-link-check for this before. I usually do a small wrapper script like this for the globbing of files https://github.com/canada-ca/digital-playbook-guide-numerique/blob/6a154f38bf406faabc492fb251ae2d09c37b987e/link-check.js. You might want an external config like https://github.com/canada-ca/digital-playbook-guide-numerique/blob/6a154f38bf406faabc492fb251ae2d09c37b987e/.markdown-link-check.json to set the site root and any links that need to be suppressed (since sometimes sites blacklist Travis-CI)

TeresaPartidaS:
Thanks! I'm actually new to this so that's of great help

nschonni:
No problem, there are probably also other solutions in different languages that you can find. I'm not sure if there is any preference for the CI setup.

TeresaPartidaS:
@nex3 Hi! I managed to integrate Travis CI and created two files to catch the errors you mentioned before, I already did the PR :D

PR #2682, opened 2019-05-02
Added continuous integration

TeresaPartidaS:
Created Travis CI file and test files to catch the errors mentioned in this [issue](https://github.com/sass/sass/issues/2670)


nex3:
Thank you so much for diving into this!

It looks like the tests are currently failing, but I can't figure out what the problem is with the tables of contents that are flagged as incorrect. Would it be possible to provide a diff so we can see exactly which lines are wrong?

.gitignore
```diff
@@ -0,0 +1,102 @@
+
+# Created by https://www.gitignore.io/api/node,visualstudiocode
```
nex3: I'd like to leave out the visual studio ignores; as a rule of thumb, IDE-specific ignores should go in your [global gitignore](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) so that every project doesn't end up having ignores for every IDE.

.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+node_js:
+  - "8.15.1"
```
nex3:
```suggestion
  - stable
```

This way we won't have to update the file as Node versions come and go.

test/toc-check.js
```diff
@@ -0,0 +1,54 @@
...
+
+var files = glob.sync("**/*.md", {ignore: ["node_modules/**/*.md"]})
+
+function removeTocHeader(str, ele, arr) {
+  return str.indexOf('Table of Contents') === -1;
+}
+
+function getCurrentToc(str, remove){
+  if(remove === undefined) remove = 0;
+  var startIndex = str.indexOf('Table of Contents') + 19;
```
nex3: Why 19?

test/toc-check.js
```diff
@@ -0,0 +1,54 @@
...
+
+    if(currentToc !== generatedToc) {
+      var diff = Diff.diffLines(currentToc, generatedToc);
+      diff.forEach(function(part){
+        var color = part.added ? 'green' :
+          part.removed ? 'red' : 'grey';
+        process.stderr.write(part.value[color]);
+      });
+
+      console.log('\n');
```
nex3: If you're writing the diff to stderr, the newline should be written to stderr as well.

test/toc-check.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function getCurrentToc(str, remove){
+  if(remove === undefined) remove = 0;
+  var startIndex = str.indexOf('Table of Contents') + 19;
+  var endIndex = str.indexOf("## ", startIndex);
+  var result = str.substring(startIndex, endIndex - remove);
+  
+  if(result[result.length -1] == '\n' || result[result.length -1] == ' '){
+    return getCurrentToc(str, remove+1);
```
nex3: I'm not sure what this recursion is for.

nex3 (changes requested):
The new output looks really good!

test/toc-check.js
TeresaPartidaS: It was an easy way to test the files and I forgot to modified it, I already fix it :D

test/toc-check.js
TeresaPartidaS: It was for removing new lines and blank spaces, I just modified it to make it more readable

.gitignore
TeresaPartidaS: Done :raised_hands:

test/toc-check.js
nex3: Could you just call [`String.prototype.trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim) on the string instead?

TeresaPartidaS: Done! Sorry, I completely forgot about that :sweat_smile:

nex3:
Almost there! I've been testing this locally, though, and it looks like `markdownLinkCheck()` isn't verifying internal anchor links. For example, if I write

```markdown
[This link](#nowhere) doesn't match any headers in this file.
```

that doesn't get flagged as an issue. That's probably the most common form of broken link we run into, though, so it's very important that it get flagged.

TeresaPartidaS:
> Almost there! I've been testing this locally, though, and it looks like `markdownLinkCheck()` isn't verifying internal anchor links. For example, if I write
> 
> ```gfm
> [This link](#nowhere) doesn't match any headers in this file.
> ```
> 
> that doesn't get flagged as an issue. That's probably the most common form of broken link we run into, though, so it's very important that it get flagged.

That's marked as an issue in the test for the TOC because the header doesn't exist

nex3:
That's not part of a table of contents, though, it's just a normal paragraph. Testing it locally, I don't get any error from the TOC test.

TeresaPartidaS:
Oh! Then I'll check that :D

TeresaPartidaS:
Hi! Sorry for the delay, I made some changes for the internal links issue :D

test/link-check.js
```diff
@@ -21,7 +22,18 @@ files.forEach(function (file) {
 
     console.log('Reading: ' + file)
 
+    var markdownToc = toc(markdown, {
+      filter: (string, _, __) => string.indexOf('Table of Contents') === -1
+    }).content
+
     results.forEach(function (result) {
+      if (result.link.match(/^#/)) {
+        if (!markdownToc.includes(result.link)) {
```
nex3: This is a clever solution!

nex3 (approved):

nex3:
Thanks so much for your contribution, Teresa! This is a huge help—you can see how many broken links and tables of contents were sitting there unrecognized!

Issue #2565, opened 2018-09-21
Stop using / as division


nex3:
**The current plan:**

1. Create an automated migration tool that can migrate users from `/`-as-division to a `divide()` function.
2. Deprecate `/`-as-division. Any time a `/` would be interpreted unambiguously as division, or an ambiguous `/` would be made unambiguous (for example by being stored in a variable), we produce a deprecation warning .
3. Add a top-level `divide()` function (`math.div()` in the new module system). This will work exactly the same way division works today, with the obvious exception that it will never be treated as a slash separator.
4. Add a new slash-separated list type to represent slash as a separator. For the time being, this will not have a literal syntax.
5. Add a `slash-list()` function (`list.slash()` in the new module system) that returns a slash-separated list. Providing the ability to work with slash-separated lists even before they have a literal syntax will help stylesheets be forwards-compatible with the new wold.
6. Wait until at least three months after the module system is released.
7. Release Dart Sass 2.0.0, which parses `/` as a list separator that binds more loosely than space but more tightly than comma.
8. Add support for [first class `calc()`](https://github.com/sass/sass/issues/2186), which will allow `/`-as-division in the form of `calc(5px / 2)`.
8. Deprecate the `slash-list()`/`list.slash()` function in favor of using the literal syntax.
9. Whenever Dart Sass 3.0.0 releases, remove `slash-list()` and `list.slash()`.

* [x] [Migrator](https://github.com/sass/migrator/issues/37)
* [x] [Proposal](https://github.com/sass/sass/blob/master/accepted/slash-separator.md)
* Deprecation, `divide()`, and `slash-list()`
  * [x] [Specs](https://github.com/sass/sass-spec/pull/1399)
  * [x] [Dart Sass](https://github.com/sass/dart-sass/issues/663)
  * ~[LibSass](https://github.com/sass/libsass/issues/2887)~
* Removal of /-as-division, deprecation of `slash-list()`
  * [x] Update deprecations (https://github.com/sass/sass/pull/4131)
  * [x] Specs (https://github.com/sass/sass-spec/pull/2086)
  * [x] [Dart Sass](https://github.com/sass/dart-sass/issues/663) (https://github.com/sass/dart-sass/pull/2649)
  * [ ] Sass Site
  * ~[LibSass](https://github.com/sass/libsass/issues/2887)~
* Removal of `slash-list()`
  * [ ] Update deprecations
  * [ ] Specs
[...]

PR #2683, opened 2019-05-02
[Slash Separator] Add a proposal for forward slash as a separator

nex3:
See #2565

### DIFF
--- a/proposal/slash-separator.md
+++ b/proposal/slash-separator.md
@@ -10,6 +10,10 @@ operator.
 
 * [Background](#background)
 * [Summary](#summary)
+* [Alternatives Considered](#alternatives-considered)
+  * [Different Operator Syntax](#different-operator-syntax)
+  * [First-Class `calc()`](#first-class-calc)
+  * [`math()` Syntax](#math-syntax)
 * [Existing Behavior](#existing-behavior)
 * [Syntax](#syntax)
 * [Semantics](#semantics)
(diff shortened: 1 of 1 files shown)


proposal/slash-separator.md
```diff
@@ -0,0 +1,415 @@
...
+3. The third stage will just remove the `slash-list()` function. This is not a
+   priority, and will be delayed until the next major version release.
+
+## Alternatives Considered
+
+> This section is non-normative.
+
+### Different Operator Syntax
+
+On other possible fix would be to change the syntax for division to another
```
nshahan: On --> One

proposal/slash-separator.md
```diff
@@ -0,0 +1,415 @@
...
+  * Let `hue`, `saturation`, and `lightness` be the three elements of `hsl`.
+
+  * Call `hsl()` with `hue`, `saturation`, `lightness`, and `alpha` as arguments
+    and return the result.
+
+* Otherwise, proceed with the existing definition of the function.
+
+## Deprecation Process
+
+The deprecation process will be divided into three phases:
```
nshahan, line 353: It could be helpful to include examples of the recommended way for end users to write division and `/` separated lists in their Sass files during each phase.

proposal/slash-separator.md
```diff
@@ -0,0 +1,415 @@
...
+## Summary
+
+> This section is non-normative.
+
+We will redefine `/` to be *only* a separator. Rather than creating an unquoted
+string (as it currently does when at least one operand isn't a number), it will
+create a slash-separated list. As such, lists will now have three possible
+separators: space, comma, and slash.
+
+Division will instead be written as a function, `divide()` (or `math.div()` in
```
nshahan, line 62: At what point is the top level `divide()` expected to be removed?

nshahan (approved):
A few comments but nicely written.

proposal/slash-separator.md
nex3: In favor of `math.div()`? At the same time as other top-level functions, which is [very tentatively set to be 1 October 2021](https://github.com/sass/sass/blob/master/accepted/module-system.md#timeline).

proposal/slash-separator.md
nex3: Good idea, done.

PR #2718, opened 2019-06-19
[Module System] Fix a bunch of minor outstanding issues

nex3:


accepted/module-system.md
```diff
@@ -1355,6 +1355,13 @@ type `type`, and an [import context](#import-context) `import`:
 * Otherwise, if `member-uses` contains a single module, return the member of
   type `type` named `name` in that module.
 
+* Otherwise, if the implementation defines a global member `member` of type
+  `type` named `name`, return that member.
+
+  > This includes the global functions and mixins defined as part of the Sass
+  > spec, and may also include other members defined through the
```
nshahan, line 1370: Are the names of global functions and mixins defined in the Sass spec protected in another part of the spec? Can modules override those methods if they want since these are returned last?

accepted/module-system.md
```diff
@@ -1015,14 +1015,14 @@ Given a source file `file`, a [configuration](#configuration) `config`, and an
...
-      flag at the root of the stylesheet, throw an error.
+      or imported by `file` contains a variable declaration named `name` with a
+      `!default` flag at the root of the stylesheet, throw an error.
 
       > Although forwarded modules are not fully loaded at this point, it's
       > still possible to statically determine where those modules are located
       > and whether they contain variables with default declarations.
       >
-      > Implementations may choose to verify this lazily, after `file` has been
+L      > Implementations may choose to verify this lazily, after `file` has been
```
nshahan: delete "L".

nshahan (approved):
A couple comments but LGTM.

accepted/module-system.md
nex3: Done.

accepted/module-system.md
nex3: No, they aren't protected. It's totally valid to define a member with the same name as a global member, and if you do it overrides the global member definition.

PR #2722, opened 2019-07-12
Allow prefixed parent selector

oles:
Issue: https://github.com/sass/sass/issues/1425

PR #2723, opened 2019-07-13
Prefixed parent selector proposal

oles:
Issue: https://github.com/sass/sass/issues/1425
Previous draft/PR: https://github.com/sass/sass/pull/2722


proposal/prefixed-parent-selector.md
```diff
@@ -0,0 +1,58 @@
+# Prefixed Parent Selector: Draft 2
```
nex3:
```suggestion
# Prefixed Parent Selector: Draft 1
```

We only update draft numbers when the *committed* version of the proposal changes; there's no need to do so for pull request review changes.

proposal/prefixed-parent-selector.md
```diff
@@ -0,0 +1,58 @@
...
+
+Everything works just as expected until a `&` is in the beginning of a nested selector.
+
+`"&" may only used at the beginning of a compound selector.` 
+
+Considering that there's been quite a few issues regarding this, and that they keep comming up, 
+and that people expect this to work, I think this would be a valuable addition to Sass.
+
+There's a work-around ([with some limitations](https://github.com/sass/sass/issues/1425#issuecomment-404462836)),
+but it's subpar. It might scare of beginners and/or cause unnecessary headache.
```
nex3: The role of the "background" section is to provide a single location that combines all the pre-existing discussion around the issue. As such, please explain the workaround in detail here, rather than requiring users to read through comments in an issue thread.

proposal/prefixed-parent-selector.md
```diff
@@ -0,0 +1,58 @@
...
+## Background
+
+When transitioning from Less or Stylus, or just learning Sass, nesting is common.
+
+Everything works just as expected until a `&` is in the beginning of a nested selector.
+
+`"&" may only used at the beginning of a compound selector.` 
+
+Considering that there's been quite a few issues regarding this, and that they keep comming up, 
+and that people expect this to work, I think this would be a valuable addition to Sass.
```
nex3: This doesn't do an effective job of making the case for why users want this behavior and why they expect this to work. I'd really like to see at least one concrete use-case written out here, as well as some discussion of users' expectation of symmetry between parent selectors with a suffix and those with a prefix.

proposal/prefixed-parent-selector.md
```diff
@@ -0,0 +1,58 @@
...
+
+There's a work-around ([with some limitations](https://github.com/sass/sass/issues/1425#issuecomment-404462836)),
+but it's subpar. It might scare of beginners and/or cause unnecessary headache.
+
+
+## Summary
+
+Prefixed parent selectors will work as expected, solving the frustration of current, and new, users of Sass.
```
nex3: This line doesn't really help to convey any new information to the reader. The summary should be informal and high-level, but it should still be focused on conveying to a user what the feature will do without anything extraneous.

proposal/prefixed-parent-selector.md
```diff
@@ -0,0 +1,58 @@
...
+
+@arkonan's [example](https://github.com/sass/sass/issues/1425#issuecomment-405921429) neatly summarises this: 
+```
+& selector { }  // works
+selector & { }  // works
+&selector { }   // works
```
nex3: Similarly, these lines about features that *aren't* prefixed parent selectors aren't particularly relevant to this feature.

proposal/prefixed-parent-selector.md
```diff
@@ -0,0 +1,58 @@
...
+selector& { }   // doesn't work
```
nex3: This is relevant, but just saying "doesn't work" doesn't convey what's being proposed. What CSS does this produce? Are there circumstances where it's disallowed? `&foo` is forbidden if the parent selector doesn't end in an identifier—are there similar cases for `foo&`?

proposal/prefixed-parent-selector.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+
+
+## Syntax
```
nex3: The syntax section should start with a [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form)-style specification of the syntax of the new feature, including how it combines with existing syntax. You can look at other proposals for examples.

proposal/prefixed-parent-selector.md
```diff
@@ -0,0 +1,58 @@
...
+}
+```
+
+to become
+
+```
+h1.title {
+  font-weight: 600;
+}
+```
```
nex3: The description of *how* the feature works should go in the Semantics section. It shouldn't be example-based—examples are for the summary—but a formal description of how the syntax is processed in general, in enough detail to implement consistently across multiple implementations.

nex3 (commented):
Thanks for continuing to iterate on this! Feel free to keep pushing new changes to this PR rather than opening new ones.

proposal/prefixed-parent-selector.md
oles: Attempted to address this in: https://github.com/sass/sass/pull/2723/commits/7631df1dfd0e4b5ca6b134834d9e1d9ded17c7d9

proposal/prefixed-parent-selector.md
oles: Attempted to address this in: https://github.com/sass/sass/pull/2723/commits/7631df1dfd0e4b5ca6b134834d9e1d9ded17c7d9

proposal/prefixed-parent-selector.md
oles: Attempted to address this in: https://github.com/sass/sass/pull/2723/commits/031a6d6ca260b20155f028ee2dd8f232f1e976b5

proposal/prefixed-parent-selector.md
oles: The BNF thing is quite something. Doesn't seem like something you can just figure out and write an example for in 30 minutes. I'll need help with this, or I'm afraid this is where my energy runs out.

oles: Though, I followed my intuition and gave it a silly go; https://github.com/sass/sass/commit/031a6d6ca260b20155f028ee2dd8f232f1e976b5#diff-734ce04a16b4d09ecbe73e775f49927bR121

nex3:
@oles I'm sorry you're having such a tough time with this :disappointed:. I'm going to close this out for now, but if anyone wants to help write the proposal I'm happy to reopen.

oles:
@nex3 no worries, I just jumped at it, thinking it should be simple enough.

I understand that even a simple thing requires a solid proposal. Hopefully my attempt will be of use either the next time I'm, or someone else, is bothered enough to do it thorough.

Thanks for all the reviews - they've been helpful and educational!

PR #2734, opened 2019-08-20
[Slash Separator] Update the proposal

nex3:


spec/functions/selector.md
```diff
@@ -0,0 +1,52 @@
...
+  * [Parse a Selector From a SassScript Object](#parse-a-selector-from-a-sassscript-object)
+
+## Procedures
+
+### Parse a Selector From a SassScript Object
+
+This algorithm takes a SassScript object `selector` and returns an abstract
+representation of a CSS selector.
+
+* Set `text` by an empty string.
```
jathak: Should this say `to be` instead of `by`?

jathak (approved):

spec/functions/selector.md
nex3: Yep!

Issue #2744, opened 2019-09-21
Support reconfigurable modules via "@forward ... with"


mirisuzanne:
*Edited by @nex3*

- [x] [Proposal](https://github.com/sass/sass/blob/master/accepted/forward-with.md)
- [x] [Specs](https://github.com/sass/sass-spec/pull/1499)
- [x] [Documentation](https://github.com/sass/sass-site/issues/383)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/846)
- [ ] [LibSass](https://github.com/sass/libsass/issues/3004)

The current plan is to add a `with` clause to `@forward`. This would work like the same clause in `@use`, except that unconfigured variables or variables set with the `!default` flag would still be configurable by the modules loading the module in which the `@forward` appears. For example:

```scss
@forward "unopinionated" with (
  $hue: 0 !default, // Can be overridden by upstream users.
  $saturation: 50% // Cannot be overridden by upstream users.
);
```

---

I've been moving my Sass packages over to the module system, but I keep hitting a problem I can't figure out. I usually have a central set of configuration variables that need to:

- be configurable by users
- be used by all the internal modules

But any time I make an internal file available to other files, I lose the ability to take user-configurations:

```
Error: This module was already loaded, so it can't be configured using "with".
```

I've tried adding duplicate `!default` settings in each file that needs, but then I get:

```
Error: Module sass/_config.scss and the new module both forward a variable named $setting.
```

As far as I can tell, everything that uses a particular default variable has to live in the same file?

Is this how it's meant to work? How would you recommend handling global package config shared among partials?

nex3:
Can you give a code example? Whatever's going on here may depend on the specifics of how you're defining and forwarding configuration.

mirisuzanne:
I'm trying to basically set it up the way you demonstrate in the proposal document, but there's a layer missing from that sample which I'm not sure how to handle:

```scss
// colors/_config.scss
$colors: 6 !default;

// colors/_generate.scss
@use 'config';
@mixin generate-colors($colors: config.$colors) { /* mixin-logic */ }

// colors/_index.scss
@forward 'config';
@forward 'generate';

// _user.scss
@use 'colors' with ( $colors: 4 );
@include colors.generate();
```

Both the `_config.scss` variables and `_generate.scss` mixins need to be forwarded to the user – but one relies on using the other.

nex3:
[...]

PR #2759, opened 2019-10-20
Reconfigurable Modules via "@forward ... with"

mirisuzanne:
This is a (draft) proposal for adding a `with` clause to the `@forward` rule, so that module configurations can be chained together, as per #2744:

```scss
// _origin.scss
$hue: 0 !default;
$saturation: 50% !default;

// _middleware.scss
@forward "origin" with (
  $hue: 330 !default, // Can be overridden by importing users.
  $saturation: 70% // Cannot be overridden by importing users.
);

// entrypoint.scss
@use "middleware" with (
  $hue: 120 // override both the origin & middleware !default values
);

// middleware.$hue == 120
// middleware.$saturation == 70%
```

### DIFF
--- a/proposal/forward-with.md
+++ b/proposal/forward-with.md
@@ -133,12 +133,44 @@ The new `WithClause` extends `@forward` to the follow grammar:
 The `@forward ... with` semantics builds on the existing proposal for
 [Executing Files][], and should be understood as modifying and expanding upon
 the existing execution process rather than being a comprehensive replacement.
-The `AsClause` logic remains unchanged, but is included here for context.
 
 [Executing Files]: ../accepted/module-system.md#executing-files
 
 Given a source file `file`, a configuration `config`, and an import context
 `import`:
+
+* Let `module` be an empty module with the same URL as `file`.
+
+* Let `uses` be an empty map from `@use` rules to [modules](#module).
+
+* When a `@use` rule `rule` is encountered:
+
+  * If `rule` has a namespace that's the same as another `@use` rule's namespace
+    in `file`, throw an error.
+
+  * Let `rule-config` be the empty configuration.
+
+  * If `rule` has a `WithClause`:
+
+    * Let `target` be the module at `rule`'s URL.
+
+    * For each `KeywordArgument` `argument` in this clause:
+
+      * If `target`'s public API does not include a variable with `argument`'s
+        name and a `!default` flag, throw an error.
+
+        > We now check for configuration errors inside each individual
+        > `withClause`, rather than waiting until the target module is execurted.
+
+      * Let `value` be the result of evaluating `argument`'s expression.
+
+      * Add a variable to `rule-config` with the same name as `argument`'s
+        identifier and with `value` as its value.
+
+  * Let `module` be the result of [loading][] the module with `rule`'s URL
+    and `rule-config`.
+
+  * Associate `rule` with `module` in `uses`.
 
 * When a `@forward` rule `rule` is encountered:
 
@@ -159,14 +191,15 @@ Given a source file `file`, a configuration `config`, and an import context
 
   * If `rule` has a `WithClause`:
 
+    * Let `target` be the module at `rule`'s URL.
+
     * For each `ForwardWithArgument` `argument` in this clause:
 
+      * If `target`'s public API does not include a variable with the same name
+        as `argument`'s identifier, and a `!default` flag, throw an error.
+
       * If a variable exists in `rule-config` with the same name as `argument`'s
-        identifier:
-
-        * If `argument` has a `!default` flag, do nothing.
-
-        * Otherwise, throw an error.
+        identifier, do nothing.
 
       * Otherwise:
 
@@ -174,3 +207,13 @@ Given a source file `file`, a configuration `config`, and an import context
 
         * Add a variable to `rule-config` with the same name as `argument`'s
           identifier, and with `value` as its value.
+
+  * Let `forwarded` be the result of [loading][] the module with `rule`'s URL
+    and `rule-config`.
+
+  * [Forward `forwarded`][forwarding] with `file` through `module`.
+
+> From this point on, the logic remains unchanged.
+
+[loading]: ../accepted/module-system.md#loading-modules
+[forwarding]: ../accepted/module-system.md#forwarding-modules
--- a/proposal/forward-with.md
+++ b/proposal/forward-with.md
@@ -131,32 +131,44 @@ The new `WithClause` extends `@forward` to the follow grammar:
 ## Semantics
 
 The `@forward ... with` semantics builds on the existing proposal for
-[Forwarding Modules][]. This proposal inserts some additional detail to the
-final step of the process:
+[Executing Files][], and should be understood as modifying and expanding upon
+the existing execution process rather than being a comprehensive replacement.
 
-[Forwarding Modules]: ../accepted/module-system.md#forwarding-modules
+[Executing Files]: ../accepted/module-system.md#executing-files
 
-> Previously: "Otherwise, add `member` to `module` with the name `name`."
+Given a source file `file`, a configuration `config`, and an import context
+`import`:
 
-* Otherwise, if `rule` has a `withClause` `with`:
+* When a `@forward` rule `rule` is encountered:
 
- * If `file` contains an `@use` rule with the same target URL and another
-   `withClause`, throw an error.
+  * For each variable `variable` in `config`:
 
-   > A module can only be configured from a single thread of `@forward`'s
-   > ending in an optional `@use`. Parallel configurations are not allowed,
-   > even if they come from the same source `file`.
+    * Let `rule-config` be an empty configuration.
 
- * If `with` includes a `KeywordArgument` `agument` with the name `name`
-   (including `$`):
+    * If `rule` has an `AsClause` with identifier `prefix`:
 
-   * If `member` does not have a `!default` flag, throw an error.
+      * If `variable`'s name begins with `prefix`, let `name` be the portion of
+        `variable`'s name after `prefix`.
 
-     > We can only configure membes that are maked as default.
+    * Otherwise, let `name` be `variable`'s name.
 
-   * Otherwise, let `value` be the value of `agument`, including the
-     `!default` flag if present.
+    * Add a variable to `rule-config` with the name `name` and with the
+      same value as `variable`.
 
-   * Add `member` to `module` with the name `name` and value `value`.
+  * Otherwise, let `rule-config` be `config`.
 
-* Otherwise, add `member` to `module` with the name `name`.
+  * If `rule` has a `WithClause`:
+
+    * For each `KeywordArgument` `argument` in this clause:
+
+      * Let `value` be the result of evaluating `argument`'s expression.
+
+      * If a variable exists in `rule-config` with the same name as `argument`'s
+        identifier:
+
+        * If `argument` does has a `!default` flag, do nothing.
+
+        * Otherwise, throw an error.
+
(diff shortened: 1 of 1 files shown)


mirisuzanne:
I wasn't confident in the technical details of Procedures & Semantics after a few attempts, but I tried to document a number of use cases and potential errors in the Summary. I'd be happy to keep working on it, but didn't want to hold things up any more – and I'd likely need input or guidance before I could get much farther.

proposal/forward-with.md
```diff
@@ -0,0 +1,170 @@
+# Reconfigurable Modules via "@forward ... with": Draft 1
```
nex3: I'd probably just call this either "Reconfigurable Modules" or "`@forward ... with`" for terseness.

proposal/forward-with.md
```diff
@@ -0,0 +1,170 @@
...
+
+*([Issues](https://github.com/sass/sass/issues/2744))*
+
+## Table of Contents
+
+- [Reconfigurable Modules via "@forward ... with": Draft 1](#reconfigurable-modules-via-%22forward--with%22-draft-1)
+  - [Table of Contents](#table-of-contents)
```
nex3: Nit: We usually don't include the title or the table of contents in the table of contents. See other proposals for examples.

proposal/forward-with.md
```diff
@@ -0,0 +1,170 @@
...
+);
+
+// entrypoint.scss
+@use "middleware" with (
+  $hue: 120 // override both the origin & middleware !default values
+);
+
+// middleware.$hue == 120
+// middleware.$saturation == 70%
+```
```
nex3: I think it's clearer to use separate code blocks for each file in a multi-file example.

proposal/forward-with.md
```diff
@@ -0,0 +1,170 @@
...
+
+// _middle2.scss
+@use "origin" with (
+  $hue: 120
+);
+
+// entrypoint.scss
+@forward "middle1";
+@use "middle2"; // ERROR
+```
```
nex3: I think you can get away without these examples. The summary should highlight the simple and common use-cases to give users an idea of how a feature works, but it doesn't need to be comprehensive; that's what the semantics section is for.

proposal/forward-with.md
```diff
@@ -0,0 +1,170 @@
...
+## Syntax
+
+The new `WithClause` extends `@forward` to the follow grammar:
+
+<x><pre>
+**ForwardRule**     ::= '@forward' QuotedString AsClause? (ShowClause | HideClause)?  WithClause?
+**AsClause**        ::= 'as' Identifier '*'
+**ShowClause**      ::= 'show' MemberName (',' MemberName)*
+**HideClause**      ::= 'hide' MemberName (',' MemberName)*
+**MemberName**      ::= '$'? Identifier
```
nex3: You don't need to repeat any production definitions that remain unchanged from the existing spec (that is, the accepted module system proposal).

nex3 (commented):
I'm happy to help out with the semantics section. What sort of guidance would you like? If you're intimidated by the blank page, try starting by reading the existing [Forwarding Modules](https://github.com/sass/sass/blob/master/accepted/module-system.md#forwarding-modules) section and describing what you'd change in that.

proposal/forward-with.md
mirisuzanne: heh, oops: I think something in my code editor did that automatically and I didn't catch it.

proposal/forward-with.md
```diff
@@ -105,66 +113,50 @@ $hue: 0 !default;
...
+   (including `$`):
+
+   * If `member` does not have a `!default` flag, throw an error.
+
+     > We can only configure membes that are maked as default.
+
+   * Otherwise, let `value` be the value of `agument`, including the
+     `!default` flag if present.
+
+   * Add `member` to `module` with the name `name` and value `value`.
```
mirisuzanne: @nex3 This is the sort of technical detail that I was getting hung up on. I'm not entirely clear if member has an available value at this point in the process, and if we can so simply change it, flags-and-all, as we pass things along. Doing it this way doesn't catch errors from external configurations, or demonstrate the configuration being applied when the forwarding module acts as an entrypoint…

(I was working from the "Executing Files" section in my first attempt, which is a bit more in-the-weeds…)

mirisuzanne (commented):
I'm not sure that I've captured all the edge cases that I mean to (which is why I added extra summary examples in the first draft). But I gave it a stab. :)

proposal/forward-with.md
```diff
@@ -0,0 +1,162 @@
...
+final step of the process:
+
+[Forwarding Modules]: ../accepted/module-system.md#forwarding-modules
+
+> Previously: "Otherwise, add `member` to `module` with the name `name`."
+
+* Otherwise, if `rule` has a `withClause` `with`:
+
+ * If `file` contains an `@use` rule with the same target URL and another
+   `withClause`, throw an error.
```
nex3: Is this necessary? Won't [Loading Modules](https://github.com/sass/sass/blob/master/accepted/module-system.md#loading-modules) throw an error anyway if a module is loaded with configuration multiple times?

proposal/forward-with.md
nex3: In this case, I think it doesn't work, because according to [the definition of "member"](https://github.com/sass/sass/blob/master/accepted/module-system.md#member) each member has a definition; so if you're adding a new value, you have to create a new member with that value. That's not what you want to do here, because you want the forwarded member to refer to the same "slot" as the original; for example, if you write:

```scss
// upstream.scss
$var: original !default;
```

```scss
// midstream.scss
@forward "upstream" with ($var: overridden);
```

You want `upstream`'s value of `$var` to be `overridden` as well as `midstream`'s value (in fact you don't want them to be separate values at all).

nex3 (commented):
I'm sorry, I think I steered you wrong when I suggested modifying Forwarding Modules, since by the time you're doing that you've already loaded the module in question. What you really want to do is make `@forward` pass in its own configuration when it loads the module, which involves modifying the `@forward` section of [Executing Files](https://github.com/sass/sass/blob/master/accepted/module-system.md#executing-files).

Your spec language looks really good though!

mirisuzanne:
@nex3 I feel a lot better about how this cleaned up now that I have it defined in file execution. Let me know if I missed something.

proposal/forward-with.md
```diff
@@ -0,0 +1,174 @@
...
+## Syntax
+
+The new `WithClause` extends `@forward` to the follow grammar:
+
+<x><pre>
+**ForwardRule**     ::= '@forward' QuotedString AsClause? (ShowClause | HideClause)?  WithClause?
+**WithClause**      ::= 'with' '('
+&#32;                     KeywordArgument (',' KeywordArgument)\* ','?
+&#32;                   ')'
+**KeywordArgument** ::= '$' Identifier ':' Expression ('!default')
```
nex3:
```suggestion
**ForwardWithArgument** ::= '$' Identifier ':' Expression '!default'?
```

Making this name more explicit avoids confusion with other `KeywordArgument` productions that don't allow `!default`.

proposal/forward-with.md
```diff
@@ -0,0 +1,174 @@
...
+the existing execution process rather than being a comprehensive replacement.
+
+[Executing Files]: ../accepted/module-system.md#executing-files
+
+Given a source file `file`, a configuration `config`, and an import context
+`import`:
+
+* When a `@forward` rule `rule` is encountered:
+
+  * For each variable `variable` in `config`:
```
nex3: I think you want this to be nested within "If `rule` has an `AsClause`"

proposal/forward-with.md
```diff
@@ -0,0 +1,174 @@
...
+    * Add a variable to `rule-config` with the name `name` and with the
+      same value as `variable`.
+
+  * Otherwise, let `rule-config` be `config`.
+
+  * If `rule` has a `WithClause`:
+
+    * For each `KeywordArgument` `argument` in this clause:
+
+      * Let `value` be the result of evaluating `argument`'s expression.
```
nex3: We probably don't want to do this if the argument is an overridden default.

proposal/forward-with.md
```diff
@@ -0,0 +1,174 @@
...
+
+      * If a variable exists in `rule-config` with the same name as `argument`'s
+        identifier:
+
+        * If `argument` does has a `!default` flag, do nothing.
+
+        * Otherwise, throw an error.
```
nex3: Currently, all the error-checking for passing an invalid config variable is handled at the beginning of [Executing Files](https://github.com/sass/sass/blob/master/accepted/module-system.md#executing-files). We probably want to handle this in the same place; if we just error here, we'll end up forbidding this:

```scss
// downstream.scss
@use "upstream" with ($value: downstream);
```

```scss
// _midstream.scss
@forward "upstream" with ($value: midstream 1);

// Because midstream declares its own `$value` variable, it's valid for
// downstream to override it even though it's not overriding upstream's.
$value: midstream 2 !default;
```

This makes me realize that there's probably a better way to do that error checking, though. Instead of having it occur at the beginning of the execution of the file and basically re-writing the logic for figuring out which names are valid, we should put it in the `@use` clause execution block and have it check whether `config` contains any variables that aren't in `module`'s public API with a `!default` flag. Then you can use the same text to verify that `@forward ... with` doesn't provide extra config too, which I think currently isn't being done.

nex3 (commented):
This is looking really good!

proposal/forward-with.md
mirisuzanne: I think I resolved the other issues. Have to run for now, but I'll dig into this one over the weekend.

mirisuzanne:
@nex3 I think this is roughly what you were suggesting. It relies on a few assumptions I wasn't sure of:

1. The public API of a not-yet-loaded module is available to check config arguments against.
2. The presence or absence of `!default` in a `@forward … with` argument is available as part of a module's public API.

I think fixing the second issue would require an update to the logic of "Forwarding Modules" – so that the member is updated before being added to the module API? I've started down that path locally - but not entirely sure I'm on the right track.

nex3 (changes requested):
This looks really good! One suggestion below and then I think we can get this submitted.

> 1\. The public API of a not-yet-loaded module is available to check config arguments against.

I think you *can* assume this, but I don't think you *need to*; you can just check against the actual `module` once it's loaded. There's no harm in executing the upstream module before checking.

> 2\. The presence or absence of `!default` in a `@forward … with` argument is available as part of a module's public API.

This you can definitely assume.

mirisuzanne:
I removed a comment that may have added confusion along those lines. I'm not sure though if I  understood the suggested change implied above. Does the logic need to change in order to allow executing the upstream module? If so, that's the aspect I don't quite have my head around yet.

nex3:
Would you mind if I just pushed a change to this branch to show what I'm thinking of?

mirisuzanne:
Sure that's fine.

nex3:
What do you think?

proposal/forward-with.md
```diff
@@ -166,6 +162,10 @@ Given a source file `file`, a configuration `config`, and an import context
   * Let `module` be the result of [loading][] the module with `rule`'s URL
     and `rule-config`.
 
+  * If `rule` has a `WithClause` that contains any variables that aren't part of
+    `module`'s public API or that weren't declared with a `!default` flag in
+    `module`, throw an error.
```
mirisuzanne, line 168: Oh right, of course - we're not waiting until execution to set the variables, that's already handled properly. We just wait until execution to throw any errors. Looks great! 

I have one more wording fix I want to cleanup

mirisuzanne:
@nex3 looks good to me!

nex3 (approved):
Thanks Miriam!

Issue #2772, opened 2019-11-06
Variables behind a @forward aren't configurable through an @import


jathak:
- [x] [Proposal](https://github.com/sass/sass/pull/2773)
- [x] [Specs](https://github.com/sass/sass-spec/pull/1497)
- [x] [Documentation](https://github.com/sass/sass-site/issues/410)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/882)
- [ ] [LibSass](https://github.com/sass/libsass/issues/3040)

---

If a library like the following:
```scss
// _library.scss (original)
$lib-color: red !default;
a { color: $lib-color; }
```
removes the `lib-` prefix when migrating to the module system, it will add an import-only stylesheet like so to preserve the manually prefixed names for downstream users still using `@import`.
```scss
// _library.scss (migrated)
$color: red !default;
a { color: $color; }

// _library.import.scss
@forward "library" as lib-*;
```
But this currently breaks downstream stylesheets like:
```scss
// downstream.scss
$lib-color: green;
@import "library";
```
which compiled to `a { color: green; }` prior to migration and `a {color: red; }` after.

It seems like this is due to `@import`-style configuration not propagating through a `@forward`.

nex3:
The proposal has landed! Since it's narrowly scoped and blocking work on migrating existing libraries, I'm going to limit the call for comments to one week unless any major issues get raised.

nex3:
Since no one has raised any issues, I'm officially marking this proposal as accepted.

nex3:
Closing this out because [LibSass is now deprecated](https://sass-lang.com/blog/libsass-is-deprecated) and we aren't expecting to add any additional features to it.

PR #2773, opened 2019-11-08
Proposal for configuring modules through imports

jathak:
This is a very rough draft of the proposal to implicitly create configurations when importing libraries that have already migrated to the module system (see #2772).

Since the semantics are just modifications of the module system spec, I just specified the sections that should be modified and stated their new contents. I'm not sure if it would be better to just directly modify that spec, but this seemed like a significant enough change that it shouldn't go through the fast-track process.


nex3:
This looks good so far. My main concern with this approach is the potential overhead of creating a new configuration for every `@import`—do you have any thoughts on how we could mitigate that?

jathak:
What if we only created the new configuration if the file being imported contains a `@forward` rule? Would that be sufficient?

nex3:
Yeah, that should work. Good thinking!

That's not something you need to include as spec text, since it's
effectively an implementation detail, but if you want you can add a
non-normative implementation advisory.

jathak:
Okay, I added that note and also clarified that a new config created by a prefixed `@forward` is implicit if the original config was.

I also expanded on the design decisions section. Is there anything else you think I should bring up there?

proposal/configuring-modules-through-imports.md
```diff
@@ -0,0 +1,225 @@
...
+issues if the same library is imported more than once.
+
+While the solution we settled on does not perfectly cover all use cases that
+worked before the library migrated to the module system, we think it strikes a
+good balance of supporting most existing use cases without hurting performance
+or making the language specification and implementation overly complicated.
+
+One potential use case that will still not work is if a downstream user of the
+library attempts to change its configuration between two imports of the same
+library, that change will be ignored. However, this is an edge case that is (a)
```
nex3: This first sentence is kind of awkward. Consider breaking it up into two sentences instead.

proposal/configuring-modules-through-imports.md
```diff
@@ -0,0 +1,225 @@
...
+probably not intended by the user, (b) relatively easy to fix by moving all
+declared configuration variables before all library imports, and (c) very
+difficult to support for a library using the module system without compromising
+the module system's [goals][].
```
nex3: I suggest being more specific about how it compromises which goals.

nex3 (commented):
A couple suggestions, but otherwise looks good!

nex3 (approved):

jathak:
Can this be merged? (I don't have write access to this repo)

Issue #851, opened 2013-07-15
Add more math functions


tylermenezes:
*(edited by @nex3)*

* [x] [Proposal](https://github.com/sass/sass/pull/2779)
* [x] [Specs](https://github.com/sass/sass-spec/pull/1508)
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/897)
* [ ] [LibSass](https://github.com/sass/libsass/issues/3044)
* [x] [Documentation](https://github.com/sass/sass-site/issues/418)

---

Sass, by default, has a very limited set of math functions. Some functions which are missing, but would be of use:
- sqrt
- pow
- mod
- log

This functionality can be added with third-party gems, but the fact that they're missing from the core just makes everything more difficult out-of-the-box.

There is very little benefit to keeping mathematical functions out of the core, because they have a standard, expected definition, so they're not really cluttering the namespace.

KittyGiraudel:
References:
- https://github.com/Team-Sass/Sassy-math
- http://compass-style.org/reference/compass/helpers/math/

chriseppstein:
:thumbsup:

Note: modulo operations are available in Sass via the `%` operator.

chriseppstein:
@nex3 do you object?

ghost:
If we're going to add a log function, there should probably be log10 and log2 functions for convenience. Ordinary log should be base-e.

nex3:
I want to wait on this until we have a better module story, since once we do it would be nice to put nice functions like this in a built-in module that can be imported with a prefix.

booleanbetrayal:
:+1: 

would love to to see expanded math functionality that mimics the compass implementation

torkiljohnsen:
:+1: Right now it's a pain to do stuff like `pow($x, $y)` if `$y` is a floating point number.

KittyGiraudel:
Any chance of seeing this open again?

Nevraeka:
@HugoGiraudel :+1: We really need the compass functionality built in at the sass level. LibSass won't add this functionality until its standard in sass. We need better consistency across the projects in that respect. @nex3 what are the updates on the module system?

nex3:
All feature work is currently on hold until Libsass reaches compatibility.

ixisio:
+1

c9s:
Hi, this is planed in C6 http://github.com/c9s/c6

ixisio:
[...]

PR #2777, opened 2019-11-21
Convert the module system proposal to stable specifications

nex3:

PR #2779, opened 2019-11-25
Propose more sass:math module functions

Awjin:
First draft for specifying Sassified versions of all the mathematical functions in the [Values and Units 4 Draft](https://drafts.csswg.org/css-values-4/#math), logarithms, and $e and $pi variables.

See #851 for context.

### DIFF
--- a/proposal/more-math-functions.md
+++ b/proposal/more-math-functions.md
@@ -6,8 +6,8 @@ This proposal adds the following members to the built-in `sass:math` module.
 
 ## Table of Contents
 * [Variables](#variables)
-  * [`$e`](#$e)
-  * [`$pi`](#$pi)
+  * [`$e`](#e)
+  * [`$pi`](#pi)
 * [Functions](#functions)
   * [`clamp()`](#clamp)
   * [`hypot()`](#hypot)
--- a/proposal/more-math-functions.md
+++ b/proposal/more-math-functions.md
@@ -1,0 +1,445 @@
+# Math Module
+
+This built-in module is available from the URL `sass:math`.
+
+## Table of Contents
+* [Variables](#variables)
+  * [`e`](#e)
+  * [`pi`](#pi)
+* [Functions](#functions)
+  * [Approximation](#approximation)
+    * [`ceil()`](#ceil)
+    * [`floor()`](#floor)
+    * [`round()`](#round)
+  * [Comparison](#comparison)
+    * [`clamp()`](#clamp)
+    * [`max()`](#max)
+    * [`min()`](#min)
+    * [`percentage()`](#percentage)
+  * [Division](#division)
+    * [`div()`](#div)
+    * [`mod()`](#mod)
+  * [Exponentiation](#exponentiation)
+    * [`log()`](#log)
+    * [`pow()`](#pow)
+    * [`sqrt()`](#sqrt)
+  * [Trigonometry](#trigonometry)
+    * [Procedures](#procedures)
+      * [Verify Inputs](#verify-inputs)
+    * [`cos()`](#cos)
+    * [`sin()`](#sin)
+    * [`tan()`](#tan)
+    * [`acos()`](#acos)
+    * [`asin()`](#asin)
+    * [`atan()`](#atan)
+    * [`atan2()`](#atan2)
+      * [Edge cases](#edge-cases)
+  * [Units](#units)
+    * [`compatible()`](#compatible)
+    * [`is-unitless()`](#is-unitless)
+    * [`unit()`](#unit)
+  * [Util](#util)
+    * [`abs()`](#abs)
+    * [`hypot()`](#hypot)
+    * [`random()`](#random)
+
+## Variables
+
+### `e`
+
+```
+$e
+```
+
+### `pi`
+
+```
+$pi
+```
+
+## Functions
+
+### Approximation
+
+#### `ceil()`
+
+```
+ceil($number)
+```
+This function is also available as a global function named `ceil()`.
+
+#### `floor()`
+
+```
+floor($number)
+```
+
+This function is also available as a global function named `floor()`.
+
+#### `round()`
+
+```
+round($number)
+```
+
+This function is also available as a global function named `round()`.
+
+### Comparison
+
+#### `clamp()`
+
+```
+clamp($min, $number, $max)
+```
+
+* Return `$min` if `$min` >= `$max`
+* Return `$min` if `$number` <= `$min`.
+* Return `$max` if `$number` => `$max`.
+* Return `$number`.
+
+#### `max()`
+
+```
+max($numbers...)
+```
+
+This function is also available as a global function named `max()`.
+
+#### `min()`
+
+```
+min($numbers...)
+```
+
+This function is also available as a global function named `min()`.
+
+#### `percentage()`
+
+```
+percentage($number)
+```
+
+This function is also available as a global function named `percentage()`.
+
+### Division
+
+#### `div()`
+
+```
+div($dividend, $divisor)
+```
+
+This function is also available as a global function named `divide()`.
+
+* Throw error if the units of `$dividend` and `$divisor` are not
+  [compatible](#compatible).
+* Return `$dividend` / `$divisor`.
+
+#### `mod()`
+
+```
+mod($dividend, $modulus)
+```
+
+* Throw error if the units of `$dividend` and `$modulus` are not
+  [compatible](#compatible).
+* Return `$dividend` % `$modulus`.
+
+### Exponentiation
+
+#### `log()`
+
+```
+log($number)
+```
+* Throw error if `$number` has units or `$number` < 0.
+* Return -Infinity if `$number` is 0.
+* Return Infinity if `$number` is infinite.
+* Return the natural log of `$number`.
+
+```
+log($number, $base)
+```
+
+* Throw error if `$base` < 0 or `$base` is 0 or `$base` is 1.
+* Return [log](#log)(`$number`) / [log](#log)(`$base`).
+
+#### `pow()`
+
+```
+pow($base, $exponent)
+```
+
+* Throw error if `$base` or `$exponent` has units.
+
+* If `$exponent` is 0:
+  * Return 1.
+
+* Else if `$exponent` is infinite:
+  * Return NaN if [abs](#abs)(`$base`) is 1.
+  * Return Infinity if [abs](#abs)(`$base`) > 1 and `$exponent` > 0, or if
+    [abs](#abs)(`$base`) < 1 and `$exponent` < 0.
+  * Return 0.
+
+* Else if `$exponent` is finite:
+  * Return NaN if `$base` < 0 and `$exponent` is not an integer.
+
+  * Return Infinity if `$base` is 0 and `$exponent` < 0, or if `$base` is
+    Infinity and `$exponent` > 0.
+  * If `$base` is -0 and `$exponent` < 0, or if `$base` is -Infinity and
+    `$exponent` > 0:
+    * Return -Infinity if `$exponent` is an odd integer.
+    * Return Infinity.
+
+  * Return 0 if `$base` is 0 and `$exponent` > 0, or if `$base` is Infinity and
+    `$exponent` < 0.
+  * If `$base` is -0 and `$exponent` > 0, or if `$base` is -Infinity and
+    `$exponent` < 0:
+    * Return -0 if `$exponent` is an odd integer.
+    * Return 0.
+
+  * Return `$base` raised to the power of `$exponent`.
+
+#### `sqrt()`
+
+```
+sqrt($number)
+```
+
+* Throw error if `$number` has units.
+* Return NaN if `$number` < 0.
+* Return -0 if `$number` is -0.
+* Return Infinity if `$number` is infinite.
+* Return the square root of `$number`.
+
+### Trigonometry
+
+#### Procedures
+
+##### Verify Inputs
+
+`cos($number), sin($number), and tan($number)` throw error if `$number` has
+units but is not a rotation value. If `$number` is unitless, it is assumed to be
+in radians.
+
+`acos($number), asin($number), and atan($number)` throw error if `$number` has
+units.
+
+Both inputs to `atan2($y, $x)` must have the same unit or be unitless. If not,
+throw error.
+
+#### `cos()`
+
+```
+cos($number)
+```
+
+* [Verify inputs](#verify-inputs).
+* Return NaN if `$number` is infinite.
+* Return cos(`$number`).
+
+#### `sin()`
+
+```
+sin($number)
+```
+
+* [Verify inputs](#verify-inputs).
+* Return NaN if `$number` is infinite.
+* Return -0 if `$number` is -0.
+* Return sin(`$number`).
+
+#### `tan()`
+
+```
+tan($number)
+```
+
+* [Verify inputs](#verify-inputs).
+* Return NaN if `$number` is infinite.
+* Return -0 if `$number` is -0.
+* If `$number` is an asymptote value:
+  * Return Infinity if `$number` is 90deg +/- 360*n* deg, where *n* is any integer.
+  * Return -Infinity if `$number` is -90deg +/- 360*n* deg, where *n* is any
+    integer.
+* Return tan(`$number`).
+
+#### `acos()`
+
+```
+acos($number)
+```
+
+* [Verify inputs](#verify-inputs).
+* Return NaN if [abs](#abs)(`$number`) > 1.
+* Return 0 if `$number` is 1.
+* Return acos(`$number`).
+
+#### `asin()`
+
+```
+asin($number)
+```
+
+* [Verify inputs](#verify-inputs).
+* Return NaN if [abs](#abs)(`$number`) > 1.
+* Return -0 if `$number` is -0.
+* Return asin(`$number`).
+
+#### `atan()`
+
+```
+atan($number)
+```
+
+* [Verify inputs](#verify-inputs).
+* Return -0 if `$number` is -0.
+* Return -90deg if `$number` is -Infinity.
+* Return 90deg if `$number` is Infinity.
+* Return atan(`$number`).
+
+#### `atan2()`
+
+`atan2($y, $x)` is distinct from `atan($y / $x)` because it preserves the
+quadrant of the point in question. For example, `atan2(1, -1)` corresponds to
+the point (-1, 1), which returns 135deg. In contrast, `atan(1 / -1)` and
+`atan(-1 / 1)` resolve first to `atan(-1)`, so both return -45deg.
+
+```
+atan2($y, $x)
+```
+
+* [Verify inputs](#verify-inputs).
+* If the inputs match one of the following edge cases, return the value for that
+  edge case. Otherwise, return normal atan2($y, $x).
+
+##### Edge cases
+
+<table>
+  <thead>
+    <tr>
+      <td colspan="2"></td>
+      <th colspan="6" style="text-align: center">X</th>
+    </tr>
+    <tr>
+      <td colspan="2"></td>
+      <th>−Infinity</th>
+      <th>-finite</th>
+      <th>-0</th>
+      <th>0</th>
+      <th>finite</th>
+      <th>Infinity</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <th rowspan="6">Y</th>
+      <th>−Infinity</th>
+      <td>-135deg</td>
+      <td>-90deg</td>
+      <td>-90deg</td>
+      <td>-90deg</td>
+      <td>-90deg</td>
+      <td>-45deg</td>
+    </tr>
+    <tr>
+      <th>-finite</th>
+      <td>-180deg</td>
+      <td>(normal)</td>
+      <td>-90deg</td>
+      <td>-90deg</td>
+      <td>(normal)</td>
+      <td>-0deg</td>
+    </tr>
+    <tr>
+      <th>-0</th>
+      <td>-180deg</td>
+      <td>-180deg</td>
+      <td>-180deg</td>
+      <td>-0deg</td>
+      <td>-0deg</td>
+      <td>-0deg</td>
+    </tr>
+    <tr>
+      <th>0</th>
+      <td>180deg</td>
+      <td>180deg</td>
+      <td>180deg</td>
+      <td>0deg</td>
+      <td>0deg</td>
+      <td>0deg</td>
+    </tr>
+    <tr>
+      <th>finite</th>
+      <td>180deg</td>
+      <td>(normal)</td>
+      <td>90deg</td>
+      <td>90deg</td>
+      <td>(normal)</td>
+      <td>0deg</td>
+    </tr>
+    <tr>
+      <th>Infinity</th>
+      <td>135deg</td>
+      <td>90deg</td>
+      <td>90deg</td>
+      <td>90deg</td>
+      <td>90deg</td>
+      <td>45deg</td>
+    </tr>
+  </tbody>
+</table>
+
+### Units
+
+#### `compatible()`
+
+```
+compatible($number1, $number2)
+```
+
+This function is also available as a global function named `comparable()`.
+
+#### `is-unitless()`
+
+```
+is-unitless($number)
+```
+
+This function is also available as a global function named `unitless()`.
+
+#### `unit()`
+
+```
+unit($number)
+```
+
+This function is also available as a global function named `unit()`.
+
+### Util
+
+#### `abs()`
+
+```
+abs($number)
+```
+
+This function is also available as a global function named `abs()`.
+
+#### `hypot()`
+
+```
+hypot($numbers...)
+```
+
+* Throw error if any argument is not a length value.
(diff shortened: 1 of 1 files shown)


proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+    * [`abs()`](#abs)
+    * [`hypot()`](#hypot)
+    * [`random()`](#random)
+
+## Variables
+
+### `e`
+
+```
+$e
```
nex3: I don't think you need to include signatures for these, since they're just variable names. I'd probably include the `$` sign in the header title though.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+    * [`is-unitless()`](#is-unitless)
+    * [`unit()`](#unit)
+  * [Util](#util)
+    * [`abs()`](#abs)
+    * [`hypot()`](#hypot)
+    * [`random()`](#random)
+
+## Variables
+
+### `e`
```
nex3: Document the values these variables should have.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+```
+$pi
+```
+
+## Functions
+
+### Approximation
+
+#### `ceil()`
```
nex3: Proposals are usually written as modifications to the existing spec, so you don't need to include functions that already exist—you can just write something like "This proposal adds the following members to the `sass:math` module" at the top.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
+# Math Module
```
nex3: Can you format the intro more like [other proposals](https://github.com/sass/sass/blob/master/proposal/import-configuration.md), with a short proposal name, draft number, and link to the GitHub issue?

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+### Comparison
+
+#### `clamp()`
+
+```
+clamp($min, $number, $max)
+```
+
+* Return `$min` if `$min` >= `$max`
```
nex3: Style nit:

```suggestion
* Return `$min` if `$min >= $max`
```

All programming or mathematical syntax should be written in code font.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+```
+div($dividend, $divisor)
+```
+
+This function is also available as a global function named `divide()`.
+
+* Throw error if the units of `$dividend` and `$divisor` are not
+  [compatible](#compatible).
+* Return `$dividend` / `$divisor`.
```
nex3: I think you can leave this out, since it's already covered by [the slash-separator proposal](https://github.com/sass/sass/blob/master/accepted/slash-separator.md#dividemathdiv-function).

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+#### `mod()`
+
+```
+mod($dividend, $modulus)
+```
+
+* Throw error if the units of `$dividend` and `$modulus` are not
+  [compatible](#compatible).
+* Return `$dividend` % `$modulus`.
```
nex3: Since `%` isn't (currently) a Sass operator, I think this operation needs to be defined in more depth, especially with respect to what units the returned number has.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+```
+
+This function is also available as a global function named `round()`.
+
+### Comparison
+
+#### `clamp()`
+
+```
+clamp($min, $number, $max)
```
nex3, line 45: How does this handle the numbers having incompatible units?

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+  [compatible](#compatible).
+* Return `$dividend` / `$divisor`.
+
+#### `mod()`
+
+```
+mod($dividend, $modulus)
+```
+
+* Throw error if the units of `$dividend` and `$modulus` are not
```
nex3: Style nit:

```suggestion
* Throw an error if the units of `$dividend` and `$modulus` are not
```

I find full sentences help specs read more clearly, especially for people unfamiliar with their formality.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+#### `log()`
+
+```
+log($number)
+```
+* Throw error if `$number` has units or `$number` < 0.
+* Return -Infinity if `$number` is 0.
+* Return Infinity if `$number` is infinite.
+* Return the natural log of `$number`.
```
nex3: You don't need to precisely define what the natural logarithm means, but it's probably worth linking to the Wikipedia article.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+```
+log($number, $base)
```
nex3: It's usually a good idea to avoid defining overloads unless it's really necessary, since they aren't representable in pure Sass and thus can be confusing to read. For example, you could define this function as `log($exponent, $base: null)` and have it return the natural logarithm if `$base` is null. That also has the benefit that a user can explicitly pass a null base if they want to trigger the default behavior.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+```
+
+* Throw error if `$base` < 0 or `$base` is 0 or `$base` is 1.
+* Return [log](#log)(`$number`) / [log](#log)(`$base`).
```
nex3: I think it would be clearer to just write "Return the natural log of `$number` divided by the natural log of `$base`" rather than referring to other Sass functions. Similarly below.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+```
+pow($base, $exponent)
+```
+
+* Throw error if `$base` or `$exponent` has units.
+
+* If `$exponent` is 0:
+  * Return 1.
+
+* Else if `$exponent` is infinite:
```
nex3:
```suggestion
* Otherwise, if `$exponent` is infinite:
```

This just matches the existing spec language.

nex3: I'd explicitly compare `$exponent` to `Infinity` and `-Infinity` here, just to avoid the term "infinite". A reader will *probably* know what it's referring to, but they could wrongfully assume that it's a more complex term of art.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+* Throw error if `$number` has units.
+* Return NaN if `$number` < 0.
+* Return -0 if `$number` is -0.
+* Return Infinity if `$number` is infinite.
+* Return the square root of `$number`.
+
+### Trigonometry
+
+#### Procedures
```
nex3: Procedures should go at the top of the proposal, as a sibling section to Functions.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+##### Verify Inputs
```
nex3: This should follow the same format as [other spec procedures](https://github.com/sass/sass/pull/2777/files?short_path=ba3ae5f#diff-ba3ae5f01809e2c7eea84ff84e951e24#user-content-determining-a-use-rules-namespace), with a prose line indicating which inputs are taken and a bulleted list of steps to take.

If the procedure differs from function to function, though, it would probably be clearer to just include each function's behavior inline in the function definition. A bit of repetition is usually fine in prose.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+`cos($number), sin($number), and tan($number)` throw error if `$number` has
+units but is not a rotation value. If `$number` is unitless, it is assumed to be
```
nex3: Define "rotation value" (linking to a CSS spec would be sufficient).

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+#### `cos()`
+
+```
+cos($number)
+```
+
+* [Verify inputs](#verify-inputs).
+* Return NaN if `$number` is infinite.
+* Return cos(`$number`).
```
nex3: These trigonometric functions should be more explicitly defined—provide links to definitions of what the functions do, as well as explanations of how to handle different units and what unit should be returned (probably unitless?).

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+#### `tan()`
+
+```
+tan($number)
+```
+
+* [Verify inputs](#verify-inputs).
+* Return NaN if `$number` is infinite.
+* Return -0 if `$number` is -0.
+* If `$number` is an asymptote value:
```
nex3: Define "asymptote value"

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+* Return -90deg if `$number` is -Infinity.
+* Return 90deg if `$number` is Infinity.
+* Return atan(`$number`).
+
+#### `atan2()`
+
+`atan2($y, $x)` is distinct from `atan($y / $x)` because it preserves the
+quadrant of the point in question. For example, `atan2(1, -1)` corresponds to
+the point (-1, 1), which returns 135deg. In contrast, `atan(1 / -1)` and
+`atan(-1 / 1)` resolve first to `atan(-1)`, so both return -45deg.
```
nex3: As a convention, we write non-normative text in block quotes:

```markdown
> `atan2($y, $x)` is distinct from `atan($y / $x)` because it preserves the
> quadrant of the point in question. For example, `atan2(1, -1)` corresponds to
> the point (-1, 1), which returns 135deg. In contrast, `atan(1 / -1)` and
> `atan(-1 / 1)` resolve first to `atan(-1)`, so both return -45deg.
```

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+```
+atan2($y, $x)
+```
+
+* [Verify inputs](#verify-inputs).
+* If the inputs match one of the following edge cases, return the value for that
+  edge case. Otherwise, return normal atan2($y, $x).
```
nex3: Is this supposed to be "normal atan"?

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+##### Edge cases
+
+<table>
```
nex3, line 249: Can you use a [markdown table](https://github.github.com/gfm/#tables-extension-) for this?

proposal/more-math-functions.md
```diff
@@ -0,0 +1,445 @@
...
+
+This function is also available as a global function named `abs()`.
+
+#### `hypot()`
+
+```
+hypot($numbers...)
+```
+
+* Throw error if any argument is not a length value.
```
nex3: Define "length value". Does this allow unitless numbers?

nex3 (changes requested):
This is a really good start!

proposal/more-math-functions.md
Awjin: Done.

proposal/more-math-functions.md
Awjin: Done.

proposal/more-math-functions.md
Awjin: Done.

proposal/more-math-functions.md
Awjin: Done.

proposal/more-math-functions.md
Awjin: Agreed, fixed throughout the document

proposal/more-math-functions.md
Awjin: I realized the word "asymptote" is kind of a red herring, I just removed this line to let the logic in the next two lines speak for themselves.

proposal/more-math-functions.md
Awjin: Removed.

proposal/more-math-functions.md
Awjin: Discussed offline: since `%` exists as a binary operation, and because it doesn't seem likely that CSS will introduce new semantics for `%`, we can just remove this `modulo` proposal.

proposal/more-math-functions.md
Awjin: This should throw an error

proposal/more-math-functions.md
Awjin: Got it, fixed.

proposal/more-math-functions.md
Awjin: Added link here and wherever needed

proposal/more-math-functions.md
Awjin: Ah, good point. Fixed.

proposal/more-math-functions.md
Awjin: Done.

proposal/more-math-functions.md
Awjin: Done.

proposal/more-math-functions.md
Awjin: Done.

proposal/more-math-functions.md
Awjin: Like you said, this is much clearer when each function has its own input verification instead of trying to DRY it up too much. I removed this section altogether.

proposal/more-math-functions.md
Awjin: Done. Linked to the precise CSS definition `angle`

proposal/more-math-functions.md
Awjin: Done.

proposal/more-math-functions.md
Awjin: Good to know, thanks

proposal/more-math-functions.md
Awjin: I meant "normal" as any non-edge-case input, but realize that's pretty unclear. Changed this to specify it as a "2-argument arctangent", and linked to the wiki.

proposal/more-math-functions.md
Awjin: I would've liked to, but markdown tables don't allow cells to span multiple rows/cols.

proposal/more-math-functions.md
Awjin: Done. Linked to the precise CSS definition `length`

proposal/more-math-functions.md
Awjin: Fixed here and throughout

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+
+[compatible]: ../spec/built_in_modules/math.md#compatible
+
+### `hypot()`
+
+```
+hypot($numbers...)
+```
+
+* Throw an error if any argument is not a [length][].
```
nex3: If we don't allow unitless numbers here, I think that makes it the *only* Sass function that requires units. That's probably more confusing of an inconsistency than we'd have if we were more lenient than plain CSS and allowed unitless numbers here.

We could even go so far as to allow any compatible units here, since the output unit is always well-defined. I'm tempted to do so, because (unlike CSS) Sass allows custom units for which vector-length calculations could make sense.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+* Return `Infinity` if any argument is `Infinity`.
+* Return the square root of the sum of the squares of each argument.
```
nex3, line 68: All of these functions should define the units the return value uses, even if it's just "as a unitless number".

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+```
+log($number, $base: null)
+```
+
+* Throw an error if `$number` has units or `$number < 0`.
+* If `$base == null`:
+  * Return `-Infinity` if `$number == 0`.
+  * Return `Infinity` if `$number == Infinity`.
+  * Return the [natural log][] of `$number`.
+* Otherwise:
```
nex3: Style nit:

```md
* Otherwise, if `$base < 0` or `$base == 0` or `$base == 1`, throw an error.
* Otherwise, return the natural log of `$number` divided by the natural log of `$base`.
```

Writing "Otherwise" for each branch of the conditional helps draw the reader's attention to the fact that there are three possible cases here.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+### Exponentiation
+
+#### `log()`
+
+```
+log($number, $base: null)
+```
+
+* Throw an error if `$number` has units or `$number < 0`.
+* If `$base == null`:
```
nex3: Style nit for consistency with existing specs:

```suggestion
* If `$base` is null:
```

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+pow($base, $exponent)
+```
+
+* Throw an error if `$base` or `$exponent` has units.
+
+* If `$exponent == 0`:
+  * Return `1`.
+
+* Otherwise, if `$exponent == Infinity`:
+  * Return `NaN` if the `absolute value of $base == 1`.
```
nex3: "Absolute value of" isn't really syntax, so it reads strange to have it code-highlighted... I'd probably either write this in prose:

```suggestion
  * Return `NaN` if the absolute value of `$base` equals `1`.
```

or avoid mentioning absolute value altogether:

```suggestion
  * Return `NaN` if `$base == 1` or `$base == -1`.
```

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+### Trigonometry
+
+#### `cos()`
+
+```
+cos($number)
+```
+
+* Throw an error if `$number` has units but is not an [angle][]. If `$number` is
+  unitless, it is assumed to be in radians.
```
nex3:
```suggestion
  unitless, it is treated as though its unit were `rad`.
```

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+* Return `NaN` if `$number == Infinity`.
+* Return the [cosine][] of `$number`, which is a unitless number between `-1`
+  and `1`, exclusive.
```
nex3: Why exclusive? Isn't `cos(pi)` -1?

I don't think you actually need to specify the range here, since it's contained in the definition of the trig functions. If you want to for clarity, put it in a block quote so it's clear that it's non-normative.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+
+```
+tan($number)
+```
+
+* Throw an error if `$number` has units but is not an angle. If `$number` is
+  unitless, it is assumed to be in radians.
+* Return `NaN` if `$number == Infinity`.
+* Return `-0` if `$number == -0`.
+* Return `Infinity` if `$number == 90 +/- (360 * n)`, where `n` is any
```
nex3:
```suggestion
* Return `Infinity` if `$number % 360deg == 90deg`.
```

It's important to specify units here, since the values 90 and 360 don't mean anything on their own if `$number` is in radians.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+* Return the [arctangent] of `$number`, in rad. It is between `-pi/2 rad` and
+  `pi/2 rad`, inclusive.
+
+[arctangent]: https://en.wikipedia.org/wiki/Inverse_trigonometric_functions#Basic_properties
+
+#### `atan2()`
+
+> `atan2($y, $x)` is distinct from `atan($y / $x)` because it preserves the
+> quadrant of the point in question. For example, `atan2(1, -1)` corresponds to
+> the point `(-1, 1)` and returns `(0.75 * pi) rad`. In contrast, `atan(1 / -1)`
```
nex3: Consider making these radian examples valid SassScript:

```suggestion
> the point `(-1, 1)` and returns `0.75rad * $pi`. In contrast, `atan(1 / -1)`
```

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+> and `atan(-1 / 1)` resolve first to `atan(-1)`, so both return (`-0.25 * pi)
+> rad`.
```
nex3: This is a really good explanation :+1:.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,325 @@
...
+
+```
+atan2($y, $x)
+```
+
+* `$y` and `$x` must either have the same unit or be unitless. If not, throw an
```
nex3: Style nit: For consistency, phrase this as either "If X, throw an error" or "Throw an error if X".

nex3: Shouldn't this just check if `$y` and `$x` have compatible units?

nex3 (changes requested):
Almost there!

proposal/more-math-functions.md
Awjin: Yeah, makes sense to me. I'll have `hypot()` accept any arguments that are all compatible with each other, and return a value that takes its unit from the leftmost argument.

proposal/more-math-functions.md
Awjin: Got it, fixed.

proposal/more-math-functions.md
Awjin: Thanks, fixed.

proposal/more-math-functions.md
Awjin: Good point. Fixed.

proposal/more-math-functions.md
Awjin: Fixed.

proposal/more-math-functions.md
Awjin: Got it. I'll go with the latter.

proposal/more-math-functions.md
Awjin: Makes sense, thanks.

proposal/more-math-functions.md
Awjin: Thanks 🙂

proposal/more-math-functions.md
Awjin: Will go with `If X, throw an error`, here and elsewhere.

Awjin: Yes, good catch!

proposal/more-math-functions.md
Awjin: Got it. I'll just lean on the trig definition

proposal/more-math-functions.md
Awjin: Fixed.

proposal/more-math-functions.md
```diff
@@ -45,26 +45,27 @@ digits: `3.141592654`.
...
-* Return `Infinity` if any argument is `Infinity`.
-* Return the square root of the sum of the squares of each argument.
-
-[length]: https://drafts.csswg.org/css-values-4/#lengths
+* If all arguments are not compatible with each other, throw an error.
+* Let the unit of the return value be the same as that of the leftmost argument
+  that has a unit. If all arguments are unitless, let the return value be
+  unitless.
+  * If any argument is `Infinity`, return `Infinity`.
+  * Return the square root of the sum of the squares of each argument.
```
nex3: Did you intend to indent these?

proposal/more-math-functions.md
```diff
@@ -45,26 +45,27 @@ digits: `3.141592654`.
...
 
-* Throw an error if any argument is not a [length][].
-* Return `Infinity` if any argument is `Infinity`.
-* Return the square root of the sum of the squares of each argument.
-
-[length]: https://drafts.csswg.org/css-values-4/#lengths
+* If all arguments are not compatible with each other, throw an error.
+* Let the unit of the return value be the same as that of the leftmost argument
+  that has a unit. If all arguments are unitless, let the return value be
+  unitless.
```
nex3: What do you think about forbidding mixing unitless numbers with numbers that have units? The lack of transitivity there is weird, and while it's probably too late to change it for `+` and `-`, we could definitely change it here.

proposal/more-math-functions.md
```diff
@@ -0,0 +1,322 @@
...
+### Trigonometry
+
+#### `cos()`
+
+```
+cos($number)
+```
+
+* If `$number` has units but is not an [angle][], throw an error.
+* If `$number` is unitless, treat it as though its unit were rad.
```
nex3: Style nit:

```suggestion
* If `$number` is unitless, treat it as though its unit were `rad`.
```

nex3 (approved):
Great work!

proposal/more-math-functions.md
Awjin: The only reason I considered allowing unitless numbers was to model the behavior after `+`/`-` 🙂 But if the lack of transitivity is undesired, I'm good with forbidding mixing unitless/units.

proposal/more-math-functions.md
Awjin: I wanted to group these under one bullet, but on second thought it doesn't add much value.

proposal/more-math-functions.md
Awjin: Fixed.

Awjin:
Thanks for the reviews!

PR #2784, opened 2019-11-27
Fix broken links

nex3:


accepted/module-system.md
```diff
@@ -725,7 +723,7 @@ and [configuration](#configuration) `config`:
 * Let `file` be the [source file](#source-file) result of [loading][loading an
   import] `url`.
 
-  [loading an import]: ../spec/at-rules/import.md#loading-an-import
+  [loading an import]: https://github.com/sass/sass/blob/2c08156248496c75678826420aafb09e2ace9041/spec/at-rules/import.md#loading-an-import
```
Awjin, line 726: Does these need to be absolute urls?

Awjin (approved):

accepted/module-system.md
nex3: Yeah, they're referring to a header that doesn't exist anymore at that location. Since the new location already has this text integrated so I think it's clearer for anyone going back and reading this later on to have it link to the version that existed at time of writing.

PR #2787, opened 2019-12-03
More math functions, Draft 1.1

Awjin:
Add background and summary sections

### DIFF
--- a/proposal/more-math-functions.md
+++ b/proposal/more-math-functions.md
@@ -1,10 +1,12 @@
-# More Math Functions: Draft 1
+# More Math Functions: Draft 1.1
 
 *[(Issue)](https://github.com/sass/sass/issues/851)*
 
 This proposal adds the following members to the built-in `sass:math` module.
 
 ## Table of Contents
+* [Background](#background)
+* [Summary](#summary)
 * [Variables](#variables)
   * [`$e`](#e)
   * [`$pi`](#pi)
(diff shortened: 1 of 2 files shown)


proposal/more-math-functions.md
```diff
@@ -25,6 +27,54 @@ This proposal adds the following members to the built-in `sass:math` module.
...
+
+## Summary
+
+> This section is non-normative.
+
+This proposal defines Sassified versions of all the mathematical functions in
+the [CSS Values and Units 4 Draft][], as well as logarithms and the constants
+`e` and `pi`. Each function is basically equivalent to its mathematical form,
+with stricter unit handling. Proper unit handling prevents these functions from
+creating meaningless units. For instance, what is the unit of `(1px)`<sup><code>2</code></sup>? A `px`<sup><code>2</code></sup>?
```
jathak: For the example, it might be better to do something with fractional exponents, since `px^2` is actually a reasonable unit of area (even if it's not really useful for CSS). It might also make more sense to just use a `^` to represent the exponent here since GitHub doesn't support proper math rendering.

Also, make sure to break this line at 80 characters

jathak (approved):

proposal/more-math-functions.md
Awjin: Good point. Changed this to `px^(1/3)`.

nex3 (approved):

PR #2790, opened 2019-12-05
More Math Functions: Draft 2

### LINKED ISSUE #2799: [Math Functions] Mark built-in module variables as unmodifiable
The [math function proposal](https://github.com/sass/sass/blob/master/proposal/more-math-functions.md) adds two variables to the built-in `sass:math` module, which are actually the first variables ever exposed by a built-in module (cool!). We probably don't want users setting `$e` and `$pi` to different values, though, so they should be unmodifiable.

Rather than special-casing those two variables in particular, I think the proposal should update the variable-assignment logic to fail if the variable comes from *any* built-in module, effectively making all built-in modules' variables unmodifiable.

Awjin:
* Variables
  * `$e` and `$pi` have 1 more digit of precision after the decimal.
  * Variables from built-in modules cannot be modified.

* `Infinity` and `-Infinity`:
  * If any argument to `hypot()` equals `-Infinity`, it returns `Infinity`.
  * The `$exponent == Infinity` case in `pow()` also holds for
    `$exponent == -Infinity`.
  * The `$number == Infinity` cases in `cos()`, `sin()`, and `tan()` also hold
    for `$number == -Infinity`.

* Input units:
  * `clamp()`'s arguments must all have compatible units, or all be unitless.
  * `log()` does not error unless the input has units, and instead delegates
    edge cases to division.

* Output units:
  * For `acos()`, `asin()`, and `atan()`, and `atan2()`, all of their outputs
    are numbers in `deg`.

Closes #2799

### DIFF
--- a/proposal/more-math-functions.md
+++ b/proposal/more-math-functions.md
@@ -62,7 +62,7 @@ unitless number.
 [angle]: https://drafts.csswg.org/css-values-4/#angles
 
 The inverse trig functions—`acos()`, `asin()`, `atan()`—accept a unitless number
-and output a SassScript number in `rad`. `atan2()` is similar, but it accepts
+and output a SassScript number in `deg`. `atan2()` is similar, but it accepts
 two unitless numbers.
 
 `clamp()` accepts three SassScript numbers with [compatible][] units: the
--- a/proposal/more-math-functions.md
+++ b/proposal/more-math-functions.md
@@ -7,6 +7,8 @@ This proposal adds the following members to the built-in `sass:math` module.
 ## Table of Contents
 * [Background](#background)
 * [Summary](#summary)
+* [Semantics](#semantics)
+  * [Built-in Module Variables](#built-in-module-variables)
 * [Variables](#variables)
   * [`$e`](#e)
   * [`$pi`](#pi)
@@ -75,6 +77,13 @@ whereas `clamp(1in, 1cm, 12in)` outputs `1in`.
 length of the `n`-dimensional vector that has components equal to each of the
 inputs. Since the inputs' units may all be different, the output takes the unit
 of the first input.
+
+## Semantics
+
+### Built-in Module Variables
+
+If a variable is provided by a built-in module, it is not modifiable. Any
+attempts to reassign such a variable results in failure.
 
 ## Variables
 
--- a/proposal/more-math-functions.md
+++ b/proposal/more-math-functions.md
@@ -1,4 +1,4 @@
-# More Math Functions: Draft 1.1
+# More Math Functions: Draft 2
 
 *[(Issue)](https://github.com/sass/sass/issues/851)*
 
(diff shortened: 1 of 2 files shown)


proposal/more-math-functions.md
```diff
@@ -175,8 +176,12 @@ tan($number)
 
 * If `$number` has units but is not an angle, throw an error.
 * If `$number` is unitless, treat it as though its unit were `rad`.
-* If `$number == Infinity`, return `NaN` as a unitless number.
+* If `$number == Infinity` or `$number == -Infinity`, return `NaN` as a unitless number.
 * If `$number == -0`, return `-0` as a unitless number.
+* If `$number` has units `rad` and is equal to `0.5rad * pi +/- 2rad * pi * n`,
+  where `n` is any integer, return `16331239353195370` as a unitless number.
```
nex3: Why these huge constants and not `Infinity`/`-Infinity`?

nex3 (approved):

proposal/more-math-functions.md
Awjin: The CSS spec basically says that `rad` can't exactly represent the asymptotic values, so "whether the result is a very large negative or positive value can depend on rounding and precise details of how numbers are internally stored. It’s recommended you don’t depend on this behavior if using such units."

We could take a stance here and just have these yield `Infinity`/`-Infinity`?

nex3: I'd prefer that, I think, yeah.

proposal/more-math-functions.md
```diff
@@ -76,17 +78,24 @@ length of the `n`-dimensional vector that has components equal to each of the
 inputs. Since the inputs' units may all be different, the output takes the unit
 of the first input.
 
+## Semantics
+
+### Built-in Module Variables
+
+If a variable is provided by a built-in module, it is not modifiable. Any
+attempts to reassign such a variable results in failure.
```
nex3: If possible, it would be good to word this as an amendment to [the existing spec language for setting variables](https://github.com/sass/sass/blob/master/spec/variables.md#executing-a-variable-declaration). How you do this specifically is up to you; one possible way is to say in prose "add this text in that place", another way is to rewrite the text with your changes highlighted [as in this spec](https://github.com/sass/sass/blob/master/accepted/import-configuration.md#importing-files).

The goal here is to make it easy for a reader to understand what the canonical language spec will look like once this proposal lands. The more ambiguity there is there, the more avenues there are for individual implementations to interpret it differently.

nex3 (changes requested):

proposal/more-math-functions.md
Awjin: > The goal here is to make it easy for a reader to understand what the canonical language spec will look like once this proposal lands.

Thanks, that put it into perspective. Updated the draft accordingly.

proposal/more-math-functions.md
```diff
@@ -76,17 +78,73 @@ length of the `n`-dimensional vector that has components equal to each of the
...
+* **Otherwise, if `resolved` is a variable from a built-in module, throw an
+  error.**
+
+* Otherwise, if `declaration` is outside of any block of statements, *or*
+  `declaration` has a `!global` flag, *or* `name` is a `NamespacedVariable`:
+
+  * ~~Let `resolved` be the result of [resolving a variable][] named `name` using
+    `file`, `uses`, and `import`.~~
+
+  [...]
```
nex3: This gets detected as a potential link reference, which is causing the tests to fail. Maybe write it `(...)` instead.

nex3 (approved):
Looks great!

proposal/more-math-functions.md
Awjin: Done.

Issue #2834, opened 2020-03-15
Add hwb() support in the sass:color module


mirisuzanne:
*Edited by @nex3*

* [x] [Proposal](https://github.com/sass/sass/pull/2835)
* [x] [Specs](https://github.com/sass/sass-spec/issues/1570)
* [x] [Documentation](https://github.com/sass/sass-site/issues/477)
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1089)
* [ ] [LibSass](https://github.com/sass/libsass/issues/3128)

---

- based on [CSS Color level 4](https://www.w3.org/TR/css-color-4/#the-hwb-notation)
- Adds Sassified functions for `hwb()` color definitions, inspection, & adjustments in the `sass:color` module
- Split out from #2831

nex3:
Oops, forgot to mention here: this has been [open for public review](https://sass-lang.com/blog/request-for-comments-hwb-functions) since 7 October, and on 21 October if no new issues have come up we'll mark it as accepted and start the release process.

nex3:
Marking this as accepted now.

PR #2835, opened 2020-03-15
Proposal for hwb support in the sass:color module

mirisuzanne:
See issue #2834

### DIFF
--- a/proposal/color-4-hwb.md
+++ b/proposal/color-4-hwb.md
@@ -132,7 +132,7 @@ All new functions are part of the `sass:color` built-in module.
   
   * Return a color with the given `red`, `green`, `blue`, and `alpha` channels.
 
-  [percent-converting]: built_in_modules/color.md#percent-converting-a-number
+  [percent-converting]: ../spec/built_in_modules/color.md#percent-converting-a-number
   [to RGB]: https://www.w3.org/TR/css-color-4/#hwb-to-rgb
 
 
@@ -270,8 +270,6 @@ This function's new definition is as follows:
   * Return the result of calling [`hsl()`][] with `hue`, `saturation`,
     `lightness`, and `alpha`.
 
-    [`hsl()`]: ../spec/functions.md#hsl-and-hsla
-
 * Otherwise, if either `$hue`, `$whiteness`, or `$blackness` aren't null:
 
   * If either `$whiteness` or `$blackness` aren't either null or numbers with
@@ -293,6 +291,8 @@ This function's new definition is as follows:
 
 * Otherwise, return a color with the same red, green, and blue channels as
   `$color` and `alpha` as its alpha channel.
+
+[`hsl()`]: ../spec/functions.md#hsl-and-hsla
 
 ### `change()`
 
--- a/proposal/color-4-hwb.md
+++ b/proposal/color-4-hwb.md
@@ -1,0 +1,173 @@
+# HWB Color Functions: Draft 1
+
+*([Issue](https://github.com/sass/sass/issues/2831))*
+
+This proposal adds a new `hwb()` color format to the `sass:color` module, along
+with inspection and adjustment options for _whiteness_ and _blackness_.
+
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+  * [Design Decisions](#design-decisions)
+* [Semantics](#semantics)
+
+## Background
+
+> This section is non-normative.
+
+The [CSS Color Module level 4][color-4], provides several new CSS formats for
+describing color, but `hwb()` stands out as part of the same `sRGB` color
+system that Sass already uses internally.
+
+[color-4]: https://www.w3.org/TR/css-color-4/
+
+## Summary
+
+> This section is non-normative.
+
+This proposal defines  aSassified versions of the `hwb()` color function
+added to [CSS Color Level 4][color-4] – along with relevant inspection and
+adjustment options. This function will only be available inside the `sass:color`
+module to avoid conflicts with the CSS syntax, and will be converted to more
+common color-name, hex, or `rgba()` syntax for output -- following the same
+logic as our current color functions.
+
+- New `color.hwb()` describes colors in the sRGB colorspace using `$hue` (defined
+  identically to the `hsl()` "hue" value), along with `$whiteness`, `$blackness`,
+  and optional `$alpha` transparency.
+- New `color.whiteness()` and `color.blackness()` functions return the respective
+  values of `w` or `b` for a given color.
+- Existing `color.adjust()`, `color.scale()`, and `color.change()` functions will
+  accept additional `$whiteness` and `$blackness` parameters before the final
+  `$alpha` parameter.
+
+### Design Decisions
+
+For backwards-compatibility reasons, both `rgb/a()` and `hsl/a()` are processed
+as Sass function in the global namespace, and output is converted to the format
+with best browser-support. The primary purpose of Sass color-function support
+is our ability to provide those conversions, but the module syntax now allows
+us to namespace new formats, so authors can opt-into the feature. With that
+in mind:
+
+- The new functions **will not** be available on the global namespace.
+- The new functions **will not** accept *special number string* or
+  *special variable string* values that can only be resolved in CSS.
+
+Unknown CSS functions are already passed through implicitly. As browser support
+grows for the native-CSS functions, authors will be able to migrate on their own
+schedule – or use both Sass & CSS functions together to progressively enhance
+their designs.
+
+Since all `color.hwb()` output will be processed for rgb output, there is no
+need to support shortcuts like `hwb($color, $alpha)`. However, it is reasonable
+to support both the space-delimited CSS syntax, as well as the comma-delimited
+Sass named-argument syntax.
+
+## Semantics
+
+### `hwb()`
+
+* ```
+  hwb($hue, $whiteness, $blackness, $alpha: 1)
+  ```
+
+  * If any of `$hue`, `$whiteness`, `$blackness`, or `$alpha` aren't numbers,
+    throw an error.
+
+  * If either of `$whiteness` or `$blackness` is not a percentage between 0% and
+    100%, throw an error.
+
+  * Let `hue` be `($hue % 360) / 6`.
+
+  * Let `whiteness` and `$blackness` be the respective values of `$whiteness`
+    and `$blackness`
+
+  * If the sum of `whiteness` and `blackness` is greater than 100%, normalize
+    both to maintain the same relative ratio, adding up to 100% combined.
+
+  * Let `red`, `green`, and `blue` be the result of converting `hue`,
+    `whiteness`, and `blackness` [to RGB][].
+
+  [to RGB]: https://www.w3.org/TR/css-color-4/#hwb-to-rgb
+
+* ```
+  hwb($channels)
+  ```
+
+  * If `$channels` is not an unbracketed space-separated list, throw an error.
+
+  * If `$channels` does not includes exactly three elements, throw an error.
+
+  * Let `hue` and `whiteness` be the first two elements of `$channels`
+
+  * If the third element of `$channels` has preserved its status as
+    two slash-separated numbers:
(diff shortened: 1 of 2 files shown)


proposal/color-4-hwb.md
```diff
@@ -0,0 +1,179 @@
...
+  * [`blackness()`](#blackness)
+  * [`adjust()`](#adjust)
+  * [`change()`](#change)
+  * [`scale()`](#scale)
+
+## Background
+
+> This section is non-normative.
+
+The [CSS Color Module level 4][color-4], provides several new CSS formats for
```
nex3: Nit

```suggestion
The [CSS Color Module level 4][color-4] provides several new CSS formats for
```

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,179 @@
...
+describing color, but `hwb()` stands out as part of the same `sRGB` color
+system that Sass already uses internally.
+
+[color-4]: https://www.w3.org/TR/css-color-4/
+
+## Summary
+
+> This section is non-normative.
+
+This proposal defines  aSassified versions of the `hwb()` color function
```
nex3: Nit

```suggestion
This proposal defines a Sassified version of the `hwb()` color function
```

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,179 @@
...
+
+  * If the sum of `whiteness` and `blackness` is greater than 100%, normalize
+    both to maintain the same relative ratio, adding up to 100% combined.
+
+  * Let `red`, `green`, and `blue` be the result of converting `hue`,
+    `whiteness`, and `blackness` [to RGB][].
+
+  [to RGB]: https://www.w3.org/TR/css-color-4/#hwb-to-rgb
+
+* ```
```
nex3, line 141: If this isn't a global function, why accept the global function color syntax?

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,179 @@
...
+    defined) as arguments and return the result.
+
+### `whiteness()`
+
+```
+whiteness($color)
+```
+
+* If `$color` is not a color, throw an error.
+* Return the `hwb()` whiteness value of `$color`.
```
nex3: I think this warrants more elaboration. Can you link to the algorithm for computing this?

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,179 @@
...
+
+  * Let `hue` be `($hue % 360) / 6`.
+
+  * Let `whiteness` and `$blackness` be the respective values of `$whiteness`
+    and `$blackness`
+
+  * If the sum of `whiteness` and `blackness` is greater than 100%, normalize
+    both to maintain the same relative ratio, adding up to 100% combined.
+
+  * Let `red`, `green`, and `blue` be the result of converting `hue`,
```
nex3, line 127: One thing we haven't explicitly specified yet is what Sass's spec representation is of colors that were originally specified in different formats. In Dart Sass, we store colors as HSL values if they were originally HSL and RGB if they were originally RGB, lazily converting to the other format if that becomes necessary. Presumably we'll do something similar for HWB. This can be visible to the user—we should be explicit about whether, for example, `whiteness(hwb(..., $x, ...))` is guaranteed to return *exactly* `$x` or merely *approximately* `$x`.

I think the actual way you've specified this, treating sRGB values as the internal representation, is reasonable... but we should add an explicit amount of flexibility to enable the sort of optimization Dart Sass does.

You've also got to specify what the function actually returns :smile:.

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,179 @@
...
+  * If either of `$whiteness` or `$blackness` is not a percentage between 0% and
+    100%, throw an error.
+
+  * Let `hue` be `($hue % 360) / 6`.
+
+  * Let `whiteness` and `$blackness` be the respective values of `$whiteness`
+    and `$blackness`
+
+  * If the sum of `whiteness` and `blackness` is greater than 100%, normalize
+    both to maintain the same relative ratio, adding up to 100% combined.
```
nex3: Please describe the process of normalizing these in more detail, ideally with a formula.

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,179 @@
...
+```
+blackness($color)
+```
+
+* If `$color` is not a color, throw an error.
+* Return the `hwb()` blackness value of `$color`.
+
+### `adjust()`
+
+This proposal adds new `$whiteness` and `$blackness` parameters to the `adjust()`
```
nex3, line 210: These functions should also be more explicit about what they actually do.

To give some background: because writing a full language spec from scratch would take way too much time, we sort of fill it in piecemeal as it becomes relevant to each new proposal. This means that proposals like this need to do a bit of extra legwork to specify at least enough of the existing behavior to clearly specify the new behavior in relation to it.

nex3 (changes requested):
Before we land this, I want to talk about our plan for `hwb()` going forward. Although it doesn't quite come out and say it explicitly, this proposal seems to imply that we won't ever add special Sass processing for the global `hwb()` function, and I want to push back on that.

Generally speaking, we're willing to add special support for new CSS syntax once it's shipped in at least one browser, since that's a pretty good guarantee that it's not going to completely change out from under us. So we probably *will* want to support global `hwb()` syntax in the future. This leads me to think that this proposal should explicitly address the following questions:

* Why add `color.hwb()` now rather than waiting for `hwb()` to land in a browser and implementing the global function when that happens?

* Once we do add support for global `hwb()`, what's the plan for `color.hwb()`? Do we deprecate it and perform a migration?

nex3:
@mirisuzanne I made some revisions that I'd love your feedback on. In particular, I changed the design decisions section pretty substantially. The actual *decisions* are the same, but the reasoning behind them and the plans going forward are not.

proposal/color-4-hwb.md
```diff
@@ -82,22 +105,37 @@ Sass named-argument syntax.
   * If any of `$hue`, `$whiteness`, `$blackness`, or `$alpha` aren't numbers,
     throw an error.
 
-  * If either of `$whiteness` or `$blackness` is not a percentage between 0% and
-    100%, throw an error.
+  * If `$hue` has any units other than `deg`, throw an error.
```
mirisuzanne, line 110: CSS accepts several different `<angle>` units. I don't know how much they are used (`deg` is certainly the most common), but we might want to match? https://www.w3.org/TR/css3-values/#angle-value

proposal/color-4-hwb.md
```diff
@@ -82,22 +105,37 @@ Sass named-argument syntax.
...
   * Let `red`, `green`, and `blue` be the result of converting `hue`,
-    `whiteness`, and `blackness` [to RGB][].
+    `saturation`, and `lightness` [to RGB][].
+
+  * Set `red`, `green`, and `blue` to their existing values multiplied by 255
+    and rounded to the nearest integers.
+
+  * Let `alpha` be the result of [percent-converting][] `$alpha` with a `max` of 1.
+  
+  * Return a color with the given `red`, `green`, `blue`, and `alpha` channels.
```
mirisuzanne, line 135: Returning values from a function is totally overrated 😂

mirisuzanne (commented):
@nex3 Changes and expansions here are great, thanks. Just one question below about hue angles…

proposal/color-4-hwb.md
nex3: We should support that, but it's tangential enough that I don't think it should be part of this proposal. I've filed https://github.com/sass/sass/issues/2904 to track that.

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,455 @@
...
+
+This algorithm takes a number `number`, a value `factor`, and a number `max`.
+It's written "scale `<number>` by `<factor>` with a `max` of `<max>`". It
+returns a number with a value between 0 and `max` and the same units as
+`number`.
+
+* If `factor` isn't a number with unit `%` between `-100%` and `100%`
+  (inclusive), throw an error.
+
+* If `factor > 0%`, return `number + (max - number) * factor / 100%`.
```
Awjin, line 93: What happens if `max < number`?

proposal/color-4-hwb.md
```diff
@@ -152,6 +216,84 @@ adjust($color,
...
+
+  * Return a color with `red`, `green`, `blue`, and `alpha` as the red, green,
+    blue, and alpha channels, respectively.
+
+* Otherwise, if either `$saturation` or `$lightness` aren't null:
+
+  * If either `$whiteness` or `$blackness` aren't null, throw an error.
+
+  * If either `$saturation` or `$lightness` aren't either null or numbers
+    between -100 and 100 (inclusive), throw an error.
```
Awjin, line 259: Here and below (lines 265, 268, 347): should these have unit `%`?

Awjin (changes requested):

proposal/color-4-hwb.md
nex3: In practice it never will be, since `number` always comes from a color channel which is guaranteed to be between 0 and `max`. I'll add a note to that effect.

proposal/color-4-hwb.md
nex3: Unfortunately, this represents the current behavior. We currently ignore units for saturation and lightness no matter what, which isn't ideal but isn't necessarily enough of a problem to warrant a deprecation pass either.

Awjin (approved):

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,457 @@
...
+  * If `$hue` has any units other than `deg`, throw an error.
+
+  * If either of `$whiteness` or `$blackness` don't have unit `%` or aren't
+    between `0%` and `100%` (inclusive), throw an error.
+
+  * Let `hue` be `($hue % 360) / 60` without units.
+
+  * Let `whiteness` be `$whiteness / 100%`.
+
+  * Let `$blackness` be `$blackness / 100%`.
```
jathak:
```suggestion
  * Let `blackness` be `$blackness / 100%`.
```

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,457 @@
...
+
+  * If `whiteness + blackness > 1`:
+
+    * Set `whiteness` to `whiteness / (whiteness + blackness)`.
+
+    * Set `blackness` to `blackness / (whiteness + blackness)`.
+
+  * Let `red`, `green`, and `blue` be the result of converting `hue`,
+    `saturation`, and `lightness` [to RGB][].
```
jathak:
```suggestion
    `whiteness`, and `blackness` [to RGB][].
```

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,457 @@
...
+
+* If `$color` isn't a color, throw an error.
+
+* Let `alpha` be `$color`'s alpha channel.
+
+* If `$alpha` isn't null:
+
+  * If `$alpha` isn't a number between -1 and 1 (inclusive), throw an error.
+
+  * Set `alpha` to `alpha + $alpha` clamped between 0 and 255.
```
jathak:
```suggestion
  * Set `alpha` to `alpha + $alpha` clamped between 0 and 1.
```

proposal/color-4-hwb.md
```diff
@@ -0,0 +1,457 @@
...
+
+  * Let `lightness` be the result of calling `lightness($color)` if
+    `$lightness` is null, or `$lightness` otherwise.
+
+  * Return the result of calling [`hsl()`][] with `hue`, `saturation`,
+    `lightness`, and `alpha`.
+
+* Otherwise, if either `$hue`, `$whiteness`, or `$blackness` aren't null:
+
+  * If either `$saturation` or `$lightness` aren't either null or numbers with
```
jathak:
```suggestion
  * If either `$whiteness` or `$blackness` aren't either null or numbers with
```

jathak (approved):

nex3 (approved):

PR #2836, opened 2020-03-16
Add nested-map support in sass:map module

mirisuzanne:
Formal proposal based on discussion in #1739

## Questions raised:

- Is it confusing that the list of `$keys...` in `map.remove()` is used to indicated _multiple keys_ rather than _nested keys_?
- Is there a reason we don't want a function that deep-merges two complex maps with nested values?

### DIFF
--- a/proposal/nested-map-functions.md
+++ b/proposal/nested-map-functions.md
@@ -9,26 +9,51 @@ setting, and getting items from nested maps.
 
 * [Background](#background)
 * [Summary](#summary)
-* [Syntax](#syntax)
+* [Functions](#functions)
   * [`get()`](#get)
   * [`has-key()`](#has-key)
   * [`set()`](#set)
   * [`merge()`](#merge)
+  * [`deep-merge()`](#deep-merge)
 
 ## Background
 
 > This section is non-normative.
 
-The current map inspection and manipulation functions don't provide any built-in
-support for managing nested maps. Projects often build thir own tooling, but
-the results are inconsistent, and often slow.
+Variables have always been a key feature of the Sass language. But these days,
+design systems and component libraries form the basis of most CSS projects --
+with well organized _design tokens_ as the foundation. While Individual token
+variables can be quite useful, the ability to group tokens into strucutred and
+meaningful relationships is essential for creating resilient systems.
+
+There are many ways to group tokens. The popular Style Dictionary recommends
+a deep nesting of _category_,  _type_, _item_, _sub-item_, and _state_. Other
+taxonomies also include concepts like _theme_, or even _operating system_. Most
+of the existing tools rely on YAML or JSON objects to achieve that nested
+structure, at the expense of other important information. YAML and JSON are not
+design languages, and do not understand fundamental CSS concepts like color or
+length.
+
+With Sass, we don't have to make that tradeoff. We already support nestable map
+structures, and the ability to interact with them programmatically -- adding or
+removing properties, accessing values, and looping over entire structures. But
+current built-in functions don't provide much support for managing nested maps.
+Projects often build thir own tooling.
+
+The results are inconsistent across projects, difficult to re-use, and often
+slow to compile. Implementing core support for nested maps could change all that.
 
 ## Summary
 
 > This section is non-normative.
 
-This proposal updates the `sass:map` module to better support inspection and
-manipulation of nested maps.
+This proposal updates existing map functions with better support for inspection
+and manipulation of nested maps, as well as adding new functions to the
+`sass:map` module. For existing legacy functions (`get()`, `has-key()`,
+`merge()`) the new behavior will be accessible through both the `sass:map`
+module, and global legacy names (`map-get()`, `map-has-key()`, `map-merge()`).
+New functions (`set()`, `deep-merge()`) will only be available inside the
+`sass:map` module.
 
 The `has-key()` and `get()` functions both accept multiple `$keys...`:
 
@@ -89,7 +114,51 @@ $nav: map.set($nav, 'color', 'hover', 'search', green);
 $nav: map.set($nav, 'color', 'hover', 'logo', orange);
 ```
 
-## Syntax
+And finally, a new `deep-merge()` function in the `sass:map` module allows
+merging two or more nested maps. This works much like the existing `merge()`
+function, but when both maps have a nested-map at the same key, those nested
+maps are also merged:
+
+```scss
+@use 'sass:map';
+
+$nav: (
+  'bg': 'gray',
+  'color': (
+    'hover': (
+      'search': yellow,
+      'home': red,
+      'filter': blue,
+    ),
+  ),
+);
+
+$update: (
+  'bg': white,
+  'color': (
+    'hover': (
+      'search': green,
+      'logo': orange,
+    )
+  )
+);
+
+$nav: map.deep-merge($nav, $update);
+
+// $nav: (
+//   'bg': white,
+//   'color': (
+//     'hover': (
+//       'search': green,
+//       'home': red,
+//       'filter': blue,
+//       'logo': orange,
+//     ),
+//   ),
+// );
+```
+
+## Functions
 
 ### `get()`
 
@@ -97,9 +166,13 @@ $nav: map.set($nav, 'color', 'hover', 'logo', orange);
 get($map, $keys...)
 ```
 
+* If `$map` is not a map, throw an error.
+
+* If `$keys` is empty, throw an error.
+
 * Let `key` be the first (or only) element in `$keys`
 
-* If `$map` does not have a key with the same name as `key`, throw an error.
+* If `$map` does not have a key equal to `key`, return `null`.
 
 * Let `value` be the value assigned to `key` in `$map`
 
@@ -116,6 +189,10 @@ get($map, $keys...)
 ```
 has-key($map, $keys...)
 ```
+
+* If `$map` is not a map, throw an error.
+
+* If `$keys` is empty, throw an error.
 
 * Let `key` be the first (or only) element in `$keys`
 
@@ -135,22 +212,22 @@ has-key($map, $keys...)
 ### `set()`
 
 ```
-set($map1, $keys..., $value)
+set($map, $args...)
 ```
 
-* If `$map1` is not a map, throw an error.
+* If `$map` is not a map, throw an error.
 
-* If fewer than three arguments are provided, throw an error.
+* Let `map` be an empty map.
 
-* Let `map` be an empty map .
+* Let `set-key` be the first item in arglist `$args`.
 
-* If there is more than one argument in arglist `$keys`:
+* Let `remaining` be a slice of all the other items in arglist `$args`.
 
-  * Let `set-key` be the first element in `$keys`.
+* If there are no items in `remaining`, throw an error.
 
-  * Let `keys` be a slice of all `$keys` elements except the first.
+* If there is more than one argument in `remaining`:
 
-  * If `$map1` has a key `current-key` with the same name as `set-key`:
+  * If `$map` has a key `current-key` that is equal to `set-key`:
 
     * Let `current` be the value of `current-key`.
 
@@ -159,46 +236,87 @@ set($map1, $keys..., $value)
     * Let `current` be an empty map.
 
   * Let `set-value` be the result of calling `set()` with `current` and expanded
-      `keys` as arguments.
+    `remaining` as arguments.
 
-    > This will error if `current` is not a map, but we stil have nested `keys`
+    > This will error if `current` is not a map, but we stil have `remaining`
 
 * Otherwise:
 
-  * Let `set-key` be the only element in `$keys`
+  * Let `set-value` be the only item in `remaining`
 
-  * Let `set-value` be the value of `$value`
-
-* Add a key with name `set-key` and value `set-value` to `map`
-
-* For each `key`, `value` pair in `$map1`
-
-  * If a key with name `key` already exists in `map`, do nothing.
-
-  * Otherwise, add a key to `map` with name `key` and value `value`.
-
-* Return the value `map`
+* Return a copy of `map` with `set-key` set to `set-value`, overriding an
+  existing value for `set-key` if one exists.
 
 ### `merge()`
 
+This proposal add a new overload to the existing `merge()` function:
+
 ```
-merge($map1, $keys..., $map2)
+merge($map1, $args...)
 ```
 
-* If only one argument is provided, throw an error.
+* If `$args` is empty, throw an error
 
-* If either the first (`$map1`) or last (`$map2`) argument is not a map, throw
-  an error.
+* Let `map2` be the last item in arglist `$args`
 
-* Let `map` be an empty map.
+* If either `$map1` or `map2` is not a map, throw an error.
 
-* For each `key`, `new` pair in `$map2`:
+* If arglist `$args` has more than one item:
 
-  * If `$map1` has a key with the same name as `key`
+  * Let `get-keys` be a slice of all except the last item in `args`.
 
-  * Let `keys` be the result of appending `key` to the argument list `$keys`.
+  * Let `sub` be the result of calling `get()` with `sub` and expanded `get-keys`
+    as arguments.
 
-  * Update `map` to be the result of calling `set()` with `map`,
-    expanded `keys`, and `value` as arguments.
+  * Let `sub-merged` be the result of calling `merge()` with `sub` and `map2`
+    as argumets.
 
-* Return the value of `map`
+  * Let `set-args` be the result of appending `sub-merged` to the list `keys`
+
+  * Return the result of calling `set()` with `$map1` and expanded `set-args`
+    as arguments.
+
+* Otherwise:
+
+  * Let `map` be a copy of `$map1`.
+
+  * For each `key`, `value` in `map2`:
+
+    * Replace `map` with the result of calling `set()` with `map`, `key`, and
+      `value` as arguments.
+
+  * Return `map`.
+
+### `deep-merge()`
+
+```
+deep-merge($maps...)
+```
+
+* If the length of `$maps` is less than two, throw an error.
+
+* If  any of the items in `$maps` are not maps, throw an error.
+
+* Let `merged` be an empty map.
+
+* For each `map` in `maps`:
+
+  * for each `key`, `value` in `map`:
+
+    * If `value` is a map:
+
+      * Let `current` be the result of calling `get()` with `merged` and `key` as
+        arguments.
+
+      * If `current` is a map:
+
+        * Let `deep` be the result of calling `deep-merge()` with `current` and
+          `value` as arguments.
+
+        * Replace `map` with the result of calling `set()` with `map`, `key`, and
+          `deep` as arguments.
+
+  * Replace `merged` with the result of calling `merge` with `merged` and `map`
+    as arguments.
+
+* Return `merged`.
--- a/proposal/nested-map-functions.md
+++ b/proposal/nested-map-functions.md
@@ -308,6 +308,8 @@ merge($map1, $args...)
 
 * If either `$map1` or `map2` is not a map, throw an error.
 
+* If `$args` has fewer than two elements, throw an error.
+
 * Let `keys` be a slice of all elements in `$args` except the last.
 
 * Let `sub` be the result of calling `get()` with `$map1` as the first
@@ -323,7 +325,8 @@ merge($map1, $args...)
   * Let `sub-merged` be `map2`.
 
 * Return the result of calling `set()` with `$map1` as the first argument,
-  followed by the contents of `keys` as separate arguments, followed by `sub`.
+  followed by the contents of `keys` as separate arguments, followed by
+  `sub-merged`.
 
 ### `deep-merge()`
 
--- a/proposal/nested-map-functions.md
+++ b/proposal/nested-map-functions.md
@@ -328,29 +328,25 @@ merge($map1, $args...)
 ### `deep-merge()`
 
 ```
-deep-merge($map1, $map2, $maps...)
+deep-merge($map1, $map2)
 ```
 
-* Let `maps` be a list containing `$map1`, `$map2`, and the elements of `$maps`.
+* If `$map1` and `$map2` are not maps, throw an error.
 
-* If any elements in `maps` are not maps, throw an error.
+* Let `merged` be a copy of `$map1`.
 
-* Let `merged` be an empty map.
+* For each `new-key`/`new-value` pair in `$map2`:
 
-* For each `map` in `maps`:
+  * If `merged` has a key `old-key` that's `==` to `new-key`:
 
-  * For each `new-key`/`new-value` pair in `map`:
+    * Let `old-value` be the value associated with `old-key` in `merged`.
 
-    * If `merged` has a key `old-key` that's `==` to `new-key`:
+    * Remove `old-key`/`old-value` from `merged`.
 
-      * Let `old-value` be the value associated with `old-key` in `merged`.
+    * If both `old-value` and `new-value` are maps, set `new-value` to the
+      result of calling `deep-merge()` with `old-value` and `new-value`.
 
-      * Remove `old-key`/`old-value` from `merged`.
-
-      * If both `old-value` and `new-value` are maps, set `new-value` to the
-        result of calling `deep-merge()` with `old-value` and `new-value`.
-
-    * Associate `new-key` with `new-value` in `merged`.
+  * Associate `new-key` with `new-value` in `merged`.
 
 * Return `merged`.
 
--- a/proposal/nested-map-functions.md
+++ b/proposal/nested-map-functions.md
@@ -3,7 +3,7 @@
 *[(Issue)](https://github.com/sass/sass/issues/1739)*
 
 This proposal updates the built-in `sass:map` module to better support merging,
-setting, and getting items from nested maps.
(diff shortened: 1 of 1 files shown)


lunelson:
> Is there a reason we don't want a function that deep-merges two complex maps with nested values?

I was about to say "IIRC because there are different strategies in deep merging", but then looking up the JS module [deep-merge](https://github.com/Raynos/deep-merge) which —IIRC— was considered a reference for this, it seems to have deprecated itself in favor of [deep-extend](https://github.com/unclechu/node-deep-extend) which seems to have a single strategy...so hmm yeah valid question 🤔 but it would need to be a new function, e.g. maybe `map-deep-merge`, otherwise it would be a breaking change as @nex3 [pointed out here](https://github.com/sass/sass/issues/1739#issuecomment-156586871)

mirisuzanne:
Yeah, I agree it would need a new syntax - but I think it might be simple to narrow down on a strategy that works for the most common use-cases. Ideally extrapolating out from the current non-recursive merge behavior:

- When both maps share a key…
- If both keys have map values, use the result of merging those maps
- Otherwise, use the value of the value from `$map2`, and discard the value from `$map1`

This behaves the way I would expect:

```scss
@function deep-merge(
  $maps...
) {
  @if not (list.length($maps) > 1) {
    @error 'Requires two or more maps for merging';
  }

  $merged: ();

  @each $map in $maps {
    @each $key, $val in $map {
      @if (meta.type-of($val) == 'map') {
        $current: map.get($merged, $key);
        @if (meta.type-of($current) == 'map') {
          $val: deep-merge($current, $val);
          $map: map.merge($map, ($key: $val));
        }
      }
    }
    $merged: map.merge($merged, $map);
  }

  @return $merged;
}

$map1: (
  a: 12,
  b: 1 2 3 4,
  c: (d: 5),
  e: (
    f: 6,
    g: 7,
    i: (a: 1, b: 2),
  ),
);

$map2: (
  a: 24,
  b: 6 7,
  c: wipe-out,
  e: (
    f: 20,
    h: 84,
    i: (a: 5, c: 3),
  ),
);

$result: deep-merge($map1, $map2);

// $result: (
//   a: 24, 
//   b: 6 7, 
//   c: wipe-out, 
//   e: (
//     f: 20, 
//     g: 7, 
//     i: (a: 5, b: 2, c: 3), 
//     h: 84
//   )
// )
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+*[(Issue)](https://github.com/sass/sass/issues/1739)*
+
+This proposal updates the built-in `sass:map` module to better support merging,
+setting, and getting items from nested maps.
+
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+* [Syntax](#syntax)
```
nex3: Nit: We usually use the "Syntax" header to describe new syntax that literally has to be parsed in a different way than it was before. I'd call this section "Functions" to match https://github.com/sass/sass/blob/master/accepted/more-math-functions.md.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+  * [`get()`](#get)
+  * [`has-key()`](#has-key)
+  * [`set()`](#set)
+  * [`merge()`](#merge)
+
+## Background
```
nex3, line 20: It's probably worth going into a bit more detail on *why* nested maps are useful, how prevalent they are in the ecosystem, and so on.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Syntax
+
+### `get()`
+
+```
+get($map, $keys...)
+```
+
+* Let `key` be the first (or only) element in `$keys`
```
nex3: What happens if `$keys` is empty? Same question for other functions.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+
+* If `$map` does not have a key with the same name as `key`, throw an error.
```
nex3: This is contrary to the current behavior, where `map.get((), 'foo')` returns `null`.

nex3: What if `$map` isn't a map? Same question for other functions.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+  * Let `keys` be all but the first element in `$keys`
+
+  * Call `has-key()` with `map` and expanded `keys` as arguments
+
+* Otherwise, return boolean `true`
+
+### `set()`
+
+```
+set($map1, $keys..., $value)
```
nex3: Since all arguments can be called by name in Sass, it's usually a good idea to name them in ways that are consistent with other similar APIs. So the first argument here should probably be `$map` rather than `$map1`.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+
+* Let `map` be an empty map .
+
+* If there is more than one argument in arglist `$keys`:
+
+  * Let `set-key` be the first element in `$keys`.
+
+  * Let `keys` be a slice of all `$keys` elements except the first.
+
+  * If `$map1` has a key `current-key` with the same name as `set-key`:
```
nex3: It's not clear to me what "the same name" means here.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+
+* Add a key with name `set-key` and value `set-value` to `map`
+
+* For each `key`, `value` pair in `$map1`
+
+  * If a key with name `key` already exists in `map`, do nothing.
+
+  * Otherwise, add a key to `map` with name `key` and value `value`.
+
+* Return the value `map`
```
nex3: I think you can simplify this to say "Return a copy of `map` with `set-key` set to `set-value`, overriding an existing value for `set-key` if one exists."

proposal/nested-map-functions.md
nex3: Sass doesn't natively allow var args to be written before a non-variable argument, so this may require a little more machinery than usual to make that work. I don't think we need to make `map.set($map: ..., $value: ...)` work for a new function, but you may need to write the signature as `map.set($map, $args...)` and explicitly slice out the value from the argument list.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+
+### `merge()`
+
+```
+merge($map1, $keys..., $map2)
```
nex3: Similar comment to `map.set()`, with the caveat that we *do* need to continue to support `map.merge($map1: ..., $map2: ...)` for backwards-compatibility. You can get away with this by just saying that this is adding a new overload, like I did in https://github.com/sass/sass/blob/master/accepted/color-4-rgb-hsl.md.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Background
+
+> This section is non-normative.
+
+The current map inspection and manipulation functions don't provide any built-in
+support for managing nested maps. Projects often build thir own tooling, but
+the results are inconsistent, and often slow.
+
+## Summary
```
nex3, line 49: It's probably a good idea to mention which of these new behaviors will be accessible from the legacy global function names (presumably all updated behaviors for existing functions), and which will be restricted to the new module system (presumably only `map.set()`).

nex3 (changes requested):
This is looking really good! I'm excited to get these landed!

nex3:
Regarding `deep-merge()`, I think your proposed semantics make sense.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,322 @@
...
+* Otherwise:
+
+  * Let `map` be a copy of `$map1`.
+
+  * For each `key`, `value` in `map2`:
+
+    * Replace `map` with the result of calling `set()` with `map`, `key`, and
+      `value` as arguments.
+
+  * Return `map`.
```
mirisuzanne: Wasn't sure how much this current two-map behavior needs to be spelled out. I maybe landed on a strange middle-ground?

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,322 @@
...
+//   ),
+// );
+```
+
+## Functions
+
+### `get()`
+
+```
+get($map, $keys...)
```
mirisuzanne: - Is it strange that this new syntax (on several functions) uses an arglist of keys to represent _nested_ keys, while the existing `map.remove()` uses the same syntax to represent _multiple_ keys? 
- Is it strange that we have no nested-key `remove()` syntax?

mirisuzanne (commented):
@nex3 updated, and left a few comments/questions of my own.

(Along the way, I implemented them all on [codepen](https://codepen.io/mirisuzanne/pen/gOaWXqo?editors=0100) for testing. Feel free to check that out if it's useful.)

proposal/nested-map-functions.md
nex3: Yes and yes. I can see a few ways to improve this moving forward:

* Deprecate the existing `map.remove()` behavior, and after a suitable period replace it with new behavior. This represents a pretty painful breaking change, though—we usually don't change behavior in a way that makes old code do something *different* than it used to unless we absolutely have to.

* We could add a new function, `map.delete()`, which behaves consistently with the rest of the new map functions. After a period of time we'd deprecate `map.remove()`, and then eventually we'd remove it entirely (or rename it `map.remove-all()`?). The major downside of this strategy is that it's completely non-obvious to a casual observer what the difference between `map.delete()` and `map.remove()` is—they would *have* to read the documentation to understand.

* We could add a new function, `map.deep-remove()`, which behaves consistently with the rest of the new map functions, but plan to leave `map.remove()` as-is.

* We could do nothing at all, and have `map.remove()` just be an odd function out.

I'm not sure which of these options is best.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,323 @@
...
+
+* Let `map2` be the last item in arglist `$args`
+
+* If either `$map1` or `map2` is not a map, throw an error.
+
+* If arglist `$args` has more than one item:
+
+  * Let `get-keys` be a slice of all except the last item in `args`.
+
+  * Let `sub` be the result of calling `get()` with `sub` and expanded `get-keys`
```
nex3:
```suggestion
  * Let `sub` be the result of calling `get()` with `$map1` and expanded `get-keys`
```

proposal/nested-map-functions.md
nex3: I think this is reasonable, but I might be more explicit rather than relying on calling other functions. Something like:

> * For each `key`, `value` in `map2`:
>
>   * Set `key` to `value` in `map`, replacing an existing value for `key` if one exists.

Same with the call to `get()` above.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,323 @@
...
+
+    * Replace `map` with the result of calling `set()` with `map`, `key`, and
+      `value` as arguments.
+
+  * Return `map`.
+
+### `deep-merge()`
+
+```
+deep-merge($maps...)
```
nex3: Do you have a concrete use-case in mind for passing multiple maps here? If not, I feel like it might make the function more unclear. You could fairly easily call `map.deep-merge(map.deep-merge(...), ...)` if you aren't going to want to merge more than two maps *super* often.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,323 @@
...
+* For each `map` in `maps`:
+
+  * for each `key`, `value` in `map`:
+
+    * If `value` is a map:
+
+      * Let `current` be the result of calling `get()` with `merged` and `key` as
+        arguments.
+
+      * If `current` is a map:
```
nex3: What if `current` isn't a map? Don't you want to set `key` to `value` in `merged`?

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+  * If `child` is not a map, return `null`.
+
+  * If `child` contains a key that's `==` to `key`, set `child` to the value
+    associated with that key. Otherwise, return `null`.
+
+* Return `child`.
+
+### `has-key()`
+
+This proposal add a new overload to the existing `has-key()` function with lower
```
mirisuzanne:
```suggestion
This proposal adds a new overload to the existing `has-key()` function with lower
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+//     ),
+//   ),
+// );
+```
+
+## Functions
+
+### `get()`
+
+This proposal add a new overload to the existing `get()` function with lower
```
mirisuzanne:
```suggestion
This proposal adds a new overload to the existing `get()` function with lower
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+
+> This means that the new overload is only called if the existing signature
+> doesn't match.
+
+```
+has-key($map, $keys...)
+```
+
+> Intuitively, `has-key($map, $key1, $key2, $key3)` is equivalent to
+> `has-key(has-key(has-key($map, $key1), $key2), $key3)` with the exception that
```
mirisuzanne: I don't think this example actually works, since each `has-key()` returns a boolean, leaving nothing for the wrapping `has-key()` to check. More accurate maybe:

```suggestion
> `has-key(get(get($map, $key1), $key2), $key3)` with the exception that
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+  * [`deep-remove()`](#deep-remove)
+
+## Background
+
+> This section is non-normative.
+
+Variables have always been a key feature of the Sass language. But these days,
+design systems and component libraries form the basis of most CSS projects --
+with well organized _design tokens_ as the foundation. While Individual token
+variables can be quite useful, the ability to group tokens into strucutred and
```
Awjin:
```suggestion
variables can be quite useful, the ability to group tokens into structured and
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+meaningful relationships is essential for creating resilient systems.
+
+There are many ways to group tokens. The popular Style Dictionary recommends
```
Awjin: Might be worth linking https://amzn.github.io/style-dictionary/#/?id=style-dictionary

proposal/nested-map-functions.md
Awjin:
```suggestion
This proposal adds a new overload to the existing `get()` function with lower
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+  * Let `new-child` be the result of calling `set()` with `child` as the first
+    argument and the elements of `remaining` as the remaining arguments.
+
+  * Associate `key` with `new-child` in `map`.
+
+  * Return `map`.
+
+### `merge()`
+
+This proposal add a new overload to the existing `merge()` function with lower
```
Awjin:
```suggestion
This proposal adds a new overload to the existing `merge()` function with lower
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+  * Associate `$key` with `$value` in `map`.
+
+  * Return `map`.
+
+* ```
+  set($map, $args...)
+  ```
+
+  > Intuitively, `set($map, $key1, $key2, $key3)` is equivalent to
+  > `set(set(set($map, $key1), $key2), $key3)` with the exception that if any
```
mirisuzanne: I don't think this example works, because the returned map needs to become the set value rather than the key…

```suggestion
  > `set($map, $key1, set((), $key2, $key3))` with the exception that if any
```

(even that might be oversimplified… but it seems to work in the basic case)

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+of the existing tools rely on YAML or JSON objects to achieve that nested
+structure, at the expense of other important information. YAML and JSON are not
+design languages, and do not understand fundamental CSS concepts like color or
+length.
+
+With Sass, we don't have to make that tradeoff. We already support nestable map
+structures, and the ability to interact with them programmatically -- adding or
+removing properties, accessing values, and looping over entire structures. But
+current built-in functions don't provide much support for managing nested maps.
+Projects often build thir own tooling.
```
jathak:
```suggestion
Projects often build their own tooling.
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+```
+
+## Functions
+
+### `get()`
+
+This proposal add a new overload to the existing `get()` function with lower
+priority than the existing signature.
+
+> This means that the new overload is only called if the existing signature
```
jathak: Does this actually need to be an overload? Wouldn't `get($map, $key)` be covered by the multiple keys version?

proposal/nested-map-functions.md
mirisuzanne:
```suggestion
This proposal adds a new overload to the existing `merge()` function with lower
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+> doesn't match.
+
+This proposal add a new overload to the existing `merge()` function:
+
+```
+merge($map1, $args...)
+```
+
+> Intuitively, `map.merge($map1, $keys..., $map2)` is equivalent to
+> `map.set($map1, $keys..., map.merge(map.get($map1, $keys...), $map2)`.
```
mirisuzanne:
```suggestion
> `map.set($map1, $keys..., map.merge(map.get($map1, $keys...), $map2))`.
```

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,391 @@
...
+deep-remove($map, $keys...)
+```
+
+> Note: This is explicitly *not* an override of `remove()`, because `remove()`
+> already accepts a variable number of arguments as a way of removing multiple
+> keys from the same map. This proposal adds a new function rather than adjust
+> the existing behavior to avoid backwards-compatibility pain.
+
+> Intuitively, `map.deep-remove($map, $keys..., $last-key)` is equivalent to
+> `map.set($map, $keys..., map.remove(map.get($map, $keys...), $last-key)`.
```
mirisuzanne, line 368:
```suggestion
> `map.set($map, $keys..., map.remove(map.get($map, $keys...), $last-key))`.
```

mirisuzanne (commented):
Overall, I think this makes sense. Tried to correct a few of the non-normative notes…

proposal/nested-map-functions.md
nex3: That's a good point. Updated.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,393 @@
...
+  * If `child` is not a map, return `false`.
+
+  * If `child` contains a key that's `==` to `key`, set `child` to the value
+    associated with that key. Otherwise, return `false`.
+
+* Return `true`.
+
+### `set()`
+
+> Note: For consistency with other functions whose multi-key overloads were
```
jathak, line 232: Since `map.get` and `map.has-key` no longer use overloads, the overload here isn't really for consistency anymore.

Would it make sense to just integrate the base case into the main `set` definition?

nex3: It's consistent with `map.merge()`, which is the only other function that has a variable list of keys *before* a named final argument.

jathak: Oh, right. That makes sense then

jathak (approved):
LGTM

nex3:
I just pushed one more change, which drops support for variable arguments from `deep-merge()`. I'm concerned that it's confusing that variable arguments work differently for it than they do for the other map functions, and I haven't heard a concrete use-case for it, so I think it's better to be conservative and leave it off for now. We can always add it back later if we hear of a strong use-case.

Awjin (approved):

mirisuzanne:
> I just pushed one more change, which drops support for variable arguments from `deep-merge()`. I'm concerned that it's confusing that variable arguments work differently for it than they do for the other map functions, and I haven't heard a concrete use-case for it, so I think it's better to be conservative and leave it off for now. We can always add it back later if we hear of a strong use-case.

That makes sense to me. I do have use-cases where I merge multiple maps, but that's very reasonable to handle in a loop without any complexity. I don't have any strong need for it to be baked-into the core functions.

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,389 @@
...
+  set($map, $args...)
+  ```
+
+  > Intuitively, `set($map, $key1, $key2, $value)` is equivalent to `set($map,
+  > $key1, set(get($map, $key1), $key2, $value))` with the exception that if any
+  > intermediate value isn't set or isn't a map it's replaced with a map.
+
+  * If `$map` is not a map, throw an error.
+
+  * If `$args` has fewer than three elements, throw an error.
```
Awjin, line 262: Should this still work with two elements, since it would be equivalent to `map.set($map, $key, $value)`?

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,389 @@
...
+
+  * Let `sub-merged` be the result of calling `merge()` with `sub` and `map2` as
+    arguments.
+
+* Otherwise:
+
+  * Let `sub-merged` be `map2`.
+
+* Return the result of calling `set()` with `$map1` as the first argument,
+  followed by the contents of `keys` as separate arguments, followed by `sub`.
```
Awjin:
```suggestion
  followed by the contents of `keys` as separate arguments, followed by `sub-merged`.
```

Should this be sub-merged?

Awjin (changes requested):

proposal/nested-map-functions.md
```diff
@@ -0,0 +1,389 @@
...
+* If `$args` is empty, return `$map1`.
+
+* Let `map2` be the last element of `$args`.
+
+* If either `$map1` or `map2` is not a map, throw an error.
+
+* Let `keys` be a slice of all elements in `$args` except the last.
+
+* Let `sub` be the result of calling `get()` with `$map1` as the first
+  argument and the contents of `keys` as the remaining arguments.
```
Awjin, line 316: What happens if `keys` is empty?

proposal/nested-map-functions.md
nex3: Either way, it's never going to be called with two elements because the prior overload takes priority. Given that, I think it's a little clearer to make its behavior totally disjoint with that overload's.

proposal/nested-map-functions.md
nex3: That's not possible here, because that would end up calling the first overload. I'll add an explicit error for it, though.

proposal/nested-map-functions.md
nex3: Yep, good catch.

Awjin (approved):

babakfp:
Hello
I think there is a little problem with the current `map.deep-merge`.

Look at this code: [COMMENT](https://github.com/sass/sass/pull/2836#issuecomment-612647102).




## Expected

```scss
$result: deep-merge($map1, $map2);
@debug $result;
/*
01-  (
02-    a: 24,
03-    b: 6 7,
04-    c: wipe-out,
05-    e: (
06-      f: 20,
07-      g: 7,
08-      i: (
09-        a: 5,
10-        b: 2,
11-        c: 3
12-      ),
13-      h: 84
14-    )
15-  )
*/
```

If `$map1` key name was not the same as `$map2` key name, that key, and value going to added to the new map.
The first argument in the user-made deep-merge function, going to read first and other actions going to happen base on that first argument. If you check the $map1, $map2, and the result, you can see that the current order of rendering in the new map. In key `e` (from lines 5 to 14) the Order is `f g i h`.


#### Switching the arguments

```scss
$result: deep-merge($map2, $map1);
@debug $result;
/*
01-  (
02-    a: 12,
03-    b: 1 2 3 4,
04-    c: (d: 5),
05-    e: (
06-      f: 6,
07-      h: 84,
08-      i: (
09-        a: 1,
10-        c: 3,
11-        b: 2
12-      ),
13-      g: 7
14-    )
15-  )
*/
```

In key `e` (from lines 5 to 14) the Order is `f h i g`.




## What it is

In the current `map.deep-merge` the order is a little different.

```scss
$result: map.deep-merge($map1, $map2);
@debug $result;
/*
01-  (
02-    a: 24,
03-    b: 6 7,
04-    c: wipe-out,
05-    e: (
06-      f: 20,
07-      h: 84,
08-      i: (
09-        a: 5,
10-        c: 3,
11-        b: 2
12-      ),
13-      g: 7
14-    )
15-  )
*/
```

In key `e` (from lines 5 to 14) the Order is `f h i g`.


#### Switching the arguments

```scss
$result: map.deep-merge($map2, $map1);
@debug $result;
/*
01-  (
02-    a: 12,
03-    b: 1 2 3 4,
04-    c: (d: 5),
05-    e: (
06-      f: 6,
07-      g: 7,
08-      i: (
09-        a: 1,
10-        b: 2,
11-        c: 3
12-      ),
13-      h: 84
14-    )
15-  )
*/
```

In key `e` (from lines 5 to 14) the Order is `f g i h`.


```
sass --version
1.27.0 compiled with dart2js 2.10.1
```

Awjin:
Good catch @babakfp. `deep-merge` doesn't guarantee the order in which elements are outputted in the final map. I added an explanation to the spec that describes this behavior: https://github.com/sass/sass/pull/2970.

Issue #2619, opened 2019-02-08
@try/@catch operator


damienwebdev:
> Moved from https://github.com/sass/sass/issues/2596

# Proposal: 
Add a try/catch operator to sass to handle `@error` directives more cleanly without hard compiler crashes.

## Use Case
As a developer, I'd like to be able to write test code to cover my sass functions. I'm unable to test all branches of my code as my `node-sass` compiler crashes if it reaches an `@error` directive. While I am glad that the `@error` worked as intended and crashed while I was using the function, I'd like to be able to catch the exception during "test" mode and make an assertion about the message that was thrown.

See: https://github.com/oddbird/true/issues/92 for an example.

### Side comments
With `sass` as popular as it is, and new sass code being written every day, I find that the number of projects that I've worked on without tests for functional code to be high. Testing is necessary for maintenance and confidence in code. As sass is becoming a de-facto "language" for writing stylesheets, it seems that the `@error` operator necessitates a partnering try/catch directive. Even if `@error` is a weakly typed error, and try/catch doesn't have exception types, having the ability to catch an exception and handle it (if only for testing purposes) allows me to be confident that the code that I write is working the way I intend.

mirisuzanne:
@damienwebdev I'd be happy to start working on a proposal here. My first thought is something like…

```scss
@try {
  // potential error
} @catch <optional-variable-for-error-message> {
  // what to do if there's an error
}
```

So in the True use-case, you could do:

```scss
@try {
  @error "that doesn't work for me";
} @catch $error {
  @if $testing {
     // mixins could output a custom property here, rather than return…
     @return $error;
  } @else {
    @error $error;
  }
}
```

There's not a great way to set a variable like `$testing` globally, but it can be passed around as needed. On the other hand, you could write the try/catch logic inside the tests, rather than the code being tested, without any need for a global variable:

```scss
@try {
  $test: my-function($bad-data);
} @catch $error {
  @include assert-equal($error, "that doesn't work for me");
}
```

Going a bit deeper: I imagine `@warn` would also catch? In that case, we might need a second variable for the type of throw? That could either happen via destructuring:


```scss
@try {
  $test: my-function($bad-data);
} @catch $error, $type {
  @include assert-equal($type, 'error');
  @include assert-equal($error, "that doesn't work for me");
}
```

I think that's simpler than adding new data-types and doing `type-of($error)`? 

Am I missing anything else here? Would there be a strong use-case for adding `@finally`?

[...]

PR #2850, opened 2020-04-28
Begin drafting try/catch proposal

mirisuzanne:
Based on https://github.com/sass/sass/issues/2619

Rough syntax (with the unit-testing use-case):

```scss
@try {
  $test: my.function('bad arg');
} @catch $error {
  @include true.assert-equal(
    $error,
    "'bad arg' is not a valid argument for 'function'"
  );
}
```

- [ ] resolve questions about what to catch, and how to shape it
- [ ] more background & use-cases
- [ ] finish the semantics

### DIFF
--- a/proposal/try-catch.md
+++ b/proposal/try-catch.md
@@ -1,0 +1,92 @@
+# Support `@try` and `@catch` flow controls for error handling: Draft 1
+
+*[(Issue)](https://github.com/sass/sass/issues/2619)*
+
+## Table of Contents
+
+## Background
+
+> This section is non-normative.
+
+In many situations, particularly when writing tests, it's useful to catch error
+and warning messages without breaking compilation. By adding `@try` and `@catch`
+at-rules similar to other control-flow syntax, authors will have more control
+over error handling, and the ability to test for error-cases.
+
+Many large Sass projects currently use wrapper functions and mixins to catch and
+manage errors, with a parameter to toggle between throwing or returning the
+error message. But that relies on the special error handling to be built into
+every individual function or mixin that might error. By adding global controls,
+error-handling can happen either where the error is generated, or where it is
+defined.
+
+## Summary
+
+> This section is non-normative.
+
+This proposal defines a pair of new at-rule directives that must be used
+together as a form of control. The `@try { ... } @catch { ... }` syntax captures
+any thrown output (error or warning messages) from inside the `@try` block, and
+allows the parser to continue compilation. The `@catch` block executes whenever
+an error or warning has been captured, allowing authors to determine how the
+error should be processed.
+
+```scss
+@try {
+  $test: my.function('bad arg');
+} @catch $error {
+  @include true.assert-equal(
+    $error,
+    "'bad arg' is not a valid argument for 'function'"
+  );
+}
+```
+
+## Syntax
+
+<x><pre>
+**TryCatchRule**    ::= '@try' '{'
+&#32;                     TryStatements
+&#32;                   '}' '@catch' exceptionVar? '{'
+&#32;                     CatchStatements
+&#32;                   '}'
+**exceptionVar**    ::= '$' Identifier
+</pre></x>
+
+## Semantics
+
+To execute a `TryCatchRule` `rule`:
+
+* Let `exceptions`
+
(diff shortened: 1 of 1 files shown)


chriseppstein:
Hey everyone, I've not been involved lately, but I hope you don't mind me swooping in and giving my thoughts on this proposal. I've always wanted to see `@try` statements in Sass, so I got excited by this.

The idea of catching a `@warn` is an interesting one, but also might be unexpected because `@warn` doesn't alter the control flow. I'd be interested to hear the use case(s) for this.

I have some questions about how `@catch` and  `@warn` would interact:

1. Because execution continues following `@warn`, what would be the behavior if several `@warn` statements are encountered while executing `TryStatements`? I can imagine several possibilities: a) the first `@warn` statement is passed to `@catch`; b) The last `@warn` statement is passed to `@catch`; c) All `@warn` statements are passed to `@catch`. d) the `@catch` is executed immediately when the `@warn` fires and may be executed several times as if it were the body of a loop. If `c`, would the value of `message` be a list in the case a warning.
2. Given that execution continues for `@warn`, in the case of several nested `@try ... @catch` statements, does the `@warn` get passed to all of them or does it stop propagating unless "rewarned".

There's a use case around deciding that an error isn't the one you want to handle and rethrowing it. Given that an error message may have error specific values interpolated into the message that could be kind of tricky. What do you think about the idea of augmenting `@error` to have an identifier as well as a message? E.g. `@error illegal-argument "Expected a string, got #{meta.inspect($arg1)}";`. Then `@catch` could provide a discriminator so that only errors matching it are caught. E.g. `@catch illegal-argument ($message) { ... }`. Or the error object could have an extra key, etc.

When rethrowing an error, I think we'll want to preserve the original source location of the error. I think that might necessitate making a new data type.

If the main use case for using `@catch` with `@warn` is to silence a message do you think it makes sense to have a `@try { ... } @silence ($warning) { ... }` syntax where you can optionally inspect the warning and issue a new `@warn` if you want to print it out. In this case, should warn get a discriminator ident similar `@error`, seems like it would be handy for the same reasons.

proposal/try-catch.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Syntax
+
+<x><pre>
+**TryCatchRule**    ::= '@try' '{'
+&#32;                     TryStatements
+&#32;                   '}' '@catch' exceptionVar? '{'
+&#32;                     CatchStatements
+&#32;                   '}'
+**exceptionVar**    ::= '$' Identifier
```
nex3, line 53: Two nits:

* Production names should be capitalized (`ExceptionVar` not `exceptionVar`).

* Add a comment below explaining that no whitespace is allowed between `'$'` and `Identifier`. Or better yet, edit `variables.md` to have a `PlainVariable` that can be re-used without having to rewrite that caveat all over the place :laughing:.

proposal/try-catch.md
```diff
@@ -0,0 +1,92 @@
...
+To execute a `TryCatchRule` `rule`:
+
+* Let `exceptions`
+
+* Execute the contents `try` of `rule`'s `TryStatements`
+
+  * If an `@error` or `@warn` rule `throw` is encountered:
+
+    * Let `exception` be a map with the same identifier as `exceptionVar`,
+      a 'name' `throw`s at-rule
```
nex3, line 67: I'm having a lot of trouble understanding this sentence.

proposal/try-catch.md
```diff
@@ -0,0 +1,92 @@
...
+error-handling can happen either where the error is generated, or where it is
+defined.
+
+## Summary
+
+> This section is non-normative.
+
+This proposal defines a pair of new at-rule directives that must be used
+together as a form of control. The `@try { ... } @catch { ... }` syntax captures
+any thrown output (error or warning messages) from inside the `@try` block, and
```
nex3, line 29: I agree that catching warnings with this is unexpected. It's not how other programming languages work, and it may do very unexpected things with the control flow.

I'd like to punt on the question of how warnings are tested for now. It's much easier and safer to devise custom warning-capturing infrastructure, since you can just add strings to a list and you don't have to worry about globally coordinating return values.

nex3 (changes requested):

Issue #2860, opened 2020-05-14
Plain css clamp()


Awjin:
- [x] [Proposal](https://github.com/sass/sass/pull/2942)
- [x] [Specs](https://github.com/sass/sass-spec/issues/1599)
- [x] [Documentation](https://github.com/sass/sass-site/issues/505)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1146)

---

The plain CSS function [`clamp()`](https://www.w3.org/TR/css-values-4/#funcdef-clamp) has shipped in Chrome, Edge, and Firefox. This function introduces a `calc()`-style context where plain CSS math is supported—e.g. `clamp(1%, 1% + 1vw, 10%)`.

To support `clamp()`, Sass must parse it specially as a `calc()`-syntax function. Fortunately, because the Sass `clamp()` function is scoped to the `sass:math` module, it does not conflict with the CSS `clamp()`. Differentiating the two is straightforward.

Although this may be a breaking change, since this function is relatively new and has no namespace conflicts, I propose we fast-track this.

nex3:
I think because this technically does introduce a backwards incompatibility (a user could be writing `clamp($foo)`, which will need to be changed to `clamp(#{$foo})`, we'll unfortunately have to avoid the fast-track process.

nex3:
The proposal has landed. Since it's relatively small, I think we'll only need to wait two weeks for feedback (unless a debate needs to happen).

PR #2863, opened 2020-05-24
Draft proposal for css clamp() support

mirisuzanne:
This is a proposal for resolving #2860 by adding `clamp(` the the list of prefixes that generate a Special Number String.

### DIFF
--- a/proposal/css-clamp.md
+++ b/proposal/css-clamp.md
@@ -1,0 +1,76 @@
+# Plain CSS `clamp()`: Draft 1
+
+*([Issue](https://github.com/sass/sass/issues/2860))*
+
+This proposal defines how Sass handles CSS's `clamp()` [math function][], by
+updating the definition of a [special number string][special] in Sass.
+
+[math function]: https://drafts.csswg.org/css-values/#comp-func
+
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+* [Definitions](#definitions)
+  * [Special Number String](#special-number-string)
+
+## Background
+
+> This section is non-normative.
+
+[CSS Values and Units Module Level 4][values] defines a CSS `clamp()` function
+which accepts three calculations: a minimum value, a central value, and a
+maximum value. All three values are treated like independant `calc()` functions
+before applying the min and max "clamps" to the result:
+
+```css
+html {
+  font-size: clamp(1em, 1em + (2vw + 2vh) / 2, 2em + 1vmin);
+}
+```
+
+This plain CSS `clamp()` function has shipped in Chrome, Edge, and Firefox.
+However, the example above currently causes an error in Sass, because the units
+are incompatable. Sass attempts to ressolve any math expressions (including
+those inside the `clamp()` function) as part of compilation. That makes it
+impossible to use the CSS `clamp()` function as intended.
+
+In order to properly support the CSS `clamp()` function, Sass must treat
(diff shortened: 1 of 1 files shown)


mirisuzanne:
Maybe we  need to take this further, and treat any unrecognized function as a special number string that we don't attempt to resolve? That would be a more major backwards-incompatible change, but also potentially set us up for future CSS math functions? There are certainly more on the way, and I expect others to accept calculation values…

proposal/css-clamp.md
```diff
@@ -0,0 +1,76 @@
...
+* [Definitions](#definitions)
+  * [Special Number String](#special-number-string)
+
+## Background
+
+> This section is non-normative.
+
+[CSS Values and Units Module Level 4][values] defines a CSS `clamp()` function
+which accepts three calculations: a minimum value, a central value, and a
+maximum value. All three values are treated like independant `calc()` functions
```
connorskees, line 23: I believe you typo'd "independent" here

proposal/css-clamp.md
```diff
@@ -0,0 +1,76 @@
...
+
+```css
+html {
+  font-size: clamp(1em, 1em + (2vw + 2vh) / 2, 2em + 1vmin);
+}
+```
+
+This plain CSS `clamp()` function has shipped in Chrome, Edge, and Firefox.
+However, the example above currently causes an error in Sass, because the units
+are incompatable. Sass attempts to ressolve any math expressions (including
```
connorskees, line 34: typo of "incompatible" and "resolve"

proposal/css-clamp.md
```diff
@@ -0,0 +1,76 @@
...
+those inside the `clamp()` function) as part of compilation. That makes it
+impossible to use the CSS `clamp()` function as intended.
+
+In order to properly support the CSS `clamp()` function, Sass must treat
+`clamp()` as a [special number string][special], similar to `calc()`.
+
+[special]: ../spec/functions.md#special-number-string
+
+Sass already provides a complimentary `clamp()` function in the `sass:math`
```
connorskees, line 43: I believe you may have meant "complementary"

proposal/css-clamp.md
```diff
@@ -0,0 +1,76 @@
...
+
+The existing `clamp()` function inside the `sass:math` module will not be
+affected by this change.
+
+## Definitions
+
+> This proposal updates the definition of a "special number string" to include
+> `clamp(` as a prefix. The new defenition reads as follows:
+
+### Special Number String
```
nex3, line 67: The definition of "special number string" is really just used for functions that might get number-like objects as inputs. It *does* need to be updated for sure, but the special parsing of these functions is the more important thing, and that probably should be specified in the Syntax section of `functions.md`.

nex3 (changes requested):
> Maybe we need to take this further, and treat any unrecognized function as a special number string that we don't attempt to resolve? That would be a more major backwards-incompatible change, but also potentially set us up for future CSS math functions? There are certainly more on the way, and I expect others to accept calculation values…

I think that's way too big a step. It'll break anyone passing raw Sass variables to plain CSS functions, which is likely to be a *lot* of code. I think it's fine to just update the set of specially-parsed function names as they land in browsers.

proposal/css-clamp.md
kvanbere: There's a typo in the second 'defenition' here.

Issue #2745, opened 2019-09-17
Provide a way to share state across multiple compilations


znck:
When using sass libraries with react/vue components and building with webpack. Every imported file (in the component) is processed individually which requires the shared libraries to reload for every file. 

Ideally, JavaScript API can create ImportCache/AsyncImportCache by importing some shared files and libraries and pass this cache to every render/compile call.

nex3:
This is something I brought up as a possibility as part of #2737. I'm retargeting it to more generally refer to a way of sharing state, because we'd need to expose this as part of the JS API that's shared between Dart Sass and Node Sass, and the latter doesn't have an identical notion of an `ImportCache`.

The reason we don't *automatically* share state is that there's always a possibility that, between different calls to `render()`/`renderSync()`, any given file has changed on the filesystem and so anything cached has become invalid. However, in a situation like webpack compiling many separate entrypoints that share a considerable amount of libraries, the client knows (or assumes) that the files won't change between those compilations.

Rather than exposing import caches to the JS API, I think it makes more sense to focus on expressing the guarantees that the user is providing the Sass implementation. For example, we could have a `Compiler` class that has `render()` and `renderSync()` as *methods*, and that explicitly assumes that the filesystem doesn't change during the lifetime of the `Compiler` instance.

We could even eventually extend this to provide the `Compiler` with notifications when files *do* change, so that it could invalidate only the cached data that's actually affected. This is essentially the model that we used in Ruby Sass as well.

jquense:
I'd be interested in helping out with this, if there is a reasonably clear path forward on what the feature would look like. This issue is near and dear to me, all our projects use css-modules with sass and it results in hundreds of individually compiled sass files  with most referencing the same one or two "theme" files

nex3:
Awesome! We haven't really had the opportunity to create a new shared JS API, so there's not a concrete process for it yet. Since it's a design that's shared across implementations, though, it probably makes sense to start with [a formal proposal](https://github.com/sass/sass/blob/master/CONTRIBUTING.md#proposal) in a similar style to a language proposal so all the stakeholders have a chance to chime in. Can you write something like that, but adapted for a JS API addition rather than a Sass language feature?

jquense:
Yes! I can try and put something together, Sort of a design doc for the proposed API? No timeline on when i can get it done but adding it to my agenda :) Thanks for the help!

nex3:
> Yes! I can try and put something together, Sort of a design doc for the proposed API?

[...]

PR #2873, opened 2020-06-16
Add public compiler api proposal

jquense:
Finally got a round to writing something up for #2745 Let me know if this format seems good and sufficient, there wasn't an example I could find doing this sort of non-language proposal

[rendered](https://github.com/jquense/sass/blob/bcfd7fdc533e14f1df6ec4cef9fbac729999af56/proposal/public-compiler-api.md)

Should also note that I am doing a bunch of hand waving for the `module`/`import` cache distinction. I've not delved into how compilers structure their module graphs internally so i'm just assuming there is some sort of cacheable `module` object

### DIFF
--- a/proposal/public-compiler-api.md
+++ b/proposal/public-compiler-api.md
@@ -1,0 +1,172 @@
+# Public Compiler API
+
+## Background
+
+> This section is non-normative.
+
+Sass compiler implementations tend to assume projects will have few entrypoint Sass files
+such that running the compiler against each without a cache is "fast enough".
+However, for projects with substantially more entrypoints, ones that depend
+on a bundler (like webpack) to do final concatenation, or using [css-modules][], there can be a significant cost to reparsing and processing unchanged files across the project.
+
+In common case of projects with many entrypoints, each entry likely imports from a
+large set of shared configuration, funtions and mixins. These dependencies often change infrequently
+and a significant amount of compile time is spent compiling the same files in the
+context of a new entrypoint file.
+
+In addition, without an ability to reuse previous work, certain project patterns with Sass's
+module system are also impossible. Consider this case using webpack:
+
+**App.js**
+
+```js
+import 'styles.scss'
+```
+
+**styles.scss**;
+
+```scss
+@use 'bootstrap' with (
+  $btn-color: red
+)
+
+html {
+  font-size: 10px;
+}
+```
+
+Further down in the app, a component may wish to reference the previously configured
+library
+
+**button.scss**
+
+```scss
+@use 'bootstrap/buttons' as buttons
+
+.my-btn {
+  color: buttons.$btn-color;
+}
+```
+
+And the js file using webpack's `import` syntax for styles:
+
+**Button.js**
+
+```js
+import 'button.scss'
+
+// some web, vue, react, etc component
+```
+
+If the user is used module caching behavior like in Node.js, they would be suprised
+that `$btn-color` is its default not `red`.
+
+[css-modules]: https://github.com/css-modules/css-modules
+
+## Summary
+
+Sass implementations should provide a constructable Compiler class whose instance exposes `render` and `renderSync` methods for compiling Scss. Method calls providing a `file` have their results (as well as the results of and dependencies) cached by absolute filename on the compiler instance. Further calls to `render(Sync)` will return the cached value if not invalidated. Calls providing `data` would only have dependencies cached. Calls providing **both** `data` and `file`
+would process `data` using the resolved absolute path of `file` as it's cache key.
+
+The compiler instance would also expose an `invalidate` method which marks a file
+and all of it's dependents as needing recompilation.
+
+## Design choices
+
+> This section is non-normative.
+
+It's important that the compiler API allow delegating cache control to the consumer.
+Bundlers like webpack, or Parcel provide their own mechanisms for caching, watching, invalidating and topographically rebuilding entrypoints based on file changes. The Sass API
+should allow integration into those mechanisms.
+
+This does not preclude the possibility of Sass implementations providing their own
+watching mechanisms. This feature could be built on top of the proposed API
+
+## API
+
+```ts
+interface CompilerClass {
+  new (): Compiler
+}
+
+interface Compiler {
+  render(options: { data?: string, file?: string, ... }): Promise<string>
+
+  renderSync(options: { data?: string, file?: string, ... }): string
+
+  invalidate(file: string): void
+}
+
+```
+
+## Semantics
+
+- When `compiler.render(Sync)` is called
+
+  - Let `import-cache` be a `Map<string, string>` on `compiler`
+
+  - Let `module-cache` be a `Map<string, module>` on `compiler`
+
+  - If options provides only `file`
+
+    - Let `data` be the contents of `file`
+
+  - If `file` is set
+
+    - Let `filepath` be the absolute path of `file`
+
+  - If `filepath` and `import-cache` has key `filepath`
+
+    - Return string value for key `filepath`
+
+  - Overwise compile `data`
+
+    - For each `@import` rule
+
+      - Let `filepath` be the absolute path of the import
+
+      - If `import-cache` has key `filepath`
+
+        - Use string value as the compilation result
+
+      - Otherwise compile `filepath` using the same and set key `filepath`
+        on `import-cache` to the result
+
+    - For each `@use` rule
+
+      - Let `filepath` be the absolute path of the import
+
+      - If `module-cache` has key `filepath`
+
+        - Use `module` to resolve local references
+
+      - Otherwise compile using the same logic and set key `filepath` on
+        `module-cache` to the result
+
+    - Let `result` be the string value of the compilation
+
+    - If `filepath`
+
+      - Let `members` be the `module` value of `data`
+
+      - Set key `filepath` on `import-cache` to `result`
+
+      - Set key `filepath` on `module-cache` to `members`
+
+    - Return `result`
+
+* When `compiler.invalidate(file)` is called
+
+  - Let `filepath` be the absolute path the `invalidate` argument `file`
+
+  - Let `dependents` be a list of the file paths of all direct and transitive **dependents** of `file`
+
+  - Delete `filepath` from `import-cache`
+
+  - Delete `filepath` from `module-cache`
+
+  - For each `dependent-filepath` in `dependents`
+
+    - Remove `dependent-filepath` from `import-cache`
+
+    - Remove `dependent-filepath` from `module-cache`


jquense:
cc @evilebottnawi and @devongovett maintainers of sass-loader and parcel respectively, since this API needs to actually work in the context of a bundler

devongovett:
This is interesting. One thing to consider, when providing a stateful API like this, is that it won't work well with tools that parallelize. For example, Parcel runs all compilation in worker threads, and we don't guarantee that recompilations happen on the same thread. This means that recompilations won't have access to data stored in memory from previous compilations. For that reason, our plugins generally need to be stateless.

In addition, I assume this would store a cache in memory internal to the SASS compiler instance? That wouldn't cover the case where you restart your bundler. Parcel (and soon webpack) supports a disk cache as well. It would be great if we could integrate that with SASS too, and provide the results of cached builds to SASS rather than the other way around.

Without knowing much about the internals of the SASS compiler, perhaps one way to solve both of these issues would be to split it into two parts: one that performs compilation to some intermediary representation for an individual file and returns dependencies that the bundler could resolve, and a second part that takes these individual pieces and combines them together into a single output file. This way, compilation of individual files could be parallelized and cached independently and would be stateless, and the combining files together could happen at the end. This is how we process JavaScript in Parcel for example: babel runs over each file individually, and at the end we combine the outputs together into the final bundle.

jquense:
> one that performs compilation to some intermediary representation for an individual file and returns dependencies that the bundler could resolve, and a second part that takes these individual pieces and combines them together into a single output file

I don't think this is feasible as the Sass compiler is currently structured, I believe it does a single pass collecting dependencies and resolving as it goes, but I could be completely wrong. 

I think there is a larger issue tho (or i could just be misunderstanding). I'm guessing Parcel runs babel over all JS files, then parses imports and constructs a graph for concating the files together? Sass would have to be the inverse, files can't be compiled until the members of the files dependencies are resolved. That is to say Sass is a language runtime as well as a CSS compiler. 

Overall I think you could do do something like parsing each file individually for dependencies. There isn't much compilation tho that can happen at that stage and the compiler would still need to share results across the entire compilation. I guess your saying that Parcel does have hooks for processing the graph post-compilation, but you'd have to collect that statelessly?


>  Parcel (and soon webpack) supports a disk cache as well. It would be great if we could integrate that with SASS too, and provide the results of cached builds to SASS rather than the other way around.

Yeah totally, this doesn't cover a persistent cache, I don't want to preclude that, so it can be an additive change later.

devongovett:
> I don't think this is feasible as the Sass compiler is currently structured, I believe it does a single pass collecting dependencies and resolving as it goes, but I could be completely wrong.

I assumed you'd need to do some of this regardless of how this is implemented right? You made reference to caching the compilation output for a file in memory in this proposal (i.e. `import-cache` and `module-cache`). Did I misunderstand or would these not be the individual file representation I was referring to?

>  I'm guessing Parcel runs babel over all JS files, then parses imports and constructs a graph for concating the files together? 

This is true, but it's a bit more complicated than that. In order to get tree shaking, we combine the individual files together using ASTs, and links all of the imported and exported variable together. This is likely similar to what SASS does internally. The final compilation would happen at this stage. Whatever work can be done at the individual file level (parsing, some transformations that don't rely on external variables) could be done in parallel and cached.

alexander-akait:
Honestly, I have a strong opinion on this issue. 

Just a few thoughts:
- should work in multi-threaded mode
- there are a lot of places to optimize (for example https://github.com/sass/dart-sass/blob/master/lib/src/importer/utils.dart#L45, we can cache filesystem calls and resolution map), so maybe most of problems with perf can be solved without caching
- should be stateless

jquense:
For transparency, my stake in has less to do with performance optimization than enabling "first class" module support (for lack of a better term), at the moment using sass modules is awkward with many entrypoints, and is functionally not much better than `@import` since you need to import all setup and configuration code in every entry. 

That said, @evilebottnawi i'd love if you could expand a bit! Unlike Parcel I do have a good working knowledge of webpack loaders and the API proposed seems like it'd be _easier_ to use in webpack by attaching the sass compiler to the webpack compiler instance. AFAIK there is no way in webpack to do the Parcel process of emitting AST representations and then combining them later (if there was wouldn't most of css-loader be unnecessary)? Also what cases is webpack even operating in a multi threaded environment? happypack?

I do think the idea of a stateless version of this would be more ideal, but it's significantly more complicated and seems like it'd involve a heavy compiler refactor, whereas the proposed approach (I think) is mostly exposing methods that already exist. e.g.

> we can cache filesystem calls and resolution map

This is already happening _inside_ calls to `render` but the cache is thrown away immediately, the proposal here offers a minimal API for reusing those caches across calls. 

It seems like suggestions for a stateless approach, at least as applies to sass, would amount to doing parsing in threads and then all evaluation and emitting later in the same thread/memory space. Functionally that isn't really any different than this as proposed except that this work _currently_ happens in a spot in bundlers designed for single file compilation. I am not at all clear on whether parallelizing parsing would be a performance win, but it seems orthogonal to the problem we are trying to solve here.

jquense:
I should leave space here for Sass folks to chime in before i dig myself into a hole trying to explain and describe mechanisms i'm not very familiar with!

nex3:
Thanks for the proposal, @jquense, and thanks for the feedback @devongovett and @evilebottnawi!

All this talk about stateless/parallelizable/incremental compilation is super interesting, but I think @jquense is right on several points:

* Enabling incremental compilation would require changes that pierce deeply into the compiler internals.
* It's not immediately clear that those changes would actually pay for themselves in terms of performance. Only parsing changes can obviously be persisted across runs, and it's not obvious that parsing is a bottleneck or that parsing a serialized AST rather than source would fix that bottleneck. Dart Sass's parser is already pretty fast.
* The idea of incremental compilation is ultimately more or less orthogonal to the use-case this proposal is targeting.

The goal here (at least as far as Parcel/Webpack are concerned) isn't really to cache progress across multiple source changes. That may be valuable, but it's a project for another day. The goal is to **share information across multiple compilations that operate on the same filesystem state**. That is, when you run the initial compilation or change some shared library later on, to share information across all entrypoints that need to be recompiled in that instant.

The hope then is that this constraint—only needing to share information across a single band of compilations—will help make it more feasible to integrate. Think about it as though you have an API like this:

```ts
sass.compileMany(inputPaths: string[]): {[path: string]: CompileResult};
```

If you pass all the Sass entrypoints to this API to compile at once, it'll go faster than if you compile them all separately. Does that help?

proposal/public-compiler-api.md
```diff
@@ -0,0 +1,172 @@
...
+However, for projects with substantially more entrypoints, ones that depend
+on a bundler (like webpack) to do final concatenation, or using [css-modules][], there can be a significant cost to reparsing and processing unchanged files across the project.
+
+In common case of projects with many entrypoints, each entry likely imports from a
+large set of shared configuration, funtions and mixins. These dependencies often change infrequently
+and a significant amount of compile time is spent compiling the same files in the
+context of a new entrypoint file.
+
+In addition, without an ability to reuse previous work, certain project patterns with Sass's
+module system are also impossible. Consider this case using webpack:
```
nex3, line 18: I'm opposed to supporting this use-case. The purpose of the module system is to make the dependencies between Sass files explicit, so you can tell exactly how a file will be interpreted by looking only at that file and the files it depends on. In your example, `button.scss` implicitly depends on the configuration defined in `styles.scss`, but there's no explicit indication of that fact in the file.

proposal/public-compiler-api.md
```diff
@@ -0,0 +1,172 @@
...
+would process `data` using the resolved absolute path of `file` as it's cache key.
+
+The compiler instance would also expose an `invalidate` method which marks a file
+and all of it's dependents as needing recompilation.
+
+## Design choices
+
+> This section is non-normative.
+
+It's important that the compiler API allow delegating cache control to the consumer.
```
nex3, line 78: I'm not totally clear on what you mean by "delegating cache control". Does that just refer to the `invalidate()` method?

proposal/public-compiler-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+```ts
+interface CompilerClass {
+  new (): Compiler
+}
+
+interface Compiler {
+  render(options: { data?: string, file?: string, ... }): Promise<string>
+
+  renderSync(options: { data?: string, file?: string, ... }): string
```
nex3, line 95: One complication here is that the `render()` and `renderSync()` functions currently supported by the JS API are effectively legacy APIs. They were inherited from Node Sass and kind of grew organically over time, but they're not really the APIs we want in the long run. @Awjin is currently working on sketching out a next-generation API.

This isn't something you need to worry about in the short term (it's fine to just say "have copies of whatever global render APIs exist with some modifications"), but before we actually release this we should figure out a plan for how to avoid building more dependencies on the legacy APIs.

proposal/public-compiler-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+```
+
+## Semantics
+
+- When `compiler.render(Sync)` is called
+
+  - Let `import-cache` be a `Map<string, string>` on `compiler`
+
+  - Let `module-cache` be a `Map<string, module>` on `compiler`
```
nex3, line 108: This confusing... it implies that these caches are only created when `render()` is called, and doesn't make clear what happens when `render()` is called multiple times.

I'd also probably avoid explicitly writing TS-style type declarations for purely internal behaviors/objects. The goal is to describe behavior, not prescribe implementation.

proposal/public-compiler-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+    - Return string value for key `filepath`
+
+  - Overwise compile `data`
+
+    - For each `@import` rule
+
+      - Let `filepath` be the absolute path of the import
+
+      - If `import-cache` has key `filepath`
```
nex3, line 128: The details of the mapping need to be fleshed out considerably here. The implication right now is that these maps have plain string keys, but that doesn't handle a number of important use-cases:

* Paths that are normalization-equivalent (`/foo/bar` versus `/foo/./bar`).
* Non-path URLs (such as for custom importers).
* Distinct paths that have the same canonical URL.

nex3 (changes requested):
Thanks for diving into this, Jason! This will be an important API going forward, and I'm really glad you're pitching in to help move it forward.

proposal/public-compiler-api.md
jquense: I also want to maintain the explicit nature of the new module system! I think whether or not this use-case fits that depends on the context and framing. If you consider each Sass entrypoint a separate "program", sharing module state like this doesn't make sense and is confusing. 

In the context of bundlers though, you aren't usually writing entry-points as distinct programs but as branches in a single "App" program. `button.scss` is functionally a "partial" in this example, e.g. I don't intend it to be it's own css file and know the bundler will combine it into the final asset. 

Of course one could write these as actual partials and have a single or limited number of hand tuned entry files  but that's not organizationally great when you have hundreds of partials in the app. It'd be better (I think) if to let the bundler app bundler create the ideal top level entrypoints based on the overall dependency graph. Ideally this enables smaller initial css downloads, and optimized outputs for above/below the fold.

If you think the above distinction makes sense, I'd argue that the proposed feature isn't all that useful for the first case. Sass is already fast enough for cases where you have a limited number of top level entrypoints. I'd suggest that the need for this feature really only arises when writing apps in the context of a bundler assuming the second framing. e.g. most every import of a style file is a partial. 

My suggestion is that the example outlined is fundamentally the same as this: 

```styles.scss
@use 'bootstrap' with (
  $btn-color: red
)
@use 'button.scss'
```
_button.scss
```
@use 'bootstrap/buttons' as buttons

.btn {...}
```

Except that to trace the interactions you need to jump through JS and Sass files, but it is no less explicit (albeit a much larger) graph. 

OK Philosophizing aside, I do think that this approach offers actual benefits:

**It removes a the need for large top level "god" files that reexport all variables, functions and mixins**

Even with the new module system you still end up with a giant "theme.scss" files that re-export everything since you need to do `@use .. with` overrides for every entrypoint, making it less useful to be able to import just the bits you need from submodules.

**It's easily to structure libraries that can be configured without duplicating CSS in the output**

As with using `@import` providing customization of libraries requires manually split out the files that produce css vs contain variables and other members. I think it'd be awesome if one could just `@use 'bootstrap' with (...)` at the app root to configure and include the bootstrap css and then be able to import the mixins and functions I need in the app without needing to manually split the `bootstrap.scss` into configuration with override file and css producing partials. (This could also be addressed by having something like Less's `@import (reference)` maybe?)

I have more thoughts and am **rambling** but it's Saturday and I should step away for now, sorry for the wall of text, I'm glad to have a chance to try and articulate why I think this is important for Sass in the context of all the crazy tooling we have in the web dev space now.

proposal/public-compiler-api.md
jquense: Yes this is definitely hand wavey. I was thinking tho at some point urls are resolved to actual file paths and that could be used as the cache key. I had not considered the custom importer case :/ How do the existing caches work in this regard?

proposal/public-compiler-api.md
jquense: yeah i guess so. I was trying to articulate that the Compiler should not be concerned with invalidating it's own cache by e.g. checking mtime's on files. It _could_ do that in a separate watch mode, but core api's should cache aggressively and require manual invalidating

proposal/public-compiler-api.md
nex3: There are a lot of different thoughts in this comment and I'm not going to try to address all of them, but at a high level, I don't think there's a compelling case for adding implicit dependencies between entrypoints. Even your example of configuring `bootstrap` and using `button` in the same file is something I would flag as unclear as a code reviewer, so doubling down on that pattern is not something I want to do.

I'll try to address the two main benefits you bring up:

> It removes a the need for large top level "god" files that reexport all variables, functions and mixins

You don't necessarily need to re-export anything, you just need to be explicit about your dependencies. In this case, you're depending on your app's local configuration, so you can write:

```scss
// button.scss
@use 'myapp/configuration'; // configures bootstrap
@use 'bootstrap/buttons';
```

This is both simple and explicit.

> It's easily to structure libraries that can be configured without duplicating CSS in the output

It's pretty easy to write a library that doesn't duplicate CSS for simple cases: just put all the CSS generation in mixins that downstream clients can choose to invoke or not. Beyond that, doing this at scale requires a proper code-splitting solution which is an explicit non-goal of the Sass module system.

proposal/public-compiler-api.md
nex3: Agreed. Can you update the document to be more explicit about this?

proposal/public-compiler-api.md
nex3: As far as the core Sass compiler is concerned, URLs are actually *never* resolved to file paths. That's entirely the domain of importers. It just happens to be the case that the default importer is one that handles `file:` URLs.

I recommend taking a look at [Dart Sass's `Importer` API](https://pub.dev/documentation/sass/latest/sass/Importer-class.html) for an idea of how importers work under the covers. The short answer is that all files have a single canonical URL which has a one-to-one relationship with file context, and that importers convert URLs as written in stylesheets into fully canonical URLs.

proposal/public-compiler-api.md
jquense: > your example of configuring bootstrap and using button in the same file is something I would flag as unclear as a code reviewer, so doubling down on that pattern is not something I want to do.

Interesting, If you don't mind how you write or structure that example then? It's seems like consumers of configured modules are always going to have no knowledge of whether they were configured or not?

> @use 'myapp/configuration'; // configures bootstrap
> @use 'bootstrap/buttons';

Definitely, you don't need to reexport everything but the cost of not doing that, is needing to remember to configure everything before accessing it in many, many files, which is error prone.

>  Beyond that, doing this at scale requires a proper code-splitting solution which is an explicit non-goal of the Sass module system.

I think we agree on that, but without an API to control the module configuration cache it's not possible for other tools (like bundlers) to implement such a system.

>  just put all the CSS generation in mixins that downstream clients can choose to invoke or not.

That is a cool approach I had not thought of! I've mostly only seen projects structured like bootstrap, which definitely does not follow that pattern


Use-case aside, I'm not actually sure how to move forward here if module configuration needs to be "reset" between entry points. It seems like there is not a lot of reusable work unless there is an easy way cache based on the configuration? I will admit that i do not know how module configuration state is managed tho, it seems like it'd be harder to know whether `with` was invoked too many times in the `compileMany` case?

proposal/public-compiler-api.md
jquense: That makes sense, so `canonicalize` would return the correct key it seems like? And if i'm reading the import cache code correctly `Importer.canonicalize` will return an empty value if it can't handle the import so it should is it enough to have the logic be something like "canonicalize the url using the first importer that returns a valid value"

proposal/public-compiler-api.md
nex3: > Interesting, If you don't mind how you write or structure that example then? It's seems like consumers of configured modules are always going to have no knowledge of whether they were configured or not?

I'd structure it like the example I gave above: explicitly using a configuration file in each entrypoint.

> Definitely, you don't need to reexport everything but the cost of not doing that, is needing to remember to configure everything before accessing it in many, many files, which is error prone.

When you're dealing with side-effects (which `@use ... with` is), they're *always* going to be error-prone. I don't think it's less error-prone for being implicit; you're just shuffling around where the errors are likely to happen.

For example, in your proposal as it stands, the behavior of a given entrypoint is highly dependent on the order in which entrypoints are compiled (which is left up to the caller). If a caller decides to compile `button.scss` before `styles.scss` then the configuration won't be available in `button.scss` for reasons that may be unclear to the author. If a caller compiles stylesheets in parallel, there may be a race condition to determine whether or not the configuration is available.

It gets even worse if some entrypoint changes the state of a shared module in some way other than `@use ... with`. If `button.scss` sets `bootstrap.$base-color: #000;` then entrypoints that run after `button.scss` will use a different color scheme than any that run before it.

> I think we agree on that, but without an API to control the module configuration cache it's not possible for other tools (like bundlers) to implement such a system.

I don't follow this logic.

> Use-case aside, I'm not actually sure how to move forward here if module configuration needs to be "reset" between entry points. It seems like there is not a lot of reusable work unless there is an easy way cache based on the configuration? I will admit that i do not know how module configuration state is managed tho, it seems like it'd be harder to know whether `with` was invoked too many times in the `compileMany` case?

I think the question of what work the compiler can re-use can be separated to a degree from the question of which guarantees the compiler API offers to its users. Defining the latter is the primary goal of this design, which will then allow the compiler to freely optimize within the constraints of those guarantees.

There are certainly some valuable optimizations we could do here, such as re-using the ASTs of files that haven't changed between compilations and even re-using the loaded modules between compilations under certain constraints.

proposal/public-compiler-api.md
nex3: Yeah, that would work. You could even gloss over the entire importer infrastructure if you wanted and just refer to the map keys as "canonical URLs".

proposal/public-compiler-api.md
jquense: > I think the question of what work the compiler can re-use can be separated to a degree from the question of which guarantees the compiler API offers to its users.

Sorry i'm not sure I understand this distinction. Isn't the question of what is resusable dependent on what requires compiling? The latter doesn't seem knowable unless the question of whether a module needs configuration applied. e,g. the result of module A is different depending on how it was imported, and seems like those results need to be cached separately, Or hand wave past it as "reevaluate as necessary" or something?

> For example, in your proposal as it stands, the behavior of a given entrypoint is highly dependent on the order in which entrypoints are compiled (which is left up to the caller)

Yes i'd consider that a feature not a bug. It'd absolutely error prone when used incorrectly, tho that's often the case with low level APIs i think. TBF i've been envisoning this API as hook for more complicated use cases and specifically intended to be used by other tools, not directly by users to compile Sass. A core assumption of mine is that is those who would benefit from this API are not using the Sass compiler directly anyway. I'm not doing a good job of championing or explaining the potential benefit here in the context of other tools, but that is were I'm coming from, Sass as part of a larger application graph.

lolmaus:
> @jquense jquense closed this 1 hour ago

Why?!

We're [trying](https://github.com/embroider-build/embroider/issues/1033) to integrate Sass into our frontend framework as a first-class citizen (e. g. have one Sass `@use` module for every frontend component).

Without a public compiler API this is impossible. 😒

jquense:
> Why?!

I don't have the time to walk this any further sorry!

lolmaus:
Oh, you're the draft RFC author! Thank you so much for your effort!! 🙇‍♂️

Can you please convert this to an issue, so that the initiative does not die? 🙏

PR #2902, opened 2020-09-05
Better define how numbers work in Sass

### LINKED ISSUE #2892: Use floating-point numbers universally
- [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/floating-point.md) (https://github.com/sass/sass/pull/3378)
- [x] [Tests](https://github.com/sass/sass-spec/issues/1819) (https://github.com/sass/sass-spec/pull/1826)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1794) (https://github.com/sass/dart-sass/pull/1802)
- [x] Update canonical spec (https://github.com/sass/sass/pull/3408)

---

Currently the to and from values for `@for` are unbounded and not formally defined.

This input is considered valid Sass,
```scss
@for $i from -999999999999999999999 to 999999999999999999999999 { ... }
``` 

In order to simplify implementations, I would like to propose the bounding of these values.

I believe the easiest solution would be to use a 32 bit signed integer in order to represent these bounds. This will provide a reasonable range of values without requiring much work on the part of an implementation.

In order to better avoid overflows, I would like to propose that the max and min values be actually defined as `i32::MAX - 1` and `i32::MIN + 1` respectively. 

More concretely, the maximum value would be `2,147,483,646` and the minimum value would be `-2,147,483,647`.

It seems that `dart-sass` today silently overflows for the to and from values at some point. Bounding these in this way would remove this issue altogether.

connorskees:
Resolves #2892

This serves to better define the edge cases regarding numbers in Sass and formally moves all numbers to be represented as 64-bit floating point real numbers. 

This is largely a description of how numbers currently work in `dart-sass`; however, this document suggests denying `NaN`, `Infinity`, and `-Infinity` as map keys. These values are currently de facto denied as map keys in `dart-sass` due to [a crash](https://github.com/sass/dart-sass/issues/1059), but `libsass` does allow them.

Interestingly, `libsass` will also treat `NaN` and `Infinity` as equal to themselves, causing `(1/0) == (2/0)` to be `true` and making it possible to use `NaN` to retrieve values from maps. Irrespective of this document, this behavior is considered a bug (or at the very least a deviation from `dart-sass`). This behavior is tracked in https://github.com/sass/libsass/issues/3126. 

Units are deliberately left out of this document, though could certainly be taken into consideration.

### DIFF
--- a/proposal/numbers.md
+++ b/proposal/numbers.md
@@ -56,8 +56,27 @@ if `a == b`, then `b == a` and if `a == b` and `b == c`, then `a == c`.
 Numbers in Sass are *not* necessarily reflexively equal. `NaN == NaN`,
 `Infinity == Infinity`, and `-Infinity == -Infinity` will always be false.
 
-In the case that two numbers use more than 10 decimal places of precision,
-both numbers are truncated to 10 decimal places and compared.
+In the case that two numbers differ only in their 11th or later digits after
+the decimal point, both numbers are truncated to 10 decimal places and compared.
+
+For example,
+
+```
+These two numbers differ *before* the 10th decimal place, so they should be considered
+unequal
+1.0000000010 == 1.0000000020
+  ---------^      ---------^
+
+These two numbers differ *at* the 10th decimal place, so they should be considered
+unequal
+1.0000000001 == 1.0000000002
+  ---------^      ---------^
+
+These two numbers differ only *beyond* the 10th decimal place, so they should be considered
+equal
+1.00000000001 == 1.00000000002
+  ---------^       ---------^
+```
 
 ### Ordering
 
--- a/proposal/numbers.md
+++ b/proposal/numbers.md
@@ -56,8 +56,7 @@ if `a == b`, then `b == a` and if `a == b` and `b == c`, then `a == c`.
 Numbers in Sass are *not* necessarily reflexively equal. `NaN == NaN`,
 `Infinity == Infinity`, and `-Infinity == -Infinity` will always be false.
 
-In the case that two numbers differ only in their 11th or later digits after
-the decimal point, both numbers are rounded to 10 decimal places and compared.
+Two numbers are considered equal if they do no differ when rounded to 10 decimal places.
 
 > For example,
 > 
--- a/proposal/numbers.md
+++ b/proposal/numbers.md
@@ -57,7 +57,7 @@ Numbers in Sass are *not* necessarily reflexively equal. `NaN == NaN`,
 `Infinity == Infinity`, and `-Infinity == -Infinity` will always be false.
 
 In the case that two numbers differ only in their 11th or later digits after
-the decimal point, both numbers are truncated to 10 decimal places and compared.
+the decimal point, both numbers are rounded to 10 decimal places and compared.
 
 For example,
 
@@ -83,7 +83,7 @@ equal
 Numbers in Sass do not have total ordering. Any comparison to `NaN` will return false.
 Otherwise, ordering follows what would be expected when comparing two numbers.
 
-Numbers with more than 10 decimal places are truncated to 10 places before comparison.
+Numbers with more than 10 decimal places are rounded to 10 places before comparison.
 
 ### Hashing
 
--- a/proposal/numbers.md
+++ b/proposal/numbers.md
@@ -9,11 +9,13 @@ This proposal formally defines all numbers in Sass to be double-precision or 64-
 - [Floating Point Numbers: Draft 1.0](#floating-point-numbers-draft-10)
   - [Table of Contents](#table-of-contents)
   - [Background](#background)
-  - [Division by Zero](#division-by-zero)
-  - [Equality](#equality)
-  - [Ordering](#ordering)
-  - [Hashing](#hashing)
-  - [Formatting](#formatting)
+  - [Summary]()
+  - [Semantics](#semantics)
+    - [Division by Zero](#division-by-zero)
+    - [Equality](#equality)
+    - [Ordering](#ordering)
+    - [Hashing](#hashing)
+    - [Formatting](#formatting)
 
 ## Background
 
@@ -29,7 +31,16 @@ and [crashes](https://github.com/sass/dart-sass/issues/1059) for `dart-sass`.
 
 This document serves to unify these edge cases and resolve existing issues within implementations.
 
-## Division by Zero
+## Semantics
+
+This proposal formally defines all numbers in Sass to be double-precision or 64-bit floating point numbers.
+
+More specifically, numbers are represented as [IEEE 754-2008 "binary64" floating point numbers](https://web.archive.org/web/20190820164556/http://www.dsc.ufcg.edu.br/~cnum/modulos/Modulo2/IEEE754_2008.pdf).
+
+The semantics enumerated here serve to more succintly describe their behavior and
+to highlight Sass-specific behavior.
+
+### Division by Zero
 
 Any positive number divided by zero will return `Infinity`
 
@@ -37,7 +48,7 @@ Zero divided by zero will return `NaN`
 
 Any negative number divided by zero will return `-Infinity`
 
-## Equality
+### Equality
 
 Numbers in Sass are symmetrically and transitively equal. That is,
 if `a == b`, then `b == a` and if `a == b` and `b == c`, then `a == c`.
@@ -48,14 +59,14 @@ Numbers in Sass are *not* necessarily reflexively equal. `NaN == NaN`,
 In the case that two numbers use more than 10 decimal places of precision,
 both numbers are truncated to 10 decimal places and compared.
 
-## Ordering
+### Ordering
 
 Numbers in Sass do not have total ordering. Any comparison to `NaN` will return false.
 Otherwise, ordering follows what would be expected when comparing two numbers.
 
 Numbers with more than 10 decimal places are truncated to 10 places before comparison.
 
-## Hashing
+### Hashing
 
 The non-reflexive nature of number equality in Sass makes it difficult to
 use floating point numbers as keys inside maps. This property means
@@ -68,7 +79,7 @@ In order to support numbers as map keys, a Sass implementation must
  - deny `NaN`, `Infinity`, and `-Infinity` as keys
  - truncate all numbers to 10 decimal places before insertion
 
-## Formatting
+### Formatting
 
 In order to emit a number:
  - Round the number to 10 decimal places
--- a/proposal/numbers.md
+++ b/proposal/numbers.md
@@ -23,7 +23,7 @@ Numbers in Sass today are ill-defined. Between implementations there
 is very little agreement around edge cases, and there are few spec tests
 exercising these cases.
 
-The behavior of `dart-sass` and `libsass`, the two officially "blessed" implementations
+The behavior of `dart-sass` and `libsass`, the two major implementations
 at the time of writing, differ greatly with regard to numeric equality, rounding, and overflows.
 
 Additionally, this lack of definition has caused both [regressions](https://github.com/sass/dart-sass/issues/807)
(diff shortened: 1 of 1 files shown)


proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
+# Floating Point Numbers: Draft 1.0
+
+*[(Issue)](https://github.com/sass/sass/issues/2892)*
+
+This proposal formally defines all numbers in Sass to be double-precision or 64-bit floating point numbers.
+
+## Table of Contents
```
jathak, line 7: This should be structured following the format listed [here](https://github.com/sass/sass/blob/master/CONTRIBUTING.md#proposal). Most of the sections here should be subsections under "Semnantics" and you should add a "Summary" section.

proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
...
+and dropping support for numbers as map keys would break backwards compatibility.
+
+In order to support numbers as map keys, a Sass implementation must
+ - deny `NaN`, `Infinity`, and `-Infinity` as keys
+ - truncate all numbers to 10 decimal places before insertion
+
+## Formatting
+
+In order to emit a real number:
+ - Round the number to 10 decimal places
```
jathak, line 101: The behavior here should probably be consistent with what's used for equality and comparison (either truncate in all cases or round in all cases). If we do choose to round here, we should specify the rounding method.

jathak (commented):
Thanks for writing this up!

A few high-level notes notes (@nex3 may also want to take a look):
- it probably makes sense to tie much of this to existing specifications for numbers in [CSS](https://www.w3.org/TR/css-values-3/#numbers), [ECMAScript](https://www.ecma-international.org/ecma-262/#sec-ecmascript-language-types-number-type), and/or [IEEE-754](https://en.wikipedia.org/wiki/IEEE_754) to ensure that our handling of edge cases matches other parts of the web ecosystem as much as possible
- since we're formalizing the internal representation of numbers and how to format them, we should probably also formalize the syntax for number literals here as well (the [CSS spec's syntax](https://www.w3.org/TR/css-syntax-3/#number-token-diagram) is a good starting place)

proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
+# Floating Point Numbers: Draft 1.0
+
+*[(Issue)](https://github.com/sass/sass/issues/2892)*
+
+This proposal formally defines all numbers in Sass to be double-precision or 64-bit floating point numbers.
```
nex3, line 5: Since the header is considered to be non-normative, this definition should be reiterated in the "Semantics" section.

proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Equality
+
+Numbers in Sass are symmetrically and transitively equal. That is,
+if `a == b`, then `b == a` and if `a == b` and `b == c`, then `a == c`.
+
+Numbers in Sass are *not* necessarily reflexively equal. `NaN == NaN`,
+`Infinity == Infinity`, and `-Infinity == -Infinity` will always be false.
+
+In the case that two real numbers use more than 10 decimal places of precision,
```
nex3: This is kind of a semantic nit, but I'm not sure the word "real" here and elsewhere is very useful. You can just talk about "numbers" directly. In fact, "real" is a little misleading because Sass can only accurately represent (a subset of) rational numbers.

nex3: > use more than 10 decimal places of precision

This phrase is a little vague. We probably want to more specifically describe this in terms of the mathematical features of the number, like "if two real numbers differ only in their 11th or later digits after the decimal point".

proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
...
+both numbers are truncated to 10 decimal places and compared.
```
nex3: It would probably be good to include non-normative examples here. (We use Markdown block quotes to indicate a non-normative that appears within a normative section.)

proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
...
+Numbers in Sass do not have total ordering. Any comparison to `NaN` will return false.
+Otherwise, ordering follows what would be expected when comparing two real numbers.
+
+Real numbers with more than 10 decimal places are truncated to 10 places before comparison.
+
+## Hashing
+
+The non-reflexive nature of number equality in Sass makes it difficult to
+use floating point numbers as keys inside maps. This property means
+that `hash(a)` does not necessarily equal `hash(a)`.
```
nex3: If I'm reading this statement right, I don't think this is necessarily correct. If the same value `a` is hashed twice, it will always produce the same value.

Maybe you mean to say "`hash(a)` doesn't necessarily equal `hash(b)` even if `a == b`", although if you're describing Sass's `==` operation that should never be true.

proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
...
+
+Despite this, it is desirable for users to be able to use numbers as map keys,
+and dropping support for numbers as map keys would break backwards compatibility.
+
+In order to support numbers as map keys, a Sass implementation must
+ - deny `NaN`, `Infinity`, and `-Infinity` as keys
```
nex3: What does "deny" mean here? If it means "creating a map with one of these keys is an error", that should be explicit.

proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
...
+ - truncate all numbers to 10 decimal places before insertion
```
nex3: This implies that when retrieving the map key, it will be truncated. This can be user-visible—for example, multiplying two numbers that are `==` by the same large number may expose differences between them.

I think it's probably better to just generally declare that map operations return values whose keys are `==`-equivalent, and leave it up to the implementation to figure out how to represent that.

proposal/numbers.md
nex3: The Dart Sass and Ruby Sass equality behavior is closer to rounding than to truncating, although in fact it's neither (which is probably undesirable since it leads to non-transitive equality).

proposal/numbers.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Formatting
+
+In order to emit a real number:
+ - Round the number to 10 decimal places
+ - Remove trailing zeros
+ - If the number is negative and non-zero, emit a negative sign, `-`
+ - If in compressed mode:
+   - If the whole number is zero, do nothing
```
nex3: I'm not sure what "the whole number" refers to here. It would probably be clearer if you used some variables; see other specs for examples.

nex3 (commented):
I'll echo Jen in saying thanks for writing this up! This has been on our to-do list for a while but it's hard to get around to it with everything else we have on our plates :smile:.

proposal/numbers.md
connorskees: Ah this is a good catch! I was thinking that `hash(NaN) != hash(NaN)`, but it seems this statement instead implies that hashing of numbers is nondeterministic. 

Reworded in 2757aa3264af8080c950b00a967c9bf4f5f78441

proposal/numbers.md
connorskees: Ah my bad! I did not see that document. Thank you -- I have reformatted the document to include "Semantics" and "Summary" sections

proposal/numbers.md
```diff
@@ -0,0 +1,134 @@
...
+These two numbers differ *at* the 10th decimal place, so they should be considered
+unequal
+1.0000000001 == 1.0000000002
+  ---------^      ---------^
+
+These two numbers differ only *beyond* the 10th decimal place, so they should be considered
+equal
+1.00000000001 == 1.00000000002
+  ---------^       ---------^
+```
```
nex3: Please put examples in Markdown quote blocks to make it clear that they're non-normative.

proposal/numbers.md
```diff
@@ -0,0 +1,134 @@
...
+### Equality
+
+Numbers in Sass are symmetrically and transitively equal. That is,
+if `a == b`, then `b == a` and if `a == b` and `b == c`, then `a == c`.
+
+Numbers in Sass are *not* necessarily reflexively equal. `NaN == NaN`,
+`Infinity == Infinity`, and `-Infinity == -Infinity` will always be false.
+
+In the case that two numbers differ only in their 11th or later digits after
+the decimal point, both numbers are rounded to 10 decimal places and compared.
```
nex3: This implies that

```
1.0000000009 != 1.0000000010
  ---------^      ---------^
```

and

```
1.0000000009 != 1.0000000000
  ---------^      ---------^
```

which seems very strange. I think it would make more sense to just say that two numbers are `==` if they're equal when rounded to 10 decimal places.

proposal/numbers.md
```diff
@@ -0,0 +1,134 @@
...
+trying to construct a map with, insert into (`map-insert`), retrieve (`map-get`) or remove (`map-remove`)
+from a map any of `NaN`, `Infinity`, or `-Infinity`.
+
+### Formatting
+
+In order to emit a number:
+ - Round the number to 10 decimal places
+ - Remove trailing zeros
+ - If the number is negative and non-zero, emit a negative sign, `-`
+ - If in compressed mode:
```
nex3: Generally we don't specify the compressed more behavior. Each implementation has the freedom to choose its output format(s) as long as the CSS semantics are identical, so we try to specify only the semantics of the output and leave the detailed formatting undefined.

nex3 (changes requested):

proposal/numbers.md
```diff
@@ -0,0 +1,134 @@
...
+The non-reflexive nature of number equality in Sass makes it difficult to
+use floating point numbers as keys inside maps. This property means
+that `hash(a)` doesn't necessarily equal `hash(b)` even if `a == b`.
+
+Despite this, it is desirable for users to be able to use numbers as map keys,
+and dropping support for numbers as map keys would break backwards compatibility.
+
+In order to support numbers as map keys, a Sass implementation must raise an error when
+trying to construct a map with, insert into (`map-insert`), retrieve (`map-get`) or remove (`map-remove`)
+from a map any of `NaN`, `Infinity`, or `-Infinity`.
```
nex3, line 96: This is a breaking change. That doesn't necessarily mean we can't do it, but I'd like to see some discussion in the spec of:

* Why we're making this an error rather than just having these values produce map entries that can't be retrieved with `map-get()`.

* What the deprecation process will look like for moving users onto this new behavior.

It's also worth noting that this would be an idiosyncratic way to handle these numbers in maps. JavaScript and C# both allow `NaN` as a key that can be accessed later with another `NaN` value. Ruby, Python, and Dart all allow it as a key but don't allow it to be looked up. I don't know of any language that throws an error when it's used as a key.

proposal/numbers.md
stof: it can also lead to non-equal numbers being output in the same way in CSS (because the output is using rounding)

proposal/numbers.md
```diff
@@ -0,0 +1,127 @@
...
+`Infinity == Infinity`, and `-Infinity == -Infinity` will always be false.
+
+Two numbers are considered equal if they do no differ when rounded to 10 decimal places.
+
+> For example,
+> 
+> These two numbers differ *before* the 10th decimal place, so they should be considered
+> unequal
+> 1.0000000010 == 1.0000000020
+>   ---------^      ---------^
```
nex3, line 66: Nit: these diagrams aren't rendering correctly. You'll need to make them code blocks.

proposal/numbers.md
```diff
@@ -0,0 +1,127 @@
...
+> 
+> These two numbers differ *at* the 10th decimal place, so they should be considered
+> unequal
+> 1.0000000001 == 1.0000000002
+>   ---------^      ---------^
+> 
+> These two numbers differ only *beyond* the 10th decimal place, so they should be considered
+> equal
+> 1.00000000001 == 1.00000000002
+>   ---------^       ---------^
```
nex3, line 76: This is misleading; only differing beyond the 10th decimal place is essentially the current behavior, since it implies that `1.00000000001 == 1.00000000009`. It's worth making clear here that `1.00000000001 != 1.00000000009` but `1.00000000001 == 0.99999999995`.

proposal/numbers.md
```diff
@@ -0,0 +1,127 @@
...
+
+Any negative number divided by zero will return `-Infinity`
+
+### Equality
+
+Numbers in Sass are symmetrically and transitively equal. That is,
+if `a == b`, then `b == a` and if `a == b` and `b == c`, then `a == c`.
+
+Numbers in Sass are *not* necessarily reflexively equal. `NaN == NaN`,
+`Infinity == Infinity`, and `-Infinity == -Infinity` will always be false.
```
nex3, line 57: These are more emergent properties of the spec than specifications themselves, so they should be in non-normative quote blocks.

proposal/numbers.md
```diff
@@ -0,0 +1,127 @@
...
+
+Two numbers are considered equal if they do no differ when rounded to 10 decimal places.
```
nex3, line 59:
```suggestion
Two numbers are considered equal if they do not differ when rounded to 10 decimal places.
```

proposal/numbers.md
```diff
@@ -0,0 +1,127 @@
...
+> 
+> These two numbers differ only *beyond* the 10th decimal place, so they should be considered
+> equal
+> 1.00000000001 == 1.00000000002
+>   ---------^       ---------^
+
+### Ordering
+
+Numbers in Sass do not have total ordering. Any comparison to `NaN` will return false.
+Otherwise, ordering follows what would be expected when comparing two numbers.
```
nex3, line 81: "what would be expected" is a bit vague, I'd say instead "the natural numerical ordering".

proposal/numbers.md
```diff
@@ -0,0 +1,127 @@
...
+
+Numbers with more than 10 decimal places are rounded to 10 places before comparison.
```
nex3, line 83: Give some (non-normative) examples here.

proposal/numbers.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Hashing
+
+The non-reflexive nature of number equality in Sass makes it difficult to
+use floating point numbers as keys inside maps. This property means
+that `hash(a)` doesn't necessarily equal `hash(b)` even if `a == b`.
```
nex3, line 89: For what numbers would `a == b` and `hash(a) != hash(b)`?

proposal/numbers.md
```diff
@@ -0,0 +1,127 @@
...
+
+Despite this, it is desirable for users to be able to use numbers as map keys,
+and dropping support for numbers as map keys would break backwards compatibility.
+
+In order to support numbers as map keys, a Sass implementation must raise an error when
+trying to construct a map with, insert into (`map-insert`), retrieve (`map-get`) or remove (`map-remove`)
+from a map any of `NaN`, `Infinity`, or `-Infinity`.
+
+### Formatting
```
nex3, line 98: I'd like to keep the way we specify output a bit more general, to leave room for implementations to produce different output that's still CSS-identical. In this case, I think it makes sense to describe the output in terms of CSS parsing:

> To serialize a number `number`:
>
> * If `number` is fuzzy-equal to an integer, return text for which the [Consume a number] algorithm will return that integer with the type "integer".
> * Otherwise, return text with at most ten decimal places for which the [Consume a number] algorithm will return a number that's fuzzy-equal to `number`.
>
> [Consume a number]: https://drafts.csswg.org/css-syntax-3/#consume-number

nex3 (changes requested):
Sorry it's taken me so long to circle back to this!

It's become apparent in the course of writing other specs that it would be useful to be able to refer explicitly to the fuzzy equality and comparison operations Sass uses, so could you add a Procedures section that defines fuzzy-equals and fuzzy-comparison operations, and then refer to those in the spec text?

stof:
@nex3 should this be closed in favor of https://github.com/sass/sass/pull/3378 ?

nex3:
Yes, good call.

PR #2905, opened 2020-09-18
Allow `map.deep-remove()` to take a single key

jathak:
This allows `map.deep-remove()` to be used in place of `map.remove()` even for shallow removals.


nex3 (approved):

Awjin (approved):

mirisuzanne (approved):
Yes, great idea - makes it much more flexible for automated calls.

PR #2921, opened 2020-10-08
Spec Feat: Add support for multiline expressions in indented syntax

alromh87:
See sass/dart-sass#280
See sass/sass#216
See sass/dart-sass#1098
See sass/sass-spec#1573

### DIFF
--- a/accepted/multiline-on-idented-syntax.md
+++ b/accepted/multiline-on-idented-syntax.md
@@ -40,67 +40,39 @@ Also worth to note that this is a very long awaited feature, issue has been open
 
 > This section is non-normative.
 
-Sass will allow multiline declaration by using backslash "\" to wrap long lines, 
-treating it as single line, 
-so should only be used where single line is used now, example:
+When using indented syntax Sass will allow multiline declaration according to parsing context, 
+efectively treating multiline input as single line when applicable, example:
 
 ```sass
 /* Mixins */
-@mixin col($cols, $mleft: 0, $mright: 0, $include-margin: false, $border: 0, \
-           $pleft: 0, $pright: 0, $include-padding: true, $extra: 0, $clear: false, \
+@mixin col($cols, $mleft: 0, $mright: 0, $include-margin: false, $border: 0,
+           $pleft: 0, $pright: 0, $include-padding: true, $extra: 0, $clear: false,
            $lead: true, $container: false)
     width: $cols
     color: red
     display: block
 ```
+Would be interpreted as:
 
 ```sass
-/* Font faces */
-@font-face
-    font-family: 'SPEdessa'
-    src: url('fonts/spedessa-webfont.eot')
-    src: url('fonts/spedessa-webfont.eot?#iefix') format('embedded-opentype'), \
-        url('fonts/spedessa-webfont.woff') format('woff'), \
-        url('fonts/spedessa-webfont.ttf') format('truetype')
-    font-weight: normal
-    font-style: normal
+@mixin col($cols, $mleft: 0, $mright: 0, $include-margin: false, $border: 0, $pleft: 0, $pright: 0, $include-padding: true, $extra: 0, $clear: false, $lead: true, $container: false)
 ```
-
-```sass
-/* List */
-$list:  (1,  "value"), \
-    (5,  "value"), \
-    (23, "value"), \
-    (85, "value"), \
-    (32, "value"), \
-    (11, "value"), \
-    (35, "value"), \
-    (89, "value")
-```
-
-```sass
-/* Maps */
-$susy: ( columns: 12, \
-  gutters: 1/3, \
-  gutter-position: after, \
-  math: fluid, \
-  output: float, \
-  flow: ltr, \
-  global-box-sizing: border-box )
-```
-
 
 ## Syntax
 
-Backslash most be followed by newline, otherwise will throw an error,
-backslash will be ignored as well as spaces after,
-making the input similar to being writen in a single line.
+When using idented syntax long declarations can be splitted into multiple lines, 
+newline is allowed after opening parenthesis, 
+opening square bracket, 
+or after a comma, 
+in this context newline will be interpreted as non-significant and the next line appened to current line effectively treating input as if it was writen as single line.
+
+Whitespaces are allowed in between the cited characters and the newline, whitespaces after the newline will be interpreted as non-significant as well.
 
 Input:
 
 <x><pre>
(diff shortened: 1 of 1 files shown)


accepted/multiline-on-idented-syntax.md
```diff
@@ -0,0 +1,111 @@
...
+  flow: ltr, \
+  global-box-sizing: border-box )
+```
+
+
+## Syntax
+
+Backslash most be followed by newline, otherwise will throw an error,
+backslash will be ignored as well as spaces after,
+making the input similar to being writen in a single line.
```
nex3: This needs to be much more formal. A proposal is a portion of a specification, and it needs to describe the new behavior unambiguously enough that any implementation could implement it without behavioral inconsistencies between them. This doesn't explain in which contexts backslashes will be interpreted, how this parsing intersects with the use of backslashes in identifier or string tokens, or what "backslash will be ignored" means.

nex3 (changes requested):
If we're going to tackle #216, I want to do it in a more robust way than just providing backslash as an opt-out for ignoring a newline. We can potentially include that as part of such a proposal, but I'd like to reach a point where most Sass files *don't need* to use a backslash. This means using the parsing context to determine when it's safe to ignore newlines in general. For example, nesting is *never* valid within matching parentheses (or square/curly braces) or immediately after a comma in a SassScript expression, so we should always interpret a newline in those contexts as non-significant.

accepted/multiline-on-idented-syntax.md
alromh87: Ok, I will try again, hope is closer now

nex3 (changes requested):
This is still not nearly detailed or specific enough. Please see [other proposals](https://github.com/sass/sass/tree/master/accepted) for examples of the style of writing we're looking for. Because this is proposing to change the fundamentals of how Sass is parsed to move it from line-by-line parsing to a more stream-oriented parser, you'll need to give a thorough explanation of how that stream-oriented parsing works from start to finish. You don't need to cover every production of course, but you will need to clearly specify not just how to determine where a statement boundary is in the input stream but what a statement boundary *means* and how it interacts with the conversion of an input stream to an AST.

To be clear, this is unlikely to be a simple proposal to write. That's why it hasn't been done before now: it's changing the fundamental nature of parsing in Sass, and that parsing is currently underspecified to begin with. I'm grateful that you're willing to give it a try, but I also caution you away from thinking it will be straightforward or easy.

alromh87:
I might be totally wrong but feel like you don't want this done or thinking it's way too complex 😅 just a long shot.

Could you be more specific on what example to look at, reading all of them seems but to much 😁

alromh87:
Didn't want to close

nex3:
I definitely do want to see this get done, but language changes—especially to such deep-seated language features—are not something that can or should happen easily or quickly. We're responsible for a language that's used by hundreds of thousands of people, and once we make a change it's *very* difficult to ever reverse it. We have strong procedures and expectations in place to help ensure that these changes are designed correctly from the outset and that they can be reliably implemented consistently across all implementations. It's not just there to make your life more difficult.

purplesyringa:
Do you think we can make the statement continuation implicit whenever the line by itself causes an early-EOF parsing error *or* when a backslash is used? For instance:

- `@mixin name($name` \n `, $other)` -- implicit continuation because a parenthesis wasn't closed.
- `$color: rgb` \n `(100, 150, 200)` -- implicit continuation because `rgb` is a CSS function, not a valid value.
- `display:` \n `none` -- implicit continuation because a value is missing.

alromh87:
@imachug I believe allowing all that cases would lead to less readable code thats why I proposed openning parenthesis and comas:
Meaning this would be allowed:
```
@mixin name(
 $name,
 $other)
```
```
@mixin name($name,
 $other)
```
But not this:
```
@mixin name($name 
 , $other)
```
Nor this:
```
@mixin name($name: 
  value, $other)
```
```
@mixin name($name
  : 
  value
, $other)
```

Also implementation for dart would be easier this way since, the main difference is when consuming whitespaces, so change would only be needed in this scenario.

purplesyringa:
> But not this:
> ```
> @mixin name($name 
>  , $other)
> ```

It is actually a rather common style in other programming languages, it's used like this:

```
some_function_with_many_arguments(arg1
                                , arg2
                                , arg3
                                , arg4
                                , arg5
                                , arg6
                                , arg7
                                , arg8
                                )
```

One of the reasons it's preferred is that adding an argument to the end of the list only adds one line without removing/adding commas to neighbour lines.

alromh87:
Not sure, maybe someone else should make the call, from the developer point of view I prefer the proposed one.

nex3:
> Do you think we can make the statement continuation implicit whenever the line by itself causes an early-EOF parsing error _or_ when a backslash is used? For instance:
> 
> * `@mixin name($name` \n `, $other)` -- implicit continuation because a parenthesis wasn't closed.

I agree that a newline should be allowed here.

> * `$color: rgb` \n `(100, 150, 200)` -- implicit continuation because `rgb` is a CSS function, not a valid value.

As far as Sass is concerned, `rgb` is a valid value. All syntactically-valid CSS identifiers are valid in Sass.

> * `display:` \n `none` -- implicit continuation because a value is missing.

You're right that a newline here wouldn't be valid, but this raises an important question about custom properties. `--display:` immediately followed by a newline isn't currently valid Sass, but `--display:` followed by a space and then a newline *is* valid because custom property values can be almost anything, including a single space. Similarly, `--font: Arial,` is valid because commas have no particular syntactic meaning in a custom property value.

This means that if we were to allow newlines after commas in normal properties, custom properties would then be parsed differently which could be confusing. On the other hand, custom properties are *always* parsed differently, so maybe it's fine. Or maybe we can add another special case and say that any text indented beneath a custom property is considered part of its value. I'm open to opinions on which option is best here.

LandonSchropp:
Just to add another use case to this, this would be very helpful when using `grid-template-areas`:

``` sass
grid-template-areas: 
  "primary-navigation primary-navigation" 
  "secondary-navigation sidebar" 
  "main sidebar" 
  "content-controls sidebar"
```

While this wouldn't be my first choice of syntax, many other languages support this by using backslashes:

```
grid-template-areas: \
  "primary-navigation primary-navigation" \
  "secondary-navigation sidebar" \
  "main sidebar" \
  "content-controls sidebar"
```

Duncan-Idaho:
Did we give up on this proposal ? I'd like to revive it.

nex3:
I'm going to close this out due to lack of response. @Duncan-Idaho, if you'd like to open a new PR for it, feel free.

Issue #2535, opened 2018-07-06
Define a standard for custom URL loaders


nex3:
Many CSS features involve referring to assets outside the stylesheet using URLs: plain CSS `@import`s, background images, and so on. It's not always obvious how these assets will map to URLs on the server, though. Currently Sass provides no built-in support for this mapping, but doing so would be very useful.

I think Sass implementations should expose the ability for users to define their own URL-resolution logic that's invoked for every `url()` and every plain-CSS `@import`. This should be done through a JS API that's shared between Dart Sass and Node Sass, as well as a Dart API for Dart Sass users who invoke it via Dart.

I also think it would be valuable to provide some built-in resolution strategies that are available via the command line and through JS/Dart as standardized resolvers. For example, we might want to provide the ability to resolve URLs to `data:` URLs that embed the document directly, or to map directories on the filesystem to URLs that will be served.

- [ ] https://github.com/sass/dart-sass/issues/195
- [ ] https://github.com/sass/libsass/issues/532

dwt:
As was discussed in #1015 I'd like to get some design considerations going here.

I know you are talking about custom URL-Resolvers for this issue, but I think there should be some serious thinking about a either a much simpler approach, or about exposing them through a much simpler interface to solve a very big subset of the use cases this is required for.

My thinking is that there are two modes of operating on included URLs - either you leave them alone (AFAIK the current approach) or you touch them. If you touch them, one very simple and very effective way, is to allow relative URLs, that are interpreted as relative to the *current* file (just as in CSS). When compiling these URLs, they should retain the this relative quality, but still modified to point to the original file.

As was mentioned this has problems when compiling sass to stdout, in which case the compiler doesn't know how to change those URLs.  Some options that are possible here:

* Warn and leave them as is, the user is required to put the referenced files in the same relative spot
* Error and abort compilation unless the user provides the information where the code will end up relative to the assets
* Allow in-file configuration by a sass directive that describes where the code will end up if compiled to stdout (maybe support two directives, so you can set a default, or override the url transformation)

Supporting custom loaders seems to require much more configuration - so maybe a simplified approach like this can be considered - especially because (I think) that it can be a really nice increment point to get to the full url loader feature, as it can be introduced by adding a sass function `file-relative-url()` (or whatever you call it that marks the urls to be specially treated). Then the URL-resolver feature you are discussing here can then later add different functions or marker arguments to this function to choose different resolver strategies.

jhnns:
[...]

PR #2927, opened 2020-10-19
Proposal for `url(...)` rewriting

DeMoorJasper:
This PR includes a proposal on how to handle rewriting url's as discussed in issue #2535 for some time now.

Hopefully this helps move things forward a bit on this issue, any feedback is very welcome.

If there's anything else I can do to help this move forward feel free to ask.

### DIFF
--- a/proposal/custom-url-resolution.md
+++ b/proposal/custom-url-resolution.md
@@ -6,6 +6,7 @@ _[(Issue)](https://github.com/sass/sass/issues/2535)_
 
 - [Background](#background)
 - [Summary](#summary)
+  - [Steps](#steps)
   - [JavaScript API](#javaScript-api)
   - [Edge cases](#edge-cases)
 - [Syntax](#syntax)
@@ -15,7 +16,7 @@ _[(Issue)](https://github.com/sass/sass/issues/2535)_
 
 > This section is non-normative.
 
-Many css features require the use of a url import to reference resources from outside the sass files, however these files also need to exist on the eventual output directory and server. To ensure the references are valid the sass API should allow for the user to provide a way to remap and/or inline these resources.
+Many css features require the use of a url import to reference resources from outside the sass files, however these files also need to exist on the eventual output directory and server. To ensure the references are valid, the sass API should allow for the user to provide a way to remap and/or inline these resources.
 
 ## Summary
 
@@ -23,39 +24,43 @@ Many css features require the use of a url import to reference resources from ou
 
 This proposal defines a standardized way to remap the url imports to the final location on the server or inline reference.
 
-This is accomplished by adding a callback function to the JavaScript API options that allows rewriting of the url references.
+This is accomplished by running url references through the importer plugin(s) and url rewriting plugin(s) if any url rewriting plugin has been defined.
+
+### Steps
+
+The steps of a url reference import:
+
+- Url reference gets extracted from the Sass file (For more information see [Syntax](#syntax))
+- This reference gets remapped to the actual location on the filesystem by the importer plugin(s)
+- The resolved reference gets passed into the rewrite url plugin pipeline which returns the output url of this resource.
+
+If a certain resource cannot be remapped the url rewrite plugin will return `null`, if all rewrite plugins return `null` it should fallback to the original value of the url reference.
 
 ### JavaScript API
 
-At the core of remapping the url imports is the JavaScript API which allows users to return a new url reference based on the sass file location and url reference.
+At the core of remapping the url imports is the JavaScript API which allows users to return a new url reference based on the filepath or content.
 
-This function can return either a promise or utilise the provided callback function. In case an error gets returned or thrown the sass compilation should fail and return this error, this will likely only happen for files that do not exist.
+The first parameter this rewriteUrl function gets passed in is the importerResult, which is an object containing an optional `file` and `content` field.
 
-When a filepath of the originating sass file is unknown, it should still call this callback but use null as the filepath, this way all url resolution is ensured and the end user can decide whether to handle this or not.
+- The `file` field is the location on disk of the given url resource, this can be used to create relative mappings or to simply read the file.
+- The `content` field is a `Blob` which contains the content of the url resource, in this case the url rewriter should not try to read the file manually and use this instead as this is what the importer returned.
 
-Callback syntax:
+This function should call the second parameter, the `done` callback, whenever it is finished rewriting the url.
+
+This callback has two parameters, `error` and `url`:
+
+- The `error` parameter can be `null` or an `Error` object, this is used when an error occurred when trying to rewrite the url.
+- The `url` parameter can be `null` or a `string`, this should be the new url as it can be found in the browser.
 
 ```TypeScript
 let sassOptions = {
   // Rewrite url references
-  rewriteUrl: (url: string, filepath: string | null, done: (error: Error, newUrl: string) => void): void => {
-    done(null, filepath ? `data:${base64(fs.readFileSync(path.join(path.basename(filepath), url)))}` : url);
-  }
-}
-```
-
-Promise syntax:
-
-```TypeScript
-let entryFilePath = '/index.scss';
-let sassOptions = {
-  // Rewrite url references
-  rewriteUrl: async (url: string, filepath: string | null): Promise<string> => {
-    if (filepath && url[0] === '.') {
-      return path.relative(entryFilePath, path.join(path.join(path.basename(filepath), url)));
-    } else {
-      return url;
+  rewriteUrl: async (importerResult: { file?: string, content?: Blob }, done: (error: Error | null, url: string | null) => void): void => {
+    if (file && !content) {
+      content = await fs.readFile(file);
     }
+
+    done(null, content ? `data:${base64(content)))}` : null);
   }
 }
 ```
@@ -84,4 +89,6 @@ This proposal does not introduce any new Syntax.
 
 ## Deprecation Process
 
-This will not directly introduce any breaking changes as this new feature will be opt-in. However tools like Parcel and WebPack will probably want to use this and in turn cause users that relied on the non existing relative url rewriting to end up with a broken codebase. However this can be worked around by these tools by falling back to resolving relative to the Sass entry point if the file does not exist, however this might be dangerous behavior and should log a warning.
+This will not directly introduce any breaking changes as this new feature will be opt-in.
+
+However tools like Parcel and WebPack will probably want to use this and in turn cause a breaking change as users are currently relying on the broken url logic of sass. However I strongly believe the impact of this will be very minor as in case a file does not exist the urlRewrite plugin can choose to just ignore this url reference.
--- a/proposal/custom-url-resolution.md
+++ b/proposal/custom-url-resolution.md
@@ -45,12 +45,14 @@ The first parameter this rewriteUrl function gets passed in is the importerResul
 - The `file` field is the location on disk of the given url resource, this can be used to create relative mappings or to simply read the file.
 - The `content` field is a `Blob` which contains the content of the url resource, in this case the url rewriter should not try to read the file manually and use this instead as this is what the importer returned.
 
-This function should call the second parameter, the `done` callback, whenever it is finished rewriting the url.
+This function can be used both asynchronously and synchronously, in case of it being asynchronous it should call the done callback, otherwise it should directly return the url synchronously (which is a string or null).
 
-This callback has two parameters, `error` and `url`:
+The done callback has two parameters, `error` and `url`:
 
 - The `error` parameter can be `null` or an `Error` object, this is used when an error occurred when trying to rewrite the url.
 - The `url` parameter can be `null` or a `string`, this should be the new url as it can be found in the browser.
+
+Asynchronous example:
 
 ```TypeScript
 // Rewrite url plugin/function
@@ -63,8 +65,26 @@ const rewriteUrlPlugin = async (importerResult: { file?: string, content?: Blob 
 }
 
 let sassOptions = {
-  // An array of all rewrite url plugins
+  // This can be an array or a single function
+  // depending on whether it's a pipeline of rewrite plugins or just a single plugin
   rewriteUrl: [rewriteUrlPlugin]
+}
+```
+
+Synchronous example:
+
+```TypeScript
+let rootFile = '/example/test.scss';
+
+let sassOptions = {
+  // Could also be an array as shown in the async example
+  rewriteUrl: (importerResult: { file?: string, content?: Blob }): string | null => {
+    if (!importerResult.file) {
+      return null;
+    }
+
+    return path.relative(rootFile, file);
+  }
 }
 ```
 
--- a/proposal/custom-url-resolution.md
+++ b/proposal/custom-url-resolution.md
@@ -115,9 +115,9 @@ let sassOptions = {
 
 ### Using Variables
 
-Variables can be used in a url reference, in this case it should get remapped based on the value that gets created after the variables have been applied to ensure we are able to remap the url as expected.
+Variables can be used in a url reference, in this case it should get remapped based on the value that gets created after the variables have been applied to ensure we are able to remap the entire url correctly.
 
-Examples:
+Examples with variables:
 
 ```Scss
 url("#{$asset-path}/image.png");
--- a/proposal/custom-url-resolution.md
+++ b/proposal/custom-url-resolution.md
@@ -10,6 +10,7 @@ _[(Issue)](https://github.com/sass/sass/issues/2535)_
   - [CLI](#cli)
     - [Possible Values](#possible-values)
   - [Edge cases](#edge-cases)
+- [Breaking Changes](#breaking-changes)
 
 ## Background
 
@@ -32,8 +33,8 @@ Callback syntax:
 ```TypeScript
 let sassOptions = {
   // Rewrite url references
-  rewriteUrl: (url: string, filepath: string, done: (newUrl: string) => void) => {
-    done(`data:${base64(url)}`);
+  rewriteUrl: (url: string, filepath: string, done: (error: Error, newUrl: string) => void) => {
+    done(null, `data:${base64(url)}`);
   }
 }
 ```
@@ -41,10 +42,15 @@ let sassOptions = {
 Promise syntax:
 
 ```TypeScript
+let entryFilePath = '/index.scss';
 let sassOptions = {
   // Rewrite url references
   rewriteUrl: async (url: string, filepath: string) => {
-    done(`data:${base64(url)}`);
+    if (url[0] === '.') {
+      return path.relative(entryFilePath, path.join(filepath, url));
+    } else {
+      return url;
+    }
   }
 }
 ```
@@ -79,4 +85,8 @@ url("./folder/#{$some-var}");
 
 #### Unknown sass filepath
 
-In case the sass compiler does not have a filepath for the originating sass file it is impossible to remap this url reference to the proper location on the server or inline this. In this case we should leave the url as is without trying to remap it in any way. 
+In case the sass compiler does not have a filepath for the originating sass file it is impossible to remap this url reference to the proper location on the server or inline this. In this case we should leave the url as is without trying to remap it in any way.
+
+## Breaking Changes
+
+This will not directly introduce any breaking changes as this new feature will be opt-in. However tools like Parcel and WebPack will probably want to use this and in turn cause users that relied on the non existing relative url rewriting to end up with a broken codebase. However this can be worked around by these tools by falling back to resolving relative to the Sass entry point if the file does not exist, however this might be dangerous behavior and should log a warning.
--- a/proposal/custom-url-resolution.md
+++ b/proposal/custom-url-resolution.md
@@ -17,8 +17,6 @@ _[(Issue)](https://github.com/sass/sass/issues/2535)_
 
 Many css features require the use of a `url()` import to reference resources from outside the sass files, however these files also need to exist on the eventual output directory and server. To ensure the references are valid the sass API should allow for the user to provide a way to remap and/or inline these resources.
 
-This will require a function in the JavaScript API as well as a cli option with some presets.
-
 ## Summary
 
 This proposal defines a standardized way to remap the url imports to the final location on the server or inline reference as well as providing some defaults for url imports remapping in the CLI.
@@ -78,3 +76,7 @@ url("#{$asset-path}/image.png");
 ```Sass
 url("./folder/#{$some-var}");
 ```
+
+#### Unknown sass filepath
+
+In case the sass compiler does not have a filepath for the originating sass file it is impossible to remap this url reference to the proper location on the server or inline this. In this case we should leave the url as is without trying to remap it in any way. 
--- a/proposal/custom-url-resolution.md
+++ b/proposal/custom-url-resolution.md
@@ -26,15 +26,19 @@ This is accomplished by adding a callback function to the JavaScript API options
 
 ### JavaScript API
 
-At the core of remapping the url imports is the JavaScript API which allows users to return a new url reference based on the sass file location and url reference. This function can return either a promise or utilise the provided callback function. In case an error gets returned or thrown the sass compilation should fail and return this error, this will likely only happen for files that do not exist.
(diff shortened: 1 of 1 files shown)


proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Steps
+
+The steps of a url reference import:
+
+- Url reference gets extracted from the Sass file (For more information see [Syntax](#syntax))
+- This reference gets remapped to the actual location on the filesystem by the importer plugin(s)
+- The resolved reference gets passed into the rewrite url plugin pipeline which returns the output url of this resource.
+
+If a certain resource cannot be remapped the url rewrite plugin will return `null`, if all rewrite plugins return `null` it should fallback to the original value of the url reference.
```
nex3: What does a `null` return value mean? Does it mean "I don't understand this URL format"? Does it mean "I couldn't find the file this URL was referring to"? We need to be very clear about these semantics because (if we do allow multiple rewriters) it's entirely possible that multiple URL rewriters will coexist in the same environment, and they need to have a shared understanding of how they're expected to operate in order to cooperate without having direct knowledge of one another.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,117 @@
...
+
+```Scss
+url("./folder/#{$some-var}");
+```
+
+## Syntax
+
+This proposal applies to the `url(...)` function, both with and without the quotation marks. This ensures this applies to the entire css specification for `url(...)`.
+
+This proposal does not introduce any new Syntax.
```
nex3: Since this doesn't affect the Sass's parsing phase, you don't need to include a Syntax section. However, you should include a Semantics section that specifies exactly what the new process of resolving a URL is.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,117 @@
...
+
+> This section is non-normative.
+
+This proposal defines a standardized way to remap the url imports to the final location on the server or inline reference.
+
+This is accomplished by running url references through the importer plugin(s) and url rewriting plugin(s) if any url rewriting plugin has been defined.
+
+### Steps
+
+The steps of a url reference import:
```
nex3: I'm not sure what "url reference import" means here. Do you mean plain-CSS `@import`s? Specifically only `@import`s that use the `url()` syntax for their URLs? Any `url()`?

A spec needs to be highly explicit, so there's no ambiguity and multiple implementations are able to implement exactly the same behavior. This is just the non-normative summary which has some more flexibility to be casual, but the expectation is that anything that's glossed over in the summary is made entirely explicit in the normative body of the proposal.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,117 @@
...
+- This reference gets remapped to the actual location on the filesystem by the importer plugin(s)
+- The resolved reference gets passed into the rewrite url plugin pipeline which returns the output url of this resource.
+
+If a certain resource cannot be remapped the url rewrite plugin will return `null`, if all rewrite plugins return `null` it should fallback to the original value of the url reference.
+
+### JavaScript API
+
+At the core of remapping the url imports is the JavaScript API which allows users to return a new url reference based on the filepath or content.
+
+The first parameter this rewriteUrl function gets passed in is the importerResult, which is an object containing an optional `file` and `content` field.
```
nex3: > importerResult

Why is this called `importerResult`? What does it have to do with importers?

> an optional `file` and `content` field.

If these fields are optional, under what circumstances are they passed or not passed?

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,117 @@
...
+
+- The `file` field is the location on disk of the given url resource, this can be used to create relative mappings or to simply read the file.
+- The `content` field is a `Blob` which contains the content of the url resource, in this case the url rewriter should not try to read the file manually and use this instead as this is what the importer returned.
```
nex3: This seems to assume that the Sass implementation already knows how to find a resource on disk given a URL. But as far as Sass is concerned, `url()`s are essentially opaque strings. The goal of a URL rewriter plugin is to encapsulate exactly that knowledge, the knowledge of the relationship between the `url()` string that appears in a file and the actual file that the user wants to refer to.

nex3 (changes requested):

DeMoorJasper:
@nex3 thanks for reviewing this proposal, I've changed it quite substantially based on your feedback.

Removed the ability to have multiple plugins in a row as it didn't really make sense after seeing your comment and thinking about it a bit more.
Also made the rewriteUrl plugin also responsible of resolving as you mentioned the importer cannot currently handle this.

Let me know if you think any other things need to change or if there's anything else I can help to move this forward.

proposal/custom-url-resolution.md
DeMoorJasper: It meant `I don't understand this URL format` but on second thought this is probably not really gonna happen, so removed this ability and the multiple plugin pipeline support.

proposal/custom-url-resolution.md
DeMoorJasper: I just meant url references. I think the wording should be less confusing now

proposal/custom-url-resolution.md
DeMoorJasper: Originally I thought we could re-use the importers for resolving and the file and content were what the importer returns.

Changed the syntax to not require the importers anymore, this way it'll probably be easier to get this implemented?

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,108 @@
...
+- These reference gets passed into the rewrite url plugin which returns the output url of this resource.
+- This returned value gets used to replace the original url reference.
+
+### JavaScript API
+
+At the core of remapping the url references is the JavaScript API which allows users to return a new url reference, based on filepath and original url reference.
+
+The first parameter of the urlRewrite function is an object with the following values:
+
+- `file`: The file path of the sass file that references the url.
```
nex3: Bceause Sass supports arbitrary importers that may completely replace the use of a physical filesystem, it doesn't think of stylesheets as having "file paths". Instead, it considers them to have "canonical URLs", which are (usually absolute) URLs that uniquely identify a stylesheet. That's probably what should be passed in here.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,108 @@
...
+- `entry`: The entry file path that has been used to call the sass compiler.
```
nex3: What's the use case for having access to the entrypoint?

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,108 @@
...
+
+```Scss
+url("#{$asset-path}/image.png");
+```
+
+```Scss
+url("./folder/#{$some-var}");
+```
+
+## Deprecation Process
```
nex3: You don't need to have a "deprecation process" section if you're not introducing any breaking changes to the Sass API.

nex3 (commented):
I was reading over https://github.com/sass/sass/issues/2535 again, and https://github.com/sass/sass/issues/2535#issuecomment-516593596 in particular where I outlined the general approach the Sass core team had decided on for this. The approach you're describing here differs substantially from the one that comment describes. What's the reasoning behind that difference?

proposal/custom-url-resolution.md
DeMoorJasper: Yes that's what it should be, didn't know it was called that

proposal/custom-url-resolution.md
DeMoorJasper: You're probably right, this is breaking but opt-in so will remove it

proposal/custom-url-resolution.md
DeMoorJasper: For outputting relative paths?

It's not really important though, as it can be worked around it by using a wrapper function or something that gets passed in additional options...

Just seemed cleaner to have it in the params, if you don't like it I can remove it

DeMoorJasper:
@nex3 originally I created a proposal based on that comment with the only real difference being using `url` instead of `sass-url` but as it would still be blocked by importer as you already mentioned the import api would also need to be changed.

I figured this approach would probably be easier to implement for now and probably solve the problem just as good for most people.

If you think this is a bad idea I could always change this to use `sass-url` again, that's the only real difference, besides the importer which can always be updated later.

I was also thinking about only rewriting url's that start with `file://` based on your comment, which I think is a wonderful idea, just not sure if this is something that should be implemented in here or be a choice for the plugins.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    return path.relative(path.dirname(entry), path.join(path.dirname(file), url));
+  }
+}
+```
+
+## Semantics
+
+This proposal defines an overload for `url()`, to allow users to rewrite url references to how it will be accessible on the browser.
```
nex3: "an overload" isn't really accurate here, since `url()` isn't a function as far as Sass is concerned and this isn't changing which arguments it takes anyway. I'd say "new behavior" instead, perhaps.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,101 @@
...
+    if (!importerResult.file) {
+      return url;
+    }
+
+    return path.relative(path.dirname(entry), path.join(path.dirname(file), url));
+  }
+}
+```
+
+## Semantics
```
nex3: Again, this section needs to be substantially expanded to describe the step-by-step process of transforming a `url()` expression or `sass-url()` function call into a resolved URL.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Summary
+
+> This section is non-normative.
+
+This proposal defines a standardized way to remap the url references to the final location on the server or inline reference.
+
+This is accomplished by running url references through the url rewriting plugin if a url rewriting plugin has been defined.
+
+### Steps
```
nex3, line 59: Since these sections are under "Summary", they're just here to provide a casual reader with a basic understanding of this proposal. Everything in here is non-normative, which means it's not required for an implementation to agree exactly with what it says. In other words, this should be casual and high-level, and everything deep and specific should go in normative sections below. This includes the JavaScript API as well.

proposal/custom-url-resolution.md
nex3: The entrypoint URL doesn't intrinsically have any relationship to the location of the CSS output file being generated. Although the current `render()` API does take an `outFile` parameter that's intended to indicate the location of the output file, it's not clear if that will be carried over to the new API (the Dart API doesn't have a corresponding parameter).

nex3 (changes requested):
Sorry for the long delay, I missed your response. For future reference, if you click the sync icon under reviewers, it will let your reviewer know the PR is ready for another look.

proposal/custom-url-resolution.md
DeMoorJasper: You're right, I'll remove it for now we can append the outFile in the future if we want to. Third party tooling like bundlers have this information already anyway

DeMoorJasper:
Thanks for the review @nex3 

I hope I've resolved all remarks you had, if I didn't feel free to leave some more feedback.

I've also been thinking about standardising the way this gets configured, as in there should probably be a list of presets that users can utilise (inline, relative, ...). The plugin that is currently introduced in this draft is definitely still required as bundlers will probably want to overwrite the default behaviour regardless to keep track of included files. (I know we do this in Parcel for other css related languages). But a standardised way of configuring this would make it possible for users to switch between webpack, parcel and just the standard sass cli with minimal changes, not sure if you think this should be part of sass or if this is something for the future?

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,128 @@
...
+
+```scss
+.upload-icon {
+  background-url: url(data:image/svg+xml;base64,Jggg==);
+}
+```
+
+## Semantics
+
+This proposal defines new behavior for `url()`, to allow users to rewrite url references to how it will be accessible on the browser.
```
nex3: I do strongly feel that we should define a new `sass-url()` function rather than overriding the existing `url()` function. It's also much simpler to specify the behavior of a function than a syntactic construct.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Steps
+
+Whenever a url reference is encountered in a sass file the following steps should be executed:
+
+- A url reference is encountered in the sass file
+- From the url function we extract the parameter value
+- Rewrite the parameter value if it contains any variables ([see using variables](#using-variables))
+- Pass the parameter value to the url rewrite plugin along with the canonical url of the current sass file
+- This plugin than returns a string, if it returns anything else or nothing throw an error ([see JavaScript API](#javascript-api))
```
nex3: What about async plugins?

nex3 (changes requested):
I think standardizing a few resolvers makes sense. These should probably be exposed as standard resolver functions, or for resolvers that need extra information like `relative`, a higher-order function that returns a resolver.

proposal/custom-url-resolution.md
DeMoorJasper: rewrote that line a bit to include async logic

proposal/custom-url-resolution.md
DeMoorJasper: alright probably a good idea, changed it

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,158 @@
...
+
+Output:
+
+```css
+.upload-icon {
+  background-url: url('data:image/svg+xml;base64,Jggg==');
+}
+```
+
+## Syntax
```
nex3: This isn't a syntactic modification anymore; you're not defining anything that changes how standard Sass syntax is parsed or interpreted. You're just defining a new global function, like `rgb()` or `inspect()`. Check out for example [More Math Functions](https://github.com/sass/sass/blob/master/accepted/more-math-functions.md) for what specifying behavior for a new function looks like.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,158 @@
...
+
+This proposal introduces logic for a new function: `sass-url("...")`, this function allow users to define a url that gets rewritten based on the config to ensure it can be loaded correctly from the browser.
+
+### Steps
+
+Whenever a url reference is encountered in a sass file the following steps should be executed:
```
nex3: "a url reference" isn't correct anymore.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+
+### Url Rewrite Presets
+
+Besides writing custom plugins, there are also presets to choose from: `none`, `inline` and `relative`. The default value is `inline`.
+
+On the cli this can be configured using the flag `--url-rewrite-preset`, example: `--url-rewrite-preset=inline`
+
+In the JavaScript configuration the flag is `urlRewritePreset`, example: `{ urlRewritePreset: "none" }`
```
nex3: Rather than having the JS configuration just use preset names, I'd prefer to have the presets be functions that can be passed to `requireUrl` (or higher-order functions that return those functions).

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,158 @@
...
+
+The inline preset also adds the file path of each inlined asset to the `stats.includedFiles` list that gets returned when running `sass.render`
+
+#### relative preset
+
+The relative preset is the most complex preset and has additional configuration options.
+
+TODO: Actually finish this part... Not sure yet how this should work, cli and JS api will probably be slightly different?
+
+The relative preset also adds the file path of each referenced asset to the `stats.includedFiles` list that gets returned when running `sass.render`
```
nex3: These should define algorithms for specifically how they transform inputs into outputs. The current descriptions are too vague.

nex3 (changes requested):

proposal/custom-url-resolution.md
DeMoorJasper: Not entirely sure how you see this. How would a cli work in this case and what is the added value?

The idea I had behind using a simple string based flag is that bundlers could actually read this standardised flag and overwrite it with their own logic as most of them have memory based filesystems for caching and testing. If we'd use functions we'd have to pass in a callback function or something that actually resolves and loads files and in the relative case also returns the final file path, so in the relative case this does not seem to add any value I think?

nex3: The CLI would still take string names. It's just the JS API that would use callbacks.

I don't like the idea of bundlers intercepting configuration that already has a specific meaning. If they wanted to define e.g. a `parcel.url.inline` function that users could pass in instead of `sass.url.inline`, that would be just about as usable and substantially more explicit.

DeMoorJasper: Oh alright, in that case I don't really think this proposal is the place to implement that, will remove it from the proposal as the functions will only be a couple lines and every tool will probably write their own anyway.

However should I still add the specification for the cli, or is that not required in this proposal?

nex3: You don't need to specify the CLI, but you should specify the built-in URL resolvers that should ship with the JS API.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,247 @@
...
+
+- `error`: This can be null or an Error object, `null` means there is no error.
+- `url`: This is the `string` that gets used to replace the sass-url reference. This cannot be null or anything else, otherwise sass will throw an error.
+
+Asynchronous example:
+
+```TypeScript
+let sassOptions = {
+  // Rewrite urls asynchronously
+  rewriteUrl: async (urlReference: { file: string, url: string }, done: (error: Error | null, url: string | null) => void): void => {
```
stof: Passing a `done` callback does not make sense to me. The result should be returned in the promise instead (an async function already returns a promise anyway)

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,247 @@
...
+    if (!urlReference.file) {
+      return url;
+    }
+
+    return path.relative(outDir, path.join(path.dirname(file), url));
+  }
+}
+```
+
+### Using Variables
```
stof: This section is totally unnecessary given that `sass-url` is defined as a function and not as a custom language construct. *All* functions are working with the value of their argument without having any way to know how it was produced.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,247 @@
...
+
+#### Inline Preset Usage
+
+To use this preset you can pass an instance of this preset in the urlRewrite option.
+
+```JS
+import * as sass from 'sass';
+
+const options = {
+  rewriteUrl: sass.url.inline()
```
stof: should be `rewriteUrl: sass.url.inline`. We want to pass the function as the value of the option, not calling it.
There is no need to make it a function returning another function IMO.

nex3:
Since this has had outstanding comments from @stof for several months, I'm going to close it out for now. I'm happy to re-open if @DeMoorJasper wants to come back to it in the future, and I'm also happy to review another PR if someone else wants to take this on.

proposal/custom-url-resolution.md
DeMoorJasper: Was just mimicing existing sass api's but will update

proposal/custom-url-resolution.md
DeMoorJasper: Right, will update

proposal/custom-url-resolution.md
DeMoorJasper: It was just to highlight another use-case or the possibility of this. But I guess that's better left for docs, will remove this

DeMoorJasper:
@nex3 sorry totally missed those comments, probably got lost in notifications as I usually just see a lot of "bump" responses to the linked issue instead of any feedback here.

But fixed the feedback on my fork, lemme know if I need to do something to get this moving forward again

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,230 @@
+# Custom url resolution: Draft 1
```
nex3, line 1: Style nit

```suggestion
# Custom URL Resolution: Draft 1
```

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,230 @@
...
+
+_[(Issue)](https://github.com/sass/sass/issues/2535)_
```
nex3, line 3: Style nit:

```suggestion
*([Issue](https://github.com/sass/sass/issues/2535))*
```

This matches the style of other proposals.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,230 @@
...
+    - [Inline Preset Implementation](#inline-preset-implementation)
+  - [The relative Preset](#the-relative-preset)
+    - [Relative Preset Usage](#relative-preset-usage)
+    - [Relative Preset Implementation](#relative-preset-implementation)
+
+## Background
+
+> This section is non-normative.
+
+Many css features require the use of a sass-url reference to reference resources from outside the sass files, however these files also need to exist on the eventual output directory and server. To ensure the references are valid, the sass API should allow for the user to provide a way to remap and/or inline these resources.
```
nex3, line 25: Please wrap lines at 80 characters.

nex3, line 25: Please use proper capitalization: "CSS", "Sass", etc. Also, code-quote any use of Sass or CSS identifiers or syntax such as `sass-url`.

nex3, line 25: The background section is supposed to describe the world _as it exists today_ to make the case for a new feature. Saying that we "require the use of a sass-url reference" doesn't make sense, because such a thing doesn't exist currently. What is the problem that exists and that needs to be solved?

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Summary
+
+> This section is non-normative.
+
+This proposal defines a standardized way to remap the sass-url references to the final location on the server or inline reference.
```
nex3, line 31: Again, you're saying "the sass-url references" without explaining at all what `sass-url` is or does.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,230 @@
...
+
+This is accomplished by running sass-url references through the url rewriting plugin if a url rewriting plugin has been defined.
```
nex3, line 33: What is "a URL-rewriting plugin"? Where and how are they defined? Are there any built-in ones?

This is just the summary so it doesn't need to be meticulously precise, but this is extremely vague. You need to give a reader a sense of the specifics of your proposal so that they can evaluate it without needing to read the entire spec, not just say "here's the problem, this proposal will solve it". Think of this as a draft of the user-facing documentation that will explain how to use the new feature.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,230 @@
...
+### Steps
+
+Whenever a sass-url reference is encountered in a sass file the following steps should be executed:
+
+- A sass-url reference is encountered in the sass file
+- From the url function we extract the parameter value
+- Rewrite the parameter value if it contains any variables ([see using variables](#using-variables))
+- Pass the parameter value to the url rewrite plugin along with the canonical url of the current sass file
+- This plugin than returns a string or calls the done callback with a string value, if it returns anything else or nothing it should throw an error ([see JavaScript API](#javascript-api))
+- The new url value that we received from the url rewrite plugin than gets used to replace the original url value
```
nex3, line 68: This is not written in a standard spec style. I recommend looking at other function definitions under spec/ and accepted/ to get a better idea of how to write this.

proposal/custom-url-resolution.md
```diff
@@ -0,0 +1,230 @@
...
+
+```css
+.upload-icon {
+  background-url: url('data:image/svg+xml;base64,Jggg==');
+}
+```
+
+## Function
+
+This proposal introduces logic for a new function: `sass-url("...")`, this function allow users to define a url that gets rewritten based on the config to ensure it can be loaded correctly from the browser.
```
nex3, line 57: This is the sort of thing that would be better in the summary. It does introduce the concept of a `sass-url` function (that's good!), but it's mostly just a nonspecific explanation of its general purpose. It's also not clear what "the config" refers to here.

nex3 (changes requested):
I've added a few comments. These should be plenty to work on for now; I'll take a look again once you've addressed them.

stof:
@DeMoorJasper are you still planning to finish this proposal ?

Issue #2904, opened 2020-09-16
Support all <angle> units in color functions


nex3:
- [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/angle-units.md) (https://github.com/sass/sass/pull/2953)

**Phase 1**

- [x] [Specs](https://github.com/sass/sass-spec/issues/1607) (https://github.com/sass/sass-spec/pull/1608)
- [x] [Documentation](https://github.com/sass/sass-site/issues/511) (https://github.com/sass/sass-site/pull/512)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1174) (https://github.com/sass/dart-sass/pull/1175)
- [x] Update canonical spec (#2974)

**Phase 2**

- [x] [Specs](https://github.com/sass/sass-spec/issues/1607) (https://github.com/sass/sass-spec/pull/1831)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1174) (https://github.com/sass/dart-sass/pull/1818)
- [x] [Documentation](https://github.com/sass/sass-site/issues/511) (https://github.com/sass/sass-site/pull/683)

**Phase 3**

- [x] Spec (https://github.com/sass/sass/pull/4139)
- [x] [Tests](https://github.com/sass/sass-spec/issues/1607) (https://github.com/sass/sass-spec/pull/2088)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1174) (https://github.com/sass/dart-sass/pull/2656)
- [ ] [Documentation](https://github.com/sass/sass-site/issues/511)

---

Currently, Sass's color functions (including both the existing functions and the functions added by https://github.com/sass/sass/pull/2835) only support `deg` as a hue unit. CSS specifies, and browsers support, a [wider variety of units](https://www.w3.org/TR/css3-values/#angles), including `grad`, `rad`, and `turn`. We should support all these units, internally converting them to `deg`.

Because `hsl()`'s and `*-color()`'s `$hue` arguments currently accept *any* unit, not just a unitless number or `deg`, this will require an intermediate deprecation step where we forbid units other than `deg` for hue values. As part of this, we can also tighten up the unit requirements for `$saturation` and `$lightness` to require either no unit or `%`.

nex3:
The proposal has landed, giving it two weeks for public comment.

nex3:
This is now waiting on Dart Sass 2.0.0

PR #2951, opened 2020-11-17
Add some basic spec text for units

nex3:
This lays some groundwork that will be useful for #2904.

### DIFF
--- a/spec/types/number.md
+++ b/spec/types/number.md
@@ -1,0 +1,81 @@
+# Numbers
+
+## Table of Contents
+
+* [Definitions](#definitions)
+  * [Conversion Factors](#conversion-factors)
+  * [Compatible Units](#compatible-units)
+* [Procedures](#procedures)
+  * [Converting a Number to a Unit](#converting-a-number-to-a-unit)
+
+## Definitions
+
+### Conversion Factors
+
+Certain units have conversion factors that define how they can be converted to
+and used with other related units. A conversion factor is itself a Sass number.
+The following conversion factors are defined:
+
+> These are intended to be kept in sync with the units in [CSS Values and Units]
+> that are fixed in relation to each other. Each conversion factor is the
+> equivalent of `1${unit}` in a canonical unit.
+
+[CSS Values and Units]: https://www.w3.org/TR/css-values-3/
+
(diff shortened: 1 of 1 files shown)


spec/types/number.md
```diff
@@ -0,0 +1,81 @@
...
+### Converting a Number to a Unit
+
+This algorithm takes a SassScript number `number` and a unit `unit`. It's
+written "convert `number` to `unit`" or "convert `number` to `unit` allowing
+unitless". It returns a number with the given unit. The `unit` argument must
+have a conversion factor.
+
+* If `number` has no units:
+
+  * If this procedure allow unitless, return `number` with unit `unit`.
```
Awjin:
```suggestion
  * If this procedure allows unitless, return `number` with unit `unit`.
```

Awjin (approved):

spec/types/number.md
```diff
@@ -0,0 +1,81 @@
...
+* `kHz`: 1000 `Hz`
+
+* `dppx`: 1 `dppx`
+* `dpi`: 1/96 `dppx`
+* `dpcm`: 2.54/96 `dppx`
+
+### Compatible Units
+
+Two units are said to be "compatible" if their [conversion factors] have the
+same unit.
```
stof, line 51: shouldn't a unit always be compatible with itself, even when it does not have a conversion factor ? For instance `%` does not have one.

spec/types/number.md
```diff
@@ -0,0 +1,81 @@
...
+unitless". It returns a number with the given unit. The `unit` argument must
+have a conversion factor.
+
+* If `number` has no units:
+
+  * If this procedure allows unitless, return `number` with unit `unit`.
+
+  * Otherwise, throw an error.
+
+* Let `number-factor` be `number`'s unit's conversion factor. If `number`
```
stof, line 70: shouldn't this procedure have a case saying that if `number`'s unit is `unit`, the number is returned as-is (even if that unit does not have a conversion factor) ?
Enforcing errors for no-op conversions even for units without a conversion factor would make the specs harder to write, as they would have to handle the case of 2 numbers having the same unit before being able to say

> convert `number1` to `number2`'s unit

PR #2970, opened 2020-12-04
Don't prescribe order of elements in deep-merged map.

Awjin:


nex3:
Since this proposal has already been implemented, I don't think it makes sense to modify it retroactively. I'd just modify the specification itself in `spec/`—probably by modifying `merge()` and making `deep-merge()` call out to `merge()` to preserve its behavior.

Also, I think we should be more explicit about exactly what behavior is and isn't guaranteed. I think it would be really strange if `map-merge($map, ())` shuffled the key orders, so maybe something like guaranteeing that unmodified keys from both maps stay in the same relative order?

Awjin:
Closing this in light of compelling use case for the opposite behavior (https://github.com/sass/dart-sass/issues/1385).

PR #2976, opened 2020-12-18
Specify invalid function names

nex3:


jathak (approved):

spec/syntax.md
```diff
@@ -26,6 +27,14 @@ A *source file* is a Sass abstract syntax tree along with an absolute URL, known
 as that file's *canonical URL*. A given canonical URL cannot be associated with
 more than one source file.
 
+### Vendor Prefix
+
+Some identifiers have with a *vendor prefix*, which is an initial substring
```
stof: should be `have a vendor prefix`, without the extra `with`

spec/at-rules/function.md
```diff
@@ -34,6 +34,12 @@ To execute a `@function` rule `rule`:
   [the current module]: ../spec.md#current-module
   [the current import context]: ../spec.md#current-import-context
 
+* If `name` is `calc`, `element`, `expression`, `url`, `and`, `or`, or `not`, or
```
stof: shouldn't this be moved up, before setting the name in the import context ?

nex3:
Good catches, thanks @stof!

PR #2977, opened 2020-12-18
Update meta.md

joshbruce:
(Long-time user, first time sending PR.)

Doing more complex work with Sass - beyond base syntax and nesting.

Found it odd in the documentation for `sass:meta` that a method for `mixin` was in the `function` section.

Moved it just in case I wasn't the only one. 

Not sure if there's a way to view the change locally as I need to get back to other things.

Thanks!


nex3:
Thanks for the thought! The organization is actually correct as-is, though... `meta.mixin-exists()` may have to do with mixins, but it *is a* function, whereas `meta.load-css()` *is a* mixin. That last section is specifically for mixins.

Issue #818, opened 2013-06-24
First-class calc


c-johnson:
* [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/first-class-calc.md)
* [x] [Specs](https://github.com/sass/sass-spec/pull/1688)
* [x] [Dart Sass](https://github.com/sass/dart-sass/pull/1452)
* [x] [Documentation](https://github.com/sass/sass-site/pull/575)
* [x] [Embedded protocol](https://github.com/sass/embedded-protocol/pull/74)
* [x] [Dart Sass embedded compiler](https://github.com/sass/dart-sass-embedded/pull/55)
* [x] [Specification](https://github.com/sass/sass/pull/3141)
* [x] JS API
   * [x] [Proposal](https://github.com/sass/sass/blob/main/proposal/calculation-api.d.ts)
   * [x] [Tests](https://github.com/sass/sass-spec/issues/1891) (https://github.com/sass/sass-spec/pull/1918)
   * [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1923) (https://github.com/sass/dart-sass/pull/1988)
   * [x] [Embedded Host](https://github.com/sass/embedded-host-node/issues/213) (https://github.com/sass/embedded-host-node/pull/237)
   * [x] Merge into canonical specification (https://github.com/sass/sass/pull/3622)

---

The following scss code does not work:

$var: 10px;
height: calc(100% - $var);

Tested with latest version (3.2.9).

Snugug:
You need to interpolate the `$var` in order to have it print out the 10px there:

``` scss
$var: 10px;
height: calc(100% - #{$var});
```

chriseppstein:
@nex3 is this intentional?

nex3:
> @nex3 is this intentional?

More or less. `calc` is considered to be a type of string literal, similarly to how `url` is treated. We could theoretically include a parser for the `calc` mini-grammar that would have explicit rules about how variables work, but then variables would work differently in that context than any other context, which may be unexpected.

I'm going to mark this as a feature request and leave it open until we come to a decision about what behavior we'd ideally like for `calc`.

ghost:
I think that calc should be treated as a special case. If you do calc(100% - $var) and $var = 5%, it should just output 95% without the calc because it's unnecessary, but if $var is 100px, it should still output the calc.

chriseppstein:
[...]

PR #2983, opened 2020-12-31
[First-Class Calc] Add an initial draft

nex3:
See #818

Note that this is not expected to be the finalized proposal, and it
has a number of outstanding TODOs. However, since it's a highly
impactful proposal, I want to get a draft committed so we can start
sending it around to stakeholders, getting feedback, and iterating.


proposal/first-class-calc.md
```diff
@@ -0,0 +1,536 @@
...
+interpolation and continuing to treat those values as opaque strings, and only
+using the new parsing logic for calcs that contain no interpolation. This is
+maximally backwards-compatible and it doesn't require interpolated calcs to be
+reparsed after interpolation.
+
+[`min()` and `max()` functions]: ../accepted/min-max.md
+
+Here are some alternatives we considered:
+
+1. Re-parsing a calc that contains interpolation once the interoplation has been
```
stof: typo: `interpolation` vs `interoplation`

proposal/first-class-calc.md
```diff
@@ -0,0 +1,536 @@
...
+
+> The `CalcArgument` production provides backwards-compatibility with the
+> historical use of interpolation to inject SassScript values into `calc()`
+> expressions. Because interpolation could inject any part of a `calc()`
+> expression regardless of syntax, for full compatibility it's necessary to
+> parse it very expansively.
+>
+> Note that the interpolation in the definition of `CalcValue` is only reachable
+> from a `CssMinMax` production, *not* from `CalcExpression`. This is
+> intentional, for backwards-compatibility with the existing `CssMinMax` syntax.x
```
stof: there's an extra `.x` here

proposal/first-class-calc.md
```diff
@@ -0,0 +1,536 @@
...
+}
+```
+
+Unless otherwise specified, when this specification creates a calc, its name is
+"calc".
+
+### Operations
+
+A calc follows the default behavior of all SassScript operations, except that it
+throws an error if used as an operand of a unary or binary `+` or `-` operation.
```
stof: what about the `*` operator ?

proposal/first-class-calc.md
```diff
@@ -0,0 +1,536 @@
...
+
+This algorithm takes a calc `calc` and returns a number or a calc.
+
+> This algorithm is intended to return a value that's CSS-semantically identical
+> to the input.
+
+* Let `arguments` be the result of [simplifying](#simplifying-a-calcvalue) each
+  of `calc`'s arguments.
+
+* If `calc`'s name is `"calc"`, `arguments` must contain a single argument. If
```
stof: This `must` here looks weird to me. It could be interpreted 2 ways:

- it triggers an error if not satisfied (the usual meaning of `must` in a spec)
- it must be valid to satisfy the `if` here.

From the explanation in the PR, I understand that you mean the second case. But that would be better written as a condition in the `if` instead.

proposal/first-class-calc.md
```diff
@@ -0,0 +1,536 @@
...
+  that argument is a number or calc, return it.
+
+* If `calc`'s name is `min`, `max`, or `clamp` and `arguments` are all numbers
+  whose units are mutually [compatible], return the result of calling
+  [`math.min()`], [`math.max()`], or `math.clamp()` (respectively) with those
```
stof, line 362: no link for `math.clamp` here ? Why ?

proposal/first-class-calc.md
```diff
@@ -0,0 +1,536 @@
...
+* Let `left` and `right` be the result of serializing the left and right values,
+  respectively.
+
+* If the operator is `"*"` or `"/"` and the left value is a `CalcOperation` with
+  operator `"+"` or `"-"`, emit `"("` followed by `left` followed by `")"`.
+  Otherwise, emit `left`.
+
+* Emit `" "`, then the operator, then `" "`.
+
+* If the operator is `"*"` or `"/"` and the right value is a `CalcOperation`
```
stof: I think this is not sound when the left or right values are interpolated `calc`s. To be safe, parenthesis should be added in such cases (similar to the TODO you have in the simplification algorithm)

proposal/first-class-calc.md
nex3: The default behavior for `*`, `/` and `%` is to throw an error. `+` and `-` are special in that they default to converting their operands to strings.

proposal/first-class-calc.md
nex3: Added a TODO.

proposal/first-class-calc.md
nex3: It actually means "the parser only allows a `calc` with name `"calc"` that has one argument". I'll make that more explicit.

proposal/first-class-calc.md
nex3: It hasn't been added to the math spec yet.

proposal/first-class-calc.md
```diff
@@ -0,0 +1,541 @@
...
+For example:
+
+* `calc(1px + 10px)` will return the number `11px`.
+
+* Similarly, if `$length` is `10px`, `calc(1px + $length)` will return `11px`.
+
+* However, `calc(1px + 10%)` will return the calc `calc(1px + 10%)`.
+
+* If `$length` is `calc(1px + 10%)`, `calc(1px + $length)` will return
+  `calc(2px + $length)`.
```
jathak:
```suggestion
  `calc(2px + 10%)`.
```

proposal/first-class-calc.md
```diff
@@ -0,0 +1,541 @@
...
+### `SpecialFunctionExpression`
+
+This proposal replaces the definition of [`SpecialFunctionName`] with the
+following:
+
+[`SpecialFunctionName`]: ../spec/syntax.md#specialfunctionexpression
+
+<x><pre>
+**SpecialFunctionName**¹      ::= VendorPrefix? ('element(' | 'expression(')
+&#32;                           | VendorPrefix `calc('
```
jathak:
```suggestion
&#32;                           | VendorPrefix 'calc('
```

proposal/first-class-calc.md
```diff
@@ -0,0 +1,541 @@
...
+
+The grammar for this production is:
+
+<x><pre>
+**CalcExpression** ::= `calc(`¹ CalcArgument ')'
+**ClampExpression** ::= `clamp(`¹ CalcArgument ',' CalcArgument ',' CalcArgument ')'
+**CalcArgument**²  ::= InterpolatedDeclarationValue | CalcSum
+**CalcSum**     ::= CalcProduct (('+' | '-')³ CalcProduct)?
+**CalcProduct** ::= CalcValue (('*' | '/') CalcValue)?
+**CalcValue**   ::= '(' CalcArgument ')
```
jathak:
```suggestion
**CalcValue**   ::= '(' CalcArgument ')'
```

jathak (approved):
LGTM overall, though I wonder if it makes sense to refer to the data type by a different name than just `calc` to avoid confusion between ones named "calc" and ones with other names (even just calling the type by the full word "calculation" could make things a bit clearer).

stof:
> LGTM overall, though I wonder if it makes sense to refer to the data type by a different name than just `calc` to avoid confusion between ones named "calc" and ones with other names (even just calling the type by the full word "calculation" could make things a bit clearer).

I tend to agree here

proposal/first-class-calc.md
```diff
@@ -0,0 +1,541 @@
...
+using the new parsing logic for calcs that contain no interpolation. This is
+maximally backwards-compatible and it doesn't require interpolated calcs to be
+reparsed after interpolation.
+
+[`min()` and `max()` functions]: ../accepted/min-max.md
+
+Here are some alternatives we considered:
+
+1. Re-parsing a calc that contains interpolation once the interpolation has been
+   resolved, and using the result as in the calc object rather than an unquoted
```
Awjin:
```suggestion
   resolved, and using the result as a calc object rather than an unquoted
```

proposal/first-class-calc.md
```diff
@@ -0,0 +1,541 @@
...
+
+  * If `left` and `right` are both numbers with [compatible] units, return
+    `left + right` or `left - right`, respectively.
+
+    > TODO: Should we throw an error here for units we can prove are actively
+    > incompatible? For example, unitless numbers are always incompatible with
+    > numbers of any units, and numbers across different known types (length +
+    > time) are also invalid.
+
+    > TODO: Should we try to simplify `calc(1px + 1% + 1px)`?
```
Awjin, line 411: For here and other similar TODOs below: what are the cons of simplifying this as much as possible? It seems to me that simplifying to `calc(2px + 1%)` would be the most logically consistent.

proposal/first-class-calc.md
```diff
@@ -0,0 +1,541 @@
...
+
+1. Re-parsing a calc that contains interpolation once the interpolation has been
+   resolved, and using the result as in the calc object rather than an unquoted
+   string. For example, `calc(#{"1px + 2px"})` would return `3px` rather than
+   `calc(1px + 2px)`. However, doing another parse at evaluation-time would add
+   substantial complexity and some amount of runtime overhead. The
+   return-on-investment would also be inherently limited, since we're planning
+   on gradually transitioning users away from interpolation in `calc()` anyway.
+
+2. Treating interpolation as a `CalcValue` that participates in the normal
```
Awjin: These are formally defined below, but it could be helpful to introduce them earlier on and use the terms throughout.

proposal/first-class-calc.md
```diff
@@ -0,0 +1,541 @@
...
+stream starts with an identifier with value `calc` (ignoring case) followed
+immediately by `(`.
+
+The grammar for this production is:
+
+<x><pre>
+**CalcExpression** ::= `calc(`¹ CalcArgument ')'
+**ClampExpression** ::= `clamp(`¹ CalcArgument ',' CalcArgument ',' CalcArgument ')'
+**CalcArgument**²  ::= InterpolatedDeclarationValue | CalcSum
+**CalcSum**     ::= CalcProduct (('+' | '-')³ CalcProduct)?
```
stof: shouldn't the right hand side actually be `CalcSum` to support `1px + 1% + 1px` ? the current syntax would require braces to achieve that as `1px + (1% + 1px)`. (making the `CalcProduct` a single `CalcValue` containing a `CalcArgument` enclosed in braces).

proposal/first-class-calc.md
```diff
@@ -0,0 +1,541 @@
...
+**CalcProduct** ::= CalcValue (('*' | '/') CalcValue)?
```
stof: same here about using `CalcProduct` in the right hand side.

proposal/first-class-calc.md
stof: This would require a more complex structure for the AST, because the `1px + 1% + 1px` would be represented as `CalcOperation(1px, '+', CalcOperation(1%, '+', 1px))` currently while simplifying such case would require recombining the sum to try another variant.

proposal/first-class-calc.md
nex3: I couldn't think of a good way to summarize them here, so I added a link to their formal definitions. Let me know if you have a better idea of how to describe them.

proposal/first-class-calc.md
nex3: You're right that this is wrong, but I actually intended to make the second clause repeated rather than using recursion to repeat them (that's the structure that's assumed in the Semantics section below).

proposal/first-class-calc.md
nex3: Alternately, it would require complexity in the simplification logic to whether `CalcOperation`s can be "flattened" usefully. The CSS spec actually does cover this sort of simplification logic, so there's prior art, but it's definitely more complex.

Awjin (approved):
LGTM

proposal/first-class-calc.md
stof: then the simplification algorithm needs to be updated, as it was describing only operations with 2 terms.

stof: Actually, `CalculationOperation` only has `left` and `right` so the necessary change is explaining how this repeated clause gets turned into the struct.

nex3: I think it works as written, the `CalcSum` and `CalcProduct` semantics sections assemble the lists into a sequence of `CalcOperations` with two operands each.

Issue #2931, opened 2020-10-27
The unit of the variable for @for is inconsistent between implementations


stof:
With this input

```scss
test {
    @for $i from 12px to 13px {
        content: unit($i);
    }
}
```

I get the following output on ruby-sass 3.7.4 and libsass 3.5.5 (these are the versions available on https://www.sassmeister.com/):

```css
test {
  content: "px";
}
```

But I get this output on dart-sass 1.26.10 and 1.26.11

```css
test {
  content: "";
}
```

IMO, the behavior of ruby-sass and libsass is more logical, as the values for `from` and `to` are handled as unit numbers (including coercing compatible units).

stof:
Looking at the behavior with different units, the behavior is even more inconsistent:

- dart-sass coerce to the unit of `to` (but then looses the unit for `$i`, making it hard to use as we might have weird numbers)
- ruby-sass coerce to the unit of `from`
- libsass does not coerce compatible units. It requires the same unit for `from` and `to`

Given that `from` and `to` are required to be integers *after* converting them to the common unit, it makes it impossible to write portable sass code when units are not the same.

stof:
Looking at `@for` in sass-spec, it always uses unitless numbers for non-error tests. There is a single test using unit, which covers the case of incompatible units (`%` and `px`) but the case of compatible units is not tested.

jathak:
Good catch! It seems like Ruby Sass's behavior here is probably the best option here, so we'll write up a [fast-track proposal](https://github.com/sass/sass/blob/master/CONTRIBUTING.md#fast-track) to formally specify this, and then update the behavior of Dart Sass.

babakfp:
> logical

Just curious, why you used `px` (for loop parameters)?

babakfp:
> With this input
> 
> ```css-scss
> test {
>     @for $i from 12px to 13px {
>         content: unit($i);
>     }
> }
> ```
> 
> I get the following output on ruby-sass 3.7.4 and libsass 3.5.5 (these are the versions available on https://www.sassmeister.com/):
> 
> ```css
> test {
>   content: "px";
> }
> ```
> 
> But I get this output on dart-sass 1.26.10 and 1.26.11
> 
> ```css
> test {
>   content: "";
> }
> ```
> 
> IMO, the behavior of ruby-sass and libsass is more logical, as the values for `from` and `to` are handled as unit numbers (including coercing compatible units).

Just curious, why you used `px` (for loop parameters)?

[...]

PR #2986, opened 2021-01-04
Spec the @for at-rule

stof:
As a basis for the fast-track proposal in #2931, this writes specs for the `@for` at-rule.

I intentionally skipped any mention of the unit conversion for `from` and `to`, because dart-sass, libsass and ruby-sass have 3 different behaviors for it (which is precisely what #2931 is meant to solve).
Implementing strictly the spec as written here would probably produce the libsass behavior (relying on the fact that adding a unitless number to a number with unit is allowed in Sass and produces a number with that unit). Specifying the ruby-sass behavior is a matter of adding a step ``Convert `to` to `from`'s unit`` before asserting that both values are integers (but currently, dart-sass does the conversion the other way, which is why I prefer keeping this line as part of the proposal instead). Spec'ing the existing dart-sass behavior would make the spec more complex for something that would have to be reverted by the proposal anyway (converting numbers to strip the unit for the loop variable).

### DIFF
--- a/spec/at-rules/for.md
+++ b/spec/at-rules/for.md
@@ -1,0 +1,61 @@
+# `@for`
+
+## Table of Contents
+
+* [Syntax](#syntax)
+* [Semantics](#semantics)
+
+## Syntax
+
+<x><pre>
+**ForRule**            ::= '@for' [PlainVariable][] FromDeclaration
+&#32;                      (ToDeclaration | ThroughDeclaration) ForBlock
+**FromDeclaration**    ::= 'from' Expression
+**ToDeclaration**      ::= 'to' Expression
+**ThroughDeclaration** ::= 'through' Expression
+**ForBlock**           ::= '{' Statements '}'
+</pre></x>
+
+[PlainVariable]: ../variables.md#syntax
+
+## Semantics
+
+To execute a `@for` rule `rule`:
+
+* Let `from` be the result of evaluating the expression in `FromDeclaration`.
+
+* If `rule` has a `ToDeclaration`:
+
+  * Let `to` be the result of evaluating the expression in `ToDeclaration`.
+
+  * Let `exclusive` be `true`.
+  
+* Otherwise:
+
+  * Let `to` be the result of evaluating the expression in `ThroughDeclaration`.
+
+  * Let `exclusive` be `false`.
+  
+* If `from` and `to` aren't numbers, throw an error.
+
+* If `from` and `to` aren't integers, throw an error.
+
+* If `from` is greater than `to`, set `direction` to `-1`. Otherwise, set
+  `direction` to `1`.
+
+* If `exclusive` is `false`, set `to` to `to + direction`.
+
+* Let `i` be `from`.
+
+* While `i` is not equal to `to`:
+
+  * Let `scope` be a new [scope].
+
+  * Add a variable with `rule`'s `VariableName` as its name and `i` as its value
+    to `scope`.
+  
+  * Execute the `ForBlock`'s statements in `scope`.
+  
+  * Set `i` to `i + direction`.
+
+  [scope]: ../variables.md#scope
--- a/spec/variables.md
+++ b/spec/variables.md
@@ -9,21 +9,22 @@
 * [Semantics](#semantics)
   * [Executing a Variable Declaration](#executing-a-variable-declaration)
   * [Evaluating a Variable](#evaluating-a-variable)
-  
+
 ## Syntax
 
 <x><pre>
-**Variable**            ::= '$' Identifier | NamespacedVariable
+**Variable**            ::= PlainVariable | NamespacedVariable
+**PlainVariable**       ::= '$' Identifier
 **NamespacedVariable**  ::= Identifier '.$' [PublicIdentifier][]
 **VariableDeclaration** ::= Variable ':' Expression ('!global' | '!default')*
 </pre></x>
 
 [PublicIdentifier]: modules.md#syntax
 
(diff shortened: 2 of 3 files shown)


spec/variables.md
```diff
@@ -9,19 +9,20 @@
...
   * [Executing a Variable Declaration](#executing-a-variable-declaration)
   * [Evaluating a Variable](#evaluating-a-variable)
-  
+
 ## Syntax
 
 <x><pre>
-**Variable**            ::= '$' Identifier | NamespacedVariable
+**Variable**            ::= VariableName | NamespacedVariable
+**VariableName**        ::= '$' Identifier
```
nex3: To make the difference clearer, I think this should probably be something like `NamespacelessVariable` or `LocalVariable`.

spec/variables.md
```diff
@@ -9,19 +9,20 @@
...
 **NamespacedVariable**  ::= Identifier '.$' [PublicIdentifier][]
 **VariableDeclaration** ::= Variable ':' Expression ('!global' | '!default')*
 </pre></x>
 
 [PublicIdentifier]: modules.md#syntax
 
 No whitespace is allowed after the `$` or before or after the `.$` in
-`Variable`. Each of `!global` and `!default` is allowed at most once in
+`VariableName`. Each of `!global` and `!default` is allowed at most once in
```
nex3: `.$` doesn't appear in `VariableName`, so this sentence should probably be rewritten slightly.

nex3 (commented):
Thanks for doing the groundwork here!

spec/variables.md
stof: After writing this, I saw that https://github.com/sass/sass/pull/2850#discussion_r432189807 suggested `PlainVariable`.

`LocalVariable` seems wrong to me, as it could be the name of a global variable.

spec/variables.md
stof: hmm, indeed. I missed the part about ``.$`` which should refer to `NamespacedVariable`

spec/at-rules/for.md
```diff
@@ -22,38 +22,38 @@
...
-* If `from` and `to` are equal, stop here
+* If `exclusive` is `false`, set `to` to `to + direction`.
 
-* Set `i` to `from`
+* Let `i` be `from`.
 
 * While `i` is not equal to `to`:
 
-  * Declare a variable with `rule`'s `VariableName` as name and value `i` in the `rule`'s `ForBlock`'s scope
+  * Let `scope` be a new [scope].
```
stof, line 52: do we actually create a new scope at each iteration ?

spec/at-rules/for.md
```diff
@@ -22,38 +22,38 @@
 
 To execute a `@for` rule `rule`:
 
-* Let `from` be the result of evaluating the expression in `FromDeclaration`
-
-* Assert that `from` is a number
```
stof: I validated the number here explicitly to match the order of error reporting in implementations (even though I wrote it the wrong way)

spec/at-rules/for.md
```diff
@@ -22,38 +22,38 @@
...
   
 * Otherwise:
 
-  * Let `to` be the result of evaluating the expression in `ThroughDeclaration`
+  * Let `to` be the result of evaluating the expression in `ThroughDeclaration`.
 
-  * Let `exclusive` be `false`
+  * Let `exclusive` be `false`.
   
-* Assert that `to` is a number
```
stof: having separate steps for number and integers was done on purpose due to the unit conversion that will happen between them (in the fast-track proposal specifying the handling of units).

stof:
@nex3 by curiosity, what is this weird `<x>` tag around `<pre>` blocks ?

spec/variables.md
stof: I will let you decide which name I should use before updating my PR.

spec/at-rules/for.md
stof: @nex3 was there anything wrong with my previous proposal ? According to the spec for variables, the curly braces of the `ForBlock` are already introducing a new scope.

nex3: I don't think it's actually detectable either way right now, but if we do ever add full closure semantics to Sass we'll want a closure to capture the *current* value of the iteration variable rather than the *final* value (that is, we'll want it to behave like `for`/`let` in JS and not like `for`/`var`). This text makes that explicit.

spec/at-rules/for.md
nex3: If a stylesheet is going to produce an error, implementations have a lot of freedom in which specific error to present to the user and how to present it. In this case, if `from` isn't a number *and* `ToDeclaration` has an evaluation error, implementations are free to surface either error first and thus to check for errors in either order. This flexibility of implementation helps keep the spec simple without ruling out optimizations or cosmetic differences in implementations.

spec/at-rules/for.md
nex3: All right, feel free to add it back.

nex3:
> @nex3 by curiosity, what is this weird `<x>` tag around `<pre>` blocks ?

A `<pre>` block on its own puts the Markdown parser into a state where inline Markdown is treated as plain text, but only if the `<pre>` isn't in another HTML element. So in order to use Markdown links and emphasis and so on in a `<pre>` context we have to wrap it in some other HTML tag. We use `<x>` as a short option that will get stripped before rendering.

nex3:
I updated my markdown-link-check fork but it still failed to fetch GitHub URLs :confused:. I'm re-running tests (again) to see if that clears it out.

nex3:
Oh, I missed #2994.

PR #2999, opened 2021-01-11
Specify the behavior of for rules regarding units

stof:
The behavior matches the behavior of ruby-sass, as discussed in the issue.

https://github.com/sass/sass/issues/2931

### DIFF
--- a/spec/at-rules/for.md
+++ b/spec/at-rules/for.md
@@ -38,6 +38,10 @@ To execute a `@for` rule `rule`:
   
 * If `from` and `to` aren't numbers, throw an error.
 
+* Let `to` be the result of [converting] `to` to `from`'s unit allowing unitless.
+
+  [converting]: ../types/number.md#converting-a-number-to-a-unit
+
 * If `from` and `to` aren't integers, throw an error.
 
 * If `from` is greater than `to`, set `direction` to `-1`. Otherwise, set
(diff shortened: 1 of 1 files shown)


spec/at-rules/for.md
```diff
@@ -53,6 +57,8 @@ To execute a `@for` rule `rule`:
 
   * Add a variable with `rule`'s `VariableName` as its name and `i` as its value
     to `scope`.
+
+    > Note that this variable will have the same unit that `from`.
```
stof, line 61: The spec relies on the fact that the `+` operator for numbers allows adding a unitless number (`direction` in this case) to any number (and preserves the original unit). This note makes this case explicit.

This is what is mostly missing in dart-sass currently, due to the fact that it implements `i` with a Dart `num` rather than making it a `SassNumber` during the iteration (and so this behavior is not automatically inherited from `SassNumber`).

stof:
Other PRs for the fast-track proposal (sass-spec and dart-sass) will follow this evening. I started the work on them but sass-spec does not have conformant specs for `@for` yet so I actually started refactoring the `@for` specs first before doing the PR with new specs for the unit coercion.

nex3:
I expected this change to be part of the fast-track proposal, since it's different than the current behavior of Dart Sass (and LibSass I believe).

stof:
@nex3 this PR is precisely the spec change for the fast-track proposal. That's really the only necessary spec change.
I now opened the sass-spec and dart-sass PR, keeping the sass-spec PR as a draft one as it depends on my PR refactoring the `@for` specs to be conformant.

nex3 (approved):
Ah, gotcha, I misunderstood—I thought you were saying the fast-track proposal was coming later. In that case, this looks good.

Issue #2829, opened 2020-02-28
:is() should behave like :matches() with @extend


nex3:
Currently, if a `:matches()` selector is extended into `:not()`, it's replaced with additional `:not()` rules. For example, `selector-extend(":not(.c)", ".c", ":matches(.d, .e)")` returns `:not(.c):not(.d):not(.e)`. However, despite `:matches()` being an alias for `:is()`, the same is not true for the `:is()` selector. It should be.

nex3:
Actually, it looks like the problem is broader—we have a fair amount of `:matches()`-specific logic, and *none* of it is currently applied to `:is()`.

nex3:
Note that `:is()` is not yet supported by that name [in any browsers](https://caniuse.com/#search=%3Ais).

stof:
`:is()` is now supported in Firefox and Safari (and Chromium has it behind a flag)

PR #3014, opened 2021-02-19
Specify more details of the extend algorithm

nex3:
This lays down some specs which can then be updated for #2829.

### DIFF
--- a/spec/at-rules/extend.md
+++ b/spec/at-rules/extend.md
@@ -241,9 +241,7 @@ a selector list `extender` and returns a selector list.
 
   * For each compound selector `compound` or combinator in `complex`:
 
-    * If it's a combinator, add it to each selector in `complex-options`.
-
-    * Let `new-complex-options` be an empty list.
+    * If it's a combinator, add it to each selector in `options`.
 
     * For each simple selector `simple` in `compound`:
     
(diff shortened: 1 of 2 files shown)


spec/at-rules/extend.md
```diff
@@ -216,6 +221,118 @@ that includes CSS for *all* modules transitively used or forwarded by
...
+
+* Let `results` be an empty selector list.
+
+* For each complex selector `complex` in `extendee`:
+
+  * Let `options` be an empty complex selector.
+
+  * For each compound selector `compound` or combinator in `complex`:
+
+    * If it's a combinator, add it to each selector in `complex-options`.
```
jathak: Should `complex-options` here be `options`?

spec/at-rules/extend.md
```diff
@@ -216,6 +221,118 @@ that includes CSS for *all* modules transitively used or forwarded by
...
+
+    * Let `new-complex-options` be an empty list.
```
jathak: This is never added to or referenced from what I can tell

jathak (commented):
The spec for extending simple selectors and the syntax looks good, but I couldn't quite parse what was going on with the complex selector case.

nex3:
> The spec for extending simple selectors and the syntax looks good, but I couldn't quite parse what was going on with the complex selector case.

I added an example, does that help?

jathak (approved):
Yes, that helps. LGTM.

PR #3026, opened 2021-03-17
Fix the spec about resolving an import

stof:
Trying to resolve the file with a .sass extension is also attempted for imports (if we haven't returned yet). So this is not a `Otherwise` of the previous condition


Awjin (approved):
Good catch, thanks.

stof:
@nex3 @Awjin this spec also misses the fact that `@import "foo"` can be resolved as `foo/index.scss`. Is this a mistake or intentional ?

PR #3027, opened 2021-03-18
Make :is() behave like :matches()

### LINKED ISSUE #2829: :is() should behave like :matches() with @extend

nex3:
This is a low-risk, low-impact fast-track proposal.

Closes #2829


jathak (approved):
LGTM. I'll try to get a PR for the implementation out in the next week or so.

PR #3049, opened 2021-05-10
Run npm upgrade

nex3:


Awjin (approved):

jathak:
I think I found the issue: https://github.com/npm/cli/issues/2610

Even with git+https explicitly specified, npm decides to use SSH instead, which requires keys on the CI.

jathak:
Okay, I used the workaround mentioned [here](https://github.com/npm/cli/issues/2610#issuecomment-795189708) to force it to use HTTPS.

@nex3: This should be good to merge now

nex3:
Thanks for debugging!

Issue #2979, opened 2020-12-22
Define a JS API for handling debug messages and warnings


nex3:
* [x] [Proposal](https://github.com/sass/sass/blob/main/proposal/js-logger.d.ts)
* [x] [Tests](https://github.com/sass/sass-spec/pull/1726)
* [x] [Documentation](https://github.com/sass/sass-site/pull/584)
* [x] [Dart Sass](https://github.com/sass/dart-sass/pull/1515)
* [ ] Node Host

---

The JS API is currently missing a means of handling messages generated by the `@debug` and `@warn` directives, as well as deprecation warnings generated by the implementation itself. Node Sass has the ability to provide a custom function named `@warn` that handles warnings, but this is neither documented nor supported by Dart Sass. Dart Sass has the ability to pass in subclasses of [`Logger`](https://pub.dev/documentation/sass/latest/sass/Logger-class.html) to handle these callbacks. We should come up with a similar `Logger`-style API for JS.

See https://github.com/sass/dart-sass/issues/585

glenn2223:
Firstly, sorry if the below isn't what you expected. However, if you can point me in the right direction, I'm more than happy to help where I can (always looking to expand the skillsets).

---

I was thinking that, as these "at rules" don't actually stop compilation, any warning and debug messages should be returned with the `Result` of `render()` and `renderSync()`. However, I can redirect my thoughts to a Dart Sass Logger API if you wish 😄 

Further to my initial thought, the adjustments I was thinking were: (also included some quick defs examples)
- New interface that takes all functionality from `SassException`
- `SassException` will then extend `Error` and `SassProblem`
- `Result` has two new properties `warningMessages` and `debugMessages`

**New interface** - SassFlag or SassProblem (sorry, not important)
````ts
export interface SassFlag {
    message: string;
    formatted: string;
    line: number;
    column: number;
    status: number;
    file: string;
}
````

**SassException**
````ts
export interface SassException extends Error, SassFlag { }
````

**Result**
````ts
export interface Result {
    css: Buffer;
    map?: Buffer;
    stats: {
        entry: string;
        start: number;
        end: number;
        duration: number;
        includedFiles: string[];
    };
    warningMessages: SassFlag[];
    debugMessages: SassFlag[];
}
````

glenn2223:
Alternate suggestion for a logger (as originally requested). Again, to prevent duplication of code I'd strip most of `SassException` and make it extend the new `SassFlag`

````ts
/*
 * MODS
 */
[...]

Issue #3056, opened 2021-05-14
New JS API entrypoint


nex3:
* [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/new-js-api.d.ts)
* [x] [Tests](https://github.com/sass/sass-spec/pull/1716)
* [x] [Documentation](https://github.com/sass/sass/pull/3183)
* [x] [Dart Sass](https://github.com/sass/dart-sass/pull/1503)
* [ ] Node Host

---

Something @Awjin and I have discussed one on one but hasn't been captured on the issue tracker yet is the idea of creating new entrypoint functions for the JS API as part of the redesign of the custom importer and function APIs (#2509 and #2510). While we could potentially support new imports and functions in the existing `render()`/`renderSync()` entrypoints, there are a number of reasons why it might be better to have an entirely new entrypoint. (I'll tentatively call this entrypoint `compile()` to match the Dart Sass API.)

* As we improve various aspects of the API, it makes it totally clear to users which aspects are new and supported (those used by `compile()`) and which are old and deprecated (those used by `render()`). Users won't be able to accidentally mix and match and then get broken when support for the deprecated versions is dropped.

* We don't have to deal with tricky questions about how aspects of the old API interact with the new one. For example, we won't have to deal with users passing both old-style and new-style importers to the same `render()` call, and we won't have to figure out which source map options take precedence in which situations.

* The names `render()` and `renderSync()` imply that the synchronous version is "less default", when in fact for Dart Sass the synchronous version is actually *much more efficient*. In the new API, we can call the entrypoints `compile()` and `compileAsync()` instead.

* We can have `compileAsync()` return a promise rather than using the old-style error/value callback without making it a breaking change.

* Cross-compilation state of the sort described in #2745 won't work well with the old importer API, since it doesn't have a strong notion of "canonical URLs" that can be stable over time. By putting the new APIs in a new entrypoint, we'll be able to expose a version of that entrypoint on a state-preserving `Compiler` class without having to figure out how to handle old-style importers.

nex3:
This proposal is now officially out for review. It will remain that way for at least a month (potentially longer depending on discussion), at which point we'll move on to implementation.

nex3:
@alexander-akait I'm particularly interested in your feedback here (and on the other proposals), as the developer of the most widely-used Sass plugin. See https://sass-lang.com/blog/request-for-comments-new-js-api for a human-friendly summary of the proposed changes.

alexander-akait:
[...]

PR #3067, opened 2021-05-24
`Logger` proposal with `.d.ts` File

glenn2223:
A proposal for a JS `Logging` mechanism plus a `.d.ts` file for shipping with the npm package

Refs: #2979, sass/dart-sass#13


glenn2223:
Before I commit it, I assume you want the `.md` file removing

glenn2223:
@nex3, I assume you want the `.md` file removing (thought I'd ask rather than just removing it)

nex3:
> I assume you want the .md file removing (thought I'd ask rather than just removing it)

Yes, let's make the `.d.ts` canonical.

glenn2223:
> > I assume you want the .md file removing (thought I'd ask rather than just removing it)
> 
> Yes, let's make the `.d.ts` canonical.

Done

nex3:
FYI, I've added a TypeScript formatter and run it on your `.d.ts` to make it easier to review.

index.d.ts
```diff
@@ -0,0 +1,373 @@
+// TypeScript Version: 4.2
+
+export type ImporterReturnType =
+  | {file: string}
+  | {contents: string}
+  | Error
+  | null;
```
nex3: `Error` isn't a valid synchronous return type for an `Importer` function (they should just throw instead), and `null` is only a valid return type for importers that are going to call their `done` callback (and even then only if they're passed to `render()`, not `renderSync()`). I think we should more clearly encode the distinction between synchronous and asynchronous importers:

```ts
export type Importer = (
  url: string,
  prev: string
) => {file: string} | {contents: string};

export type AsyncImporter = Importer | (
  url: string,
  prev: string,
  done: (result: {file: string} | {contents: string} | Error)
) => void;
```

Then you can make `renderSync()` take `Importer`s and `render()` take `AsyncImporter`s.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+
+export type Importer = (
+  url: string,
+  prev: string,
+  done: (data: ImporterReturnType) => void
+) => ImporterReturnType | void;
+
+export interface LogData {
```
nex3: Add some interface-level documentation as well.

Remember that this isn't just normal API documentation; this is also a specification for how implementation behave. Multiple implementations will need to support this API, and it's important that they do so as consistently as possible. There should be no room for ambiguity in how a given chunk of JS + Sass will work.

(This is mostly important for the logger-related interfaces, because that's a new API that needs to be implemented from scratch. The existing APIs that you're defining types here for would *ideally* be fully specified as well, but it's a lot less important because they're generally already implemented and because they'll be supplanted by #3056 before too long anyway.)

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+  /**
+   * The message as-is in the file/data
+   */
+  message: string;
```
nex3: This should probably be passed as a positional parameter, to indicate that it's the most important part of each call.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+
+  /**
+   * A set of information related to the location of the error
+   */
+  source: SourceSpan;
```
nex3: This should be optional for `warn`. There are some warnings that don't have corresponding spans (such as a warning about how `render()` itself is being called).

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+
+  /**
+   * Is it a deprecation message
+   */
+  deprecation: boolean;
```
nex3: This field doesn't make sense for the `debug` function. There are enough differences between the two signatures that you should probably just make them both take inline options, rather than trying to share a single type that's passed to both.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+}
+
+export interface Logger {
+  /**
+   * A function that is called each time `@warn` is encountered
```
nex3: This is not a comprehensive specification. `warn()` may also be called for warnings generated by the implementation itself, and in fact the implementation *must* pass all warnings to `warn()` if it's specified.

Which gets to another point: what should the implementation do if one of these functions is *not* specified and it needs to emit a warning or a `@debug` message?

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+  file?: string;
+
+  /**
+   * A string to pass to compile.
+   *
+   * It is recommended that you use `includePaths` in conjunction with this so that sass can find files when using the @import directive.
+   *
+   * @default null
+   */
+  data?: string;
```
nex3: Although each of these is individually optional, at least one of them and optionally both *must* be passed by the caller. You can see [here](https://github.com/sass/embedded-host-node/blob/661fa06465e8cae1ecc518e11dc6dae24dccb573/lib/src/node-sass/render.ts#L13-L44) how @awjin represents this for the Node embedded host.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+   */
+  deprecation: boolean;
+}
+
+export interface Logger {
+  /**
+   * A function that is called each time `@warn` is encountered
+   * @param data All data relative to the `@warn` line
+   */
+  warn?(data: LogData): void;
```
nex3: `warn()` should also take an optional stack trace.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+
+  /**
+   * A function that is called each time `@debug` is encountered
+   * @param data All data relative to the `@debug` line
+   */
+  debug?(data: LogData): void;
+}
+
+export interface Options {
```
nex3: You'll probably need to split this into `Options` and `AsyncOptions` as well, which both inherit from `SharedOptions`.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+  /**
+   * Path to a file to compile.
+   *
+   * @default null
```
nex3: It's not obvious how this `@default` tag maps onto the behavior that's required for an implementation.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+  importer?: Importer | Importer[];
+
+  /**
+   * Holds a collection of custom functions that may be invoked by the sass files being compiled.
+   *
+   * @default undefined
+   */
+  functions?: {
+    [key: string]: (...args: types.SassType[]) => types.SassType | void;
+  };
```
nex3: Like importers, functions can be either synchronous or asynchronous. If they're asynchronous, the last argument is a `done()` callback.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+
+  /**
+   * The filename of error.
+   *
+   * In case file option was not set (in favour of `data`), this will reflect the value `stdin`.
+   */
+  file: string;
+}
+
+export interface SourceLocation {
```
nex3: Like the `Logger` classes, these source interfaces should be thoroughly specified.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+}
+
+export function render(
+  options: Options,
+  callback: (exception: SassException, result: Result) => void
+): void;
+export function renderSync(options: Options): Result;
+
+export namespace types {
+  abstract class SassType {}
```
nex3: Dart Sass doesn't export any object with this name, and I don't *think* Node Sass does either.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+
+  interface Null extends SassType {
+    NULL: Null;
+  }
+
+  const Null: Null;
```
nex3: Because you're specifying the `Null` interface separately from the corresponding constant, I'm not sure if TypeScript will understand that `types.Null.NULL` is an instance of `types.Null`. You may need to just make this a `class` without a (visible) constructor.

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+
+  class Number implements SassType {
```
nex3: The linter is complaining that these are unused. Do you need to export them?

index.d.ts
```diff
@@ -0,0 +1,373 @@
...
+    setR(value: number): void;
+    getG(): number;
+    setG(value: number): void;
+    getB(): number;
+    setB(value: number): void;
+    getA(): number;
+    setA(value: number): void;
+  }
+
+  class List<T extends SassType = SassType> implements SassType {
```
nex3: Sass doesn't consider its lists or maps to be typed, so I don't think this API should either.

nex3 (changes requested):
Thanks for taking the time to write out the entire API!

index.d.ts
glenn2223: So they should just be `List<unknown>` and `Map<unknown>`, is that right?

index.d.ts
glenn2223: I gave the option for `render` and `renderSync` to support both async and sync calls. From what you're saying here, it can't so I'll update

index.d.ts
glenn2223: Added some documentation to each line - ready for review

index.d.ts
glenn2223: Done

index.d.ts
glenn2223: Separated them into `DebugData` and `WarnData`

index.d.ts
glenn2223: Thanks for the sample - every day is a learning day 😁

index.d.ts
glenn2223: This way just an easy way to add it to the `functions` parameter - it's not actually exported. 

The alternative is `types.Null | types.Number | types.String | ....`

How would you like to proceed?

index.d.ts
glenn2223: Sorry, I don't follow

glenn2223:
Did you want the interfaces to be types? It's the same but they're more strict ([reference](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces))

nex3:
Thanks @glenn2223! I can take it from here, starting with splitting out the TS definition of the existing API into the `spec` directory so that the logger proposal can go into the`logger` directory.

nex3:
Closing this out in favor of https://github.com/sass/sass/pull/3074.

PR #3069, opened 2021-05-27
Add proposal for new JS API.

Awjin:
This proposal specifies the behavior of the compile functions in the new JS API.

Partially addresses https://github.com/sass/sass/issues/3056.

### DIFF
--- a/proposal/new-js-api.d.ts
+++ b/proposal/new-js-api.d.ts
@@ -204,8 +204,8 @@ export interface CompileResult {
  * Compiles the Sass file at `path`.
  *
  * To perform the compilation, the compiler executes the [compiling a path]
- * procedure on `path`. The compiler must respect the configuration specified by
- * the `options` object.
+ * procedure on `path`, with `options.loadPaths` as the import paths. The
+ * compiler must respect the configuration specified by the `options` object.
  *
  * [compiling a path]: ../spec/spec.md#compiling-a-path
  *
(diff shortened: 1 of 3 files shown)


proposal/new-js-api.d.ts
```diff
@@ -200,59 +201,85 @@ export interface CompileResult {
...
+ * object composed as follows:
+ *
+ * - Let `css` be the CSS emitted by the Sass compilation. Set
+ *   `CompileResult.css` to `css`.
+ *
+ * - Set `CompileResult.includedFiles` to a list of unique canonical URLs of
+ *   source files [loaded] during the compilation. The order of URLs is not
+ *   guaranteed.
+ *   - If `options.url` is set, include it in the list.
+ *   - Otherwise, do not include `source`.
```
nex3:
```suggestion
 *   - Otherwise, do not include a URL for `source`.
```

proposal/new-js-api.d.ts
```diff
@@ -0,0 +1,297 @@
...
+  css: string;
+  includedUrls: URL[];
+  sourceMap?: RawSourceMap;
+}
+
+/**
+ * Compiles the Sass file at `path`.
+ *
+ * To perform the compilation, the compiler executes the [compiling a path]
+ * procedure on `path`. The compiler must respect the configuration specified by
```
nex3: I would also say "with `options.loadPaths` as the import paths" here.

nex3 (approved):

PR #3074, opened 2021-06-04
Add a proposal for a JS Logger API

nex3:
See sass/sass#2979


proposal/js-logger.d.ts
```diff
@@ -0,0 +1,206 @@
...
+ * (whether from a `@warn` or `@debug` rule or from a Sass-internal warning),
+ * they do so by printing that warning directly to the standard error stream.
+ * This is a good default in that it's likely to surface the warnings to the end
+ * user, but it's not adequate for all use-cases.
+ *
+ * For example, a build system might want to integrate Sass's messages into its
+ * own messaging infrastructure. An IDE that's driving Sass through the JS API
+ * might want to highlight warnings in the file the user is editing. If Sass is
+ * ever run in the browser, the web page might want to surface warnings as
+ * components on the page. As such, there's need for an API to provide API users
```
Awjin: Nit:

```suggestion
 * components on the page. As such, there's need for an API to provide users
```

proposal/js-logger.d.ts
```diff
@@ -0,0 +1,206 @@
...
+   *     trace.
+   *   * Invoke `warn` with a string describing the warning and `options`.
+   *
+   * If this field is defined, the compiler must not surface warnings in any way
+   * other than inokving `warn`.
+   */
+  warn?(
+    message: string,
+    options: {
+      deprecation: boolean;
```
Awjin, line 111: Could this be optional and default to `false`?

proposal/js-logger.d.ts
```diff
@@ -0,0 +1,206 @@
...
+   *     Otherwise, set `options.deprecation` to `false`.
+   *   * If this warning is associated with a specific span of a Sass
+   *     stylesheet, set `options.span` to a `SourceSpan` that covers that span.
+   *   * If this warning occurred during execution of a stylesheet, set
+   *     `options.stack` to a string representation of the current Sass stack
+   *     trace.
+   *   * Invoke `warn` with a string describing the warning and `options`.
+   *
+   * If this field is defined, the compiler must not surface warnings in any way
+   * other than inokving `warn`.
```
Awjin:
```suggestion
   * other than invoking `warn`.
```

proposal/js-logger.d.ts
```diff
@@ -0,0 +1,206 @@
...
+   * `@debug` rule using the following procedure:
+   *
+   * * Let `value` be the result of evaluating the rule's expression.
+   * * Let `message` be `value`'s text if it's a string, or the result of
+   *   serializing `value` if it's not.
+   * * Invoke `debug` with `message` and an object with `span` set to the span
+   *   covering the `@debug` rule and its expression.
+   *
+   * If this field is defined, the compiler must not surface debug messages in
+   * any way other than inokving `debug`.
```
Awjin:
```suggestion
   * any way other than invoking `debug`.
```

proposal/js-logger.d.ts
```diff
@@ -0,0 +1,206 @@
...
+   * Additional source text surrounding this span.
+   *
+   * If this isn't empty, it must contain `text`. Furthermore, `text` must begin
+   * at column `start.column` of a line in `context`.
+   *
+   * > This usually contains the full lines the span begins and ends on if the
+   * > span itself doesn't cover the full lines, but the specific scope is up to
+   * > the compiler.
+   */
+  context: string;
```
Awjin: Would marking this optional be more explicit than letting it be an empty string?

proposal/js-logger.d.ts
nex3: Since this is an argument that's being passed *to* user code, I'd rather have it always set so that callbacks don't need to think about the distinction between `undefined` and `false`.

proposal/js-logger.d.ts
nex3: Yeah, good point. I think that was a carryover from the embedded protocol.

Awjin (approved):

PR #3076, opened 2021-06-11
Add proposal for Function and Values API

### LINKED ISSUE #2510: Overhaul JS Function API
* [x] [Proposal](https://github.com/sass/sass/blob/main/proposal/new-function-and-values-api.d.ts)
* [x] [Tests](https://github.com/sass/sass-spec/pull/1744)
* [x] [Documentation](https://github.com/sass/sass/pull/3183)
* [x] [Dart Sass](https://github.com/sass/dart-sass/pull/1558)
* [x] Node Host

---

The JavaScript function API shared by Dart Sass and Node Sass needs an overhaul. The current API has a number of issues:

* Values are mutable. This goes against the grain of how values work within Sass, and makes it difficult to implement the API efficiently in Dart Sass.

* It's hard to verify argument types with a nice error. Compare to Dart Sass's [`assert*()` functions](https://www.dartdocs.org/documentation/sass/latest/sass/Value/assertColor.html).

* It's hard to follow Sass's conventions for dealing with objects, such as:
  * Treating `false` and `null` as falsey and everything else as truthy.
  * Treating all values as lists, and treating maps as lists of pairs.
  * Treating empty lists and empty maps as the same value.
  * Using 1-based indexes for lists and strings, and negative indexes to count from the back.
  * Indexing the Unicode codepoints of strings rather than UTF-16 code units.

* There's no representation of first-class function values.

* There's no representation of an argument list value that may contain keyword arguments.

* Maps are represented as a list of pairs, without any way of accessing a value using its key other than iterating through the entire list.

* Numbers represent their units as a single string in an undocumented format, making it very difficult to work with them if they have more than just a single numerator unit.

* The `this` context includes a bunch of undocumented information without a clear use-case.

We should design a new API that addresses these issues and makes it easy to write importers that work similarly to built-in Sass functions.

Awjin:
This proposal specifies the behavior of custom functions and Sass values in the new JS API.

Fixes https://github.com/sass/sass/issues/2510 and partially addresses #3056.

### DIFF
--- a/proposal/new-function-and-values-api.d.ts
+++ b/proposal/new-function-and-values-api.d.ts
@@ -7,17 +7,76 @@
  *
  * > This section is non-normative.
  *
- * ## Summary
+ * Sass's current JavaScript API was inherited from Node Sass, which developed
+ * over time in an _ad hoc_ manner. The functions API in particular has a number
+ * of notable issues:
  *
- * > This section is non-normative.
+ * - Values are mutable. This goes against the grain of how Sass values work,
+ *   and makes it difficult to implement the API efficiently in Dart Sass.
+ *
+ * - It's hard to verify argument types with assertions.
+ *
+ * - It's hard to follow Sass's conventions for dealing with objects, such as:
+ *
+ *   - Treating `false` and `null` as falsey and everything else as truthy.
+ *   - Treating all values as lists, and treating maps as lists of pairs.
+ *   - Treating empty lists and empty maps as the same value.
+ *   - Using 1-based indexes for lists and strings, and negative indexes to
+ *     count from the back.
+ *   - Indexing the Unicode codepoints of strings rather than UTF-16 code units.
+ *   - There's no representation of first-class function values.
+ *
+ * - There's no representation of an argument list value that may contain
+ *   keyword arguments.
+ *
+ * - Maps are represented as a list of pairs, without any way of accessing a
+ *   value using its key other than iterating through the entire list.
+ *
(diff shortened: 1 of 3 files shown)


proposal/new-function-and-values-api.d.ts
```diff
@@ -0,0 +1,868 @@
...
+
+/**
+ * This definition updates the one in the [New JavaScript API proposal].
+ *
+ * [New JavaScript API proposal]: ./new-js-api.d.ts
+ */
+declare module './new-js-api' {
+  interface Options<sync extends 'sync' | 'async'> {
+    /**
+     * When the compiler encounters a function with a signature that does not
```
nex3:
```suggestion
     * When the compiler encounters a global function call with a signature that does not
```

proposal/new-function-and-values-api.d.ts
```diff
@@ -0,0 +1,868 @@
...
+     * match that of a built-in function, but matches a key in this map, it must
+     * call the associated `CustomFunctionCallback`.
```
nex3:
```suggestion
     * call the associated `CustomFunctionCallback` and return its result.
```

nex3: I think it's also worth explicitly saying "throw an error if this returns anything other than a `Value`". Normally we just say "failing to match type declarations produces undefined behavior", but this specific call feels likely enough to do something wrong (such as use an old-API function) that I think it's worth being explicit.

nex3 (approved):
Great work!

PR #3077, opened 2021-06-11
Rename CompileResult.includedUrls to CompileResult.loadedUrls

Awjin:


proposal/new-js-api.changes.md
```diff
@@ -0,0 +1,9 @@
+## Draft 1.1
```
nex3: This should increment the major version, since it affects the intended semantics of the proposal.

nex3 (approved):

Issue #2509, opened 2018-04-27
Overhaul JS Importer API


nex3:
* [x] [Proposal](https://github.com/sass/sass/blob/main/proposal/new-js-importer.d.ts)
* [x] [Tests](https://github.com/sass/sass-spec/pull/1727)
* [x] [Documentation](https://github.com/sass/sass/pull/3183)
* [x] [Dart Sass](https://github.com/sass/dart-sass/pull/1522)
* [x] Node Host

---

The JavaScript importer API shared by Dart Sass and Node Sass needs an overhaul. The current API has a number of issues:

* It doesn't respect URL semantics. In particular, it doesn't provide any guarantee that `@import "./bar"` within a stylesheet with URL `foo` means the same thing as `@import "foo/bar"`, even when resolved by the same importer.

* It doesn't have a notion of resolving an imported URL to an absolute URL—for example, an npm-aware importer might resolve `~foo` to `file:///home/nex3/app/node_modules/foo/_index.scss`. This makes canonicalization of imported stylesheets impossible, and means that importers are fundamentally less expressive than built-in filesystem imports. It also means that the URLs reported in source maps and other APIs are unlikely to be resolvable.

* It doesn't support the indented syntax.

* Its notion of "redirecting" to a file import has some strange corner cases—for example, a relative path redirect can be resolved relative to the file that contains the import, relative to the working directory, *or* relative to an import path.

* It doesn't seem to do much error checking for invalid responses.

* It uses the string `"stdin"` to refer to files passed by data, even when they don't come from standard input. This could potentially conflict with a real file named `"stdin"`.

* The `this` context includes a bunch of undocumented information without a clear use-case.

We should design a new API that addresses these issues and makes it easy to write importers that work similarly to native filesystem imports. I strongly favor writing the API in such a way that filesystem imports themselves can be modeled as an importer.

ArmorDarks:
I hope that `~foo` is just an example, since `~` isn't part of the NPM loader, but a something that people came up while doing custom Webpack loaders.

nex3:
Yes, it's just an example.

nex3:
This proposal is now officially out for review. It will remain that way for at least a month (potentially longer depending on discussion), at which point we'll move on to implementation.

nex3:
It's been a month and there have been no objections, so I'm going to mark the proposal as accepted and start merging it into the main spec directory.

nex3:
[...]

PR #3081, opened 2021-06-16
[New JS Importer] Add a proposal

nex3:
See #2509

### DIFF
--- a/proposal/new-js-importer.d.ts
+++ b/proposal/new-js-importer.d.ts
@@ -1,0 +1,322 @@
+/**
+ * # New JS Importer: Draft 1
+ *
+ * Issue: https://github.com/sass/sass/issues/2509
+ *
+ * Note: this issue builds on the New JavaScript API proposal.
+ *
+ * ## Background
+ *
+ * > This section is non-normative.
+ *
+ * Sass's current JavaScript API was inherited from Node Sass, which developed
+ * it over time in an _ad hoc_ manner. The importer API in particular has a
+ * number of notable issues:
+ *
+ * - It doesn't respect URL semantics. In particular, it doesn't provide any
+ *   guarantee that `@import "./bar"` within a stylesheet with URL `foo` means
+ *   the same thing as `@import "foo/bar"`, even when resolved by the same
+ *   importer.
+ *
+ * - It doesn't have a notion of resolving an imported URL to an absolute
+ *   URL—for example, a Webpack importer can't resolve `~foo` to
+ *   `file:///home/nex3/app/node_modules/foo/_index.scss`. This makes
+ *   canonicalization of imported stylesheets impossible, and means that
+ *   importers are fundamentally less expressive than built-in filesystem
+ *   imports. It also means that the URLs reported in source maps and other APIs
+ *   are unlikely to be resolvable.
+ *
+ * - It doesn't support the indented syntax for stylesheets that aren't on disk.
+ *
+ * - Its notion of "redirecting" to a file import has some strange corner
+ *   cases—for example, a relative path redirect can be resolved relative to the
+ *   file that contains the import, relative to the working directory, _or_
+ *   relative to an import path.
+ *
+ * - It has limited error checking for invalid responses.
+ *
+ * - It uses the string `"stdin"` to refer to files passed by data, even when
+ *   they don't come from standard input. This could potentially conflict with a
+ *   real file named `"stdin"`.
+ *
+ * - The `this` context includes a bunch of undocumented information without a
+ *   clear use-case.
+ *
+ * Given all of this, the API needs an overhaul. Dart Sass's Dart API has [an
+ * `Importer` API] (itself based on Ruby Sass's API) that can serve as a useful
+ * model.
+ *
+ * [an `importer` api]: https://pub.dev/documentation/sass/latest/sass/Importer-class.html
+ *
+ * ## Summary
+ *
+ * > This section is non-normative.
+ *
+ * Rather than modeling importers as single-function callbacks, we'll model them
+ * as objects that expose multiple methods. The most important of these methods
+ * are those that _canonicalize_ a URL and _load_ a canonical URL.
+ *
+ * ### Canonicalizing
+ *
+ * The first step determines the canonical URL for a stylesheet. Each stylesheet
+ * has exactly one canonical URL that in turn refers to exactly one stylesheet.
+ * The canonical URL must be absolute, including a scheme, but the specific
+ * structure is up to the importer. In most cases, the stylesheet in question
+ * will exist on disk and the importer will just return a `file://` URL for it.
+ *
+ * The `canonicalize()` method takes a URL string that may be either relative or
+ * absolute. If the importer recognizes that URL, it returns a corresponding
+ * absolute URL (including a scheme). This is the _canonical URL_ for the
+ * stylesheet in question. Although the input URL may omit a file extension or
+ * an initial underscore, the canonical URL must be fully resolved.
+ *
+ * For a stylesheet that's loaded from the filesystem, the canonical URL will be
+ * the absolute `file:` URL of the physical file on disk. If it's generated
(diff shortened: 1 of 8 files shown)


proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+ * Issue: https://github.com/sass/sass/issues/2509
+ *
+ * Note: this issue builds on the New JavaScript API proposal.
+ *
+ * ## Background
+ *
+ * > This section is non-normative.
+ *
+ * Sass's current JavaScript API was inherited from Node Sass, which developed
+ * it over time in an _ad hoc_ manner. The importer API in particular has a
```
Awjin:
```suggestion
 * over time in an _ad hoc_ manner. The importer API in particular has a
```

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+ * will exist on disk and the importer will just return a `file://` URL for it.
+ *
+ * The `canonicalize()` method takes a URL string that may be either relative or
+ * absolute. If the importer recognizes that URL, it returns a corresponding
+ * absolute URL (including a scheme). This is the _canonical URL_ for the
+ * stylesheet in question. Although the input URL may omit a file extension or
+ * an initial underscore, the canonical URL must be fully resolved.
+ *
+ * For a stylesheet that's loaded from the filesystem, the canonical URL will be
+ * the absolute `file:` URL of the physical file on disk. If it's generated
```
Awjin, line 74: We should be consistent with this wording throughout (e.g. versus "`file://` URL" on line 65).

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+ * in-memory, the importer should choose a custom URL scheme to guarantee that
```
Awjin, line 75: It might help to add an example of a custom scheme.

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+      }
+  );
+
+/**
+ * This interface represents an [importer]. When the importer is invoked with a
+ * string `string`:
+ *
+ * [importer]: ../spec/modules.md#importer
+ *
+ * * Let `fromImport` be `true` if the importer is being run for an `@import`
```
Awjin: Super nitty but IMO hyphens make lists more readable in jsdocs

```suggestion
 * - Let `fromImport` be `true` if the importer is being run for an `@import`
```

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+  findFileUrl(
+    url: string,
+    options: {fromImport: boolean}
+  ): FileImporterResult | null;
+
+  canonicalize?: never;
+}
+
+/**
+ * This interface represents an [importer]. It has the same beavior as
```
Awjin:
```suggestion
 * This interface represents an [importer]. It has the same behavior as
```

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+ * `SyncFileImporter`, except that if `findFileUrl` returns a `Promise` it waits
+ * for it to complete and uses its value as the return value. If a `Promise`
```
Awjin, line 178: Should `findFileUrl` just return the `Promise` instead of waiting for it to resolve?

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+  findFileUrl(
+    url: string,
+    options: {fromImport: boolean}
+  ): Promise<FileImporterResult | null> | FileImporterResult | null;
+
+  canonicalize?: never;
+}
+
+export interface FileImporterResult {
+  /** The partially-resolved `file:` URL of a file on disk. */
```
Awjin: Could you expand on the meaning of "partially-resolved"? It's mentioned in the Summary section too, so might be useful to formally define it.

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+ *
+ * * Let `syntax` be:
+ *   * "scss" if `url` ends in `.scss`.
+ *   * "indented" if `url` ends in `.sass`.
+ *   * "css" if `url` ends in `.css`.
+ *
+ *   > The algorithm for resolving a `file:` URL guarantees that `url` will have
+ *   > one of these extensions.
+ *
+ * * Return `text`, `syntax`, and `resolved`.
```
Awjin: How do these get transformed into `FileImporterResult`?

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+ *
+ * * Let `result` be the result of calling `load` with `url`.
+ *
+ * * If `result` is null, return null.
+ *
+ * * If `result.syntax` is `"scss"`, `"css"`:
+ *
+ *   * Let `syntax` be `result.syntax`.
+ *
+ * * Otherwise, if `result.syntax` is `"sass"`:
```
Awjin: I'm regretting naming this syntax `sass` in the `compile` API. WDYT about changing the API to use `indented` throughout?

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+ * [resolving `url`]: ../spec/modules.md#resolving-a-file-url
+ */
+interface SyncImporter {
+  canonicalize(url: string, options: {fromImport: boolean}): URL | null;
+  load(canonicalUrl: URL): ImporterResult | null;
+  findFileUrl?: never;
+}
+
+/**
+ * This interface represents an [importer]. It has the same beavior as
```
Awjin:
```suggestion
 * This interface represents an [importer]. It has the same behavior as
```

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+   * spans in `css`.
+   */
+  sourceMapUrl?: URL;
+}
+
+/**
+ * > This type is exported so that TypeScript users can explicitly write a class
+ * > that `implements FileImporter`.
+ */
+export type FileImporter<sync extends 'sync' | 'async'> = sync extends 'async'
```
Awjin, line 287: wdyt about exporting `FileImporter` and `AsyncFileImporter` directly instead?

proposal/new-js-importer.d.ts
```diff
@@ -0,0 +1,322 @@
...
+ *   * "indented" if `url` ends in `.sass`.
+ *   * "css" if `url` ends in `.css`.
+ *
+ *   > The algorithm for resolving a `file:` URL guarantees that `url` will have
+ *   > one of these extensions.
+ *
+ * * Return `text`, `syntax`, and `resolved`.
+ *
+ * [resolving `url`]: ../spec/modules.md#resolving-a-file-url
+ */
```
Awjin, line 165: Consider speccing how `sourceMapUrl` is constructed differently from `url`.

Awjin (changes requested):

proposal/new-js-importer.d.ts
nex3: This is describing how to implement the spec's notion of an "importer", which doesn't know what a `Promise` is, so I think we need to explicitly de-promiseify it before it's returned.

proposal/new-js-importer.d.ts
nex3: They don't; this isn't an `Importer` in the JS sense, it's an importer in the sense of the definition in `modules.md`.

proposal/new-js-importer.d.ts
nex3: I think if we're going to expose the generic type anywhere, we should use it consistently everywhere.

proposal/new-js-importer.d.ts
nex3: As above, this isn't actually a JS `Importer`.

Awjin (approved):

PR #3085, opened 2021-06-21
Fix docs for `index` case in URL resolution

Rafer45:
The description here does not match the implementation in https://github.com/sass/dart-sass/blob/7e371666f4f4554bb4e78de3087fec8177074b29/lib/src/importer/utils.dart#L62
For demonstration, see https://gist.github.com/Rafer45/3f0bce5ce2519192282564056144fc38


nex3 (approved):
Thanks!

PR #3107, opened 2021-08-05
[New JS Function] Make a number of small improvements

nex3:


proposal/new-function-and-values-api.d.ts
```diff
@@ -263,7 +278,7 @@ export abstract class Value {
    * Must be the same for `Value`s that are equal to each other according to the
    * `==` SassScript operator.
    */
-  hashCode(): string;
+  hashCode(): number;
```
Awjin, line 281: We could also change this to a getter.

Awjin (changes requested):
Another improvement that could be worth making here—`convert/coerceValueToMatch` can be more simply defined as returning the value of the result of `convert/coerceToMatch`

proposal/new-function-and-values-api.d.ts
nex3: I think that violates the `immutable` package protocol.

nex3 (commented):
> Another improvement that could be worth making here—`convert/coerceValueToMatch` can be more simply defined as returning the value of the result of `convert/coerceToMatch`

Done.

Awjin (approved):

PR #3112, opened 2021-08-06
Rate limit Markdown link checks

Awjin:


accepted/extend-specificity.md
```diff
@@ -64,7 +64,7 @@ some cases.
 
 #### First Law of Extend: `spec(extend(S, A, B)[0]) >= spec(S)`
 
-This is not always the behavior in Sass, either in master or in stable; this is
+This is not always the behavior in Sass, either in main or in stable; this is
```
nex3: Since this is a historical document, I think it should remain as-is.

test/link-check.mjs
```diff
@@ -6,29 +6,38 @@ import * as urlModule from 'url'
 import markdownLinkCheck from 'markdown-link-check'
```
nex3: It's a good idea to enforce style on the files in `test/`, but if we do so we should probably go whole hog and convert them to TypeScript formatted with GTS. It would be really weird to hold them to a style that's different from the one used in the rest of the repo.

nex3 (changes requested):

test/link-check.mjs
Awjin: I tried converting these to TS but it added a bunch of brittle machinery (declaration types missing for some packages, ts-node choking on ESM https://github.com/TypeStrong/ts-node/issues/1007). So instead I converted these to JS, and included them in the directories formatted by GTS.

.eslintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "./node_modules/gts/"
+  "extends": "./node_modules/gts/",
+  "parserOptions": {
+    "ecmaVersion": 2020
+  }
```
nex3: What does this do?

package.json
```diff
@@ -1,13 +1,15 @@
 {
   "name": "sass",
-  "//": [
-    "This is used to track dependencies for the Travis CI tests run on this ",
-    "repository. For the official Sass npm package, see ",
-    "https://npmjs.org/package/sass"
-  ],
```
nex3: We should probably keep this.

nex3 (approved):

.eslintrc.json
Awjin: This needs to be specified to support the `import.meta` call here: https://github.com/sass/sass/pull/3112/files#diff-fe16e804620b4c53d6bcf6fbfe0e6580e9b2b3a89f05f120ccf914e971238884R31

package.json
Awjin: right, good catch

nex3:
It looks like these checks are still failing intermittently. In addition to the rate limiting, it may be a good idea to add retry logic with exponential backoff.

Awjin:
Yeah, sounds good.

.eslintrc.json
Awjin: Looks like the `import.meta` wasn't even necessary, so I removed it.

Awjin:
I looked deeper and it's not clear we need to add retry logic. I think the issue was that we weren't properly waiting for `markdownLinkCheck`'s built-in retry logic to finish up before firing off another http request.

The script `await`s it properly now. Re-ran CI a few times and it looks good. If CI gets flaky again we should reconsider adding retry logic.

PR #3123, opened 2021-08-18
Apply the Phase 1 slash-separator changes to the specification

nex3:


spec/built-in-modules/math.md
```diff
@@ -457,10 +457,21 @@ This function is also available as a global function named `unit()`.
 math.div($number1, $number2)
```
jathak: This should probably just be `div` to be consistent with other functions that don't include the namespace

spec/built-in-modules/math.md
```diff
@@ -457,10 +457,21 @@ This function is also available as a global function named `unit()`.
...
-* If `$number1` and `$number2` have units, but the units are not [compatible],
-  throw an error.
-* Return the result of dividing `$number1` by `$number2`.
+* If `$number1` is a color and `$number2` is either a number or a color, throw an
+  error.
+* Otherwise, if `$number2` is a number and `$number2` is a color, throw an error.
+* Otherwise, if either of `$number1` or `$number2` are not numbers, return an
+  unquoted string whose contents is the result of serializing `$number1`
+  followed by `"/"` followed by the result of serializing `$number2`.
+* Otherwise, let `value` be .
```
jathak: Is this line supposed to end here?

jathak (approved):

PR #3139, opened 2021-09-10
Clarify the definition of compatible units

nex3:


Awjin (approved):
Thanks, this is much clearer.

Issue #3142, opened 2021-09-14
First-Class Calc: min/max are backwards-incompatible


SnakeO:
```
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: 0.25rem and 0 are incompatible.
  ╷
9 │       $return: append($return, max($value, 0));
  │                                    ^^^^^^ 0.25rem
  │                                            ━ 0
  ╵
  node_modules/bootstrap/scss/mixins/_border-radius.scss 9:36   valid-radius()
  node_modules/bootstrap/scss/mixins/_border-radius.scss 19:20  border-radius()
  node_modules/bootstrap/scss/_images.scss 18:3                 @import
  node_modules/bootstrap/scss/bootstrap.scss 14:9               @import
```

nex3:
Sorry for the trouble, folks! This is the result of an unintentional backwards incompatibility in first-class calc support: we're now parsing `min()` and `max()` expressions as calculations if they're valid calculations, but we didn't take into account the fact that calculations are a bit stricter about units than the old global `min()` and `max()` functions. We'll need to take some time to design a more compatible way of handling this, but for the time being I'm going to partially revert this behavior:

* `min()` and `max()` functions that appear outside of calculations will be parsed the same way they were in 1.39.2, as global function calls if they contain any Sass-specific arguments and as unquoted strings if they don't.
* `min()` and `max()` functions *within* calculations will continue to be parsed as calculations, since they could never have been parsed as global function calls in the first place.

Jeffery-Young:
same error, could rollback it first?

nex3:
I've unpublished the npm package. On other platforms, there's not really a meaningful way to "roll back", so you'll need to wait for 1.40.1, which should be published within the hour.

nex3:
Sass 1.40.1 has been released with the fix. I'm going to keep this open to track the progress of re-enabling `min()` and `max()` calculations.

maoxiaoke:
> Sass 1.40.1 has been released with the fix. I'm going to keep this open to track the progress of re-enabling `min()` and `max()` calculations.

Seems not working!
![image](https://user-images.githubusercontent.com/13417006/133235480-488d51d2-5a83-4a07-a5be-e15b66f68371.png)

[...]

PR #3146, opened 2021-09-14
[First-Class Calc] Temporarily drop support for min/max calculations

nex3:
See sass/sass#3142


nex3:
This is an emergency-track proposal. The implementation change (https://github.com/sass/sass-spec/pull/1702) and test change (https://github.com/sass/sass-spec/pull/1702) have already landed.

Awjin (approved):

PR #3151, opened 2021-09-15
Make `options.quotes` optional.

Awjin:
This is better aligned with how optional options are handled by all
other Values (e.g. SassNumber, SassList).


nex3:
If we don't have these `@default` annotations, the spec text itself should describe what to do if the option isn't passed.

nex3 (approved):

PR #3154, opened 2021-09-16
[First-Class Calc] Preserve parentheses around var() functions

### LINKED ISSUE #3153: Parentheses left out in compiled CSS since 1.40.0 (causing order of operations problems)
I bumped into an issue that seems to be present since 1.40.0 and is still present in the latest 1.41.0.

Given the folowing `scss`:
```scss
my-element {
  &::before{
    padding-bottom: calc(100% / (var(--aspect-ratio, 16/9)));
  }
```

This is compiled starting from 1.40.0 as:
```css
my-element::before {
  padding-bottom:calc(100%/var(--aspect-ratio, 16/9));
}
```

Notice the missing `()` parentheses around the `var(...)` statement. When `--aspect-ratio` is not defined and the fallback value is used this causes the outcome to be `100%/16/9`. Due to the missing parentheses and the order of operations, the calculation is wrong.

Before 1.40.0 this was compiled correctly to:
```css
my-element::before {
  padding-bottom:calc(100%/(var(--aspect-ratio, 16/9)));
}
```

Notice the parentheses around `var(...)` are present here. When `--aspect-ratio` would not be defined the outcome would (correctly) be `100%/(16/9)`.

nex3:
Closes #3153

### DIFF
--- a/spec/types/calculation.md
+++ b/spec/types/calculation.md
@@ -22,6 +22,7 @@
   * [`CalcSum`](#calcsum)
   * [`CalcProduct`](#calcproduct)
   * [`CalcValue`](#calcvalue)
+  * [`ParenthesizedVar`](#parenthesizedvar)
 
 ## Syntax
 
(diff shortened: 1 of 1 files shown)


nex3:
This is on the emergency track.

jathak (approved):
LGTM, just needs the TOC entry for `ParenthesizedVar`.

spec/types/calculation.md
```diff
@@ -341,3 +347,32 @@ To evaluate a `CalcValue` production `value` into a `CalculationValue` object:
...
+### `ParenthesizedVar`
+
+> If a `var()` is written directly within parentheses, it's necessary to
+> preserve those parentheses. CSS resolves `var()` by literally replacing the
+> function with the value of the variable and *then* parsing the surrounding
+> context.
+>
+> For example, if `--ratio: 2/3`, `calc(1 / (var(--ratio)))` is parsed as
+> `calc(1 / (2/3)) = calc(3/2)` but `calc(1 / var(--ratio))` is parsed as
+> `calc(1 / 2/3) = calc(2/3)`.
```
stof, line 360: should be `= calc(1/6)`

spec/types/calculation.md
```diff
@@ -341,3 +347,32 @@ To evaluate a `CalcValue` production `value` into a `CalculationValue` object:
...
+  [`NamespacedIdentifier`] and with `value`'s `ArgumentInvocation`.
+
+  [`FunctionCall`]: ../functions.md#syntax
+  [`NamespacedIdentifier`]: ../modules.md#syntax
+
+* Let `result` be the result of evaluating `function`.
+
+* If `result` is a number or a calculation, return it.
+
+  > This could happen if the user defines a `var` function in Sass.
```
stof, line 374: isn't that forbidden ?

spec/types/calculation.md
nex3: Thanks! https://github.com/sass/sass/pull/3157

spec/types/calculation.md
nex3: No, only names that have special parsing rules (`url`, `and`, `not`, etc) are forbidden as function names. Users are allowed to override plain CSS functions if they want.

PR #3158, opened 2021-09-22
Give compilers more flexibility in handling Options.alertColor

nex3:


spec/js-api/options.d.ts
```diff
@@ -50,7 +50,7 @@ export interface Options<sync extends 'sync' | 'async'> {
    * > Compilers are not obligated to use terminal colors if they are irrelevant
    * > in the output environment.
    *
-   * @default false
+   * @default true
```
Awjin: I don't think the docs are implying that this should be default true, but rather that even if this is set `true`, the compiler isn't obliged to use colors.

Regardless, we should change the default if that makes more sense. IIRC Dart Sass is default false

nex3: That's a fair point. Dart Sass does default to `false`, but in retrospect I think it probably should default to `hasTerminal`. Maybe a better way to frame this is "implementations can choose the default based on output environment" and just expect users to set it explicitly if they feel strongly.

Awjin (approved):

PR #3159, opened 2021-09-22
Clarify when the JS API exception class should and shouldn't be used

nex3:
This allows us to make the stackTrace and span fields non-nullable,
since they'll always be meaningful when an exception is thrown within
a compilation.

~This is a fast-track proposal.~

### DIFF
--- a/spec/js-api/exception.d.ts
+++ b/spec/js-api/exception.d.ts
@@ -1,6 +1,10 @@
 import {SourceSpan} from './logger';
 
-/** The error thrown by the compiler when a Sass compilation fails. */
+/**
+ * The error thrown by the compiler when a Sass compilation fails. This should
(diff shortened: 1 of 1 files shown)


Awjin (approved):

nex3:
I'm actually going to remove the designation of this as a fast-track proposal. Since the JS API isn't implemented anywhere yet, we really shouldn't have integrated it into the `spec/` directory in the first place according to [the proposal process](https://github.com/sass/sass/blob/main/CONTRIBUTING.md#proposal). My instinct was to do so in order to get the type declarations in one easily-referenced place, but it muddies the water about what's *official spec* and what's still mutable due to the fact that implementation always influences design. Next time we do a change like this, we'll need to figure out a better way to model this.

PR #3168, opened 2021-09-28
Add an explicit note about free interpolation following numbers

stof:
Refs https://github.com/sass/dart-sass/issues/1506


jathak (approved):
Thanks!

PR #3173, opened 2021-09-30
Test that typedoc builds with no warnings

nex3:


jathak (approved):
LGTM aside from the double quotes Prettier is complaining about

PR #3174, opened 2021-10-02
 Document the legacy API and Logger

nex3:


js-api-doc/README.md
```diff
@@ -0,0 +1,60 @@
...
+as a native extension in the [`node-sass` package].
+
+[Dart Sass]: https://sass-lang.com/dart-sass
+[LibSass]: https://sass-lang.com/libsass
+[`sass` package]: https://www.npmjs.com/package/sass
+[`node-sass` package]: https://www.npmjs.com/package/node-sass
+
+## Usage
+
+The JavaScript API has two entrypoints for compiling a Sass to CSS. Each one can
```
Awjin:
```suggestion
The JavaScript API has two entrypoints for compiling Sass to CSS. Each one can
```

js-api-doc/README.md
```diff
@@ -0,0 +1,60 @@
...
+compile either a Sass file by passing in [[LegacyFileOptions]] or a string of
+Sass code by passing in a [[LegacyStringOptions]].
+
+* [[renderSync]] runs synchronously. It's **by far the fastest option** when
+  using Dart Sass, but at the cost of only supporting synchronous {@link
+  LegacyImporter | importer} and {@link LegacyFunction | function} plugins.
+  
+  ```js
+  var sass = require('sass'); // or require('node-sass');
```
Awjin: Nit (here and throughout): Probably makes sense to use `const`/`let` instead of `var`.

js-api-doc/compile.d.ts
```diff
@@ -9,18 +9,22 @@ export interface CompileResult {
   sourceMap?: RawSourceMap;
 }
 
+/** @hidden */
```
Awjin, line 12: Is this for temporarily hiding types from the website?

js-api-doc/legacy/exception.d.ts
```diff
@@ -1,8 +1,50 @@
+/**
+ * The exception type thrown by [[renderSync]] and passed as the error to
+ * [[render]]'s callback.
+ */
 export interface LegacyException extends Error {
+  /**
+   * The error message. For Dart Sass, when possible this includes a highlighted
+   * indication of where in the source file the error occurred as well as the
+   * Sass stack
+   * trace.
```
Awjin:
```suggestion
   * Sass stack trace.
```

js-api-doc/legacy/function.d.ts
```diff
@@ -1,94 +1,661 @@
...
+ *
+ * An asynchronous function must return `undefined`. Its final argument will
+ * always be a callback, which it should call with the result of the function
+ * once it's done running.
+ *
+ * If this throws an error, Sass will treat that as the function failing with
+ * that error message.
+ *
+ * ```js
+ * const result = sass.renderSync({
```
Awjin: This should be the async version.

js-api-doc/legacy/function.d.ts
```diff
@@ -1,94 +1,661 @@
...
+     *
+     * // string is `"Helvetica Neue"`.
+     * string.getValue(); // "Helvetica Neue"
+     *
+     * // string is `\1F46D`.
+     * string.getValue(); // "\\1F46D"
+     *
+     * // string is `"\1F46D"`.
+     * string.getValue(); // "👭"
+     * string.setValue(value) permalink
```
Awjin: Probably shouldn't use a deprecated feature here:

```suggestion
```

js-api-doc/legacy/function.d.ts
```diff
@@ -1,94 +1,661 @@
...
+     * This means that `96px` and `1in` will return different values, even
+     * though they represent the same length.
+     *
+     * </div>
+     *
+     * @example
+     *
+     * ```js
+     * var number = new sass.types.Number(10, "px");
+     * console.log(number.getValue()); // 10
```
Awjin: Nit (here and below): Other examples (color/string) are using `value.getValue(); // 'val'` instead of `console.log(value.getValue()); // 'val'`. Consider aligning onto one style.

js-api-doc/legacy/function.d.ts
```diff
@@ -1,94 +1,661 @@
...
+     * </div>
+     *
+     * @example
+     *
+     * ```js
+     * var list = new sass.types.List(3);
+     * list.setValue(0, new sass.types.Number(10, "px"));
+     * list.setValue(1, new sass.types.Number(15, "px"));
+     * list.setValue(2, new sass.types.Number(32, "px"));
+     * list; // 10px, 15px, 32px
```
Awjin, line 456: Nit (here and below, and same for maps):

```suggestion
     * console.log(list); // 10px, 15px, 32px
```
or
```suggestion
     * list.toString(); // 10px, 15px, 32px
```

js-api-doc/legacy/importer.d.ts
```diff
@@ -1,22 +1,145 @@
...
+   * [import-only files](https://sass-lang.com/documentation/at-rules/import#import-only-files).
+   */
   fromImport: boolean;
 }
 
+/**
+ * The result of running a [[LegacyImporter]]. It must be one of the following
+ * types:
+ *
+ * * An object with the key contents whose value is the contents of a stylesheet
```
Awjin:
```suggestion
 * * An object with the key `contents` whose value is the contents of a stylesheet
```

js-api-doc/legacy/importer.d.ts
```diff
@@ -1,22 +1,145 @@
...
+ *   (in SCSS syntax). This causes Sass to load that stylesheet’s contents.
+ *
+ * * An object with the key file whose value is a path on disk. This causes Sass
```
Awjin:
```suggestion
 * * An object with the key `file` whose value is a path on disk. This causes Sass
```

spec/js-api/value/index.d.ts
```diff
@@ -34,6 +34,8 @@ export const sassNull: Value;
  * > Sass value it represents.
  */
 export abstract class Value implements ValueObject {
+  protected constructor();
```
Awjin, line 37: Why is this needed?

js-api-doc/legacy/options.d.ts
```diff
@@ -3,46 +3,692 @@ import {LegacyImporter} from './importer';
...
+   *
+   * @defaultValue `false`
+   * @category Source Maps
+   */
   omitSourceMapUrl?: boolean;
+
+  /**
+   * The location that Sass expects the generated CSS to be saved to. It’s used
+   * to determine the URL used to link from the generated CSS to the source map,
+   * and from the source map to t
```
Awjin:
```suggestion
   * and from the source map to the Sass source files.
```
?

js-api-doc/legacy/options.d.ts
```diff
@@ -3,46 +3,692 @@ import {LegacyImporter} from './importer';
...
+   *   <em>always</em> work. It shouldn’t be possible for some configuration
+   *   somewhere else to break it.
+   * </div>
+   *
+   * Additional handler(s) for loading files when a [`@use`
+   * rule](https://sass-lang.com/documentation/at-rules/use) or an [`@import`
+   * rule](https://sass-lang.com/documentation/at-rules/import) is encountered.
+   * It can either be a single [[LegacyImporter]] function, or an array of
+   * [[LegacyImporter]]s.
+   *
```
Awjin, line 361:
```suggestion
```

Awjin (approved):

js-api-doc/README.md
nex3: Good call. Done.

js-api-doc/compile.d.ts
nex3: Exactly. I don't want the new API docs up there until they're actually available somewhere, but I *do* want the logger docs available since it's available from the old API.

js-api-doc/legacy/function.d.ts
nex3: I think this one's all right—the comment indicates the Sass value, which isn't necessarily the result users will get from calling `.toString()`, it's just the notional value associated with the object.

spec/js-api/value/index.d.ts
nex3: Typedoc doesn't know to automatically treat abstract constructors as protected. You're right that this seems like an oversight, though—I've filed https://github.com/TypeStrong/typedoc/issues/1729 to track the issue.

PR #3177, opened 2021-10-05
Document the logger API

nex3:
I missed this as part of #3174


js-api-doc/logger/index.d.ts
```diff
@@ -3,7 +3,48 @@ import {SourceSpan} from './source_span';
...
+ * An object that can be passed to [[LegacySharedOptions.logger]] to control how
+ * Sass emits warnings and debug messages.
+ *
+ * @example
+ *
+ * ```js
+ * const fs = require('fs');
+ * const sass = require('sass');
+ *
+ * const log = "";
```
Awjin:
```suggestion
 * let log = "";
```

Awjin (approved):

PR #3183, opened 2021-10-08
Document the new JS API

nex3:
This shouldn't be merged until the new API is supported by Dart Sass.

### DIFF
--- a/js-api-doc/README.md
+++ b/js-api-doc/README.md
@@ -1,15 +1,75 @@
-Both [Dart Sass] and [LibSass] support the same JavaScript API. Dart Sass is
-distributed as the pure-Javascript [`sass` package], and LibSass is distributed
-as a native extension in the [`node-sass` package].
+The [`sass` package] on npm is a pure-JavaScript package built from the [Dart
+Sass] implementation. In addition to Dart Sass's [command-line interface], it
+provides a JavaScript API that can be used to drive Sass compilations from
+JavaScript. It even allows an application to control {@link Options.importers |
+how stylesheets are loaded} and {@link Options.functions | define custom
+functions}.
 
-[Dart Sass]: https://sass-lang.com/dart-sass
-[LibSass]: https://sass-lang.com/libsass
 [`sass` package]: https://www.npmjs.com/package/sass
-[`node-sass` package]: https://www.npmjs.com/package/node-sass
+[Dart Sass]: https://sass-lang.com/dart-sass
+[command-line interface]: https://sass-lang.com/documentation/cli/dart-sass
 
 ## Usage
 
-The JavaScript API has two entrypoints for compiling Sass to CSS. Each one can
+The JavaScript API provides two entrypoints for compiling Sass to CSS, each of
+which has a synchronous variant that returns a plain [[CompileResult]] and an
+asynchronous variant that returns a `Promise`. **The asynchronous variants are
+much slower,** but they allow custom importers and functions to run
+asynchronously.
+
+* [[compile]] and [[compileAsync]] take a path to a Sass file and return the
+  result of compiling that file to CSS. 
+
+  ```js
+  const sass = require('sass');
+
+  const result = sass.compile("style.scss");
+  console.log(result.css);
+  ```
+
+* [[compileString]] and [[compileStringAsync]] take a string that represents the
+  contents of a Sass file and return the result of compiling that file to CSS.
+
+  ```js
+  const sass = require('sass');
+
+  const result = sass.compileString(`
+  h1 {
+    font-size: 40px;
+    code {
+      font-face: Roboto Mono;
+    }
+  }`);
+  console.log(result.css);
+  ```
+
+## Integrations
+
+Most popular Node.js build systems have integrations available for the JS API:
+
+* Webpack uses the [`sass-loader` package].
+* Gulp uses the [`gulp-sass` package].
+* Broccoli uses the [`broccoli-sass-source-maps` package].
+* Ember uses the [`ember-cli-sass` package].
+* Grunt uses the [`grunt-sass` package].
+
+[`sass-loader` package]: https://www.npmjs.com/package/sass-loader
+[`gulp-sass` package]: https://www.npmjs.com/package/gulp-sass
+[`broccoli-sass-source-maps` package]: https://www.npmjs.com/package/broccoli-sass-source-maps
+[`ember-cli-sass` package]: https://www.npmjs.com/package/ember-cli-sass
+[`grunt-sass` package]: https://www.npmjs.com/package/grunt-sass
+
+## Legacy API
+
+The `sass` package also supports an older API. Although this API is deprecated,
+it will continue to be supported until the release of version 2.0.0 of the
+`sass` package. The legacy API is also supported by the [`node-sass` package],
+which is a native extension wrapper for the deprecated [LibSass] implementation.
+
+[`node-sass` package]: https://www.npmjs.com/package/node-sass
+[LibSass]: https://sass-lang.com/libsass
+
+The legacy API has two entrypoints for compiling Sass to CSS. Each one can
 compile either a Sass file by passing in [[LegacyFileOptions]] or a string of
 Sass code by passing in a [[LegacyStringOptions]].
 
@@ -41,20 +101,3 @@ Sass code by passing in a [[LegacyStringOptions]].
     }
   });
   ```
-
-## Integrations
-
-Most popular Node.js build systems have integrations available for the JS API:
-
-* Webpack uses the [`sass-loader` package].
-* Gulp uses the [`gulp-sass` package].
-* Broccoli uses the [`broccoli-sass-source-maps` package].
-* Ember uses the [`ember-cli-sass` package].
-* Grunt uses the [`grunt-sass` package].
-
-[`sass-loader` package]: https://www.npmjs.com/package/sass-loader
-[`gulp-sass` package]: https://www.npmjs.com/package/gulp-sass
-[`broccoli-sass-source-maps` package]: https://www.npmjs.com/package/broccoli-sass-source-maps
-[`ember-cli-sass` package]: https://www.npmjs.com/package/ember-cli-sass
-[`grunt-sass` package]: https://www.npmjs.com/package/grunt-sass
-
--- a/js-api-doc/compile.d.ts
+++ b/js-api-doc/compile.d.ts
@@ -3,28 +3,150 @@ import {RawSourceMap} from 'source-map-js';
 
 import {Options, StringOptions} from './options';
 
+/**
+ * The result of compiling Sass to CSS. Returned by [[compile]],
+ * [[compileAsync]], [[compileString]], and [[compileStringAsync]].
+ *
+ * @category Compile
+ */
 export interface CompileResult {
+  /**
+   * The generated CSS.
+   *
+   * Note that this *never* includes a `sourceMapUrl` comment—it's up to the
+   * caller to determine where to save the source map and how to link to it from
+   * the stylesheet.
+   */
   css: string;
+
+  /**
+   * The canonical URLs of all the stylesheets that were loaded during the
+   * Sass compilation. The order of these URLs is not guaranteed.
+   */
   loadedUrls: URL[];
+
+  /**
+   * The object representation of the source map that maps locations in the
+   * generated CSS back to locations in the Sass source code.
+   *
+   * This typically uses absolute `file:` URLs to refer to Sass files, although
+   * this can be controlled by having a custom [[Importer]] return
+   * [[ImporterResult.sourceMapUrl]].
+   *
+   * This is set if and only if [[Options.sourceMap]] is `true`.
+   */
   sourceMap?: RawSourceMap;
 }
 
-/** @hidden */
+/**
+ * Synchronously compiles the Sass file at `path` to CSS. If it succeeds it
+ * returns a [[CompileResult]], and if it fails it throws an [[Exception]].
+ *
+ * This only allows synchronous [[Importer]]s and [[CustomFunction]]s.
+ *
+ * @example
+ *
+ * ```js
+ * const sass = require('sass');
+ *
+ * const result = sass.compile("style.scss");
+ * console.log(result.css);
+ * ```
+ *
+ * @category Compile
+ * @compatibility dart: "1.44.0", node: false
+ */
 export function compile(path: string, options?: Options<'sync'>): CompileResult;
 
-/** @hidden */
+/**
+ * Asynchronously compiles the Sass file at `path` to CSS. Returns a promise
+ * that resolves with a [[CompileResult]] if it succeeds and rejects with an
+ * [[Exception]] if it fails.
+ *
+ * This only allows synchronous or asynchronous [[Importer]]s and
+ * [[CustomFunction]]s.
+ *
+ * **Heads up!** When using Dart Sass, **[[compile]] is almost twice as fast as
+ * [[compileAsync]]**, due to the overhead of making the entire evaluation
+ * process asynchronous.
+ *
+ * @example
+ *
+ * ```js
+ * const sass = require('sass');
+ *
+ * const result = await sass.compileAsync("style.scss");
+ * console.log(result.css);
+ * ```
+ *
+ * @category Compile
+ * @compatibility dart: "1.44.0", node: false
+ */
 export function compileAsync(
   path: string,
   options?: Options<'async'>
 ): Promise<CompileResult>;
 
-/** @hidden */
+/**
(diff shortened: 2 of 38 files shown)


js-api-doc/exception.d.ts
```diff
@@ -1,13 +1,41 @@
...
+   */
   message: string;
 
+  /**
+   * A textual description of what went wrong.
+   *
+   * Unlike [[message]], this does *not* include representations of [[span]] or
+   * [[sassStack]].
+   */
   sassMessage: string;
```
stof: @nex3 should those be typed as readonly properties ? AFAICT, dart-sass implements them as getters

js-api-doc/README.md
```diff
@@ -1,15 +1,75 @@
...
+much slower,** but they allow custom importers and functions to run
+asynchronously.
+
+* [[compile]] and [[compileAsync]] take a path to a Sass file and return the
+  result of compiling that file to CSS. 
+
+  ```js
+  const sass = require('sass');
+
+  const result = sass.compile("style.css");
```
Awjin:
```suggestion
  const result = sass.compile("style.scss");
```

js-api-doc/value/string.d.ts
```diff
@@ -1,21 +1,87 @@
...
+   * JavaScript, `"n😊b".charCodeAt(1)` returns `0xD83D`, whereas in Sass
+   * `string.slice("n😊b", 1, 1)` returns `"😊"`.
+   *
+   * This function converts Sass's code point indices to JavaScript's code unit
+   * indices. This means it's O(n) in the length of `text`.
+   *
+   * @throws `Error` if `sassIndex` isn't a number, if that number isn't an
+   * integer, or if that integer isn't a valid index for this string.
+   *
+   * @throws `Error` If `sassIndex` isn't a number, if that number isn't an
```
Awjin:
```suggestion
   * @throws `Error` if `sassIndex` isn't a number, if that number isn't an
```

spec/js-api/value/boolean.d.ts
```diff
@@ -7,6 +7,8 @@ export const sassTrue: SassBoolean;
 export const sassFalse: SassBoolean;
 
 /** The JS API representation of a Sass boolean. */
-export interface SassBoolean extends Value {
+export class SassBoolean extends Value {
+  private constructor();
+
```
Awjin, line 12: Why is this changed to a class?

js-api-doc/value/list.d.ts
```diff
@@ -2,25 +2,57 @@ import {List} from 'immutable';
 
 import {Value} from './index';
 
+/**
+ * Possible separators used by Sass lists. The special separator `null` is only
+ * used for lists with fewer than two elements, and indicates that the separator
+ * has not yet been decided for this list.
```
Awjin, line 8: Is this still true for the new API? (the constructor defaults to comma separator) I think this should be null only for scalar values

js-api-doc/value/list.d.ts
```diff
@@ -2,25 +2,57 @@ import {List} from 'immutable';
...
+   * `false`.
+   */
   static empty(options?: {
-    /** @default null */
     separator?: ListSeparator;
-    /** @default false */
     brackets?: boolean;
   }): SassList;
 
+  /** @hidden */
```
Awjin, line 52: Does this need to stay hidden?

js-api-doc/value/map.d.ts
```diff
@@ -2,12 +2,29 @@ import {OrderedMap} from 'immutable';
...
+  /** Creates a new empty map. */
   static empty(): SassMap;
 
+  /**
+   * Returns the contents of this map as an immutable [[OrderedMap]] from the
+   * [`immutable` package](https://immutable-js.com/).
+   */
   get contents(): OrderedMap<Value, Value>;
 
+  /** @hidden */
```
Awjin, line 39: Does this need to stay hidden?

js-api-doc/value/index.d.ts
```diff
@@ -17,40 +17,144 @@ export {SassMap} from './map';
...
+   */
   tryMap(): OrderedMap<Value, Value> | null;
 
+  /** Returns whether `this` represents the same value as `other`. */
   equals(other: Value): boolean;
 
+  /** Returns a hash code that can be used to store `this` in a hash map. */
   hashCode(): number;
 
+  /** @hidden */
```
Awjin, line 171: Does this need to stay hidden?

js-api-doc/logger/source_span.d.ts
```diff
@@ -2,14 +2,35 @@ import {URL} from 'url';
...
+ * A span of text within a source file.
+ *
+ * @category Logger
+ */
 export interface SourceSpan {
+  /** The beginning of this span, inclusive. */
   start: SourceLocation;
 
+  /**
+   * The end of this span covers, exclusive.
```
Awjin:
```suggestion
   * The end of this span, exclusive.
```

js-api-doc/exception.d.ts
```diff
@@ -1,13 +1,41 @@
...
 
+  /**
+   * A human-friendly representation of the Sass stack trace at the point of
+   * error.
+   */
   sassStack: string;
 
+  /** The location the error ocurred in the Sass file that triggered it. */
```
Awjin:
```suggestion
  /** The location the error occurred in the Sass file that triggered it. */
```

js-api-doc/value/number.d.ts
```diff
@@ -2,83 +2,308 @@ import {List} from 'immutable';
...
+   * that integer. Otherwise, throws an error.
+   *
+   * @param name - The name of the function argument `this` came from (without
+   * the `$`) if it came from an argument. Used for error reporting.
+   */
   assertInt(name?: string): number;
 
+  /**
+   * Returns [[value]] if it's within `min` and `max`, or if it's equal to `min`
+   * or `max` according to Sass's equality returns [[value]] clamped to that
```
Awjin: nit: The wording here is a bit confusing

js-api-doc/importer.d.ts
```diff
@@ -3,32 +3,293 @@ import {URL} from 'url';
...
+   *
+   * If this importer doesn't recognize the URL, it should return `null` instead
+   * to allow other importers or {@link Options.loadPaths | load paths} to
+   * handle it.
+   *
+   * This may also return a `Promise`, but if it does the importer may only be
+   * passed to [[compileAsync]] and [[compileStringAsync]], not [[compile]] or
+   * [[compileString]].
+   *
+   * @throws any - If this importer recognizes the URL but determines that
```
Awjin: Here and below, why is this typed as `any` vs `Error`?

js-api-doc/importer.d.ts
```diff
@@ -3,32 +3,293 @@ import {URL} from 'url';
...
+   *   if the extension is explicitly specified, a stylesheet with the opposite
+   *   extension is allowed to exist.
+   *
+   * - If none of the possible paths is valid, the importer should perform the
+   *   same resolution on the URL followed by `/index`. In the example above,
+   *   it would look for:
+   *   - `foo/bar/baz/_index.sass`
+   *   - `foo/bar/baz/index.sass`
+   *   - `foo/bar/baz/_index.scss`
+   *   - `foo/bar/baz/index.scss`
```
Awjin: Nit: Is order meaningful here? If not consistency with previous example could be nice:

```suggestion
   *   - `foo/bar/baz/index.sass`
   *   - `foo/bar/baz/index.scss`
   *   - `foo/bar/baz/_index.sass`
   *   - `foo/bar/baz/_index.scss`
```

js-api-doc/importer.d.ts
```diff
@@ -3,32 +3,293 @@ import {URL} from 'url';
...
+ *       };
+ *     }
+ *   }]
+ * });
+ * ```
+ *
+ * @category Importer
+ */
 export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
+  /**
```
Awjin, line 152: Not a review comment per se, but I really liked how clear this section is!

js-api-doc/importer.d.ts
```diff
@@ -3,32 +3,293 @@ import {URL} from 'url';
...
+   * Loads the Sass text for the given `canonicalUrl`, or returns `null` if this
+   * importer can't find the stylesheet it refers to.
+   *
+   * If the importer finds a stylesheet at [url] but it fails to load for some
+   * reason, or if [url] is uniquely associated with this importer but doesn't
+   * refer to a real stylesheet, the importer may throw an exception that will
+   * be wrapped by Sass. If the exception object has a `message` property, it
+   * will be used as the wrapped exception's message; otherwise, the exception
+   * object's `toString()` will be used. This means it's safe for importers to
+   * throw plain strings.
```
Awjin: Ah... this answers my earlier question about throwing `any`. Might be worth duplicating this in the other jsdocs.

js-api-doc/importer.d.ts
```diff
@@ -3,32 +3,293 @@ import {URL} from 'url';
...
+   *
+   * @param canonicalUrl - The canonical URL of the stylesheet to load. This is
+   * guaranteed to come froma call to [[canonicalize]], although not every call
```
Awjin:
```suggestion
   * guaranteed to come from a call to [[canonicalize]], although not every call
```

js-api-doc/importer.d.ts
```diff
@@ -3,32 +3,293 @@ import {URL} from 'url';
...
+   * passed to [[compileAsync]] and [[compileStringAsync]], not [[compile]] or
+   * [[compileString]].
+   *
+   * @throws any - If there's an error while loading `canonicalUrl` This error will be
+   * wrapped in an [[Exception]] associated with the offending load rule.
+   */
   load(canonicalUrl: URL): PromiseOr<ImporterResult | null, sync>;
 
+  /** @hidden */
   findFileUrl?: never;
```
Awjin, line 272: Could you clarify why this field exists? (same with `FileImporter.canonicalize`)

js-api-doc/options.d.ts
```diff
@@ -5,44 +5,403 @@ import {Logger} from './logger';
...
+ * @param args - An array of arguments passed by the function's caller. If the
+ * function takes [arbitrary
+ * arguments](https://sass-lang.com/documentation/at-rules/function#taking-arbitrary-arguments),
+ * the last element will be a [[SassArgumentList]].
+ *
+ * @returns The function's result. This may be in the form of a `Promise`, but
+ * if it is the function may only be passed to [[compileAsync]] and
+ * [[compileStringAsync]], not [[compile]] or [[compileString]].
+ *
+ * @throws any - This function may throw an error, which the Sass compiler will
```
Awjin, line 69: Why not `Error`?

js-api-doc/options.d.ts
```diff
@@ -5,44 +5,403 @@ import {Logger} from './logger';
...
+   *   `Value` objects to more specific types. For values that were passed
+   *   directly as arguments, pass in the argument name as well. This ensures
+   *   that the user gets good error messages when they pass in the wrong type
+   *   to your function.
+   *
+   * * Individual classes may have more specific `assert*` methods, like
+   *   [[SassNumber.assertInt]], which should be used when possible.
+   *
+   * * In Sass, every value counts as a list. Rather than trying to detect the
+   *   [[SassList]] type, and should use [[Value.asList]] to treat all values as
```
Awjin:
```suggestion
   *   [[SassList]] type, you should use [[Value.asList]] to treat all values as
```

js-api-doc/options.d.ts
```diff
@@ -5,44 +5,403 @@ import {Logger} from './logger';
...
+   *   [[SassString.sassIndexToStringIndex]] can be used to do this
+   *   automatically.
+   *
+   * * String indexes in Sass refer to Unicode code points while JavaScript
+   *   string indices refer to UTF-16 code units. For example, the character
+   *   U+1F60A SMILING FACE WITH SMILING EYES is a single Unicode code point but
+   *   is represented in UTF-16 as two code units (`0xD83D` and `0xDE0A`). So in
+   *   JavaScript, `"a😊b".charCodeAt(1)` returns `0xD83D`, whereas in Sass
+   *   `str-slice("a😊b", 1, 1)` returns `"😊"`. Functions should follow Sass's
+   *   convention. The [[SassString.sassIndexToStringIndex]]can be used to do
```
Awjin:
```suggestion
   *   convention. [[SassString.sassIndexToStringIndex]] can be used to do
```

js-api-doc/options.d.ts
```diff
@@ -5,44 +5,403 @@ import {Logger} from './logger';
...
+   * [[importer]]. If the importer doesn't recognize it, it's then passed to
+   * [[importers]] and [[loadPaths]].
+   *
+   * @category Input
+   */
+  importer: Importer<sync> | FileImporter<sync>;
+
+  /**
+   * The canonical URL of the entrypoint stylesheet. If this is passed along
+   * with [[impoter]], it's used to resolve relative loads in the entrypoint
```
Awjin:
```suggestion
   * with [[importer]], it's used to resolve relative loads in the entrypoint
```

js-api-doc/compile.d.ts
```diff
@@ -3,28 +3,150 @@ import {RawSourceMap} from 'source-map-js';
...
+/**
+ * Synchronously compiles the Sass file at `path` to CSS. If it succeeds it
+ * returns a [[CompileResult]], and if it fails it throws an [[Exception]].
+ *
+ * This only allows synchronous [[Importer]]s and [[CustomFunction]]s.
+ *
+ * @example
+ *
+ * ```js
+ * const sass = require('sass');
```
Awjin, line 50: nit: Should we use ESM here and throughout?

```suggestion
 * import * as sass from 'sass';
```

Awjin (changes requested):

js-api-doc/importer.d.ts
stof: This make the type checker complain if you pass an object implementing both `canonicalize` and `findFileUrl` (it is forbidden to try being both a FileImporter and an Importer).

spec/js-api/options.d.ts
```diff
@@ -110,23 +110,28 @@ export interface Options<sync extends 'sync' | 'async'> {
   verbose?: boolean;
 }
 
-/** Additional options specific to compiling from a string. */
-export type StringOptions<sync extends 'sync' | 'async'> = Options<sync> & {
+export interface StringOptionsWithoutImporter<sync extends 'sync' | 'async'>
```
stof, line 124: should this type have `importer?: never`  to ensure that passing an `importer` option gets typed as `StringOptionsWithImporter` and reports a missing `url` if needed.

js-api-doc/compile.d.ts
nex3: Since afaik the latest version of Node still chokes on this without special configuration, I think it's not yet time to do it by default. People who are using ESM will know how to translate better than unfamiliar people. At some point we should make the switch, though.

js-api-doc/exception.d.ts
nex3: Yeah, good call.

js-api-doc/importer.d.ts
nex3: Good thought, I've updated all three to use `@throws` annotations with more detail.

js-api-doc/value/index.d.ts
nex3: It doesn't need to, but it's kind of unsightly in the generated documentation and doesn't really provide any additional information since `Object.toString()` already exists.

js-api-doc/value/list.d.ts
nex3: Yes, it just means that it's *possible* to use the `null` separator for single-element lists—any list may have a non-null separator, but only 0- or 1-element lists may have a null one.

js-api-doc/value/list.d.ts
nex3: Similarly to `Value.toString()`, it doesn't provide any additional documentation beyond the superclass's method so it's better to just omit it. (I wish TypeDoc omitted undocumented overrides by default, but sadly it does not.)

spec/js-api/value/boolean.d.ts
nex3: To match other types, and so that `instanceof SassBoolean` works if necessary.

spec/js-api/options.d.ts
nex3: It turns out this doesn't work because `StringOptionsWithImporter` extends `StringOptionsWithoutImporter`. But that superclass relationship should provide pretty decent error messages on its own.

Awjin (approved):

PR #3200, opened 2021-11-11
Make the new JS API a bit more idiomatic

nex3:
This takes more advantage of overloads rather than using factory
constructors or methods with different names.

### DIFF
--- a/js-api-doc/value/color.d.ts
+++ b/js-api-doc/value/color.d.ts
@@ -1,26 +1,26 @@
 import {Value} from './index';
 
 export class SassColor extends Value {
-  static rgb(
-    red: number,
-    green: number,
-    blue: number,
-    alpha?: number
-  ): SassColor;
+  constructor(options: {
+    red: number;
+    green: number;
+    blue: number;
+    alpha?: number;
+  });
 
-  static hsl(
-    hue: number,
-    saturation: number,
-    lightness: number,
-    alpha?: number
-  ): SassColor;
+  constructor(options: {
+    hue: number;
+    saturation: number;
+    lightness: number;
+    alpha?: number;
+  });
 
-  static hwb(
-    hue: number,
-    whiteness: number,
-    blackness: number,
-    alpha?: number
-  ): SassColor;
+  constructor(options: {
+    hue: number;
+    whiteness: number;
+    blackness: number;
+    alpha?: number;
+  });
 
   get red(): number;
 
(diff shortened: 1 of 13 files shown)


nex3:
I'm basing this on the JS API doc branch so that you can review with documentation, but if it looks good I'll rebase onto main and update the tests and the embedded host.

Awjin (approved):

js-api-doc/value/color.d.ts
```diff
@@ -1,26 +1,26 @@
...
+    alpha?: number;
+  });
 
-  static hsl(
-    hue: number,
-    saturation: number,
-    lightness: number,
-    alpha?: number
-  ): SassColor;
+  constructor(options: {
```
stof, line 11: Given that the 3 overloads all have a single argument and the same return type (none as this is a constructor), the idiomatic way to write such typescript types is to use a union type instead of overloads: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-union-types

js-api-doc/value/color.d.ts
```diff
@@ -40,26 +40,24 @@ export class SassColor extends Value {
 
   get alpha(): number;
 
-  changeRgb(options: {
+  change(options: {
```
stof, line 43: The idiomatic way to write this is also to use a union rather than overloads (btw, that's how you did it in embedded-host-node)

js-api-doc/value/number.d.ts
```diff
@@ -5,13 +5,13 @@ import {Value} from './index';
 export class SassNumber extends Value {
   constructor(value: number, unit?: string);
 
-  static withUnits(
+  constructor(
```
stof, line 6: should also be a union type on the second argument

spec/js-api/value/number.d.ts
```diff
@@ -11,28 +11,34 @@ export class SassNumber extends Value {
...
-   * Creates a Sass number:
-   *
-   * - Set `internal` to a Sass number with value set to `value`, numerator
-   *   units set to `options.numeratorUnits` (if passed), and denominator units
-   *   set to `options.denominatorUnits` (if passed).
-   * - Set `internal` to the result of `simplify`ing `internal`.
-   * - Return `this`.
-   */
-  static withUnits(
+  constructor(
```
stof, line 33: should also be a union type

stof (changes requested):

js-api-doc/value/color.d.ts
nex3: As discussed elsewhere, constructor overloads are necessary here in order to disallow `new Color({hue: ..., red: ...})`.

PR #3226, opened 2021-12-27
Add the sourceMapIncludeSources option in the new JS API

### LINKED ISSUE #3197: [JS API] Add an sourceMapEmbedSources option in the sourcemap API
The CLI has a `embed-sources` option for the source-map to control whether the sources are embedded. The dart API allows to access this feature because it exposes the SingleMapping object directly which has the feature. the legacy JS API also has an option for that
However, the new JS API does not seem to expose such feature. and the issue on the embedded protocol has been closed without feedback: https://github.com/sass/embedded-protocol/issues/46

stof:
Closes https://github.com/sass/sass/issues/3197

### DIFF
--- a/js-api-doc/options.d.ts
+++ b/js-api-doc/options.d.ts
@@ -276,6 +276,16 @@ export interface Options<sync extends 'sync' | 'async'> {
   sourceMap?: boolean;
 
   /**
+   * Whether Sass should include the sources in the generated source map.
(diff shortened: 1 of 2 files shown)


stof:
@nex3 @Awjin is anything missing to move forward with this ?

jathak (approved):

PR #3238, opened 2022-01-05
Add type annotations for the sass.types.Error class

nex3:


js-api-doc/legacy/function.d.ts
```diff
@@ -79,9 +80,73 @@ export type LegacySyncFunction = (
  * APIs. Use [[CustomFunction]] with [[compile]], [[compileString]],
  * [[compileAsync]], and [[compileStringAsync]] instead.
  */
-export type LegacyAsyncFunction = (
-  this: LegacyPluginThis,
-  ...args: [...LegacyValue[], (result: LegacyValue) => void]
+export type LegacyAsyncFunction =
+  | ((this: LegacyPluginThis, done: (result: LegacyValue) => void) => void)
```
stof: shouldn't this use `LegacyAsyncFunctionDone` as well here ?

nex3: Good catch!

jathak (approved):

PR #3268, opened 2022-03-09
Fix FileImporter example documentation to `require` correct module

### LINKED ISSUE #3267: Incorrect url module `require`d in FileImporter documentation
A very small change, and only to documentation.

I spotted a mistake in the example code for [FileImporter](https://sass-lang.com/documentation/js-api/interfaces/FileImporter), where the first line imports [`fileURLToPath`](https://nodejs.org/api/url.html#urlfileurltopathurl) from Node's `url` module. The implementation actually uses [`pathToFileURL`](), so the first line should change from:

https://github.com/sass/sass/blob/f355f602fc15f55b0a0a795ebe6eb819963e08a5/js-api-doc/importer.d.ts#L26

to:

```js
const {pathToFileURL} = require('url');
```

I initially thought this required a change on `sass/sass-site`, and followed the contribution guidelines there, but after finding the change is needed here and on reading the contributing guidelines for this repo that things are different, I haven't submitted my PR, but can do if it's easier. I have the change in a [fork](https://github.com/codemacabre/sass/tree/file-importer-docs-require-fix).

codemacabre:
Fixes #3267.

This PR updates the example code in the [FileImporter](https://sass-lang.com/documentation/js-api/interfaces/FileImporter) documentation, to import Node's `url` module [`pathToFileURL`](https://github.com/sass/sass/issues/3267) instead of [`fileURLToPath`](https://nodejs.org/api/url.html#urlfileurltopathurl).

### DIFF
--- a/js-api-doc/importer.d.ts
+++ b/js-api-doc/importer.d.ts
@@ -23,7 +23,7 @@ import {PromiseOr} from './util/promise_or';
  * @example
  *
  * ```js
- * const {fileURLToPath} = require('url');
+ * const {pathToFileURL} = require('url');
  *
  * sass.compile('style.scss', {
  *   importers: [{


nex3:
Thanks for the fix!

PR #3272, opened 2022-03-15
Don't simplify calcs in supports declarations

### LINKED ISSUE #3259: Calculations in supports queries should not be simplified
Right now

```scss
@supports (top: max(0%)) {
  ...
}
```
simplifies to

```css
@supports (top: 0%) {
  ...
}
```

which is incorrect.

jathak:

Closes #3259.

### DIFF
--- a/spec/types/calculation.md
+++ b/spec/types/calculation.md
@@ -105,7 +105,7 @@ type CalculationValue =
   | CalculationInterpolation
   | CalculationOperation
   | Calculation;
-  
+
 interface CalculationInterpolation {
   value: string;
 }
@@ -192,6 +192,9 @@ This algorithm takes a calculation `calc` and returns a number or a calculation.
 * Let `arguments` be the result of [simplifying](#simplifying-a-calculationvalue) each
   of `calc`'s arguments.
 
+* If `calc` is contained within a supports declaration, return a calculation
+  with the same name as `calc` and `arguments` as its arguments.
+
 * If `calc`'s name is `"calc"`, the syntax guarantees that `arguments` contain
   only a single argument. If that argument is a number or calculation, return
   it.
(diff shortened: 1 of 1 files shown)


spec/types/calculation.md
```diff
@@ -192,6 +192,9 @@ This algorithm takes a calculation `calc` and returns a number or a calculation.
 * Let `arguments` be the result of [simplifying](#simplifying-a-calculationvalue) each
   of `calc`'s arguments.
 
+* If `calc` is contained within a supports declaration, return a calculation
```
nex3: "is contained within a supports declaration" is ambiguous... it could apply to calculations that appear within interpolations, or it could even be read to apply to calculations that are in other statements nested inside a `@supports` rule. It's better to refer to specific parser productions, like those in [this proposal](https://github.com/sass/sass/blob/main/accepted/supports-general.md#syntax).

For example, you could say "If `calc` was parsed from an expression within a `SupportsDeclaration`'s `Expression` but outside any interpolation".

nex3: It's a bit confusing that this bullet point comes after simplifying `calc`'s arguments, but doesn't use the `arguments` variable. I might move it to the beginning and say "If ..., return `calc` as-is."

spec/types/calculation.md
```diff
@@ -257,8 +260,11 @@ This algorithm takes a `CalculationValue` `value` and returns a
 
   [simplifying]: #simplifying-a-calculation
 
-* Otherwise, `value` must be a `CalculationOperation`. Let `left` and `right` be
-  the result of simplifying `value.left` and `value.right`, respectively.
+* Otherwise, `value` must be a `CalculationOperation`. If within a supports
+  declaration, return `value` as-is.
+
+* Otherwise, let `left` and `right` be the result of simplifying `value.left`
+  and `value.right`, respectively.
```
nex3: Do we ever invoke this procedure from anywhere other than "Simplifying a Calculation"? If not, you probably don't need to adjust it.

nex3 (changes requested):

spec/types/calculation.md
jathak: Done

jathak: It was using the `arguments` variable, but only because I was mixing up this procedure with the actual Dart Sass implementation (which does some simplification while evaluating a `CalculationExpression` into a `Calculation`). Done.

spec/types/calculation.md
jathak: Done

nex3 (approved):

PR #3278, opened 2022-04-03
Update spec for relative imports in `compileString()`

### LINKED ISSUE #1657: `url` documentation does not match js-api-spec
https://sass-lang.com/documentation/js-api/interfaces/StringOptionsWithoutImporter#url

> The canonical URL of the entrypoint stylesheet. If this isn't passed along with `StringOptionsWithoutImporter.importer`, it's optional and **only used for error reporting**.

The problem is that the wording `only used for error reporting` is inaccurate as specifying `url` would allow imports to be resolved relative to to this `url`. In some cases setting `url` is not desirable, e.g. in an asset pipeline where files need to be preprocessed before imported by sass. It looks like the only way to prevent sass to resolve imports relative to the `url` (provided for error reporting only) is to not specify `url` at all or specifying a non-file url that sass cannot resolve.

``` js
const sass = require('sass')
const fs = require('fs')
const path = require('path')
const url = require('url')

fs.writeFileSync('child.scss', 'a{b:c}');

console.log(
  sass.compileString('@import "child.scss";', {
    // without url option, the import fails
    // with url option, the import works
    url: url.pathToFileURL(path.join(process.cwd(), 'main.scss'))
  }).css
)
```

ntkme:
The current documentation for the `url` field in `StringOptionsWithoutImporter` is incorrect that it does not match the behavior defined in [js-api-spec](https://github.com/sass/sass-spec/blob/126d17ff9f77372b6f68cb0d43dd357165519c93/js-api-spec/compile.test.ts#L124-L133) and actual implementation:

``` js
    it('url is used to resolve relative loads', () =>
      sandbox(dir => {
        dir.write({'foo/bar/_other.scss': 'a {b: c}'});

        expect(
          compileString('@use "other";', {
            url: dir.url('foo/bar/style.scss'),
          }).css
        ).toBe('a {\n  b: c;\n}');
      }));
```

- Fixes https://github.com/sass/dart-sass/issues/1657.

### DIFF
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -141,7 +141,8 @@ It runs as follows:
     > This importer will only ever be passed absolute URLs, so its base won't
     > matter.
 
-  * If `url` is not a `file:` URL, set `importer` to be a no-op importer.
+  * If `url` is not a `file:` URL, set `importer` to be a function that always
+    returns null.
 
 * Let `file` be the [source file][] with `ast`, canonical URL `url`, and
   importer `importer`.
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -133,7 +133,15 @@ It runs as follows:
     > value, implementations should help users understand the source of the string
     > if possible.
 
-* If `importer` is null, set it to a function that always returns null.
+* If `importer` is null:
+
+  * If `url` is a `file:` URL, set `importer` to be a [filesystem importer] with an
+    arbitrary `base`.
+
+    > This importer will only ever be passed absolute URLs, so its base won't
+    > matter.
+
+  * If `url` is not a `file:` URL, set `importer` to be a no-op importer.
 
 * Let `file` be the [source file][] with `ast`, canonical URL `url`, and
   importer `importer`.
--- a/js-api-doc/options.d.ts
+++ b/js-api-doc/options.d.ts
@@ -333,8 +333,8 @@ export interface Options<sync extends 'sync' | 'async'> {
  * Options that can be passed to [[compileString]] or [[compileStringAsync]].
  *
  * If the [[StringOptionsWithImporter.importer]] field isn't passed, the
- * entrypoint file can't load files relative to itself and the [[url]] field is
- * optional.
+ * entrypoint file can load files relative to itself if a `file://` URL is
+ * passed to the [[url]] field.
  *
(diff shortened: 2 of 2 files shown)


nex3:
This behavior doesn't match the behavior [specified here](https://github.com/sass/sass/blob/main/spec/spec.md#compiling-a-string). (Specifically, the part that says "If `url` is not null and `importer` is null, set `importer` to a function that always returns null.") However, after talking it over we've decided that the current actual behavior is likely to be more intuitive for users, so we should stick with it and update the documentation.

**tl;dr**: Please update [the "Compiling a String" procedure](https://github.com/sass/sass/blob/main/spec/spec.md#compiling-a-string) to reflect this behavior as well.

spec/spec.md
```diff
@@ -133,7 +133,15 @@ It runs as follows:
...
-* If `importer` is null, set it to a function that always returns null.
+* If `importer` is null:
+
+  * If `url` is a `file:` URL, set `importer` to be a [filesystem importer] with an
+    arbitrary `base`.
+
+    > This importer will only ever be passed absolute URLs, so its base won't
+    > matter.
+
+  * If `url` is not a `file:` URL, set `importer` to be a no-op importer.
```
nex3: Since we haven't defined the term "no-op" importer, it's probably better to continue to spell out "a function that always returns null".

nex3 (changes requested):

nex3:
I'm treating this as a fast-track proposal. Although the *spec change itself* is backwards-incompatible, the new specification wording matches the existing behavior of the reference implementation so this is not considered a backwards-incompatible change. I'll leave this open for two workdays from now in case anyone has any objections to this change.

ntkme:
List of related PRs to make implementations consistent with the spec update:
- https://github.com/sass/sass-spec/pull/1785
  - Update JS API spec.
- https://github.com/sass/dart-sass/pull/1664
  - Fix dart-sass to set NoOpImporter when `url` is provided and is not a file: URL.
- https://github.com/sass/dart-sass-embedded/pull/83
  - Fix dart-sass-embedded to set NoOpImporter when `url` is not a file: URL.
- https://github.com/sass/embedded-host-node/pull/128
  - Fix embedded-host-node to force a FileSystemImporter for legacy API when `filename` is not provided.

@nex3 PTAL. Thanks!

Issue #1950, opened 2016-01-05
string.split() function


xzyfer:
* [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/string-split.md) (https://github.com/sass/sass/pull/3284)
* [x] [Tests](https://github.com/sass/sass-spec/issues/1832) (https://github.com/sass/sass-spec/pull/1841)
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1821) (https://github.com/sass/dart-sass/pull/1839)
* [x] [Documentation](https://github.com/sass/sass-site/issues/684) (https://github.com/sass/sass-site/pull/701)
* [x] Update canonical spec (https://github.com/sass/sass/pull/3502)

---

It would be really useful to have a `str-split($string, $delimiter:" ")` function akin to [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split), or [Ruby](http://ruby-doc.org/core-2.2.0/String.html#method-i-split).

The primary use case I run into is splitting selector strings i.e.

```
.foo .bar { /* code */ }
```

It's rather difficult to get `.bar` from `&` without boilerplate string functions or importing something like sassdash.

A another use case I run into is remove part of selector. As it stands `selector-replace` does not allow you to remove part of a selector (i.e. replace it with `""` or `null`). Being able to easily split selector strings would makes this much easier.

cimmanon:
Selectors aren't strings.  They're lists of lists.  If you want to remove or capture part of a selector, you use a loop.

```
.foo .bar, .one .two {
    @each $sel in & {
        @debug nth($sel, -1);
    }
}
```

Output:

```
>> DEBUG: .bar
>> DEBUG: .two
```

xzyfer:
> Selectors aren't strings. They're lists of lists.

Yes I understand.

> If you want to remove or capture part of a selector, you use a loop.

You're correct.

---

As it turns out, in my experimentation, where I was mistaken was that even if there is one selector,`&` is still a nested list. This is where I was getting tripped up.

i.e.

```
.foo .bar { 
  a: length(&); // 1
  b: length(nth(&, 1)); // 2
}
```

I was expecting `a` to be `b`. I understand why it works this way, it was just a detail I had missed.

xzyfer:
[...]

PR #3284, opened 2022-04-06
Proposal for string.split 

dvdherron:
Proposal for adding `split()` to the `Strings` module  based on Issue #1950


proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+
+- ```scss
+  $font: "Helvetica"
+  string.split($font, ''); // "H" "e" "l" "v" "e" "t" "i" "c" "a"
+  ```
+
+
+## Semantics
+
+This proposal adds the ability to split a `string` to the
```
jathak: Nit: This sentence feels redundant with the summary. It's probably best to just remove it and have the semantics section start the `string.split()` semantics.

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+`sass:string` module.
+
+### `string.split()`
+
+```
+split($string, $separator, $limit)
+```
+
+* If `$string` is not a `string`, throw an error.
```
jathak: Types shouldn't be in backticks (here and elsewhere).

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+  string.split($font, ''); // "H" "e" "l" "v" "e" "t" "i" "c" "a"
+  ```
+
+
+## Semantics
+
+This proposal adds the ability to split a `string` to the
+`sass:string` module.
+
+### `string.split()`
```
jathak: Nit: For consistency with other proposals, this header should just be `split()`.

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+
+```
+split($string, $separator, $limit)
+```
+
+* If `$string` is not a `string`, throw an error.
+
+* If `$string` is an empty `string`(`''`), or $separator` is `null`,
```
jathak: We should just throw an error if the separator is null.

It also looks like you missed a backtick before `$separator`. Giving an example of an empty string is probably also unnecessary.

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Semantics
+
+This proposal adds the ability to split a `string` to the
+`sass:string` module.
+
+### `string.split()`
+
+```
+split($string, $separator, $limit)
```
jathak:
```suggestion
split($string, $separator, $limit: null)
```

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+* If `$string` is not a `string`, throw an error.
+
+* If `$string` is an empty `string`(`''`), or $separator` is `null`, 
+return a `list` with `$string` as the only item.
+
+* If `$limit` is 0 or `null`, return an empty `list`.
```
jathak:
```suggestion
* If `$limit` is 0, return an empty `list`.
```

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+
+* If `$separator` is empty (`''`), return a `list` of each individual 
+character in `$string`.
+
+* If `$limit` is a value other than an `integer` or `null`, return an
```
jathak: This should error instead.

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+* If `$limit` is a value other than an `integer` or `null`, return an 
+empty list.
+
+* Let `split-list` be an empty `list`.
+
+* Let `length` be the value of calling `string.length($string)`.
+
+* Let `index` be the value of calling `string.index($string, $separator)`.
+
+* If `$limit == > 0`:
```
jathak: We already know that `$limit` can't be zero at this point (or we would have returned an empty list above). It's probably better to just make passing a negative number an error as well, and have this section just set `$limit` to the length if it's null.

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+
+  * Let `limit` be the value of `$limit`.
+
+  * Otherwise, let `limit` be the value of `string.length($string)`.
+
+* While `list.length(split-list)` is less than `$limit`:
+
+    * Call `string.index($string, $separator)` to find the first
```
jathak: Phrase this as "Let `index` be the result of calling..." so you can refer to `index` below.

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+
+  * Otherwise, let `limit` be the value of `string.length($string)`.
+
+* While `list.length(split-list)` is less than `$limit`:
+
+    * Call `string.index($string, $separator)` to find the first 
+    instance of the `$separator`.
+
+    * Let `current-substring` be the returned value of calling 
+    `string.slice($string, 1, `index` - 1)`.
```
jathak:
```suggestion
    `string.slice($string, 1, index - 1)`.
```

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+
+    * Append `current-substring` to the end of `split-list`.
+    
+    * Set `$string` to `string.slice($string, index + string.length($separator)`.
```
jathak:
```suggestion
    * Set `$string` to `string.slice($string, index + string.length($separator))`.
```

jathak (changes requested):
This looks good overall; most of my comments are about changes in what arguments are allowed and some formatting tweaks.

Thanks!

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+
+A third argument can limit the number of `strings` 
+returned in the list:
+
+- ```scss
+  string.split($fonts, ', ', 2); // "Helvetica Neue" "Helvetica"
+  ```
+
+
+An empty `$separator` returns all characters in the original `string`:
```
stof: I suggest adding explicitly that this splits by Unicode codepoints (as that's what sass strings count as characters), not by grapheme clusters or by UTF-16 codeunits (Javascript splits by codepoints as its indexes are also based on codepoints)

proposal/string-split.md
```diff
@@ -0,0 +1,110 @@
...
+  - ```scss
+    $fonts: "Helvetica Neue, Helvetica, Arial";
+    string.split($fonts, ', '); // "Helvetica Neue" "Helvetica" "Arial"
+    ```
+
+A third argument can limit the number of `strings` 
+returned in the list:
+
+- ```scss
+  string.split($fonts, ', ', 2); // "Helvetica Neue" "Helvetica"
```
stof: should the limit truncate the end of the string, or should it take all the remainder in the last part ? Javascript is one of the few languages implementing the truncation (which is easy to achieve otherwise by doing an unlimited split and slicing the array)

proposal/string-split.md
dvdherron: Removed this sentence.

proposal/string-split.md
dvdherron: Removed these, here and elsewhere.

proposal/string-split.md
dvdherron: Good catch. I made both of the changes you suggested here.

proposal/string-split.md
dvdherron: Fixed. Changed this to explicitly reference Unicode code points in the list of examples and in the semantics section.

proposal/string-split.md
dvdherron: Changed this heading to `split()`

proposal/string-split.md
dvdherron: A null separator now throws an error. And I fixed the backtick before `$separator` and removed the empty string example

proposal/string-split.md
dvdherron: The default value of $limit is changed to null.

proposal/string-split.md
dvdherron: This throws an error now.

proposal/string-split.md
dvdherron: This was the second reference, so I changed the first mention of the value of `index` above this.

proposal/string-split.md
dvdherron: @stof Are you saying, should this work by 1) making all the splits and then slicing the array at the limit or 2) only splitting up to the limit and then stopping? 

Do you think there would be downsides to taking one approach or the other?

proposal/string-split.md
dvdherron: fixed

proposal/string-split.md
dvdherron: fixed

proposal/string-split.md
dvdherron: Removed `or null` here

proposal/string-split.md
stof: Well, to me, it would be more useful if `limit` would work by splitting the string on the `limit -1` first separators, with the last item containing all the remaining work.
Your current proposal does the opposite (truncating the remaining parts), and this could be achieved through an unlimited split and then slicing the list (but I just figured out that `list.slice` does not exist in Sass).

I remember reading an article a few weeks ago comparing the behavior of the string splitting API of lots of different languages in that regard (but unfortunately, I did not manage to find it back for now to share it here. But AFAIR, most languages are providing the non-truncating version of the API (Javascript being one of the exceptions).

Also, a benefit of the non-truncation is that we don't loose content when splitting the string, so joining the list items with the separator (not exposed as a function in `sass:list` either btw) always produces the original string.

proposal/string-split.md
```diff
@@ -0,0 +1,105 @@
...
+* [Background](#background)
+* [Summary](#summary)
+* [Semantics](#semantics)
+  * [`split()`](#split)
+
+## Background
+
+> This section is non-normative.
+
+The `sass:string` module contains several functions for
```
jathak: Nit: Can you reflow this to 80-character lines? These first few paragraphs use shorter lines than necessary and a few of the lines down below are longer than 80 characters.

proposal/string-split.md
```diff
@@ -0,0 +1,105 @@
...
+
+### `split()`
+
+```
+split($string, $separator, $limit: null)
+```
+
+* If `$string` is not a string, throw an error.
+
+* If `$string` is an empty string, return a list with `$string` as the only item.
```
jathak: We should handle all the error cases before returning any actual value. `string.split("", null)` should error.

proposal/string-split.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Semantics
+
+### `split()`
+
+```
+split($string, $separator, $limit: null)
+```
+
+* If `$string` is not a string, throw an error.
```
jathak, line 66: We should also check that `$separator` is a string.

proposal/string-split.md
```diff
@@ -0,0 +1,105 @@
...
+
+* If `$string` is an empty string, return a list with `$string` as the only item.
+
+* If `$separator` is `null`, throw an error.
+
+* If `$separator` is empty (`''`), return a list consisting of each Unicode code point in `$string`.
```
jathak: This should still respect `$limit`.

proposal/string-split.md
```diff
@@ -0,0 +1,105 @@
...
+
+* If `$limit` is a negative number, throw an error. 
+
+* If `$limit` is 0, return an empty list.
+
+* Let `split-list` be an empty list.
+
+* Let `length` be the result of calling `string.length($string)`.
+
+* Let `index` be the result of calling `string.index($string, $separator)`.
```
jathak: Instead of trying to call this line like a function, this line should replace the `Call index to find...` line inside the while loop, since `index` is not referenced anywhere outside the loop.

proposal/string-split.md
jathak: I'd lean towards matching the JS behavior here. @nex3: Do you have any thoughts on this?

jathak (commented):
Sorry for the delay on re-reviewing this!

proposal/string-split.md
```diff
@@ -0,0 +1,105 @@
...
+string.split($font, ''); // "H" "e" "l" "v" "e" "t" "i" "c" "a"
+```
+
+
+## Semantics
+
+### `split()`
+
+```
+split($string, $separator, $limit: null)
```
roydukkey, line 63: Since we are returning a list here, should we allow configuration of the list's separator and brackets? Such as:

```
split ($string, $delimiter, $limit: null, $separator: auto, $bracketed: auto)
```

dvdherron:
> Sorry for the delay on re-reviewing this!

All good @jathak. I'm back in the office in another week-and-a-half or so and I'll adress your latest review then!

proposal/string-split.md
dvdherron: Reflowed. All lines should be the correct length now.

proposal/string-split.md
dvdherron: Moved the error cases before any returned values and made a null `separator` error.

proposal/string-split.md
dvdherron: If `$separator` is not a string it errors now.

proposal/string-split.md
dvdherron: Moved this logic into the main `while` loop to respect `$limit`.

proposal/string-split.md
dvdherron: Moved this line inside of the `while` loop.

proposal/string-split.md
nex3: I agree with @stof here. I think JS's behavior is a misfeature, and we should match other languages in preserving the entire remainder of the string.

proposal/string-split.md
jathak: Adding configuration here isn't really worth the complexity. For other functions that return a list without having an existing one to base the separator and brackets on, we always return an unbracketed, comma-separated list.

We should specify that explicitly in this proposal to match the specs of those other functions (e.g. [meta.calc-args()](https://github.com/sass/sass/blob/main/spec/built-in-modules/meta.md#calc-args)).

proposal/string-split.md
jathak: In that case, let's change this to do that.

jathak (changes requested):
Just realized that I left comments, but never submitted a formal review. The changes necessary now are to change the limit parameter so it adds the remaining text to the last item, and to explicitly specify that it returns an unbracketed, comma-separated list.

dvdherron:
> Just realized that I left comments, but never submitted a formal review. The changes necessary now are to change the limit parameter so it adds the remaining text to the last item, and to explicitly specify that it returns an unbracketed, comma-separated list.

Thanks @jathak . I'm currently working on edits based on re-working how `limit` works and your review. I should have something to show by the week's end or early next week at the latest.

proposal/string-split.md
dvdherron: @jathak I've updated this section (and the examples) to show `limit` working as noted in the conversation above. The function should preserve the remainder of the string now. 

@mirisuzanne made a function based on these steps that you can see [here](https://www.sassmeister.com/gist/19ec36ad3ee106ba98ad8af31a3577a4)

dvdherron:
> Just realized that I left comments, but never submitted a formal review. The changes necessary now are to change the limit parameter so it adds the remaining text to the last item, and to explicitly specify that it returns an unbracketed, comma-separated list.

@jathak I just wanted to give you a heads up that I will be away starting next week and back around early August. I can respond to your upcoming review then, or if something is urgent, feel free to reach out to the [OddBird](https://github.com/oddbird) team in the meantime.

proposal/string-split.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+* If `$string` is not a string, throw an error.
+
+* If `$separator` is `null` or is not a string, throw an error.
```
jathak:
```suggestion
* If `$separator` is not a string, throw an error.
```

`null` is already not a string, so we don't need to specifically call it out here.

proposal/string-split.md
```diff
@@ -0,0 +1,131 @@
...
+
+* Let `length` be the result of calling `string.length($string)`.
+
+* Let `limit` be the value of `$limit`.
+
+* Otherwise, if `$limit` is `null`, set `limit` to the value of `length`.
+
+* Let `split-counter` equal 0.
+
+* While `split-counter` is `<=` `limit` and `length` is `>` 0:
```
jathak:
```suggestion
* While `split-counter <= limit` and `length > 0`:
```

proposal/string-split.md
```diff
@@ -0,0 +1,131 @@
...
+
+  * If `split-counter` `==` `limit`:
```
jathak:
```suggestion
  * If `split-counter == limit`:
```

proposal/string-split.md
```diff
@@ -0,0 +1,131 @@
...
+
+* If `$limit` is 0, return a list with `$string` as the only item.
+
+* Let `split-list` be an empty list.
+
+* Let `length` be the result of calling `string.length($string)`.
+
+* Let `limit` be the value of `$limit`.
+
+* Otherwise, if `$limit` is `null`, set `limit` to the value of `length`.
```
jathak: It's probably simpler to just say:
```
* If `$limit` is `null`, set `$limit` to the value of `length`.
```
and then refer to `$limit` below.

proposal/string-split.md
```diff
@@ -0,0 +1,131 @@
...
+          `string.slice($string, 1, index - 1)`.
+
+        * Append `current-substring` to `split-list`.
+    
+        * Set `$string` to 
+          `string.slice($string, index + string.length($separator))`.
+
+        * Increase `split-counter` by 1.
+
+  * Set `length` to `string.length($string)`.
```
jathak: Since we have to update it after each change to `$string` anyway, and `length` is only referenced once each loop, I think it makes more sense to eliminate the `length` variable entirely and just call `string.length($string)` when we need it.

jathak (commented):
Sorry for the long delay on reviewing this!

The logic here all looks good, my only remaining comments are formatting/eliminating redundancies.

proposal/string-split.md
dvdherron: Makes sense. I removed the `length` variable and inserted calls to `string.length($string)` where needed.

proposal/string-split.md
dvdherron: fixed

proposal/string-split.md
dvdherron: fixed

proposal/string-split.md
dvdherron: removed the `null` reference here.

proposal/string-split.md
dvdherron: fixed

dvdherron:
Hi @jathak! These changes should cover the last bit of cleanup you suggested. Ready for another look whenever you get a chance.

jathak (approved):
Thank you for all your work on this! Sorry it's taken so long to get approved

PR #3289, opened 2022-04-16
Fix the algorithm computing the default use namespace

stof:
The described algorithm contains a mistake. It should use the `basename` defined in the previous step to compute the module name, not the full path.

dart-sass already implements the proper behavior using the basename, so this makes the spec match the reference implementation.


jathak (approved):
Thanks!

jathak:
Waiting for the (unrelated) link referents check failure to be fixed in a separate PR, and then I'll merge this

stof:
@jathak this check failure is now fixed

Issue #3285, opened 2022-04-13
Support for cascade layers @import urll()


thomaslaker:
https://developer.mozilla.org/en-US/docs/Web/CSS/@layer

Currently unable to use cascade layers @import url() in sass although the latest version of all major browsers support them.

Example can be seen here: https://codepen.io/tomlaker/pen/gOodjvY

PR #3290, opened 2022-04-16
Update specs to mention CSS layer conditions for CSS imports

stof:
Refs #3285

### DIFF
--- a/spec/at-rules/import.md
+++ b/spec/at-rules/import.md
@@ -15,15 +15,24 @@ still supported for backwards-compatibility.
 
 <x><pre>
 **ImportRule**                ::= '@import' ImportArgument (',' ImportArgument)*
-**ImportArgument**            ::= ImportUrl ImportLayer? ImportSupportsDeclaration? [MediaQueryList][]?
-**ImportLayer**                 ::= 'layer' ('(' InterpolatedIdentifier ('.' InterpolatedIdentifier)* ')')?
+**ImportArgument**            ::= ImportUrl ImportModifier*
+**ImportModifier**            ::= ImportFunction | ImportSupports | [MediaQueryList]
+**ImportSupports**            ::= 'supports(' SupportsDeclaration ')'
+**ImportFunction**            ::= [InterpolatedIdentifier]¹ '(' InterpolatedDeclarationValue? ')'
 **ImportUrl**                 ::= QuotedString | [InterpolatedUrl][]
-**ImportSupportsDeclaration** ::= 'supports(' SupportsDeclaration ')'
 </pre></x>
 
 [InterpolatedIdentifier]: ../syntax.md#InterpolatedIdentifier
 [InterpolatedUrl]: ../syntax.md#InterpolatedUrl
 [MediaQueryList]: media.md#syntax
+
+1: This identifier may not be the `"supports"`. No whitespace is allowed between
+   it and the following `(`.
+
+> Note that this parses `@import "..." layer` differently than the CSS standard:
+> in CSS, `layer` is a CSS layering keyword but Sass parses it as part of a
+> media query. This doesn't pose a problem in practice because Sass's semantics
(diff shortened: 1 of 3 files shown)


nex3:
Since this is a new feature, it should explicitly define "layer query". That means it's probably time to write out explicit BNF syntax for the `@import` rule.

nex3:
Actually, thinking about this more: it may be a better idea to future-proof by allowing *any* sequence of identifiers or functions after an `@import`. That way we won't have to explicitly add support for any future modifiers like this unless they use some highly unexpected syntax.

stof:
Well, supports condition has a special handling, and media queries are parsed as well (which requires them to have a special handling as well). And given that the order of import conditions is fixed in the CSS spec and is what allows to find the beginning of the media query, I fear that we cannot have a mix of allowing anything and having special handling for some of them.

stof:
however, I agree that writing the BNF for the import rule could be a good idea. I'll look into it.

nex3:
> Well, supports condition has a special handling, and media queries are parsed as well (which requires them to have a special handling as well). And given that the order of import conditions is fixed in the CSS spec and is what allows to find the beginning of the media query, I fear that we cannot have a mix of allowing anything and having special handling for some of them.

I think if we allow any sequence of identifiers, function calls, `supports()` functions, or media query lists it should work—we'll have to say that `supports()` functions aren't parsed as normal functions, and we won't give as rigorous of checks as to correct order, but I'd rather not have Sass in the business of correctness checks anyway.

stof:
@nex3 but identifiers vs media query lists are ambiguous, as `screen` is both of them for instance.

nex3:
That's fine, though, since they both produce the same output in the end and nothing depends on a given identifier being parsed as one versus the other. We can just say "in cases of ambiguity parse it as a plain identifier".

nex3:
FYI I'd still like to make this parsing more general.

nex3:
@stof I went ahead and generalized the parsing. Let me know what you think, and whether you have time to update the specs accordingly.

spec/at-rules/import.md
```diff
@@ -15,16 +15,25 @@ still supported for backwards-compatibility.
...
 [InterpolatedUrl]: ../syntax.md#InterpolatedUrl
 [MediaQueryList]: media.md#syntax
 
+1: This identifier may not be the `"supports"`. No whitespace is allowed between
+   it and the following `(`.
+
+> Note that this parses `@import "..." layer` differently than the CSS standard:
+> in CSS, `layer` is a CSS layering keyword but Sass parses it as part of a
+> media query. This doesn't pose a problem in practice because Sass's semantics
+> never depend on how import modifiers are parsed.
```
stof: this will be an issue if dart-sass keeps parsing the media query list to produce the CSS AST, because `layer screen` will not be a valid media query list, and would throw in that parsing.

stof:
@nex3 I updated the sass-spec PR, replacing all wrong_order error specs with specs that intentionally requires supporting out-of-order modifiers at the Sass level.

spec/at-rules/import.md
nex3: That's a problem for future me :smile:

nex3 (approved):

nex3:
While implementing this, I realized there's are some more ambiguities:
1. `@import "..." a b(c)` could be parsed as either:
   1. `MediaQuery "a", ImportFunction "b(c)"`
   2. `MediaQuery "a b", MediaQuery "(c)"`
2. `@import "..." a and(b)` could be parsed as either:
   1. `MediaQuery "a", ImportFunction "and(b)"`
   2. `MediaQuery "a and(b)"`

1i is correct but implementing that way is difficult in a greedy parser; plus, implementing it naively could break 2ii (which is correct in that case). I'll probably have to add some more complexity to the spec to fix this.

jathak (approved):

PR #3293, opened 2022-04-18
Mark the deep-merge-order spec as accepted

nex3:
This has been open for comment for four weeks.


stof:
shouldn't this also update https://github.com/sass/sass/blob/main/spec/built-in-modules/map.md#deep-merge to update the algorithm in the spec of the function ?

jathak (approved):

nex3:
Good call, done.

PR #3294, opened 2022-04-19
Add the syntax definition for the import rule

stof:
This adds the BNF for the existing support of `@import`. Once it is merged, I will update #3290 to update that BNF with the layer condition.

I took 2 shortcuts here:
- I referenced `SupportsDeclaration` without declaring it (declaring it would be the scope of the BNF for the `@supports` rule, which is not yet in the spec). Note that there are precedents for that (for instance, `ArgumentDeclaration` is used in several places but not declared either)
- for `InterpolatedDynamicUrl`, I used a `unescaped url contents` pointing to the CSS spec about url-token, even though this is only a subpart of the `url-token` diagram and not a reusable diagram on its own (unlike `escape`)

### DIFF
--- a/spec/at-rules/import.md
+++ b/spec/at-rules/import.md
@@ -6,6 +6,23 @@ still supported for backwards-compatibility.
 
 [`@use` rule]: use.md
 
+## Table of Contents
+
+* [Syntax](#syntax)
+* [Semantics](#semantics)
+
+## Syntax
+
+<x><pre>
+**ImportRule**                ::= '@import' ImportArgument (',' ImportArgument)*
+**ImportArgument**            ::= ImportUrl ImportSupportsDeclaration? [MediaQueryList][]?
+**ImportUrl**                 ::= QuotedString | [InterpolatedUrl][]
+**ImportSupportsDeclaration** ::= 'supports(' SupportsDeclaration ')'
+</pre></x>
+
+[InterpolatedUrl]: ../syntax.md#InterpolatedUrl
+[MediaQueryList]: media.md#syntax
+
 ## Semantics
 
 To execute an `@import` rule `rule`:
@@ -16,8 +33,9 @@ To execute an `@import` rule `rule`:
 
     * `argument`'s URL string begins with `http://` or `https://`.
     * `argument`'s URL string ends with `.css`.
-    * `argument`'s URL string is syntactically defined as a `url()`.
-    * `argument` has a media query and/or a supports query.
+    * `argument`'s URL is an `InterpolatedUrl`.
+    * `argument` has an `ImportSupportsDeclaration`.
+    * `argument` has a `MediaQueryList`.
 
     > Note that this means that imports that explicitly end with `.css` are
     > treated as plain CSS `@import` rules, rather than importing stylesheets as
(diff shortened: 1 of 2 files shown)


spec/syntax.md
```diff
@@ -50,6 +51,17 @@ as the *unprefixed identifier*.
 
 No whitespace is allowed between components of an `InterpolatedIdentifier`.
 
+### `InterpolatedDynamicUrl`
+
+<x><pre>
+**InterpolatedDynamicUrl**         ::= 'url(' (QuotedString | InterpolatedDynamicUnquotedUrl) ')' | InterpolatedDynamicUnquotedUrl
```
nex3: What's up with the `| InterpolatedDynamicUnquotedUrl` at the end here?

spec/syntax.md
```diff
@@ -50,6 +51,17 @@ as the *unprefixed identifier*.
 
 No whitespace is allowed between components of an `InterpolatedIdentifier`.
 
+### `InterpolatedDynamicUrl`
```
nex3: I'm not sure what "Dynamic" means here. Is it intended to disambiguate an interpolated URL that can't contain a SassScript expression from one that can?

spec/at-rules/import.md
```diff
@@ -6,6 +6,23 @@ still supported for backwards-compatibility.
...
+## Table of Contents
+
+* [Syntax](#syntax)
+* [Semantics](#semantics)
+
+## Syntax
+
+<x><pre>
+**ImportRule**                ::= '@import' ImportArgument (',' ImportArgument)*
+**ImportArgument**            ::= ImportUrl ImportSupportsDeclaration? [MediaQueryList][]?
```
nex3, line 18: It's probably a good idea to update the semantics below to explicitly refer to `ImportSupportsDeclaration` and `MediaQueryList` rather than the looser "supports query" and "media query".

nex3 (changes requested):
Thanks for doing this!

spec/syntax.md
stof: well, `dynamicUrl` is the way it is called in the dart-sass source code.

spec/syntax.md
stof: that's a typo, left-over from a change I made

stof:
@nex3 updated

spec/at-rules/import.md
```diff
@@ -17,7 +34,8 @@ To execute an `@import` rule `rule`:
     * `argument`'s URL string begins with `http://` or `https://`.
     * `argument`'s URL string ends with `.css`.
     * `argument`'s URL string is syntactically defined as a `url()`.
-    * `argument` has a media query and/or a supports query.
+    * `argument` has an ImportSupportsDeclaration.
+    * `argument` has a MediaQueryList.
```
nex3: Nit

```suggestion
    * `argument` has an `ImportSupportsDeclaration`.
    * `argument` has a `MediaQueryList`.
```

spec/syntax.md
nex3: That's for implementation-level reasons (specifically, to disambiguate it from the way we parse "static" URLs in plain-CSS stylesheets). I think just saying `InterpolatedUrl` here is fine.

nex3 (changes requested):

stof:
@nex3 updated

nex3 (approved):

PR #3311, opened 2022-05-10
Update hsl/hwb specs to reflex current CSS logic

### LINKED ISSUE #3298: Legacy typo in the Sass spec for hsl/hwb?
The `spec/functions.md` definition of `hsl()` has this logic for managing hue in a conversion to rgb:

```md
* Let `hue` be the result of [converting] `$hue` to `deg` allowing unitless.

* Set `hue` to `(hue % 360deg) / 60deg`.

* ... (some satuarion/lightness handling)

* Let `red`, `green`, and `blue` be the result of converting `hue`,
    `saturation`, and `lightness` [to RGB][].

[converting]: types/number.md#converting-a-number-to-a-unit
 [to RGB]: https://www.w3.org/TR/css-color-4/#hsl-to-rgb
```

There is similar logic in the `hwb` accepted proposal. The second point there looks wrong to me, and like it might be leftover from a previous draft of the CSS spec? The `(hue % 360deg)` part is harmless, but already handled in the linked pseudocode for conversion. However, I can't find any reason to divide by `60` given the current conversion functions. It seems like that should be removed? Or am I missing something?

jathak:
Fixes #3298.

### DIFF
--- a/accepted/color-4-hwb.md
+++ b/accepted/color-4-hwb.md
@@ -78,7 +78,7 @@ are left to a future proposal.
 
 ## Procedures
 
-### Scaling a Number 
+### Scaling a Number
 
 This algorithm takes a number `number`, a value `factor`, and a number `max`.
 It's written "scale `<number>` by `<factor>` with a `max` of `<max>`". It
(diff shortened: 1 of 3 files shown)


nex3 (approved):

mirisuzanne (commented):
lgtm!

PR #3336, opened 2022-06-16
Explicitly define `FunctionExpression`

### LINKED ISSUE #3245: CSS variable (custom property) with fallback as nothing compiles to no fallback
- [ ] [Spec](https://github.com/sass/sass/pull/3338)
- [ ] [Tests](https://github.com/sass/sass-spec/pull/1794)
- [ ] [Dart Sass](https://github.com/sass/dart-sass/pull/1723)

---

https://www.sassmeister.com/gist/64d412cd7fcc65d2fd09d10c89cb2e0f

```css
div {
    transform: translate(100px) var(--transform-y, );
}
```

The `, ` at the end is valid CSS and is no mistake and should not be removed after compiling.

---
#### Read more about the use-case in this article:

https://css-tricks.com/optional-custom-property-values-trick/

#### W3 ref:
https://www.w3.org/TR/css-variables-1/#using-variables

> In an exception to the usual comma elision rules, which require commas to be omitted when they’re not separating values, a bare comma, with nothing following it, must be treated as valid in var(), indicating an empty fallback value.

nex3:
This sets the stage for modifying it to fix #3245

### DIFF
--- a/spec/functions.md
+++ b/spec/functions.md
@@ -45,6 +45,28 @@ matching is case-insensitive.
 ## Syntax
 
 <x><pre>
+**FunctionExpression**¹ ::= [CssMinMax]
+&#32;                     | [SpecialFunctionExpression]
+&#32;                     | [CalculationExpression]
+&#32;                     | FunctionCall
+**FunctionCall**²       ::= [NamespacedIdentifier] ArgumentInvocation
+</pre></x>
+
+[CssMinMax]: types/calculation.md#cssminmax
+[SpecialFunctionExpression]: syntax.md#specialfunctionexpression
+[CalculationExpression]: types/calculation.md#calculationexpression
+[NamespacedIdentifier]: modules.md#syntax
+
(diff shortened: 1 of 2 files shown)


spec/syntax.md
```diff
@@ -71,6 +72,8 @@ No whitespace is allowed between components of an `InterpolatedUnquotedUrlConten
 [identifier code point]: https://drafts.csswg.org/css-syntax-3/#identifier-code-point
 [escape]: https://drafts.csswg.org/css-syntax-3/#escape-diagram
 
+### `FunctionExpression`
```
Goodwine: Is there a link missing in this section pointing to spec/functions.md? Or is this not meant to be added here?

Goodwine (changes requested):

spec/syntax.md
nex3: Oh, this is just a leftover from my previous structure. Removed.

Goodwine (approved):

Issue #3340, opened 2022-06-18
Drop support for non-standard combinator arrangements


nex3:
* [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/bogus-combinators.md) (https://github.com/sass/sass/pull/3342)

**Deprecation**

* [x] [Tests](https://github.com/sass/sass-spec/issues/1797) (https://github.com/sass/sass-spec/pull/1803)
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1727) (https://github.com/sass/dart-sass/pull/1740)
* [x] [Documentation](https://github.com/sass/sass-site/issues/647) (https://github.com/sass/sass-site/pull/654)
* [x] Update canonical spec (https://github.com/sass/sass/pull/3362)

**Removal**

* [ ] [Tests](https://github.com/sass/sass-spec/issues/1797)
* [ ] [Dart Sass](https://github.com/sass/dart-sass/issues/1727)
* [ ] [Documentation](https://github.com/sass/sass-site/issues/647)

---

Currently, Sass is *very* liberal when it comes to where explicit selector combinators can be written—much more so than CSS itself. We allow:

* Multiple combinators between compound selectors (`a > + b`).
* Combinators at the beginning of selectors (`> a`).
* Combinators at the end of selectors (`a >`).

The latter two are useful when nesting style rules, but the former has no known use at all. Historically, we've said that we support these because they enable some browser hacks, but this seems to be purely hypothetical: [browserhacks.com] is the most comprehensive source of hacks I've found, and it doesn't list any hacks that are enabled by this support, even among its "legacy hacks".

[browserhacks.com]: http://browserhacks.com

In addition, supporting these selectors imposes substantial complexity on Sass implementations. They cause bugs (https://github.com/sass/dart-sass/issues/1053) and make the data model more complex than it needs to be. This in turn makes bugs like https://github.com/sass/sass/issues/1807 that don't involve non-standard combinators more difficult to resolve.

Fortunately, these bogus combinators are unlikely to be widely-used in practice and have no effect in real-world browsers, which gives us considerable freedom to drop support for them without needing to do a lengthy deprecation period. We'll need to be careful to preserve the useful aspects of their behavior, and we'll probably want to make these produce errors eventually, but for the time being we can just cause them to omit selectors. In particular:

* If any of a style rule's selectors contains a leading, trailing, or multiple combinator after nesting is resolved, omit that style rule from the generated CSS and emit a deprecation warning.

[...]

PR #3342, opened 2022-06-18
[Bogus Combinators] Create a proposal

nex3:
See sass/sass#3340

### DIFF
--- a/proposal/bogus-combinators.md
+++ b/proposal/bogus-combinators.md
@@ -1,0 +1,203 @@
+# Bogus Combinators: Draft 1
+
+*([Issue](https://github.com/sass/sass/issues/3340))*
+
+This proposal increases the strictness with which Sass parses and resolves
+non-standard uses of selector combinators. In particular, it forbids the use of
+multiple combinators in a row (such as `div + ~ a`) and limits the use of
+leading combinators (such as `> a`) and trailing combinators (such as `a >`) to
+selector nesting.
+
+## Table of Contents
+
+* [Foo](#foo)
+
+## Background
+
+> This section is non-normative.
+
+Currently, Sass is very liberal when it comes to where explicit selector
+combinators can be written—much more so than CSS itself. It allows:
+
+* Multiple combinators between compound selectors (`a > + b`).
+* Combinators at the beginning of selectors (`> a`).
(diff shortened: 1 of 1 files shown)


proposal/bogus-combinators.md
```diff
@@ -0,0 +1,223 @@
...
+
+[visible combinator]: #visible-combinator 
+
+### Complex Selector Component
+
+~~A *complex selector component* is a compound selector as well as a sequence of
+zero or more [visible combinators].~~
+
+A *complex selector component* is a compound selector as well as a single
+[combinator].
```
stof, line 134: should the combinator be optional ? Or will this use a descendant combinator ?

nex3: It will use a descendant combinator (since it doesn't specify *visible* combinator).

proposal/bogus-combinators.md
```diff
@@ -0,0 +1,223 @@
+# Bogus Combinators: Draft 1
```
jathak, line 1: Is "bogus" an existing term for these sorts of combinators? If not, it might be worth adding a note early on that we're defining a new term.

proposal/bogus-combinators.md
```diff
@@ -0,0 +1,223 @@
...
+
+[Evaluating a Style Rule]: ../spec/style-rules.md#semantics
+
+* If `selector` is [bogus], throw an error.
+
+[bogus]: #bogus-selector
+
+## Functions
+
+For the `selector.append()`, `selector.extend()`, `selector.is-superselector()`,
```
jathak, line 187: Is `selector.extend()` supposed to be included here? #3340 seems to suggest it shouldn't be.

proposal/bogus-combinators.md
nex3: No, there is no existing term since they're not valid according to the CSS spec. I'll add something to the opening paragraph.

proposal/bogus-combinators.md
nex3: Yes; the main body of the proposal specifies the *final* form of the spec, in which bogus combinators are forbidden everywhere. The "Deprecation Process" section specifies the intermediate state during which extend functions will merely issue warnings.

jathak (approved):

Issue #2538, opened 2018-07-12
Support full media query conditions


nex3:
* [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/media-logic.md)

**Deprecation**

* [x] [Tests](https://github.com/sass/sass-spec/issues/1798) (https://github.com/sass/sass-spec/pull/1807)
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1728) (https://github.com/sass/dart-sass/pull/1749)
* [x] [Documentation](https://github.com/sass/sass-site/issues/655) (https://github.com/sass/sass-site/pull/656)
* [x] Update canonical spec (https://github.com/sass/sass/pull/3365)

**Final**

* [x] [Tests](https://github.com/sass/sass-spec/issues/1798) (https://github.com/sass/sass-spec/pull/1833)
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1728) (https://github.com/sass/dart-sass/pull/1822)
* [x] [Documentation](https://github.com/sass/sass-site/issues/655) (https://github.com/sass/sass-site/pull/685)

---

The [Media Queries Level 4](https://www.w3.org/TR/mediaqueries-4/#media-conditions) spec defines syntax for full boolean algebra among media features. At time of writing, I believe no browsers support this syntax, but once one does Sass should add support as well.

In addition to parsing additional forms, this will affect Sass's media query merging logic. For example, merging `not print` and `(min-width: 600px)` should produce `not print and not (min-width: 600px)` rather than failing entirely.

stof:
For reference, this is supported in Firefox 63+.
And looking at the Chromium issue tracker, they have some in-progress work done (probably related to their implementation of container queries, as they rely on that range syntax too)

stof:
Those are announced to be shipping in Chrome 104, which should make the awaited browser support being there as Firefox already shipped them.
https://caniuse.com/mdn-css_at-rules_media_range_syntax

nex3:
I have a PR out for a proposal: https://github.com/sass/sass/pull/3344

nex3:
The proposal has landed. Since it's relatively straightforward and unlikely to change, I'll give it two weeks for public comment.

nex3:
The initial deprecation phase has landed. We'll wait at least three months to give users a chance to migrate away from the deprecated syntax, then land full support.

PR #3344, opened 2022-06-22
[Media Logic] Add a proposal

nex3:
See #2538

### DIFF
--- a/proposal/media-logic.md
+++ b/proposal/media-logic.md
@@ -1,0 +1,129 @@
+# Media Logic: Draft 1
+
+*([Issue](https://github.com/sass/sass/issues/2538))*
+
+This proposal adds support for the full [Media Queries Level 4] syntax for media
+conditions, including arbitrary boolean logic using `and`, `or`, and `not`.
+
+[Media Queries Level 4]: https://www.w3.org/TR/mediaqueries-4/#media-conditions
+
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+* [Syntax](#syntax)
+  * [`MediaQuery`](#mediaquery)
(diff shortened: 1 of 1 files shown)


stof (approved):

proposal/media-logic.md
```diff
@@ -0,0 +1,129 @@
...
+&#32;                | MediaOrInterp (MediaAnd* | MediaOr*)
+&#32;                | MediaType ('and' MediaNot | MediaAnd*)
+**MediaType**      ::= [InterpolatedIdentifier] [InterpolatedIdentifier]¹?
+**MediaNot**       ::= 'not' MediaOrInterp
+**MediaAnd**       ::= 'and' MediaOrInterp
+**MediaOr**        ::= 'or' MediaOrInterp
+**MediaOrInterp**  ::= Interpolation | MediaInParens
+**MediaInParens**  ::= '(' Expression² ')'
+&#32;                | '(' Expression² [\<mf-comparison>] Expression² ')'
+&#32;                | '(' Expression² [\<mf-lt>] Expression² [\<mf-lt>] Expression² ')'
```
Goodwine, line 73: Only question here is that on the csswg spec, <mf-range> defines
`value < name < value` for example

So it makes me think here instead of
`expr < expr < expr`
It maybe should be
`expr < identOrInterpolat < expr`

But this may just be me misunderstanding something

nex3: This part of the syntax is already locked in on the existing spec, but also I think people would expect to be able to use a variable here since to represent the property name since they can already use it as the left-hand side of an `<mf-plain>`.

Goodwine (approved):

Issue #3242, opened 2022-01-08
Add the charset parameter to the new JS API


nex3:
https://github.com/sass/sass/pull/3126 added a `charset` parameter to the legacy JS API, but it didn't make it into the new API. It should be added.

PR #3346, opened 2022-06-24
Add charset option to js-api

ntkme:
https://github.com/sass/sass/issues/3242


Goodwine (approved):

Goodwine (changes requested):
By the way, could you also add a test for sass/sass-spec
Thanks

Goodwine (approved):

PR #3357, opened 2022-07-13
Note configurations are shared across all files in a module by instance given a uniquely-identifying opaque ID

Goodwine:
See:
- https://github.com/sass/sass-spec/pull/1796
- https://github.com/sass/dart-sass/pull/1739

### DIFF
--- a/spec/at-rules/forward.md
+++ b/spec/at-rules/forward.md
@@ -48,9 +48,10 @@ To execute a `@forward` rule `rule`:
 
 * If `rule` has an `AsClause` with identifier `prefix`:
 
-  * Let `rule-config` be an empty [configuration][].
+  * Let `rule-config` be an empty [configuration] inheriting the opaque ID
+    from [the current configuration].
 
-  * For each variable `variable` in [the current configuration][]:
+  * For each variable `variable` in the current configuration:
 
     * If `variable`'s name begins with `prefix`:
 
@@ -66,7 +67,7 @@ To execute a `@forward` rule `rule`:
 
 * If `rule` has a `WithClause`:
 
-  * Set `rule-config` to a copy of itself.
+  * Set `rule-config` to a copy of itself including its opaque ID.
 
   * For each `ForwardWithArgument` `argument` in this clause:
 
@@ -96,7 +97,7 @@ To execute a `@forward` rule `rule`:
 * For every member `member` in `forwarded`:
 
   * Let `name` be `member`'s name.
-  
+
   * If `rule` has an `AsClause` `as`, prepend `as`'s identifier to `name` (after
     the `$` if `member` is a variable).
 
--- a/spec/at-rules/forward.md
+++ b/spec/at-rules/forward.md
@@ -48,8 +48,8 @@ To execute a `@forward` rule `rule`:
 
 * If `rule` has an `AsClause` with identifier `prefix`:
 
-  * Let `rule-config` be an empty [configuration] inheriting the opaque ID
-    from [the current configuration].
+  * Let `rule-config` be an empty [configuration] with the same opaque ID as
+    [the current configuration].
 
   * For each variable `variable` in the current configuration:
 
--- a/spec/modules.md
+++ b/spec/modules.md
@@ -260,11 +260,13 @@ This algorithm takes a string `argument` and [configuration](#configuration)
 
 * If `file` is null, throw an error.
 
-* If `file` has already been [executed][]:
+* If `file` has already been [executed] by the [Loading a Module] procedure:
 
   [executed]: spec.md#executing-a-file
+  [Loading a Module]: #loading-a-module
 
-  * If `config` has a different ID than before and is not empty, throw an error.
+  * If `config` is not empty and has a different ID than the configuration that
+    was passed the first time `file` was executed, throw an error.
 
     > An ID may be reused in a new configuration via [`@forwards .. with`].
 
--- a/spec/modules.md
+++ b/spec/modules.md
@@ -263,16 +263,18 @@ This algorithm takes a string `argument` and [configuration](#configuration)
 * If `file` has already been [executed] by the [Loading a Module] procedure:
 
   [executed]: spec.md#executing-a-file
-  [Loading a Module]: #loading-a-module
 
   * If `config` is not empty and has a different ID than the configuration that
-    was passed the first time `file` was executed, throw an error.
+    was passed the first time `file` was executed by the [Loading a Module]
+    procedure, throw an error.
 
     > An ID may be reused in a new configuration via [`@forwards .. with`].
 
     [`@forwards .. with`]: ../spec/at-rules/forward.md#semantics
 
   * Otherwise, return the module that execution produced.
+
+  [Loading a Module]: #loading-a-module
 
 * If `file` is currently being executed, throw an error.
 
--- a/spec/modules.md
+++ b/spec/modules.md
@@ -38,7 +38,7 @@ Two members are considered identical if they have the same name, type, source
 location, and were defined in or forwarded from the same original module.
 
 > Each member type has its own namespace in Sass, so for example the mixin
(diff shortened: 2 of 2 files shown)


nex3:
Since the module system proposal has already been implemented, this should edit the spec itself (in `spec/module-system.md`).

Goodwine:
Gotcha, I actually wasn't sure following the contributing process, I have updated the pr :)

spec/modules.md
```diff
@@ -52,9 +52,13 @@ An *empty CSS tree* contains no statements.
 ### Configuration
 
 A *configuration* is a map from [variable](variables.md) names to SassScript
-values. An *empty configuration* contains no entries.
+values. An *empty configuration* contains no entries. Each *configuration* has
+a uniquely-identifying opaque ID.
```
nex3: Nit: I'd include the ID in the first sentence, since it's (now) part of the structure of a configuration:

```suggestion
values and an opaque ID. An *empty configuration* contains no entries.
```

I'd omit "uniquely-identifying" because this is so far just describing the _structure_ of a configuration, not its _semantics_.

spec/modules.md
```diff
@@ -52,9 +52,13 @@ An *empty CSS tree* contains no statements.
...
 
-[source file]: syntax.md#source-file
+Each *configuration* has a uniquely-identifying opaque ID, sharing the same
+instance across all [source files][] in a [module](#module).
```
nex3: The first half of this is just reiterating the previous paragraph, and it's not clear exactly which new configurations the second part applies to. I'd say instead something like, "Each new configuration has a unique ID by default." Then in the text of the spec, you can say something like "Create a new configuration with the same ID as before" if you want them to share identities.

Goodwine: done, I think.. :)

spec/spec.md
```diff
@@ -47,6 +47,9 @@ innermost active invocation of [Executing a File](#executing-a-file).
 
 [configuration]: modules.md#configuration
 
+ > Each new *configuration* has a unique ID by default. Create a new
+ > configuration with the same ID to share identities.
```
nex3: Block quotes are used for non-normative information—things that help the reader understand _why_ the spec works the way it does, or clarifies an interaction that's implicit in the normative spec text. They shouldn't be used to to specify how Sass behaves.

I don't think this is the right place to specify when configurations do or don't share IDs, either. The "Current Configuration" definition explains which configuration object is current, but it's not related to the process of creating configurations. I think instead you want to say in the definition of "Configuration":

> A new configuration has a unique ID unless otherwise specified.

...and then in the semantics for the `@forward` rule:

> ...an empty configuration with the same ID as the current configuration.

You also need to specify what to _do_ with the IDs. Currently the "Loading a Module" procedure still says "If `config` is not empty, throw an error", but that's no longer what we want.

nex3 (changes requested):

nex3:
Another thing we should bear in mind here is that different `@forward` rules can actually _change_ configured values. For example, the following should still throw an error:

```hrx
<===> input.scss
@use "downstream" with ($var1: from input);

<===> _downstream.scss
@forward "upstream" with ($var2: from downstream 1);
@forward "upstream" with ($var2: from downstream 2);

<===> _upstream.scss
$var1: from upstream !default;
$var2: from upstream !default;
```

...so simply checking that two configurations come from the same original configuration isn't sufficient.

Goodwine:
It does fail as expected because while `@use .. with` seems to be OK with it (although I'm not 100% sure), the `@forward .. with` does not seems to be OK with it, guessing because they are not the same nodespan

https://github.com/sass/sass-spec/pull/1796/commits/908b6a0c8bfca6559577b8641ebcfdb49d2b733e

spec/spec.md
Goodwine: I think I know what you mean here and on the comment below.. it may be working that way on the implementation but it's definitely not clear in here, let me update this and double check the impl actually does this :)

Goodwine: I have updated the PR but I didn't use this behavior:
> an empty configuration with the same ID as the current configuration.

Instead I used:
> The `rule-config` opaque ID will be generated deterministically for the location it was declared at.

I guess the former behavior would decouple the config ID generation from the file location 🤔 and I just tested it and the example you shared below (but changing `$var2: from downstream 1 !default`) worked

I found the implementation to be slightly more complex (having to carry the opaqueId around) but I guess it works.. so let me re-update the PR

Goodwine: Done.

Goodwine:
PTAL

spec/modules.md
```diff
@@ -271,7 +273,11 @@ This algorithm takes a string `argument` and [configuration](#configuration)
   > This disallows circular `@use`s, which ensures that modules can't be used
   > until they're fully initialized.
 
-* Otherwise, return the result of [executing][] `file` with `config` and a new
+* Otherwise
+
+  * Keep track of which `config` was used for the `file` by its opaque ID.
```
nex3: You don't need to explicitly specify when and where to track stuff in the spec—you can just refer to values and, as long as it's unambiguously specified what you're referring to, it's up to the implementation to figure out how to actually ensure the information is available when necessary.

In other words, specs only need to say _what to do_, they don't have to be concerned with _how to do it_.

spec/modules.md
```diff
@@ -262,7 +264,7 @@ This algorithm takes a string `argument` and [configuration](#configuration)
 
   [executed]: spec.md#executing-a-file
 
-  * If `config` is not empty, throw an error.
+  * If `config` has a different ID than before and is not empty, throw an error.
```
nex3: Currently, nothing in this spec says to create a config with anything other than a unique ID, so this is really just saying "if `config` is a different object than the previous one". If that's all we need to specify, then we can do away with the "opaque ID" concept entirely and just say that directly. But if we want to consider different configuration objects to be "the same"—for example if they're created with `@forward as` or `@forward with`—we need to explicitly say "create a new configuration with the same ID as ..." when they're created.

Goodwine: I have added a note on `ForwardWithArgument` but it's a bit far so I added a note below, so info is not too far apart

spec/at-rules/forward.md
```diff
@@ -48,9 +48,10 @@ To execute a `@forward` rule `rule`:
 
 * If `rule` has an `AsClause` with identifier `prefix`:
 
-  * Let `rule-config` be an empty [configuration][].
+  * Let `rule-config` be an empty [configuration] inheriting the opaque ID
+    from [the current configuration].
```
nex3: Nit: I think "with the same opaque ID as" is maybe a little clearer, since "inheriting" is kind of an overloaded term.

spec/modules.md
```diff
@@ -262,7 +264,11 @@ This algorithm takes a string `argument` and [configuration](#configuration)
...
```
nex3: I think "than the configuration that was passed the first time `file` was executed" would be clearer.

Also, this is a pre-existing issue, but you might as well fix it while you're here: I think the previous bullet point should say "If `file` has already been executed **by the Loading a Module procedure**", since it's valid to load a file with `@import` and also with `@use`.

spec/modules.md
```diff
@@ -258,11 +260,17 @@ This algorithm takes a string `argument` and [configuration](#configuration)
...
 
-* If `file` has already been [executed][]:
+* If `file` has already been [executed] by the [Loading a Module] procedure:
 
   [executed]: spec.md#executing-a-file
+  [Loading a Module]: #loading-a-module
 
-  * If `config` is not empty, throw an error.
+  * If `config` is not empty and has a different ID than the configuration that
+    was passed the first time `file` was executed, throw an error.
```
nex3: "was executed by the Loading a Module procedure"

nex3 (approved):
One more small change, then looks good! Thanks for bearing with me through so many iterations.

Goodwine:
Thanks for the reviews!

Issue #1890, opened 2015-11-10
Units dropped from random() $limit


chriseppstein:
* [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/random-with-units.md)

Deprecation

* [x] [Tests](https://github.com/sass/sass-spec/issues/1810) https://github.com/sass/sass-spec/pull/1815
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1756) https://github.com/sass/dart-sass/pull/1779
* [x] [Docs](https://github.com/sass/sass-site/issues/659) https://github.com/sass/sass-site/pull/664
* [x] Update canonical spec https://github.com/sass/sass/pull/3377

Final

* [ ] [Tests](https://github.com/sass/sass-spec/issues/1810)
* [ ] [Dart Sass](https://github.com/sass/dart-sass/issues/1756)
* [ ] [Docs](https://github.com/sass/sass-site/issues/659)

---

_(Original:)_

The unit passed with the limit should be the unit on the returned value. Instead the unit is dropped and never considered nor is an error raised.

Step 1:
Deprecate passing units to `random()` as the `$limit`.

Step 2:
Make passing a unit to `random()` as the `$limit` an error.

Step 3:
Allow units to be passed in, returning them with the random result.

jcutrell:
Is the desire to have `random()` be redefined to use `random($limit, $unit)`, where `$unit` is optional?

nex3:
@jcutrell No, the desire is to (ultimately) have `math.random(10px)` return a number with the unit `px`.

Goodwine:
Blocking until feedback arrives from proposal

https://github.com/sass/sass/blob/main/proposal/random-with-units.md

Goodwine:
Updated main comment with:

[Proposal](https://github.com/sass/sass/blob/main/proposal/random-with-units.md)

Deprecation

* [Tests](https://github.com/sass/sass-spec/issues/1810)
* [Dart Sass](https://github.com/sass/dart-sass/issues/1756)
* [Docs](https://github.com/sass/sass-site/issues/659)
* Update canonical spec

Final

* [Tests](https://github.com/sass/sass-spec/issues/1810)
* [Dart Sass](https://github.com/sass/dart-sass/issues/1756)
* [Docs](https://github.com/sass/sass-site/issues/659)

nex3:
Now that the proposal has landed, we'll give this a month for public comment since it's a low-controversy breaking change.

[...]

Issue #533, opened 2012-09-30
Get unitless value


flops:
Is there any way to parseInt the values and get unitless value ?
All i`ve found that i only can get is it unitless value or not... 

For example:

``` scss
@function clear-units($value){
    @if type-of($value) == "number" {
        @if (unitless($value)) {
            @return $value;
        } @else if unit($value) == "em"{
            @return $value / 1em;
        } @else if unit($value) == "px" {
            @return $value / 1px;
        } @else if unit($value) == "pt" {
            @return $value / 1pt;
        }
    } @else {
        @warn "Not a number value: #{$value}";
        @return $value;
    }
}
```

made this function. But as you can see all units must be declared in if statement, and i don`t really like the realization.

pdaoust:
yes, it does seem a bit verbose. I wrote a function like that myself... It would be nice to see this rolled into SASS core, along with a cast() function that converts one unit into another, without doing arithmetic on it (with the exception of %, which should be divided by 100).

nex3:
There's a built-in [`unitless()` function](http://sass-lang.com/docs/yardoc/Sass/Script/Functions.html#unitless-instance_method) that does this.

flops:
Not even close.
unitless - gives you Bool.
Function in example returns value without units, to convert em to px for example.

nex3:
Sorry, you're correct. Reopening.

robwierzbowski:
In the meanwhile, if you don't want to cycle through all of those ifs:

``` scss
@function strip-units($number) {
  @return $number / ($number * 0 + 1);
 }
```

http://stackoverflow.com/questions/12328259/how-do-you-strip-the-unit-from-any-number-in-sass/12335841#12335841

pdaoust:
@robwierzbowski Thanks for the tip; I'll incorporate it into my own stuff and see if it works!

chriskempson:
Would love to see this in SASS core.

KittyGiraudel:
Function from @robwierzbowski works great. I'd love to see this native. It occurred to me it comes in handy sometimes.

chriseppstein:
I'd like to see use cases where this is a good idea. I have doubts.

chriseppstein:
I think too many people don't know how to work with units and will just use this to do bad things instead of proper math. :(

KittyGiraudel:
[...]

PR #3361, opened 2022-07-20
Proposal: Define random() behavior for numbers with units

Goodwine:
Proposal for:
- https://github.com/sass/sass/issues/1890

### DIFF
--- a/proposal/random-with-units.md
+++ b/proposal/random-with-units.md
@@ -1,0 +1,101 @@
+# Random: Draft 1
+
+*([Issue](https://github.com/sass/sass/issues/1890))*
+
+This proposal defines the behavior of the built-in [`random()`][random]
+function.
+
+[random]: ../spec/built-in-modules/math.md#random
+
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+  * [Design Decisions](#design-decisions)
+* [Semantics](#semantics)
+* [Deprecation Process](#deprecation-process)
+
+## Background
+
+> This section is non-normative.
+
+Sass contains a built-in `random()` function which is [documented on the public
+site][random public], but the behavior itself is **not** [specified][].
+
+Currently the function parameter `$limit` (defaults to `null`) changes the
+behavior from returning a decimal in range `[0, 1]` to returning an integer in
+range `[1, $limit]`. An error is thrown when passing a non-numeric value.
+
+It is undefined what happens when the `$limit` parameter contains a unit (e.g.
+`5px` or `8em`), and the current behavior is unexpected as `random()` will
+[drop the units and percent symbol][issue].
+For example: `random(42px) => 28` (there is no `px`).
+
+[random public]: https://sass-lang.com/documentation/modules/math#random
+[specified]: ../spec/built-in-modules/math.md#random
+[issue]: https://github.com/sass/sass/issues/1890
+
+## Summary
+
+> This section is non-normative.
+
+The built-in `random($limit: null)` function will keep its current behavior but
+it will now return units whenever a number with units or percentage is given as
+an argument.
+
+### Design Decisions
+
+This proposal keeps the existing syntax but changes the current semantics, and
+we therefore consider it a breaking change.
+
+An alternative that has also been suggested adding an optional second parameter
+to specify units, for example `random(42, 'px')` even though this can also have
+unexpected behavior such as dealing with `random(42em, 'px')` which under the
+current behavior it would strip the number from the units and replace it with
+different ones.
+
+We decided the alternative would be more confusing and instead will pursue with
+allowing numbers with units to be passed through and implement the expected
+behavior on the next major version. For now we'll include a deprecation warning
+whenever a number with units or percentages is passed as an argument.
+
+One potential source of friction is that Sass doesn't include a built-in utility
+to strip away units so users relying on the current behavior for [stripping
+units] would need to implement it themselves.
+
+[stripping units]: https://stackoverflow.com/a/12335841
+
+## Semantics
+
+The `random()` function can take an optional parameter `$limit` which defaults
+to `null`.
+
+* If `$limit` is `null` then return a floating-point number in range `[0, 1]`.
+
+  > Example: `random() => 0.1337001337`
+
+* If `$limit` is an **integer** [number][] greater than zero:
+
+  * Generate an integer number in range `[1, $limit]`
+
+  * If `$limit` [is unitless][], return the generated number without units.
+
+    > Example: `random(123) => 87`
+
+  * If `$limit` has [units][], return the generated number with the same units.
+
+    > Examples: `random(123px) => 97px` and `random(500%) => 238%`
+
+* Otherwise throw an error.
+
+[number]: https://sass-lang.com/documentation/values/numbers
+[is unitless]: ../spec/built-in-modules/math.md#is-unitless
+[units]: https://sass-lang.com/documentation/values/numbers#units
+
+## Deprecation Process
+
+Given some users may be relying on the existing Dart Sass implementation which
+strips off the units, this will be a breaking change for Dart Sass v1.
+
+We will emit deprecation warnings for any use of `random($limit)` where the
+`$limit` argument evaluates to a number with units.
--- a/proposal/random-with-units.md
+++ b/proposal/random-with-units.md
@@ -1,9 +1,10 @@
-# Random: Draft 1
+# Random With Units: Draft 1
 
 *([Issue](https://github.com/sass/sass/issues/1890))*
 
-This proposal defines the behavior of the built-in [`random()`][random]
-function.
+This proposal modifies the behavior of the built-in [`math.random()`][random]
+function to return a number with matching units to the numeric argument it
+received.
 
 [random]: ../spec/built-in-modules/math.md#random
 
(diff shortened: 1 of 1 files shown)


proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
+# Random: Draft 1
```
nex3: The proposal name should match the filename:

```suggestion
# Random With Units: Draft 1
```

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+
+*([Issue](https://github.com/sass/sass/issues/1890))*
+
+This proposal defines the behavior of the built-in [`random()`][random]
```
nex3: Nit: "Defines" suggests that this defines an entirely new function. It would be good to be a bit more specific about what behavior it's changing.

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+  * [Design Decisions](#design-decisions)
+* [Semantics](#semantics)
+* [Deprecation Process](#deprecation-process)
+
+## Background
+
+> This section is non-normative.
+
+Sass contains a built-in `random()` function which is [documented on the public
+site][random public], but the behavior itself is **not** [specified][].
```
nex3:
```suggestion
site][random public], but the behavior itself is **not** [specified].
```

Markdown doesn't require a trailing `[]` for links, and we generally omit it when it's not required to disambiguate. (There's some older documentation that still includes it, so I don't blame you for being confused :sweat_smile:)

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+range `[1, $limit]`. An error is thrown when passing a non-numeric value.
+
+It is undefined what happens when the `$limit` parameter contains a unit (e.g.
+`5px` or `8em`), and the current behavior is unexpected as `random()` will
+[drop the units and percent symbol][issue].
+For example: `random(42px) => 28` (there is no `px`).
+
+[random public]: https://sass-lang.com/documentation/modules/math#random
+[specified]: ../spec/built-in-modules/math.md#random
+[issue]: https://github.com/sass/sass/issues/1890
```
nex3: The background and summary should be written more for the audience of Sass authors than people who are working on the spec. For this audience, it doesn't really matter what's in the spec (and in fact, it can generally be assumed that most aspects of the language _aren't_ formally specified yet). Instead, try focusing your explanation on what the `random()` function actually does in practice, and why that's undesirable.

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Summary
+
+> This section is non-normative.
+
+The built-in `random($limit: null)` function will keep its current behavior but
+it will now return units whenever a number with units or percentage is given as
```
nex3: "return units" is a bit confusing. Maybe something like "When it's passed a number with units, the number it returns will have matching units"?

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+### Design Decisions
+
+This proposal keeps the existing syntax but changes the current semantics, and
+we therefore consider it a breaking change.
+
+An alternative that has also been suggested adding an optional second parameter
+to specify units, for example `random(42, 'px')` even though this can also have
+unexpected behavior such as dealing with `random(42em, 'px')` which under the
+current behavior it would strip the number from the units and replace it with
+different ones.
```
nex3: This is a very long sentence. Try splitting it up into two or three.

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+it will now return units whenever a number with units or percentage is given as
+an argument.
+
+### Design Decisions
+
+This proposal keeps the existing syntax but changes the current semantics, and
+we therefore consider it a breaking change.
+
+An alternative that has also been suggested adding an optional second parameter
+to specify units, for example `random(42, 'px')` even though this can also have
```
nex3: Nit:

```suggestion
to specify units, for example `random(42, 'px')`, even though this can also have
```

Asides like "for example" should have commas both before and after them. The commas function to indicate the scope of the aside, like parentheses but for an aside that's more closely related to the current topic.

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+unexpected behavior such as dealing with `random(42em, 'px')` which under the
+current behavior it would strip the number from the units and replace it with
```
nex3:
```suggestion
current behavior would strip the number from the units and replace it with
```

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+different ones.
+
+We decided the alternative would be more confusing and instead will pursue with
+allowing numbers with units to be passed through and implement the expected
```
nex3: "pursue with allowing" -> "allow"

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+behavior on the next major version. For now we'll include a deprecation warning
+whenever a number with units or percentages is passed as an argument.
+
+One potential source of friction is that Sass doesn't include a built-in utility
+to strip away units so users relying on the current behavior for [stripping
+units] would need to implement it themselves.
```
nex3: Since this is a separate design decision, it and the previous one should probably have their own subheadings.

It's also probably worth noting that we intentionally discourage users from stripping units from numbers, so this isn't considered a major concern.

proposal/random-with-units.md
nex3: Nowadays we try to refer to functions using their namespaces:

```suggestion
This proposal defines the behavior of the built-in [`math.random()`][random]
```

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Background
+
+> This section is non-normative.
+
+Sass contains a built-in `random()` function which is [documented on the public
+site][random public], but the behavior itself is **not** [specified][].
+
+Currently the function parameter `$limit` (defaults to `null`) changes the
+behavior from returning a decimal in range `[0, 1]` to returning an integer in
```
nex3: The range for the default `math.random()` is actually `[0, 1)` (that is, it can't return `1` itself).

proposal/random-with-units.md
```diff
@@ -0,0 +1,101 @@
...
+units] would need to implement it themselves.
+
+[stripping units]: https://stackoverflow.com/a/12335841
+
+## Semantics
+
+The `random()` function can take an optional parameter `$limit` which defaults
+to `null`.
+
+* If `$limit` is `null` then return a floating-point number in range `[0, 1]`.
```
nex3:
```suggestion
* If `$limit` is `null` then return a floating-point number in range `[0, 1)`.
```

nex3 (changes requested):

proposal/random-with-units.md
Goodwine: I was actually curious about it but wasn't sure so my intention of pointing it out as ranges instead of "between x and y" was to clarify this 😄

proposal/random-with-units.md
Goodwine: wording is hard lol

proposal/random-with-units.md
Goodwine: Right, so I had wording like that, but then realized that in a way it was never "defined" here and only in the Sass website, but I can update it to clarify it's modifying the current behavior

proposal/random-with-units.md
Goodwine: separated and clarified, I added a link to this comment not sure if there's a better option 🤔 :
- https://github.com/sass/sass/issues/533#issuecomment-52531596

proposal/random-with-units.md
```diff
@@ -0,0 +1,103 @@
...
+> This section is non-normative.
+
+Sass provides a built-in [`math.random()` function][random] which takes an
+optional numeric parameter `$limit` (defaults to `null`).
+
+When `null` is passed it returns a decimal in range `[0, 1)`. When an integer
+greater than zero is passed it returns a number in range `[1, $limit)`.
+Otherwise it throws an error.
+
+However a numeric integer can include units (e.g. `5px` or `8em`) and the
```
nex3: Nit:

```suggestion
However, a numeric integer can include units (e.g. `5px` or `8em`) and the
```

"However" should be followed by a comma if it's introducing a sentence that would be valid on its own.

proposal/random-with-units.md
```diff
@@ -0,0 +1,103 @@
...
+unit-based arithmetic.
+
+[stripping units an anti-pattern]: https://github.com/sass/sass/issues/533#issuecomment-52531596
+
+## Semantics
+
+The `math.random()` function can take an optional parameter `$limit` which
+defaults to `null`.
+
+* If `$limit` is `null` then return a floating-point number in range `[0, 1)`.
```
nex3: Probably good to specify that the number is random:

```suggestion
* If `$limit` is `null` then return a pseudo-random floating-point number in range `[0, 1)`.
```

Also below.

nex3: Nit: "in the range", here and below

proposal/random-with-units.md
```diff
@@ -0,0 +1,103 @@
...
+
+  * Generate an integer number in range `[1, $limit)`
+
+  * If `$limit` [is unitless], return the generated number without units.
+
+    > Example: `math.random(123) => 87`
+
+  * If `$limit` has [units], return the generated number with the same units.
+
+    > Examples: `math.random(123px) => 97px` and `math.random(500%) => 238%`
```
nex3: You can probably consolidate these into just "Return the generated number with the same units as `$limit`", since the same units as unitless is unitless

nex3 (approved):

proposal/random-with-units.md
Goodwine: Merged all three bullets

PR #3362, opened 2022-07-20
[Bogus Combinators] Update the specification

nex3:


spec/style-rules.md
```diff
@@ -37,6 +37,10 @@ To execute a style rule `rule`:
 
 * Execute each child `child` of `rule`.
 
+* If `css` contains any children and `selector` is [bogus], throw an error.
```
jathak, line 40: Should this be updated now even though the current state is just a deprecation warning?

nex3: Yeah—as far as the spec is concerned, deprecation warnings and intermediate states are essentially an implementation detail.

jathak (approved):

Issue #3366, opened 2022-07-26
Sass `compileString` unable to import stylesheets


jhildenbiddle:
Node version: 18.4.0
Sass version: 1.54.0
OS: macOS 12.4 (M1/ARM)

I am unable to use the [compileString()](https://sass-lang.com/documentation/js-api/modules#compileString) method to process CSS that imports other stylesheets via `@use` or `@forward`. I have tried absolute paths and numerous relative path variations. No luck. FWIW, the same text contained in a file can be processed using the [compile()](https://sass-lang.com/documentation/js-api/modules#compile) method and the Sass CLI without issue.

**Steps to reproduce**:

Create the following three files:

```scss
// File: test.scss

@use "test-import";

.my-class {
    color: red;
}
```

```scss
// File: test-import.scss

.my-import {
    color: blue;
}
```

```js
// File: test.js

const fs = require('fs');
const path = require('path');
const sass = require('sass');

const srcPath = path.resolve(__dirname, 'test.scss');
const srcText = fs.readFileSync(srcPath, 'utf8');

try {
    const result = sass.compile(srcPath);
    console.log('sass.compile() success');
    console.log(result);
}
catch(err) {
    console.error('sass.compile() fail');
    throw(err);
}

try {
    const test2 = sass.compileString(srcText);
    console.log('sass.compileString() success');
    console.log(result);
}
catch(err) {
    console.error('sass.compileString() fail');
    throw(err);
}
```

Run the `test.js` file:

```bash
node ./test.js
```

The `sass.compile()` test will complete successfully and display the results object to the console. The CSS generated is as expected:

```css
.my-import {
  color: blue;
}

.my-class {
  color: red;
}
```

The `sass.compileString()` test will fail and display the following in the console:

```bash
sass.compileString() fail
/Users/john/Dev/project-name/test.js:27
    throw(err);
    ^

sass.Exception [Error]: Can't find stylesheet to import.
  ╷
1 │ @use "test-import";
  │ ^^^^^^^^^^^^^^^^^^
  ╵
  - 1:1  root stylesheet
    at Object.wrapException (/Users/john/Dev/project-name/node_modules/sass/sass.dart.js:1246:17)
    at _EvaluateVisitor1._evaluate0$_loadStylesheet$4$baseUrl$forImport (/Users/john/Dev/project-name/node_modules/sass/sass.dart.js:80868:19)
[...]

PR #3373, opened 2022-08-08
Clarify that compile and compileString accept optional options

Goodwine:
Related to:
- #3366


nex3 (approved):
Nice! I particularly like the link to the specific options type each function takes, since that's been confusing to users in the past.

PR #3377, opened 2022-08-16
Mark accepted: Random with Units Proposal

Goodwine:
See - https://github.com/sass/sass/issues/1890

### DIFF
--- a/accepted/random-with-units.md
+++ b/accepted/random-with-units.md
@@ -1,4 +1,4 @@
-# Random With Units: Draft 1
+# Random With Units
 
 *([Issue](https://github.com/sass/sass/issues/1890))*
 
--- a/accepted/random-with-units.md
+++ b/accepted/random-with-units.md
@@ -1,4 +1,4 @@
-# Random With Units
+# Random With Units: Draft 1
 
 *([Issue](https://github.com/sass/sass/issues/1890))*
 
--- a/accepted/random-with-units.md
+++ b/accepted/random-with-units.md
@@ -1,6 +1,6 @@
-# Random With Units: Draft 1
+# Random With Units: Draft 1.1
 
-*([Issue](https://github.com/sass/sass/issues/1890))*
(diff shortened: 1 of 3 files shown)


accepted/random-with-units.md
```diff
@@ -1,4 +1,4 @@
-# Random With Units: Draft 1
```
nex3, line 1: We typically keep the draft indicator even after the specification has been marked accepted, since it's useful to track the version number.

nex3 (changes requested):

spec/built-in-modules/math.md
```diff
@@ -486,4 +486,25 @@ This function is also available as a global function named `percentage()`.
 random($limit: null)
 ```
 
-This function is also available as a global function named `random()`.
+The `random()` function can take an optional parameter `$limit` which defaults
```
nex3: You don't need to write this out explicitly; the signature above already specifies it.

spec/built-in-modules/math.md
```diff
@@ -486,4 +486,25 @@ This function is also available as a global function named `percentage()`.
...
+to `null`.
+
+* If `$limit` is `null` then return a pseudo-random floating-point number in the
+  range `[0, 1)`.
```
nex3: A floating point number isn't a Sass number—this should be "a unitless number whose value is ..."

spec/built-in-modules/math.md
```diff
@@ -486,4 +486,25 @@ This function is also available as a global function named `percentage()`.
...
+
+  > Example: `math.random() => 0.1337001337`
+
+* If `$limit` is an **integer** [number] greater than zero:
+
+  * Return a pseudo-random integer in the range `[1, $limit)` with the same
```
nex3: I believe this is actually `[1, $limit]`.

nex3 (changes requested):

nex3 (approved):

PR #3408, opened 2022-10-03
[Floating Point] Integrate the proposal into the spec

### LINKED ISSUE #2892: Use floating-point numbers universally

nex3:
Closes https://github.com/sass/sass/issues/2892

### DIFF
--- a/spec/built-in-modules/math.md
+++ b/spec/built-in-modules/math.md
@@ -14,6 +14,11 @@ though some have special handling of units.
 * [Variables](#variables)
   * [`$e`](#e)
   * [`$pi`](#pi)
+  * [`$epsilon`](#epsilon)
+  * [`$max-safe-integer`](#max-safe-integer)
+  * [`$min-safe-integer`](#min-safe-integer)
+  * [`$max-number`](#max-number)
+  * [`$min-number`](#min-number)
 * [Functions](#functions)
   * [Bounding Functions](#bounding-functions)
     * [`ceil()`](#ceil)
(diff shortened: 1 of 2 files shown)


spec/built-in-modules/math.md
```diff
@@ -71,6 +119,10 @@ ceil($number)
 
 This function is also available as a global function named `ceil()`.
 
+* Return a number whose value is the result of
+  `convertToIntegerTowardPositive($number.value)` as defined by [IEEE 754 2019],
```
jathak, line 123: The link for `[IEEE 754 2019]` isn't defined anywhere in this file.

jathak (approved):

spec/built-in-modules/math.md
nex3: Oops!

Issue #3424, opened 2022-10-12
Color Spaces: color.adjust() should handle out-of-range values for HSL/HWB channels


nex3:
The color spaces proposal (https://github.com/sass/sass/issues/2831) currently doesn't check the range of any arguments for the `color.adjust()` function. Normally, this is fine—most color spaces fully preserve even out-of-gamut values. However, HSL and HWB _don't_ do so, so we should handle that somehow.

Currently, `color.adjust()` forbids arguments outside of `[-100%, 100%]` for these arguments, but if the resulting channel value would be out-of-range it's simply clamped to `100%` or `0%`. We could preserve this behavior, or we could match other color functions in allowing _any_ argument while using the clamping behavior to preserve the guarantee that HSL and HWB never represent out-of-gamut colors.

nex3:
Just to make sure I'm understanding correctly, this is now handled in two ways, right?

1. Normalizing Color Channels now explicitly _clamps_ channels for HSL and HWB. This means that `hsl()`, `hwb()`, `color.change()`, and `color.adjust()` clamp out-of-bounds channels.

2. Converting colors into `hsl()` and `hwb()` now automatically gamut-maps the RGB equivalent before doing the final conversion.

mirisuzanne:
That's right, yes.

mirisuzanne:
(I can make that more clear in the changelog)

mirisuzanne:
Resolved in draft 1.2

Issue #3418, opened 2022-10-08
Color Spaces: Should a and b be scalable?


nex3:
The color spaces proposal (https://github.com/sass/sass/issues/2831) currently explicitly marks the a and b channels of the lab and oklab color spaces as not scalable, since their minimum values are negative mirrors of their positive values instead of being zero. However, we _could_ meaningfully define scaling here as a linear scale between min and max instead of between zero and max.

In mathematical terms, we'd say `scale(original, factor) = original + (max - original) * factor` if `factor > 0` and `scale(original, factor) = original + (original - min) * factor` if `factor < 0`. This preserves the behavior that scaling a number by positive X% brings it X% closer to the maximum and scaling it by negative X% brings it X% closer to the minimum.

mirisuzanne:
What I found strange with these is that they map `%` from `-100%` to `100%` - so it's not clear to me if an adjustment of `100%` should move it the entire distance, or half the distance between the min and max. Is it confusing if a change of `100%` moves the value `200%`?

nex3:
The `%` in `color.scale()` refers to a different thing than the `%` in a color's channel. The former refers to the distance between the current value and the minimum/maximum, while the latter refers to the distance between 0 and the minimum/maximum. So I think it's fine to say that 100% (scale factor) can increase the channel by 200% (channel value).

mirisuzanne:
Resolved in draft 1.2

Issue #3417, opened 2022-10-08
Color Spaces: Support legacy behavior for color.change/adjust/scale


nex3:
Like all color functions, `color.change()`, `.adjust()`, and `.scale()` historically treated RGB, HSL, and HWB colors as interchangeable. As currently specified in the color spaces proposal (https://github.com/sass/sass/issues/2831), though, any color channels that aren't in a color's space throw errors. To preserve backwards-compatibility, this should allow the any RGB, HSL, and HWB channels to be modified for any legacy colors (at least when `$space` isn't set).

mirisuzanne:
Resolved in draft 1.2

Issue #3415, opened 2022-10-05
Color Spaces: color.invert() handles $weight in reverse


nex3:
The color spaces proposal (https://github.com/sass/sass/issues/2831) defines [`color.invert()`](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md#colorinvert) with a definition that ends with a call to `color.mix(invert, color, $weight, mix-space)`. But since `$weight` defaults to `100%`, this means that by default it will return _the original color_ rather than the inverted color. I think it should instead say `color.mix(invert, color, 100% - $weight, mix-space)`.

Also worth noting that the signature doesn't include `$weight`.

nex3:
Another issue with the same function: currently, it's invalid to run `color.invert($color, $space: display-p3)` even though this has a well-defined output, because `display-p3` isn't a supported interpolation space. I think the definition should have explicit early-exit clauses if `$weight == 0%` or `$weight == 100%` so that you can pass in those color spaces when mixing isn't necessary.

mirisuzanne:
According to [the current documentation for `color.mix`](https://sass-lang.com/documentation/modules/color#mix)

> A larger weight indicates that more of `$color1` should be used, and a smaller weight indicates that more of `$color2` should be used.

Since `invert` is `$color1` in this case, a default weight of `100%` returns the inverted color. 

Fixing the other issues here, but I don't think the weight is actually reversed.

mirisuzanne:
(I do think our handling of weight is confusing, and not what I expect)

nex3:
> Fixing the other issues here, but I don't think the weight is actually reversed.

Ah, you're right, my mistake.

mirisuzanne:
> because `display-p3` isn't a supported interpolation space.

There are a few ways we could address this:

- Make the limitation explicit (probably overkill)
- Use either `srgb` or `linear-srgb` as the 'closest' option (still rgb)
- Use `oklab`, which is defined by CSS as the 'best default' option when none is specified
- Skip the call to `mix`, and just do a weighted averaging of input/output channels?

nex3:
Another option would be to simply expand on the CSS spec and support `mix()` with all known color spaces, but I'm reticent to do that when I don't have a clear understanding of why CSS only supports a subset of them in the first place.

I think for now, my preference is to say that if you pass an unmixable color space like `display-p3`, `$weight` must be 100% or 0% and so we never call `mix()`.

mirisuzanne:
[...]

PR #3426, opened 2022-10-13
CSS Color Level 4, New Color Spaces: Draft 1.2

mirisuzanne:
Address issues from the first draft of the color proposal:

- [x] #3400 
- [x] #3401 
- [x] #3404 
- [x] #3405 
- [x] #3409 
- [x] #3410 
- [x] #3411 
- [x] #3413 (not sure this needs any changes)
- [x] #3414
- [x] #3415 (needs discussion for handling eg `display-p3`)
- [x] #3417 
- [x] #3418 (needs discussion)
- [x] #3424 
- [x] #3443
- [x] #3444 
- [x] #3454
- [x] #3455


proposal/color-4-new-spaces.changes.md
```diff
@@ -1,3 +1,47 @@
+## Draft 1.2
+
+* Clamp `hsl` saturation & lightness when generating `hsl` colors, and as part
+  of the color conversion process
```
nex3: Color conversion gamut-maps rather than clamping, right?

proposal/color-4-new-spaces.md
```diff
@@ -548,7 +549,7 @@ The known color spaces and their channels are:
       > Percentages `[0%,100%]` map to the `[0,100]` range.
 
   * `a`, `b`:
-    * gamut: un-bounded
+    * gamut: un-bounded, scalable
```
nex3: Is it even wroth explicitly listing channels as "scalable", given that they're now scalable if and only if they're not polar-angle?

proposal/color-4-new-spaces.md
```diff
@@ -770,17 +769,37 @@ To serialize a non-legacy color `color`:
...
 ## Procedures
 
+### Normalizing a Color Space Name
+
+This procedure accepts a color space `name`, throws an error if the name is
+invalid, and returns a normalized version of the name when possible.
+
+* If `name` is not an unquoted string, throw an error.
+
+* If `name` is a [dashed ident], return `name`.
```
nex3: We should omit this for now, right?

proposal/color-4-new-spaces.md
```diff
@@ -770,17 +769,37 @@ To serialize a non-legacy color `color`:
...
 
-* If `name` is not an unquoted string, throw an error.
+* Set `name` to the result of [normalizing a color space name] named `name`.
 
-* If `name` is the name of a [known color space], ignoring case, return the
-  matching [known color space].
+* If `name` is the name of a [known color space], return the matching [known
+  color space].
 
 * Otherwise, return `null`.
```
nex3: Do we still need this procedure if we're _only_ dealing with known color spaces?

proposal/color-4-new-spaces.md
```diff
@@ -1161,7 +1213,13 @@ input colors.
 * If `weight` is null, set `weight` to `0.5`.
 
 * Otherwise, set `weight` to the result of [percent-converting] `weight` with a
-  max of 1, and then clamping the value between 0 and 1 (inclusive).
+  max of 1.
+
+* If `weight` is greater than 1 or less than 0, throw an error.
```
nex3: Probably a good idea to explicitly write `weight > 1` and `weight < 0`, so that it's clear that this uses Sass's fuzzy comparison logic.

proposal/color-4-new-spaces.md
```diff
@@ -1332,19 +1392,26 @@ to-space($color, $space)
...
-* If `origin-space == $space`, return `$color`.
+  > The `color.space()` function always returns a normalized color space name.
 
-  > This allows unknown spaces, as long as they match the origin space.
+* If `origin-space == space`, return `$color`.
 
-* Let `space` be the result of [looking up a known color space] named `$space`.
+  > This allows unknown spaces, as long as they match the origin space. While
+  > the comparison is case-sensitive, all non-custom spaces have been
+  > normalized to lower-case at this point.
```
nex3: This is no longer relevant, right?

proposal/color-4-new-spaces.md
```diff
@@ -1635,25 +1710,25 @@ This function is also available as a global function named `adjust-color()`.
...
+    * Let `space` be the result of [normalizing a color space name] `$space`.
 
-    * Let `color` be the result of [converting] `$color` to `space`.
+    * If `space != origin-space`, set `color` to the result of calling
+      `color.to-space(color, space)`.
 
-* Otherwise:
+      > The `color.to-space()` function will error if conversion isn't possible.
+      > If conversion isn't required, we can perform changes on colors in an
+      > unknown color space.
```
nex3: Again, I think this isn't relevant anymore.

proposal/color-4-new-spaces.md
```diff
@@ -1683,14 +1758,27 @@ This function is also available as a global function named `adjust-color()`.
...
+
+      * Let `legacy-color` be the result of [converting] `color` to `hwb`.
+
+    * Otherwise, throw an error.
 
-    > This ensures that `adjust` is percent-converted when necessary, and
-    > throws errors when the adjustment is not a valid value for the space.
+    * Set `channels` to be a list of `legacy-color`'s channels.
 
   * Let `channel` be the channel in `channels` with name or index of `key`.
```
nex3: I think this should be "the value of the channel", so distinguish it from the channel's metadata.

proposal/color-4-new-spaces.md
```diff
@@ -1702,12 +1790,17 @@ This function is also available as a global function named `adjust-color()`.
 
   * Set `channel` to `channel + adjust`.
 
+    > This will throw an error if `adjust` and `channel` are not compatible.
```
nex3: If `channel` is just the numeric value of the channel, then I don't think this addition will work automatically if `adjust` is a percentage, unless it's being converted to a plain value somewhere else I haven't noticed.

proposal/color-4-new-spaces.md
```diff
@@ -1810,29 +1925,31 @@ This function is also available as a global function named `complement()`.
...
-invert($color, $space: null)
+invert($color,
+  $weight: 100%,
+  $space: null)
 ```
 
 This function is also available as a global function named `invert()`.
 
-* If `$color` is not a color, throw an error.
+* If `$color` is not a color in a [known color space], throw an error.
```
nex3: It's no longer possible to be a color that's _not_ in a known color space

nex3 (changes requested):

proposal/color-4-new-spaces.md
mirisuzanne: I think so. It now handles several error cases, including the error if no known color space is found. We could split that out into each place it's used, but that seems like unnecessary duplication?

nex3 (changes requested):
Empty review so Miriam can re-mark this as "ready to review" at her leisure

nex3:
Merging this now, we can update the deprecated functions in a later PR

PR #3428, opened 2022-10-20
Check for github action updates weekly

jlosito:
This will use dependabot to check updates for any github action which this project uses.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


jathak:
Thanks!

PR #3446, opened 2022-11-15
Proposal: first-class support for CSS Functions

jerivas:
This proposal changes all CSS Functions from being parsed as unquoted strings to being parsed in-depth and producing a new data type known as a "css-function". This data type holds the name and arguments of the function.

### DIFF
--- a/proposal/first-class-css-functions.md
+++ b/proposal/first-class-css-functions.md
@@ -1,0 +1,142 @@
+# First-Class CSS Functions
+
+*([Issue](https://github.com/sass/sass/issues/3176))*
+
+## Table of Contents
+
+- [First-Class CSS Functions](#first-class-css-functions)
+  - [Table of Contents](#table-of-contents)
+  - [Background](#background)
+  - [Summary](#summary)
+  - [Syntax](#syntax)
+  - [Types](#types)
+    - [Serialization](#serialization)
+      - [CssFunction](#cssfunction)
+  - [Semantics](#semantics)
+  - [Functions](#functions)
+    - [`meta.type-of()`](#metatype-of)
+    - [`meta.fn-name()`](#metafn-name)
+    - [`meta.fn-args()`](#metafn-args)
+
+## Background
+
+> This section is non-normative.
+
+Sass currently treats non-[calculation] [CSS Functions] as fully opaque,
+allowing almost any sequence of tokens within the parentheses and evaluating it
+to an unquoted string. Once a function is used it can't be inspected or
+manipulated in any way.
+
+Notice that interpolation in CSS Functions is already supported and therefore
+not included in this proposal:
+
+```scss
+$foo: 1px;
+width: var(--foo, $foo); // Output: width: var(--foo, 1px);
+```
+
+[calculation]: https://sass-lang.com/documentation/values/calculations
+[CSS Functions]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions
+
+## Summary
+
+> This section is non-normative.
+
+This proposal changes all CSS Functions from being parsed as unquoted strings to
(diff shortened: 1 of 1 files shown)


proposal/first-class-css-functions.md
```diff
@@ -0,0 +1,142 @@
...
+- `meta.fn-args($foo)` will return `unquote("--foo")`
+- `meta.fn-args($bar)` will return `unquote("--bar"), 8px`
+- `meta.fn-args($baz)` will return `unquote("--my-pkg-baz"), 8px`
+- `meta.fn-name($rotation)` will return `"rotate"`
+- `meta.fn-args($rotation)` will return `0.5turn`
+
+Note that if an argument is a number or a nested calculation, it’s returned as
+that type. Otherwise, it’s returned as an unquoted string.
+
+## Syntax
```
jerivas, line 71: Looking for guidance on the Syntax section

proposal/first-class-css-functions.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+### Serialization
+
+#### CssFunction
+
+To serialize a CSS Function, emit its name followed by "(", then each of its
+arguments separated by ",", then ")".
+
+## Semantics
```
jerivas, line 97: Looking for guidance on the Semantics section

proposal/first-class-css-functions.md
```diff
@@ -0,0 +1,142 @@
...
+
+* If `$fn` is not a CSS Function, throw an error.
+
+* Let `args` be an empty list.
+
+* For each argument `arg` in `$fn`'s arguments:
+
+  * If `arg` is a number or a calculation, add it to `args`.
+
+  * Otherwise, [serialize](#serialization) `arg` and add the result to `args` as
```
asyncliz, line 139: How would this handle the newer space and slash-separated `rgb($red $green $blue / $alpha)` syntax?

proposal/first-class-css-functions.md
nex3: Since this proposal doesn't actually change the grammar of CSS functions, you don't need a syntax section. You'll just refer to the syntax [already defined for functions](https://github.com/sass/sass/blob/main/spec/functions.md#syntax).

proposal/first-class-css-functions.md
```diff
@@ -0,0 +1,142 @@
...
+```ts
+interface CssFunction {
+  name: string;
+  arguments: CssFunctionArgument[];
+}
+
+type CssFunctionArgument =
+  | Number
+  | UnquotedString
+  | Calculation;
```
nex3, line 87: Why these three specific values? Shouldn't a plain CSS function allow any value?

proposal/first-class-css-functions.md
nex3: This is where you describe how to create a CSS function value given the syntax. You can think of this as describing the difference between [the current semantics](https://github.com/sass/sass/blob/main/spec/functions.md#functioncall) and the way you want it to work.

proposal/first-class-css-functions.md
nex3: Those would just be considered single-argument functions whose first argument is a list (slash-separated in this case once we finish deprecating `/`-as-division).

nex3 (changes requested):
One important thing to note here: for backwards-compatibility, I think it's very important that we continue to have CSS functions _act like_ strings, since the only way existing users have to extract information about them is through string functions. Specifically, this means that all the string functions should have a clause added to them that says "if the argument is a CSS function, serialize it and then do the string function on the result".

nex3:
Ideally, I'd also like to have a JS API type added for this as well. We've been falling behind on ensuring the JS API supports all Sass value types, so I'd like to avoid furthering that problem here.

Issue #3465, opened 2022-12-20
Color Spaces: Should we deprecate alpha?


nex3:
The [Color Spaces proposal](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md) currently deprecates all channel accesor functions _except_ `alpha()`/`color.alpha()`. Was this an oversight or did we have some justification for it?

mirisuzanne:
Good question. I don't remember if we discussed that. It does seem like there's a valid distinction we could make there - since every color is guaranteed to have an alpha value, no matter what space it is in. Getting direct access to alpha could be a useful shortcut for authors – but that's all it would be. And the more useful shortcut would allow adjusting/setting alpha. So I don't think I'd make a strong case for keeping it around.

nex3:
I think we should just deprecate it. No sense leaving it around and not the rest of them.

Issue #3464, opened 2022-12-20
Color Spaces: Unquoted channel names play poorly with RGB


nex3:
The [Color Spaces proposal](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md) currently requires channel names to be unquoted strings for functions like `color.channel()` and `color.is-powerless()`. While this matches the requirement that color spaces be unquoted, it raises a serious problem: the channel names `red`, `green`, and `blue` for the various predefined color spaces will be parsed by Sass as color values rather than unquoted strings, making `color.channel($color, red)` throw an error. While this could be worked around with `color.channel($color, unquote("red"))`, that's obviously infeasible in practice.

I see three basic approaches we could take here:

1. Require quoted strings instead. This avoids any ambiguity issues, but it's inconsistent with color space names which require unquoted strings and _aren't_ ambiguous with predefined colors. We could also take this a step further and change color spaces to require quoted strings as well for consistency, but then that's inconsistent with how they're written in plain CSS (particularly the `color()` function and the `<color-interpolation-method>` production, the latter of which we want to accept as-is for the `mix` function).

2. Switch to CSS's  channel terminology. Colors 5 defines [relative color syntax](https://www.w3.org/TR/css-color-5/#relative-colors) which makes explicit reference to channel names, but (other than `alpha`) they're different than the names we use. In particular, they're all single letters: `r`, `g`, `b`; `l`, `c`, `h`; and so on. While this is less obvious to a reader, there's real value in being consistent with CSS in addition to avoiding ambiguity with color names. If we did go this route, though, we'd need to consider whether to deprecate the existing long-form argument names for `color.adjust()`, `.scale()`, and `.change()`.

3. Hack around the issue entirely by defining `#f00`, `#0f0`, and `#00f` as value aliases of the channel names `red`, `green`, and `blue`. This is extremely messy so I'm inclined to dislike it, but it may be the path of least resistance.

mirisuzanne:
I would probably lean towards options 2 here, tho 3 is tempting. 

2. I like matching CSS - and while it's less clear about what each channels represents, it's more directly related to the color name. `rgb()` has `r`, `g`, and `b` channels. I don't love either option when it comes to deprecation: breaking changes suck, and maintaining legacy syntax is a pain. But it would be a change that the migrator can automate pretty well. 

3. I agree it doesn't feel great as an implementer - but kinda clever in hiding the implementation complexity from authors. Dunno if it's worth it, tho.

[...]

Issue #3455, opened 2022-12-03
[Color Spaces] Explicitly specify deprecated color functions


nex3:
The [current color spaces proposal PR](https://github.com/sass/sass/pull/3426) doesn't explicitly specify how to handle most of the deprecated color functions. It's reasonable to assume they work more or less the same way they always did, but as I get into the weeds of the implementation I'm realizing that because the internal representation of colors has changed so much these functions should probably also get explicitly defined.

Some particular questions I've run into:

* Should functions like `adjust-hue()` return a color in the original space or in HSL? (Probably original to match our general design principle that only `color.to-space()` changes the color space.)

* Should `hue()` return the hue `% 360deg`? We currently _store_ the hue as written, but taking the mod here would preserve some edge-case backwards compatibility.

* Should channel-access functions like `red()`, `hue()`, and so on even allow legacy colors? My current implementation forbids them.

mirisuzanne:
> Should channel-access functions like red(), hue(), and so on even allow legacy colors? My current implementation forbids them.

Do you mean the opposite of this? I would expect that they _only_ allow legacy colors?

mirisuzanne:
I take it you're looking for full definitions of all these functions? Or at least updates to their current definition? That might not be a small undertaking, but I can start on it.
  
> * Should functions like `adjust-hue()` return a color in the original space or in HSL? (Probably original to match our general design principle that only `color.to-space()` changes the color space.)

I don't know - for deprecated legacy functions, wouldn't it be more important to maintain legacy behavior?

> * Should `hue()` return the hue `% 360deg`? We currently _store_ the hue as written, but taking the mod here would preserve some edge-case backwards compatibility.

This makes sense because it does maintain legacy behavior, rather than following the new approach. Seems strange that we would update one to the new behavior, but keep the other one legacy?

> * Should channel-access functions like `red()`, `hue()`, and so on even allow legacy colors? My current implementation forbids them.

I guess I'm confused about how the deprecation process works. Why would these now error on legacy colors (and non-legacy colors, I assume?) while the other functions mentioned here continue to support their legacy behavior, or get updated to match the new behavior? I'm not seeing the different logic behind handing these all in different ways.

mirisuzanne:
- define in terms of aliases to new functions
- ensure that rgb values are rounded

PR #3477, opened 2023-01-12
Colors Proposal, Draft 1.3

dvdherron:
- [x] https://github.com/sass/sass/issues/3464
- [x] https://github.com/sass/sass/issues/3465
- [x] https://github.com/sass/sass/issues/3455
  - [x] color.red(), red()
  - [x] color.green(), green()
  - [x] color.blue(), blue()
  - [x] color.hue(), hue()
  - [x] color.saturation(), saturation()
  - [x] color.lightness(), lightness()
  - [x] color.whiteness()
  - [x] color.blackness()
  - [x] color.alpha()
  - [x] adjust-hue()
  - [x] saturate()
  - [x] desaturate()
  - [x] transparentize()
  - [x] opacify(), fade-in()
  - [x] lighten()
  - [x] darken()


proposal/color-4-new-spaces.md
```diff
@@ -2327,31 +2345,229 @@ plain CSS function named `"hsl"` that function is named `"hsla"` instead.
...
-[legacy color]: #legacy-color
+  * Return the result of calling `color.channel($color, 'red')`.
+
+* ```
+  red($color)
+  ```
+
+  * If `$color` is not a [legacy color], throw an error.
+
+  * Return the result of calling `color.channel($color, 'red')`.
```
mirisuzanne: I think we might need to pass in a space for all these `color.channel()` aliases. That way we don't end up with an error asking for the 'red' channel of an 'hsl' color.

```suggestion
  * Return the result of calling `color.channel($color, 'red', 'rgb')`.
```

(I don't think we can add any internal automation to `color.channel()` associating channel names to default spaces. The only channel names that are specific to a single space are `whiteness` and `blackness`. But in these alias functions we do know that we want the legacy space associated with the channel name.)

proposal/color-4-new-spaces.md
```diff
@@ -2327,31 +2345,229 @@ plain CSS function named `"hsl"` that function is named `"hsla"` instead.
...
+
+### `color.hue()`, `hue()`
+
+* ```
+  color.hue($color)
+  ```
+
+  * If `$color` is not a [legacy color], throw an error.
+
+  * Return the result of calling `color.channel($color, 'hue')`.
```
mirisuzanne: For these, I don't think it technically matters if we choose `hsl` or `hwb` as the space - since they are both clamped to srgb, and share the same hue wheel. Probably `hsl` is the more obvious choice.

proposal/color-4-new-spaces.md
```diff
@@ -2327,31 +2345,229 @@ plain CSS function named `"hsl"` that function is named `"hsla"` instead.
...
+
+### `color.alpha()`
+
+* ```
+  color.alpha($color)
+  ```
+
+  * If `$color` is not a [legacy color], throw an error.
+
+  * Return the result of calling `color.channel($color, 'alpha')`.
```
mirisuzanne, line 2452: Space doesn't affect this, so not needed here.

proposal/color-4-new-spaces.md
```diff
@@ -2327,31 +2345,229 @@ plain CSS function named `"hsl"` that function is named `"hsla"` instead.
...
+
+### `adjust-hue()`
+
+* ```
+  adjust-hue($color, $amount)
+  ```
+
+  * If `$color` is not a [legacy color], throw an error.
+
+  * Return the result of calling `color.adjust($color, $hue: $amount)`.
```
mirisuzanne: I suppose these also likely need spaces provided. Everything that isn't just dealing with alpha.

proposal/color-4-new-spaces.md
```diff
@@ -2327,31 +2345,229 @@ plain CSS function named `"hsl"` that function is named `"hsla"` instead.
...
-* `color.saturation()`/`saturation()`
-* `color.lightness()`/`lightness()`
-* `color.whiteness()`
-* `color.blackness()`
+deprecated in favor of the new `color.channel()` function. Legacy global color
+functions are also deprecated. These functions always throw errors. During
+the deprecation process, they act as alias functions described below.
 
-Legacy global color functions are also deprecated:
+### `color.red()`, `red()`
```
nex3, line 2352: You don't really need to explicitly define this for both the module and non-module versions; you can just steal the terminology [from the existing spec](https://github.com/sass/sass/blob/main/spec/built-in-modules/color.md#complement): "This function is also available as a global function named `...`."

mirisuzanne:
@nex3 @dvdherron - I went ahead and addressed the review comments here. I think this is ready to merge, unless there are additional concerns?

nex3 (approved):

nex3:
@mirisuzanne Can you update the table of contents as well?

mirisuzanne:
@nex3 As far as I can tell, the TOC is up to date. But I know different tools generate different in-document links, and I'm not sure how to match the expectations of this linter. Even the output in the CI error doesn't seem right. Can you document what tool is being used for the failing CI, and how to test against it (or even better - generate a TOC from it) locally?

nex3:
Sorry, you just need to run `npx tool/update-toc.ts`. We should probably make the check print out instructions for that!

mirisuzanne:
@nex3 When I run that command, I don't see any changes. It doesn't look like it requires an argument to be passed in. Is there something else I might be missing?

nex3:
Sorry, I got the command wrong :weary:. It's `npx ts-node tool/update-toc.ts`.

mirisuzanne:
Pushed the change, but I think there's an issue with the TOC it generates. There are now two distinct sections with the `#color` ID.

PR #3488, opened 2023-02-06
Improve the output of the ToC-check test.

nex3:
* Include the command for fixing the ToC
* Print a patch rather than the entire ToC.
* Always enable colors on CI.


nex3:
@mirisuzanne This should mitigate future pain :sweat_smile:

Goodwine (approved):

PR #3497, opened 2023-02-20
Color Proposal, Draft 1.4

### LINKED ISSUE #3474: Color Spaces: Track CSS decision on interpolation spaces
We need to make sure this CSS issue is resolved before we ship support in Sass: https://github.com/w3c/csswg-drafts/issues/7907

### LINKED ISSUE #3489: Match CSS behavior for missing channels in conversions
The CSS Spec has been updated to handle missing channels in the relative-color syntax, using the same logic as interpolation: https://github.com/w3c/csswg-drafts/issues/7771

In the current proposal, we define that behavior for interpolations - but not for conversion, and therefor also not for channel adjustments. Since CSS doesn't have 'conversion' as a user-facing feature outside interpolation/relative-colors, I think it's now safe to consider this part of the conversion algorithm, and implement it anywhere color spaces are converted.

(As part of this, we should also make sure to track edge cases like conversion within a space. Since the conversion logic explicitly marks powerless channels as 'missing' - a conversion of `color.to-space(hsl(0 0% 0%))` should return `hsl(none 0% 0%)`)

mirisuzanne:
- [x] https://github.com/w3c/csswg-drafts/issues/7750
- [x] fixes #3474
- [x] fixes #3489
- [x] https://github.com/w3c/csswg-drafts/issues/7924
- [x] https://github.com/w3c/csswg-drafts/issues/7900 / https://github.com/w3c/csswg-drafts/issues/8231

### DIFF
--- a/proposal/color-4-new-spaces.changes.md
+++ b/proposal/color-4-new-spaces.changes.md
@@ -1,5 +1,12 @@
 ## Draft 1.4
 
+* Colors conversion is only performed when necessary. Previously, colors could
+  be converted into their current space.
+
+* Color conversion procedure explicitly handles 'carrying forward' missing
+  channels when converting to a space with an analogous component. This was
+  previously only applied to interpolation.
+
 * Allow all color spaces to be used for hue interpolation.
 
 * Remove `specified` hue interpolation method, and normalize hues to be in the
--- a/proposal/color-4-new-spaces.changes.md
+++ b/proposal/color-4-new-spaces.changes.md
@@ -1,3 +1,10 @@
+## Draft 1.4
+
+* Allow all color spaces to be used for hue interpolation.
+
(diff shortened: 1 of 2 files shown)


proposal/color-4-new-spaces.changes.md
```diff
@@ -1,3 +1,22 @@
+## Draft 1.4
+
+* All `lightness` channels are now clamped in the `[0,100]` range.
+
+* Missing channels and mixed numbers/percentages are not allowed in legacy
+  comma-separated `rgb`/`rgba` or `hsl`/`hsla` syntaxes.
```
nex3: We already allow mixed numbers/percentages in this syntax, so we need to continue to do so for backwards compat even if CSS forbids it.

nex3 (changes requested):

nex3 (approved):

PR #3500, opened 2023-03-01
fix: add node types to prod dependencies

43081j:
You ship your own types now, but actually depend on node's types in those types.

`@types/node` is currently a dev dependency, which means we don't pull it down automatically when we install `sass`. That means the sass types will fail to compile, since the node types it depends on will be missing.

It should really be a production dependency.

Slight pain since it means non-ts consumers will have to pull it down, but it is the way. If you're publishing typescript types, your type dependencies need to be declared as regular dependencies.

let me know your thoughts

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,9 +4,11 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "sass",
       "dependencies": {
         "@types/diff": "^5.0.1",
         "@types/glob": "^7.1.4",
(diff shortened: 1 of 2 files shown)


nex3:
This is not the repository for Dart Sass. This is the Sass language repo, which tracks the language specification. You're probably looking for https://github.com/sass/dart-sass.

43081j:
Whoop yeah my mistake

No worries, I'll go figure out what's going on over in the other repo.

Issue #3258, opened 2022-02-08
Support constants in calc()


nex3:
- [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/calc-constants.md)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1900) (https://github.com/sass/dart-sass/pull/1922)
- [x] [Tests](https://github.com/sass/sass-spec/issues/1885) (https://github.com/sass/sass-spec/pull/1890)
- [x] [Documentation](https://github.com/sass/sass-site/issues/700) (https://github.com/sass/sass-site/pull/707)
- [x] Update canonical spec (https://github.com/sass/sass/pull/3519)

---

Values and Units 4 defines `pi`, `e`, `infinity`, `-infinity`, and `NaN` [constants for use in `calc()` expressions](https://drafts.csswg.org/css-values-4/#calc-constants). We should support these as they land in browsers. The current status (as of 8 February 2022) is:

* [x] `infinity`: [landing in Chrome 99](https://blog.chromium.org/2022/02/chrome-99-css-cascade-layers-new-picker.html?m=1)
* [x] `-infinity`: [landing in Chrome 99](https://blog.chromium.org/2022/02/chrome-99-css-cascade-layers-new-picker.html?m=1)
* [x] `NaN`: [landing in Chrome 99](https://blog.chromium.org/2022/02/chrome-99-css-cascade-layers-new-picker.html?m=1)
* [x] `pi`: unsupported anywhere as far as I can tell
* [x] `e`: unsupported anywhere as far as I can tell

For the degenerate constants (`infinity`, `-infinity`, and `NaN`) we should not only support parsing them in `calc()` expressions but also serializing Sass numbers with those values *to* `calc()` expressions.

luiscla27:
I think this enhancement is _a kind of_ dup of  #2954 

(I was looking for [general purpose constants](https://github.com/sass/sass/issues/3398) but found yours and the suggested dup instead)

nex3:
They're different—https://github.com/sass/sass/issues/2954 would add support for `math.$infinity`, `math.$negative-infinity`, and `math.$nan` (`math.$pi` and `math.$e` already exist). This would add support for `calc(infinity)`, `calc(-infinity)`, `calc(NaN)`, `calc(pi)`, and `calc(e)`.

nex3:
[...]

PR #3501, opened 2023-03-02
[Calc Constants] Write a proposal

nex3:
See sass/sass#3258

### DIFF
--- a/proposal/calc-constants.md
+++ b/proposal/calc-constants.md
@@ -1,0 +1,183 @@
+# Calculation Constants: Draft 1
+
+*([Issue](https://github.com/sass/sass/issues/3258))*
+
+This proposal adds support for constant names in CSS calculations.
+
+## Table of Contents
+
+* [foo]()
+
+## Background
+
+> This section is non-normative.
+
+Since Sass now fully parses calculations and resolves them where necessary, it's
+necessary to keep up with new syntactic features added in plain CSS. CSS Values
+4 adds support for several [numeric constants], written as identifiers in the
+calculation, which Sass should support as well.
+
+[numeric constants]: https://www.w3.org/TR/css-values-4/#calc-constants
+
+## Summary
+
+Calculations may now contain arbitrary identifiers in any value position. The
+specific identifiers `pi`, `e`, `infinity`, `-infinity`, and `NaN` will be
+parsed to their corresponding double values.
+
+In addition, this creates for the first time a clear and standard CSS way of
+representing degenerate numeric constants like `NaN` and the infinities. This
+allows us to serialize a Sass number with value `Infinity`, `-Infinity`, or
+`NaN` to `calc(infinity)`, `calc(-inifinity)`, and `calc(NaN)`, respectively.
+
+### Design Decisions
+
+#### Arbitrary Identifiers
+
+We're choosing to support arbitrary identifiers in calculations, rather than
+just those specified in CSS Values 4, for forwards-compatibility. While it's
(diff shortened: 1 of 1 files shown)


proposal/calc-constants.md
```diff
@@ -0,0 +1,202 @@
...
+  [degenerate]: #degenerate-number
+  [convert it to a calculation]: #converting-a-number-to-a-calculation
+
+* Otherwise, serialize the number as normal.
+
+#### `CalculationOperation`
+
+Add another possible condition for parenthesizing the right value:
+
+* the operator is `"+"`, `"-"`, or `"/"` and the right value is a degenerate
```
Goodwine, line 146: Why does the right value for multiplication not need to be parenthesized? I kind of expected `+` and `*` to be roughly the same 🤔 

Also nit: Uppercase for the first word

proposal/calc-constants.md
```diff
@@ -0,0 +1,202 @@
...
+#### Arbitrary Identifiers
+
+We're choosing to support arbitrary identifiers in calculations, rather than
+just those specified in CSS Values 4, for forwards-compatibility. While it's
+unclear whether the `<calc-constant>` production itself will be expanded, it's
+likely that calculations will begin supporting additional non-constant values
+depending on context (such as in relative color functions like `rgb(from
+var(--color) r g calc(b * 1.5))`).
+
+### Serializing Complex Units
```
Goodwine, line 63: Complex units being things like `px*px` or `em/%`?

proposal/calc-constants.md
```diff
@@ -0,0 +1,202 @@
...
+
+* For each unit `unit` in `number`'s numerator units:
+
+  * Set `value` to a `CalculationOperation` with `operator` set to `'*'`, `left`
+    set to `value`, and `right` set to a number with value 1 and unit `unit`.
+
+* For each unit `unit` in `number`'s denominator units:
+
+  * Set `value` to a `CalculationOperation` with `operator` set to `'/'`, `left`
+    set to `value`, and `right` set to a number with value 1 and unit `unit`.
```
Goodwine, line 195: I imagine this procedure is here only for reference because the section above about numbers says an error would be thrown. Is there any other case that would trigger the "convert a number to a calculation" besides the number being degenerate?

Goodwine (commented):
Looks good, but I have a couple questions

proposal/calc-constants.md
nex3: Exactly.

proposal/calc-constants.md
nex3: > Why does the right value for multiplication not need to be parenthesized? I kind of expected `+` and `*` to be roughly the same

It has to do with the order of operations. Operators with the same precedence are evaluated left-to-right, `*` and `/` have the same precedence, and degenerate numbers always produce multiplication and division, so a degenerate number on the left-hand side of a multiplication is always guaranteed to be evaluated first without parentheses.

Similarly, `+` and `-` have _lower_ precedence than `*` and `/`, so they're always evaluated later... meaning that a degenerate number on either side of them will be evaluated first.

> Also nit: Uppercase for the first word

This is going into a comma-separated list, which is why I didn't uppercase it.

proposal/calc-constants.md
nex3: Not currently, but we'll probably need to do so later on once browsers start supporting complex units in calculations.

Goodwine (approved):

PR #3508, opened 2023-03-13
Document the progress of deprecating `@import`

ntkme:
This PR updates the documentation tracking the progress of deprecating `@import`, to state that we have reached the 80% adoption target for Dart Sass in order to move this deprecation from "future" to "current".

cc @nex3 @jathak


nex3:
I've filed https://github.com/sass/sass/issues/3513 to track this process.

PR #3521, opened 2023-03-24
[JS Calculations] Add a proposal

nex3:


jathak (approved):

proposal/calculation-api.d.ts
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * Creates a value that represents `calc(argument)` expression.
+   *
+   * - If `argument` is or transitively contains a quoted `SassString`, throw an
+   *   error.
+   *
+   * - Let `calculation` be a calculation with name `"calc"` and a `argument` as
+   *   its single argument.
+   *
+   * - Return the result of [simplifying] `calculation`.
```
stof, line 60: Should we enforce that this returns a simplified calculation ? It would force to duplicate the simplification algorithm in all embedded hosts

proposal/calculation-api.d.ts
```diff
@@ -0,0 +1,193 @@
...
+  /** `internal`'s `operator` field. */
+  get operator(): CalculationOperator;
+
+  /** `internal`'s `left` field. */
+  get left(): CalculationValue;
+
+  /** `internal`'s `right` field. */
+  get right(): CalculationValue;
+
+  /** Whether `internal` is equal to `other.internal` in Sass. */
```
stof, line 162: the description looks weird to me, as there is no `internal` in this type definition.

nex3: See https://github.com/sass/sass/blob/32a7db88ee8d320ee92e6e0deb4f85e21696ba77/spec/js-api/value/index.d.ts#L32-L34

Issue #3520, opened 2023-03-24
Add JS API for deprecations


jathak:
* [x] Proposal (https://github.com/sass/sass/pull/3522) - Open for public comment until 4/19
* [x] Tests
* [x] Documentation
* [x] Dart Sass
* [x] Node Host

---

Tracking issue for adding deprecations support to the JS API, matching the behavior of the `--fatal-deprecation` (https://github.com/sass/dart-sass/issues/633), `--future-deprecation` (https://github.com/sass/dart-sass/issues/1363), and `--silence-deprecation` (https://github.com/sass/dart-sass/issues/1918) CLI flags.

stof:
is there any progress on implementing that accepted proposal or was it forgotten ?

nex3:
It's in @jathak's list of tasks, but that list is very long and this hasn't hit the top of the priority queue yet.

PR #3522, opened 2023-03-24
Add JS deprecation API proposal

jathak:
See #3520.

This should generally be similar to the Dart API, but with added support for silencing deprecations and an `obsoleteIn` field (both of which I will add to the CLI and Dart API alongside the implementation for this).

This also adds support for loggers getting the actual `Deprecation` object in addition to the boolean flag, which I _believe_ should be a non-breaking change if my TypeScript understanding is correct. I'd like to add this support to the Dart API too, but that _will_ require a breaking change, so it will need to wait for Dart Sass 2.0.0.

Let me know if missing anything here.

### DIFF
--- a/proposal/js-deprecations.d.ts
+++ b/proposal/js-deprecations.d.ts
@@ -1,0 +1,302 @@
+/**
+ * # JavaScript Deprecations API: Draft 1
+ *
+ * *([Issue](https://github.com/sass/sass/issues/3520))*
+ *
+ * ## Background
+ *
+ * > This section is non-normative.
+ *
+ * We recently added support to Dart Sass that allowed users to opt in to
+ * treating deprecation warnings as errors (on a per-deprecation basis), as
+ * well as opting in early to certain future deprecations. This is currently
+ * supported on the command line and via the Dart API, but we'd like to extend
+ * this support to the JS API as well.
+ *
+ * We would also like to add support for silencing a particular deprecation's
+ * warnings, primarily to enable a gentler process for deprecating `@import`.
+ *
+ * ## Summary
+ *
+ * > This section is non-normative.
+ *
+ * This proposal adds two new interfaces to the JS API (`Deprecation` and
+ * `Version`), three new optional properties on `Options` (`fatalDeprecations`,
+ * `silenceDeprecations`, and `futureDeprecations`), a new parameter on
+ * `Logger.warn` (`options.deprecationType`) two type aliases (`DeprecationOrId`
+ * and `DeprecationStatus`) and a new object `deprecations` that contains the
+ * various `Deprecation` objects.
+ *
+ * All language-wide deprecations are specified in `deprecations`, and any new
+ * deprecations added in the future should update the specification accordingly.
+ * However, the `deprecations` object also allows individual implementations to
+ * add their own, implementation-specific deprecations.
+ *
+ * Every `Deprecation` has a unique `id`, one of four `status` values, and
+ * (optionally) a human-readable `description`. Depending on the status, each
+ * deprecation may also have a `deprecatedIn` version and an `obsoleteIn`
+ * version that specify the compiler versions the deprecation became active
+ * and became obsolete in, respectively.
+ *
+ * ### Design Decisions
+ *
+ * #### Exposing the Full `Deprecation` Interface
+ *
+ * One alternative to specifying a full `Deprecation` interface is to just have
+ * the relevant APIs take in string IDs. We considered this, but concluded that
+ * each deprecation has additional metadata that users of the API may wish to
+ * access (for example, a bundler may wish to surface the `description` and
+ * `deprecatedIn` version to its users).
+ *
+ * #### Formally Specifying the Deprecations
+ *
+ * We chose to make the list of deprecations part of the specification itself,
+ * as this ensures that the language-wide deprecations are consistent across
+ * implementations. However, if an implementation wishes to add a deprecation
+ * that applies only to itself, it may still do so.
+ *
+ * Additionally, while a deprecation's status is part of the specification, we
+ * chose to leave the `deprecatedIn` and `obsoleteIn` versions of each
+ * deprecation out of the specification. As the two current implementers of this
+ * API are both based on Dart Sass, these versions are _currently_ consistent
+ * across implementations in practice, potential future implementers should not
+ * need to be tied to Dart Sass's versioning.
+ *
+ * #### Errors and Warnings for Invalid Deprecations
+ *
+ * For all of the deprecation options, we chose to have the API throw an error
+ * if a string ID is passed for a deprecation that does not exist. This will
+ * ensure that typos are caught early and is consistent with the behavior on the
+ * command line.
+ *
+ * For `fatalDeprecations`, we chose to make passing in a future deprecation
+ * that's not also passed to `futureDeprecations` an error, since, in a
+ * sense, future deprecations aren't really deprecations yet unless they're
+ * enabled via `futureDeprecations`. This also prevents someone who just passes
+ * all deprecations to `fatalDeprecations` from accidentally opting into a
+ * deprecation that hasn't been released yet. This is not yet consistent with
+ * the behavior on the command line / Dart API, but we plan to update them to
+ * match this specification.
+ *
+ * For obsolete deprecations passed to `fatalDeprecations`, we chose to emit a
+ * warning so users can know to update their configuration, but we allow
+ * compilation to proceed without an error, since an obsolete deprecation means
+ * the underlying behavior has now been removed, which should match the behavior
+ * of making that deprecation fatal while it was still active.
+ *
+ * For `silenceDeprecations`, we chose to make passing anything other than
+ * an active deprecation an error to prevent users from unintentionally
+ * over-silencing. Future deprecations are already effectively silenced, so
+ * users shouldn't be able to pass it here until it's actually active. If the
+ * deprecation a user was silencing is now obsolete, that means the breaking
+ * change has now been made, so silencing has no effect, and users who were
+ * silencing it may not be ready for the actual breaking change. There is no
+ * way to distinguish between different user-authored deprecations, so they
+ * should not be able to be silenced as a group.
+ *
+ * For `futureDeprecations`, we only emit a warning for non-future deprecations,
+ * since future deprecations all eventually become active. The warning will let
+ * users know to clean up their configuration, but there should be no behavior
+ * change.
+ */
+
+/* ## API */
+
+import {SourceSpan} from '../spec/js-api';
+
+declare module '../spec/js-api' {
+  interface Options<sync extends 'sync' | 'async'> {
+    /**
+     * A set of deprecations to treat as fatal.
+     *
+     * If a version is provided, then all deprecations released in that
+     * version or earlier will be treated as fatal. If a string is provided, it
+     * will resolve to the deprecation with that ID, or error if none exists.
+     *
+     * This will throw an error if a future deprecation is included here, unless
+     * that future deprecation is also passed to `futureDeprecations`. This
+     * will emit a warning if an obsolete deprecation is included here. */
+    fatalDeprecations?: (Deprecation | string | Version)[];
+
+    /**
+     * A set of active deprecations to ignore.
+     *
+     * If a string is provided, it will resolve to the deprecation with that ID,
+     * or error if none exists.
+     *
+     * This will throw an error if a future, obsolete, or user-authored
+     * deprecation is included here.
+     */
+    silenceDeprecations?: (Deprecation | string)[];
+
+    /**
+     * A set of future deprecations to opt into early.
+     *
+     * If a string is provided, it will resolve to the deprecation with that ID,
+     * or error if none exists.
+     *
+     * This will emit a warning if a non-future deprecation is included here.
+     */
+    futureDeprecations?: (Deprecation | string)[];
+  }
+
+  interface Logger {
+    /**
+     * Update the third sub-bullet of bullet two to read:
+     *
+     * If this warning is caused by behavior that used to be allowed but will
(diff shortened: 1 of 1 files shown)


proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+ * `Version`), three new optional properties on `Options` (`fatalDeprecations`,
+ * `silenceDeprecations`, and `futureDeprecations`), a new parameter on
+ * `Logger.warn` (`options.deprecationType`) two type aliases (`DeprecationOrId`
+ * and `DeprecationStatus`) and a new object `deprecations` that contains the
+ * various `Deprecation` objects.
+ *
+ * All language-wide deprecations are specified in `deprecations`, and any new
+ * deprecations added in the future should update the specification accordingly.
+ * However, the `deprecations` object also allows individual implementations to
+ * add their own, implementation-specific deprecations.
```
nex3: I think even implementation-specific deprecations should be added to this spec, since otherwise we'll either end up in a situation where typos don't produce errors or where some implementations produce errors for deprecations that are valid in other implementations.

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+ * all deprecations to `fatalDeprecations` from accidentally opting into a
+ * deprecation that hasn't been released yet. This is not yet consistent with
+ * the behavior on the command line / Dart API, but we plan to update them to
+ * match this specification.
+ *
+ * For obsolete deprecations passed to `fatalDeprecations`, we chose to emit a
+ * warning so users can know to update their configuration, but we allow
+ * compilation to proceed without an error, since an obsolete deprecation means
+ * the underlying behavior has now been removed, which should match the behavior
+ * of making that deprecation fatal while it was still active.
```
nex3: Is the expectation that deprecations will remain in place even after a major version release? (I think "yes" makes sense here because the nature of deprecations means that different implementations may remove them at different times, but this is worth documenting I think.)

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+   * The compiler version this feature was fully removed in, making the
+   * deprecation obsolete.
+   *
+   * This is null for active and future deprecations.
+   */
+  obsoleteIn: status extends 'obsolete' ? Version : null;
+}
+
+/** A semantic version of the compiler. */
+export interface Version {
```
nex3: I'm concerned that this custom version struct will be annoying for users to construct by hand relative to just writing out the version number. It's probably not worth depending on the JS `semver` library since it's not ES-compatible and it has its own transitive dependencies, but maybe it would be worth making this a class with a constructor that can parse a string.

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * A set of deprecations to treat as fatal.
+     *
+     * If a version is provided, then all deprecations released in that
+     * version or earlier will be treated as fatal. If a string is provided, it
+     * will resolve to the deprecation with that ID, or error if none exists.
+     *
+     * This will throw an error if a future deprecation is included here, unless
+     * that future deprecation is also passed to `futureDeprecations`. This
+     * will emit a warning if an obsolete deprecation is included here. */
```
nex3: Remember that this is effectively spec text rather than user-facing documentation. The audience is implementation authors who are looking to consistently and precisely implement this behavior. To that end, I think it's better to write it imperatively as a description of how the compilation logic behaves when the option is passed, along with explicit logic for converting strings to deprecations and handling errors.

This goes for the member comments below as well.

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+    fatalDeprecations?: (Deprecation | string | Version)[];
```
nex3: You can use `keyof _Deprecation` rather than `string` here to provide better static checking in TypeScript and better autocompletion.

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+     * be disallowed in the future, set `options.deprecation` to `true` and
+     * set `options.deprecationType` to the relevant `Deprecation`. Otherwise,
+     * set `options.deprecation` to `false` and leave `options.deprecationType`
+     * undefined.
+     */
+    warn?(
+      message: string,
+      options: {
+        deprecation: boolean;
+        deprecationType?: Deprecation;
```
nex3, line 185: I thought for the Dart API we decided that it was better to not allow user-authored warnings to have specific deprecation types, and to lump them all under "user deprecations"

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+   * An object containing all of the deprecations recognized by this
+   * implementation as properties, as well as a `find` function to get the
+   * deprecation for a given ID.
+   */
+  export const deprecations: Readonly<
+    _Deprecations & {[key: string]: Deprecation}
+  >;
+}
+
+interface _Deprecations {
```
nex3: Naming this with a leading underscore is likely to be confusing. We generate JS docs from the TypeScript interfaces, so this name will be publicly visible even if users can't access it as a type.

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+  strictUnary: Deprecation<'strict-unary', 'active'>;
+
+  /** Deprecation for passing invalid units to certain built-in functions. */
+  functionUnits: Deprecation<'function-units', 'active'>;
+
+  /** Deprecation for `@import` rules. */
+  importRules: Deprecation<'import', 'future'>;
+
+  /** Used for deprecations coming from user-authored code. */
+  userAuthored: Deprecation<'user-authored', 'user'>;
```
nex3: It's a bit of a pain, but it's probably worth explicitly saying here "update the proposal for XYZ to say that it emits deprecations with id QRS" so that the IDs end up colocated with the text that describes when they're emitted.

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+  id extends string = string,
+  status extends DeprecationStatus = DeprecationStatus
+> {
+  /**
+   * An ID for this deprecation.
+   *
+   * For language-wide deprecations specified in the API, this will be a short,
+   * alphanumeric, kebab-case identifier. Implementation-specific deprecations
+   * should be prefixed with an implementation identifier followed by a slash to
+   * avoid collision with future language-wide deprecations.
```
nex3: I'm not sure how feasible it is to distinguish impl-specific deprecations by prefix. Currently _all_ deprecations are essentially Dart Sass only, since there are no other substantially-compatible impls. Even if there were, which impls actually need to deprecate behavior can be contingent on whether/when they implement the original behavior—when we were first rolling out Dart Sass, we just implemented the new non-deprecated behaviors for a number of "standard" Ruby Sass deprecations.

I think generally it's probably a good idea to just have all deprecations share a namespace, and ask implementation authors to collaborate on a central set of IDs. I can't imagine there will ever be so many implementations of such independence that that becomes onerous.

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+ * that applies only to itself, it may still do so.
+ *
+ * Additionally, while a deprecation's status is part of the specification, we
+ * chose to leave the `deprecatedIn` and `obsoleteIn` versions of each
+ * deprecation out of the specification. As the two current implementers of this
+ * API are both based on Dart Sass, these versions are _currently_ consistent
+ * across implementations in practice, potential future implementers should not
+ * need to be tied to Dart Sass's versioning.
+ *
+ * #### Errors and Warnings for Invalid Deprecations
```
nex3: We should consider the possibility of users writing configurations that are meant to apply to multiple implementations. Which deprecations have which status is something that's particularly likely to have skew across implementations, so I'm inclined to avoid throwing errors in situations where the user looks like they're passing in more deprecations than necessary to one of these options.

proposal/js-deprecations.d.ts
```diff
@@ -0,0 +1,302 @@
...
+  description?: string;
+
+  /**
+   * The compiler version this feature was first deprecated in.
+   *
+   * This is implementation-dependent, so versions are not guaranteed to be
+   * consistent between different compilers. For future deprecations, or those
+   * originating from user-authored code, this is null.
+   */
+  deprecatedIn: status extends 'future' | 'user' ? null : Version;
```
nex3, line 306: Loving these type conditions 😊

nex3 (changes requested):

proposal/js-deprecations.d.ts
jathak: Done

proposal/js-deprecations.d.ts
jathak: Yes, that was my intention with adding the concept of obsolete deprecations. Added clarification for this above.

proposal/js-deprecations.d.ts
jathak: Done

proposal/js-deprecations.d.ts
jathak: The idea here was to allow user-authored Loggers to have access to the deprecation types used by the compiler, not to allow users to create their own deprecation types.

I think I've now updated the `Deprecation` interface to be restricted only to the IDs provided in `Deprecations` to prevent the latter.

proposal/js-deprecations.d.ts
jathak: Okay, I've updated everything to be a warning, and added specification for the precedence of the three options if a deprecation is passed to more than one.

proposal/js-deprecations.d.ts
jathak: I'm not quite understanding what you mean by this. Do you mean updating other proposal texts within this PR, or adding a general comment here to describe the process for adding new deprecations, or something else?

proposal/js-deprecations.d.ts
jathak: Removed implementation-specific deprecations.

proposal/js-deprecations.d.ts
jathak: Done

proposal/js-deprecations.d.ts
jathak: Done

proposal/js-deprecations.d.ts
jathak: Done

proposal/js-deprecations.d.ts
nex3: I mean in the spec comments for each of these deprecation types, literally writing "update the proposal for XYZ to say that it emits deprecations with id QRS". (Language proposals are generally intended to specify modifications to existing spec text.)

proposal/js-deprecations.d.ts
```diff
@@ -109,35 +117,54 @@ declare module '../spec/js-api' {
...
-     * This will throw an error if a future deprecation is included here, unless
-     * that future deprecation is also passed to `futureDeprecations`. This
-     * will emit a warning if an obsolete deprecation is included here. */
-    fatalDeprecations?: (Deprecation | string | Version)[];
+     * The compiler must error if a future deprecation is included here, unless
+     * that future deprecation is also passed to `futureDeprecations`. It must
+     * emit a warning if an obsolete deprecation is included here.
+     *
+     * If a deprecation is passed both here and to `silenceDeprecations`, a
+     * warning must be emitted, but making it fatal must take precedence.
```
nex3: "making the deprecation fatal" helps disambiguate the referent

proposal/js-deprecations.d.ts
```diff
@@ -109,35 +117,54 @@ declare module '../spec/js-api' {
...
      * A set of deprecations to treat as fatal.
      *
-     * If a version is provided, then all deprecations released in that
-     * version or earlier will be treated as fatal. If a string is provided, it
-     * will resolve to the deprecation with that ID, or error if none exists.
+     * If a deprecation warning of any provided type is encountered during
+     * compilation, the compiler must error instead.
+     *
+     * The compiler should convert any string passed here to a `Deprecation`
+     * by indexing `Deprecations`.
```
nex3: Nit:

```suggestion
     * by indexing `deprecations`.
```

...since this is the actual field with concrete values.

proposal/js-deprecations.d.ts
```diff
@@ -109,35 +117,54 @@ declare module '../spec/js-api' {
...
-     * or error if none exists.
+     * If a deprecation warning of any provided type is encountered during
+     * compilation, the compiler must ignore it.
      *
-     * This will throw an error if a future, obsolete, or user-authored
-     * deprecation is included here.
+     * The compiler should convert any string passed here to a `Deprecation`
+     * by indexing `Deprecations`.
+     *
+     * The compiler must error if an obsolete or user-authored is included here.
```
nex3: This sentence is missing its object. Also, might be worth being explicit that "user-authored" means specifically `deprecations['user-authored']`.

nex3 (approved):

proposal/js-deprecations.d.ts
jathak: Done.

proposal/js-deprecations.d.ts
jathak: Done

proposal/js-deprecations.d.ts
jathak: Done

proposal/js-deprecations.d.ts
jathak: Okay, I added comments for each of the ones that had a matching proposal. `call-string`, `elseif`, `moz-document`, `relative-canonical`, `new-global`, and `color-module-compat` didn't as far as I could find, so I'm not sure if there's something else I should add here.

Issue #3507, opened 2023-03-13
CSS invert and SASS invert are conflicting.


ByPikod:
```css
.input {

    --self-background: var(--input-background);
    --self-forecolor: var(--input-forecolor);

    background: var(--self-background);
    color: var(--self-forecolor);

    padding: var(--input-padding);
    transition: all 0.2s;
    border-radius: var(--input-border-radius);

    outline: none;
    border: none;
    
    &:focus {
        --self-background: var(--input-hover-background);
        --self-forecolor: var(--input-hover-forecolor);
    }

    &::placeholder {
        color: invert(var(--self-forecolor));
    }

}
```

![image](https://user-images.githubusercontent.com/46069238/224629051-4d557138-d1b8-46b7-8b6a-ebfe0ed2eb0d.png)

That might be a misunderstanding of mine since I am just learning SASS. But seems like, CSS invert and SASS invert are conflicting. 

Here is a workaround I found:
```css
&::placeholder {
     color: #{"invert(var(--self-forecolor))"};
}
```

Goodwine:
You are trying to use the CSS `invert()` function in a `color:` property. CSS `invert()` can only be used in a `filter:` declaration.

That said, the error is indeed thrown even within a filter attribute.

I'll take a look at this, but I wanted to let you know that `color: invert(...)` doesn't work in CSS.

ByPikod:
My bad. Thanks for letting me know.

Goodwine:
The use of `filter: invert(var(--foo))` still throws an error when it shouldn't tho, so I'll reopen this to look at that

rororomeda:
+1

Goodwine:
I have an implementation at https://github.com/sass/dart-sass/pull/1926, however I need to do a language change proposal. FYI this issue affects a few other native CSS filter functions besides `invert()`. Those are: `grayscale()`, `scale()` and `saturate()`.

PR #3528, opened 2023-04-04
Allow global functions 'invert()', 'grayscale()', 'opacity()', and 'saturate()' to use 'var()'

Goodwine:
When these functions use `var()` they output the native CSS function with the same name and args.
The saturate spec was expanded to match the existing implementation.

See implementation: https://github.com/sass/dart-sass/pull/1926
Issue https://github.com/sass/sass/issues/3507

### DIFF
--- a/spec/built-in-modules/color.md
+++ b/spec/built-in-modules/color.md
@@ -193,9 +193,9 @@ This function is also available as a global function named `opacity()`.
 >
 > [global functions]: ../functions.md#alpha
 
-* If `$color` is a [special number] and this function is called as the global
-  `opacity()` function, return a plain CSS function string with the name
-  `"opacity"` and the argument `$color`.
+* If `$color` is a number or a [special number], and this function is called as
+  the global `opacity()` function, return a plain CSS function string with the
+  name `"opacity"` and the argument `$color`.
 
 * Otherwise, if `$color` is not a color, throw an error.
 
@@ -380,9 +380,9 @@ grayscale($color)
 
 This function is also available as a global function named `grayscale()`.
 
-* If `$color` is a [special number] and this function is called as a global
-  function, return a plain CSS function string with the name `"grayscale"`
-  and the argument `$color`.
+* If `$color` is a number or a [special number], and this function is called as
+  a global function, return a plain CSS function string with the name
+  `"grayscale"` and the argument `$color`.
 
 * Otherwise, if `$color` is not a color, throw an error.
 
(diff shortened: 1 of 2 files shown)


spec/built-in-modules/color.md
```diff
@@ -193,9 +193,9 @@ This function is also available as a global function named `opacity()`.
 >
 > [global functions]: ../functions.md#alpha
 
-* If `$color` is a number and this function is called as the global `opacity()`
-  function, return a plain CSS function string with the name `"opacity"` and the
-  argument `$color`.
+* If `$color` is a number or a [special variable string], and this function is
```
nex3: I think we should do this logic for all special numbers, not _just_ special variable strings.

spec/built-in-modules/color.md
```diff
@@ -593,7 +594,24 @@ This function is also available as a global function named `mix()`.
 saturate($color, $amount)
 ```
 
-* Throw an error.
+* If this function is called as a global function:
+
+  * If this function is called with a single argument `$amount`:
```
nex3: Currently, the signature for this function requires two arguments. You'll need to add a separate signature that only takes one argument. (See `rgb()` as an example of an overloaded function.)

spec/built-in-modules/color.md
```diff
@@ -593,7 +594,24 @@ This function is also available as a global function named `mix()`.
...
+
+  * Otherwise:
+
+    * If `$color` isn't a color, throw an error.
+
+    * If `$amount` isn't a number, throw an error.
+
+    * Return the result of calling `changeHsl($saturation)`, where `$saturation`
+      is the `$color`'s saturation incremented by `$amount` clamped between the
+      inclusive range `[0, 100]`.
```
nex3: Leaving this out was intentional—the old behavior is deprecated, and the canonical spec only includes non-deprecated behavior.

spec/built-in-modules/color.md
```diff
@@ -711,3 +729,5 @@ whiteness($color)
   > RGB channels and convert back when `whiteness()` or `blackness()` is called,
   > where another may keep around their original HWB values and return those
   > as-is.
+
+[special variable string]: ../spec.md#special-variable-string
```
nex3: We generally put reference links after the first reference so they're easier to clean up later on. Also, I think this term is defined in `functions.md`.

nex3 (changes requested):

spec/built-in-modules/color.md
Goodwine: ohh.... I actually wanted to do it that way but I wasn't sure XD

nex3 (approved):

Issue #3504, opened 2023-03-09
Support additional calculation functions


nex3:
- [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/calc-functions.md) (https://github.com/sass/sass/pull/3540)
- **Original Implementation** (these changes incorrectly caused a breaking change and had to be reverted; see https://github.com/sass/sass/issues/3504#issuecomment-1682737824)
  - [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1934) (https://github.com/sass/dart-sass/pull/1970)
  - [x] [Tests](https://github.com/sass/sass-spec/issues/1899) (https://github.com/sass/sass-spec/pull/1912)
  - [x] [Documentation](https://github.com/sass/sass-site/issues/719) (https://github.com/sass/sass-site/pull/755)
- **Revert**
  - [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1934) (https://github.com/sass/dart-sass/pull/2067)
  - [x] [Tests](https://github.com/sass/sass-spec/issues/1899) (https://github.com/sass/sass-spec/pull/1931)
  - [x] [Documentation](https://github.com/sass/sass-site/issues/719) (https://github.com/sass/sass-site/pull/804)
- **Roll Forward**
  - [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1934) (https://github.com/sass/dart-sass/pull/2076)
  - [x] [Tests](https://github.com/sass/sass-spec/issues/1899) (https://github.com/sass/sass-spec/pull/1934)
  - [x] [Documentation](https://github.com/sass/sass-site/issues/719) (https://github.com/sass/sass-site/pull/831)
  - [x] Update canonical spec (https://github.com/sass/sass/pull/3693)
- **Breaking Changes**
  - [ ] [Dart Sass](https://github.com/sass/dart-sass/issues/1934)
  - [ ] [Tests](https://github.com/sass/sass-spec/issues/1899)
  - [ ] [Documentation](https://github.com/sass/sass-site/issues/719)

Trigonometric calculation functions have landed in [Firefox and Safari](https://developer.mozilla.org/en-US/docs/Web/CSS/sin#browser_compatibility), and the rest of the calculation functions seem to be live in Safari. We should support these in Sass.

nex3:
The proposal is out for review. Since it involves some complexity with the handling of cases like `abs(10%)` and involves a small breaking change to the behavior of `%`, I'm going to give it a month for public comment.

[...]

PR #3540, opened 2023-04-12
[Calc Functions] Add a proposal

nex3:
See sass/sass#3504


proposal/calc-functions.md
```diff
@@ -0,0 +1,363 @@
...
+* If `calc`'s name is `"sign"` and `arguments` contains exactly a single number
+  `number` with [known units]:
+
+  * If `number`'s value is positive, return `1`.
+  * If `number`'s value is negative, return `-1`.
+  * Otherwise, return a unitless number with the same value as `number`.
+  
+    > In this case, `number` is either positive zero, negative zero, or NaN.
+  
+  > To match CSS's behavior, these computations _don't_ use fuzzy comparisons.
```
Goodwine, line 283: Does "fuzzy comparisons" here mean this?
https://github.com/sass/sass/blob/main/spec/types/number.md#fuzzy-equality

Also, I'm slightly confused by what it means for this to "not use fuzzy comparisons" when determining whether a number is positive or negative

proposal/calc-functions.md
```diff
@@ -0,0 +1,363 @@
...
+
+* If `calc`'s name is `"atan2"`, `"pow"`, or `"log"`:
+
+  * If `calc`'s name isn't `"log"`, `arguments` has only one element, and it's
```
Goodwine, line 287: Should these 3 functions be split into 2 separate rules, one for `log` and the other for `pow` and `atan2`? It's a bit odd how these are merged but the condition below separates them again

Also, as written, IIUC `log(1)` matches the first condition so it doesn't throw an error. But then it fails the `contains two numbers` condition 🤔

proposal/calc-functions.md
```diff
@@ -0,0 +1,363 @@
...
+
+This algorithm takes a calculation `calc` and returns a number or a calculation.
+
+> This algorithm is intended to return a value that's CSS-semantically identical
+> to the input.
+
+* If `calc` was parsed from an expression within a `SupportsDeclaration`'s
+  `Expression`, but outside any interpolation, return a `calc` as-is.
+
+* Let `arguments` be the result of [simplifying](#simplifying-a-calculationvalue) each
```
Goodwine: this link seems to be broken (or at least it's not rendering anything, I imagine it will work when copied over to the `calculation.md` file, but perhaps it should be fixed here too?

proposal/calc-functions.md
```diff
@@ -0,0 +1,363 @@
...
+
+      [definitely-incompatible]: ../spec/types/number.md#possibly-compatible-numbers
+
+    * If `dividend` and `modulus` are mutually [compatible], let `result` be the
+      result of `dividend % modulus`.
+
+      [compatible]: ../spec/types/number.md#compatible-units
+
+    * If `calc`'s name is `"rem"`, and if `dividend` is positive and `modulus`
+      is negative or vice versa, set `result` to `mod - dividend`.
```
Goodwine: The `mod` in `mod - dividend` should be `modulus` instead.

Goodwine: Is `-` correct here? If mod is 10 and dividend is -10 this will return `20` instead of `0` 🤔, should it be `+` instead given that both have different sign?

proposal/calc-functions.md
```diff
@@ -0,0 +1,363 @@
...
+  
+    * Otherwise, set `number` and `step` to the two arguments respectively and
+      `strategy` to an unquoted string with value `"nearest"`.
+
+  * Otherwise, if the single argument isn't an unquoted string or
+    `CalculationInterpolation`, throw an error.
+
+  * If `strategy`, `number`, and `step` are set:
+
+    * If `strategy` isn't a [special variable string], nor is it an unquoted
```
Goodwine, line 348: It looks like `[special variable string]` is not linkified

proposal/calc-functions.md
```diff
@@ -0,0 +1,363 @@
...
+      isn't an unquoted string or `CalculationInterpolation`, throw an error.
+
+      > Normally we allow unquoted strings anywhere in a calculation, but this
+      > helps catch the likely error of a user accidentally writing `round(up,
+      > 10px)` without realizing that it needs a third argument.
+  
+    * Otherwise, set `number` and `step` to the two arguments respectively and
+      `strategy` to an unquoted string with value `"nearest"`.
+
+  * Otherwise, if the single argument isn't an unquoted string or
```
Goodwine, line 343: IIUC this means that a single argument for `round()` can be a single unquoted string. So what would `round(foo)` do?

proposal/calc-functions.md
```diff
@@ -0,0 +1,363 @@
...
+    `CalculationInterpolation`, throw an error.
+
+  * If `strategy`, `number`, and `step` are set:
+
+    * If `strategy` isn't a [special variable string], nor is it an unquoted
+      string or interpolation with value `"nearest"`, `"up"`, `"down"`, or
+      `"to-zero"`, throw an error.
+
+    * If `strategy` is an unquoted string or interpolation and both `number` and
```
Goodwine, line 352: Is this missing an `Otherwise, throw an error`? 
I noticed some other branches in the algorithm don't "handle all cases" either, does that mean that any "unhandled" case will be caught by the final condition:

> Otherwise, return a calculation with the same name as `calc` and `arguments`
  as its arguments.
  
?

proposal/calc-functions.md
```diff
@@ -0,0 +1,363 @@
...
+`10%` because the number is positive, but in plain CSS it could return the
+equivalent of `-10%` since percentages are resolved before calculations. To
+handle this, we'll deprecate the global `abs()` function with a percentage and
+recommend users explicitly write `math.abs()` or `abs(#{})` instead.
+
+## Definitions
+
+### Known Units
+
+A number has *known units* unless it has unit `%`.
```
Goodwine, line 88: Is a unitless number considered to have known units? Asking because of the text added on the "Deprecation Process" section.

Goodwine (changes requested):

proposal/calc-functions.md
Goodwine: Or perhaps it should be `result - dividend`?

Goodwine: Or perhaps this should be `+ abs(dividend)`? (assuming `result` is negative)

proposal/calc-functions.md
nex3: > Does "fuzzy comparisons" here mean this? https://github.com/sass/sass/blob/main/spec/types/number.md#fuzzy-equality

Yep

> Also, I'm slightly confused by what it means for this to "not use fuzzy comparisons" when determining whether a number is positive or negative

Fuzzy comparisons round numbers to the nearest 1e-11, so any positive double less than 1e-11 would be considered fuzzy-equal to 0 rather than positive.

proposal/calc-functions.md
nex3: Yes, because it doesn't have unit `%` :smiley:.

proposal/calc-functions.md
nex3: Fixed. Looks like the automated link checker is broken, so I've filed https://github.com/sass/sass/issues/3541 to track that.

proposal/calc-functions.md
nex3: `log(1)` is handled above with the rest of the single-argument functions. `log(var(--foo))` will just fall through to the final clause and return a calculation.

I could split out the two-argument `log()` here, but I think that would make this already-long-winded procedure even more verbose so I'd prefer to keep it terser where possible.

proposal/calc-functions.md
nex3: You were right that `mod` should be `result`. For the case of `rem(10, -10)` we should have an explicit exception for the result being exactly 0.

proposal/calc-functions.md
nex3: Nothing, but it's not feasible for us to detect all the invalid cases. (The relevant cases are `round(var(--foo))` where the variable may contain multiple comma-separated values, or `round(#{"1, 2"})` where the string itself may contain multiple comma-separated values.)

proposal/calc-functions.md
nex3: That's right. Everything just falls through to that unless explicitly specified otherwise, because there are a bunch of cases that can only be handled in the browser because they require rendering knowledge to resolve.

proposal/calc-functions.md
Goodwine: yup, I jsut realized that until the very end 😅

proposal/calc-functions.md
```diff
@@ -145,6 +188,43 @@ Add the following production:
...
+
+* Let `c1` and `c2` be the result of [matching units] for `n1` and `n2` allowing
+  unitless.
+
+* **If `c2` is infinity and has a different sign than `c2` (including
+  oppositely-signed zero), return NaN with the same units as `c1`.**
+
+  > This matches the behavior of CSS's `mod()` function.
+
+* Let `remainder` be a number whose value is the result of `remainder(c1.value,
```
Goodwine, line 213: I think this should say, "Let `result` be a number", no? Or where is `result` defined?

(reading the current spec, it looks like that may be an existing bug?)

proposal/calc-functions.md
```diff
@@ -222,15 +303,20 @@ This algorithm takes a calculation `calc` and returns a number or a calculation.
...
-      is negative or vice versa, set `result` to `mod - dividend`.
+        [compatible]: ../spec/types/number.md#compatible-units
 
-    * Return `result`.
+      * If `calc`'s name is `"rem"`, and if `dividend` is positive and `modulus`
+        is negative or vice versa:
+
+        * If `modulus` is infinite, return `dividend`.
+        * If `result` [exactly equals] 0, return `-result`.
+        * Otherwise, return `result - dividend`.
```
Goodwine, line 317: I see I found this confusing before (the subtraction) but that's because of my lack of understanding of how % works when the signs are opposite (and because dart `%` behaves differently I think)

IIUC:
```
70 % 20 = 10
-70 % 20 = 10
70 % -20 = -10
-70 % -20 = -10
```
Is that correct?

Goodwine (approved):
there's one nit and one question but otherwise LGTM

proposal/calc-functions.md
nex3: Updated `result` references to be `remainder`.

proposal/calc-functions.md
nex3: That's right. In Sass and CSS, the sign of the result of modulo is always the same as the sign of the right-hand side.

Issue #2848, opened 2020-04-27
Styles produced by the `use`d modules appear out of order related to the comments


emirotin:
- [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/ordered-comments.md) (https://github.com/sass/sass/pull/3548)
- [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1939) (https://github.com/sass/dart-sass/pull/1989)
- [x] [Tests](https://github.com/sass/sass-spec/issues/1901) (https://github.com/sass/sass-spec/pull/1919)

---

# Intro

I've encountered a problem when `use`ing a library, specifically MDC.

Before going into details I should mention that:
- I understand that `use` is intended for "no side effects" modules — to only get the exported variables, mixins, etc., not the yielded styles,
- I understand that `use` statements must go first and are meant to resemble the programming language static imports.

Given that I have a valid use-case which I describe in the end of this description.

---

# The issue

The CSS yielded by the `use`d module is output at the beginning of the file even before any multiline comments which may precede the `use` line.

Repro: https://github.com/emirotin/sass-comments-issue (run `npm install && npm run build`).

---

# My use-case

I'm using MDC following the recommended way to use its styles:

```
@use "@material/theme" with (
  $primary: #9c27b0,
  $secondary: #76ff03,
  $background: #fff,
);
@use "material-components-web";
```

From https://material.io/develop/web/docs/theming/

`material-components-web` package does produce lots of styles with of course lots of CSS to handle edge cases in various browsers.

I have stylelint enabled and it has a lot of warnings for these styles. I only care about the linting for my own styles, as well as I cannot influence the MDC styles so I want it wrapped in stylelint-disabled/enabled comments.

Apparently it's not possible. I think I can workaround it with `@import` (it causes a different, unrelated issue at the moment) but I think keeping the preceding comments can be desired in other cases as well.

Awjin:
I assume you're seeing the following input—
```
/* stylelint-disable */
@use "foo";
/* stylelint-enable */
```
emit the following output—
```
foo { ... }
/* stylelint-disable */
/* stylelint-enable */
```
This is intended behavior, since the CSS from a dependency must be emitted first. It would be really complicated to add an exception for comments that come before an `@use` rule.

> I have stylelint enabled and it has a lot of warnings for these styles. I only care about the linting for my own styles, as well as I cannot influence the MDC styles so I want it wrapped in stylelint-disabled/enabled comments.

[...]

PR #3548, opened 2023-04-17
[Ordered Comments] Add a proposal

nex3:
See #2848

### DIFF
--- a/proposal/ordered-comments.md
+++ b/proposal/ordered-comments.md
@@ -90,7 +90,7 @@ a {
 ```scss
 // In linked order
 /* before @use */
-/* before @forward
+/* before @forward */
 a {
   b: c;
 }
--- a/proposal/ordered-comments.md
+++ b/proposal/ordered-comments.md
@@ -158,6 +158,8 @@ Replace the line of [the `@import` semantics] that reads:
 * Add an `@import` with the evaluated modifiers to [the current module]'s
   CSS AST.
 
+  [the current module]: ../spec/spec.md#current-module
+
 with one that reads:
 
 * Add an `@import` with the evaluated modifiers to [the current module]'s CSS
--- a/proposal/ordered-comments.md
+++ b/proposal/ordered-comments.md
@@ -131,12 +131,12 @@ definition of the traversing procedure with the following:
     statements in `domestic`'s CSS tree that contains only comments and
     `@import` rules *and* that ends with an `@import` rule.
 
-  * Insert a copy of `initial-imports` in `css` after the last `@import` rule,
-    or after the longest initial subsequence of comments in `css` if it doesn't
-    contain any `@import` rules.
+  * Insert a copy of `initial-imports` in `css` after the longest initial
+    subsequence of comments and `@import` rules in `css`.
 
-    > If there are no comments in `css`, this initial subsequence is empty and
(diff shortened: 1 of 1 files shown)


proposal/ordered-comments.md
```diff
@@ -0,0 +1,170 @@
...
+  * Let `initial-imports` be the longest initial subsequence of top-level
+    statements in `domestic`'s CSS tree that contains only comments and
+    `@import` rules *and* that ends with an `@import` rule.
+
+  * Insert a copy of `initial-imports` in `css` after the last `@import` rule,
+    or after the longest initial subsequence of comments in `css` if it doesn't
+    contain any `@import` rules.
+
+    > If there are no comments in `css`, this initial subsequence is empty and
+    > `initial-imports` is inserted at the beginning of `css`.
```
stof: This note is confusing to me. After reading this bullet point 3 times, I finally understood that the note only refers to the second half of the bullet point (after the `or`). The first 2 reads had me starring at it thinking that it was totally wrong due to `@import` rules. Maybe there is a way to clarify it.

proposal/ordered-comments.md
```diff
@@ -0,0 +1,170 @@
...
+    * Traverse `upstream`.
+
+    > Because this traverses modules depth-first, it emits CSS in reverse
+    > topological order.
+
+  * Let `initial-imports` be the longest initial subsequence of top-level
+    statements in `domestic`'s CSS tree that contains only comments and
+    `@import` rules *and* that ends with an `@import` rule.
+
+  * Insert a copy of `initial-imports` in `css` after the last `@import` rule,
```
stof: what happens if there are some comments after that last `@import` rule ? should we insert the copy before it (as currently specified) or after it (which would be consistent with what happens the first time we need to insert imports) ?

nex3: Good point, we should probably just insert after the longest initial subsequence of comments and `@import`s.

proposal/ordered-comments.md
```diff
@@ -0,0 +1,170 @@
...
+      a copy of that comment to `css` and then mark it.
+
+    * Traverse `upstream`.
+
+    > Because this traverses modules depth-first, it emits CSS in reverse
+    > topological order.
+
+  * Let `initial-imports` be the longest initial subsequence of top-level
+    statements in `domestic`'s CSS tree that contains only comments and
+    `@import` rules *and* that ends with an `@import` rule.
```
stof, line 132: Given that we take the long initial subsequence of comments and `@import` rules in `css` in the next step when looking for the insertion point, should we enforce ending with an `@import` here ?

nex3: I think we should put this below comments, because otherwise if the entrypoint file contains no `@import`s they'll end up going before _everything_ in that file, including potentially a license comment, which wouldn't be ideal.

proposal/ordered-comments.md
```diff
@@ -0,0 +1,170 @@
...
+  b: c;
+}
+
+/* before @forward */
+```
+
+```scss
+// In linked order
+/* before @use */
+/* before @forward
```
jathak:
```suggestion
/* before @forward */
```

jathak (approved):

PR #3552, opened 2023-04-20
Use literate programming for JS API specs

nex3:


jathak (approved):
Is the plan to do something similar for the user-facing documentation in js-api-doc, or is that not feasible since we need the actual documentation to be in comments so it's accessible to TypeDoc?

nex3:
My plan is to leave js-api-doc as-is, for the reason you describe. It also follows a much more typical pattern of "documenting specific types" rather than the spec-style pattern of "specifying behavior which is sometimes but not always attached to types".

jgerigmeyer:
@nex3 I think this PR (specifically the removal of `spec/js-api/index.d.ts`) is breaking CI currently on dart-sass and sass-spec repos:

- https://github.com/sass/sass-spec/actions/runs/4824362331
- https://github.com/sass/dart-sass/actions/runs/4824310028

nex3:
Oops, forgot to land corresponding updates elsewhere. I bet it's also breaking sass-site.

nex3:
@jgerigmeyer Should be fixed now, let me know if it's not.

jgerigmeyer:
@nex3 That worked -- thanks!

PR #3574, opened 2023-05-17
Pull the embedded protocol definition into the spec repo

nex3:


CHANGELOG.md
jathak, line 1: Are this and VERSION supposed to be here given EMBEDDED_PROTOCOL_CHANGELOG and EMBEDDED_PROTOCOL_VERSION are a thing?

nex3: Nope!

embedded_sass.proto
jathak, line 1: This is also duplicated here and in `spec/`

jathak (approved):

Issue #3579, opened 2023-04-21
Breaking Change Proposal: Move compilation_id out of the protobuf to improve performance in multi-thread/multi-isolate context


ntkme:
* [x] Embedded protocol (https://github.com/sass/sass/pull/3599)
* [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1980) (https://github.com/sass/dart-sass/pull/1981)
* [x] [Node.js Embedded Host](https://github.com/sass/embedded-host-node/issues/223) (https://github.com/sass/embedded-host-node/pull/224)

---

Today we have the wire data encoded as:

```
[ varint proto_length ][ proto ]
```

This is a propose for a breaking change to remove `compilation_id` field inside proto, and replace it with a top level `session_id` like below to serve the same purpose:

```
[ varint session_id ][ varint proto_length ][ proto ]
```

The motivation is to improve I/O performance in a multi-thread context, by offloading message encoding and decoding to different CPU cores running different threads/isolates to reduce the pressure on the main isolate/thread due to potential heavy I/O multiplexing on stdio. With this change, both host, and compiler would be able to pass raw proto to a thread/isolate that is currently working on a specific compilation, and the decoding of the proto can happen on the child thread's CPU cycle, giving the main thread more free CPU cycle to purely process multiplexed I/O.

`session_id` should be in the range of `uint32`, the reason to choose `varint` as its encoding instead is:

1. For most of cases it would be shorter
2. Avoid the possible confusion on endianness of uint32 on wire.

ntkme:
More specifically, these can be removed and replaced with a single `session_id` on wire:

- CompileRequest.id
- VersionRequest.id
- ProtocolError.id
- CompileResponse.id
- VersionResponse.id
- LogEvent.compilation_id
- CanonicalizeRequest.compilation_id
- ImportRequest.compilation_id
- FileImportRequest.compilation_id
- FunctionCallRequest.compilation_id

These should remain, to reserve the potential of a compiler implementation sending multiple of same kind concurrently in a single compilation:
- CanonicalizeRequest.id
- ImportRequest.id
- FileImportRequest.id
- FunctionCallRequest.id
- CanonicalizeResponse.id
- ImportResponse.id
- FileImportResponse.id
- FunctionCallResponse.id

nex3:
[...]

Issue #3577, opened 2023-01-24
Make loaded_files available for failed compilations


nex3:
For hosts implementing filesystem watcher APIs, it's important to know which files to watch when determining when to retry a compilation. The `CompileSuccess.loaded_urls` field exposes this information, but it's not available for failed compilations which leaves hosts unable to determine what to watch if the initial compilation fails. We should move it to `CompileResult.loaded_urls` (although for backwards compatibility we should leave `CompileSuccess.loaded_urls` as well).

OlaviSau:
Adding context here, since this seems to be the correct place / upcoming discussion.

nex3:
> I think the best solution here would probably be to move loaded_urls from CompileSuccess to CompileResult, so it's available even when compilation fails. But in the interim, you can also define a custom loader that acts as a filesystem loader while also recording all the files that get touched in the compilation process. See [the spec](https://github.com/sass/sass/blob/main/spec/modules.md#resolving-a-file-url) for details on how filesystem loading should work.

my response:
> @nex3 I like both of your ideas, but I actually tried the custom loader and found that it only uses the loader when it can't find the file using normal means, thus it does not produce the full file list. If that was changed and it would be possible to provide files manually - it could offer other benefits as well.

> The reason being is that watch without control over loading is inefficient; on each compile - all the files are loaded again on each compile, even though most changes only involve a single file. I am not sure how much this really affects performance with modern SSD's but it's definitely a possible improvement area.

OlaviSau:
I'll add a bit more info - I tried both Importer and FileImporter and I even debugged a bit on the sass host and it does not even receive a message about files that can be resolved without assistance. The intent seems to be that the importer is only for non standard imports.

This behavior seems to match the spec as well, notice how importers trigger after the normal resolution did not return null. https://github.com/sass/sass/blob/main/spec/modules.md#loading-a-source-file

ntkme:
> We should move it to `CompileResult.loaded_urls` (although for backwards compatibility we should leave `CompileSuccess.loaded_urls` as well).

As this protocol is internal between host and compiler and not visible to end users, making breaking changes only impacts a very small group of developers and does not impact end users. Therefore, I'd rather just take the breaking change and go to protocol 2.0 if necessary.

nex3:
> I like both of your ideas, but I actually tried the custom loader and found that it only uses the loader when it can't find the file using normal means, thus it does not produce the full file list. If that was changed and it would be possible to provide files manually - it could offer other benefits as well.

[...]

Issue #3575, opened 2021-01-22
Use proto3 optional fields


stof:
proto3 has experimental support for `optional` on scalar fields as of 3.12. In the current 3.14 release, this is still guarded by the `--experimental_allow_proto3_optional`. However, all code generators I checked (all the ones listed in [the protobuf readme](https://github.com/protocolbuffers/protobuf#protobuf-runtime-installation) plus the [swift one](https://github.com/apple/swift-protobuf)) have already implemented support for the feature. So there is a chance that the feature gets unguarded.

If the feature becomes available unguarded soon enough compared to the stable release, it might make sense to use that feature for optional fields (for instance `LogEvent.stack_trace`)

nex3:
I'll talk to the protobuf team and see if I can get a sense of how likely this is to become stable, and how quickly that might happen.

stof:
As of protoc 3.15.0, the feature is available without an experimental flag.

nex3:
I think we can do this backwards-compatibly, since optional fields are encoded the same way on the wire as the "default values" that we currently use to signal that fields don't exist. We probably want to explicitly specify that both endpoints should treat "default values" the same way they treat absent values until we make a breaking version release, though.

PR #3582, opened 2023-05-19
[Embedded Protocol 2] Add proposal

nex3:
See #3579, #3575, #3577

### DIFF
--- a/proposal/embedded-protocol-2.md
+++ b/proposal/embedded-protocol-2.md
@@ -151,7 +151,7 @@ Replace the paragraph that beings "Implementations must guarantee that they use
 a unique `id` for every request" with:
 
 Each endpoint must guarantee that each request's `id` doesn't match the `id` of
-any other outstanding with the same [compilation ID] request from that endpoint.
+any other outstanding request with the same [compilation ID] from that endpoint.
 The same `id` may be used for an inbound request and an outbound request, and
 the same `id` may be used for two requests with different compilation IDs. The
 host must similarly guarantee that a `CompileRequest`'s compilation ID doesn't
--- a/proposal/embedded-protocol-2.md
+++ b/proposal/embedded-protocol-2.md
@@ -53,6 +53,46 @@ This proposal makes three breaking changes to the embedded Sass protocol:
 * Move `CompileSuccess.loaded_urls` to `CompileResult.loaded_urls` so it's
   available even when compilation fails.
 
+### Design Decisions
+
+#### Length Before Compilation ID
+
+This proposal includes the compilation ID for each request _after_ the length,
+and includes the size of the compilation ID in the total length.
+
+Another possible approach would be to have the compilation ID first, then the
+length, and have the length list only the length of the protocol buffer itself.
+This approach would have the benefit of being somewhat easier to decode, since
+the only length-delimited chunk of input would be parsed as a unit by the
+protocol buffer parser rather than needing to be sliced out of the (compilation
+ID, protocol buffer) pair.
+
+However, the embedded protocol is intended to be transport-independent, and not
+all transports will necessarily require an explicit length encoding at all.
+Although today all use of the embedded protocol is over stdin/stdout which
+requires a length delimiter, it's likely that we'll eventually add WASM support
+as well which won't.
+
+We want to make it easy for the same compiler and host codebases to support
+multiple transports. Making the length-delimiting process a layer that can be
+transparently applied to the same message blobs that are used in other transport
+layers makes it easier for the underlying endpoint logic to just work with those
+blobs regardless of where they came from.
+
+For languages that can parse protocol buffers from a subsequence of binary
+content without copying, it should be easy to slice off the compilation ID and
+parse the remaining binary data. For those that can't, the true length of the
+protocol buffer can be determined by taking the given length and subtracting the
+length of the compilation ID, which is given by the following table:
+
+| compilation ID          | length in bytes |
+|-------------------------|-----------------|
+| [0, 128)                | 1               |
+| [128, 16384)            | 2               |
+| [16384, 2097152)        | 3               |
+| [2097152, 268435456)    | 4               |
+| [268435456, 4294967296) | 5               |
+
 ## Overview
 
 ### Packet Structure
--- a/proposal/embedded-protocol-2.md
+++ b/proposal/embedded-protocol-2.md
@@ -111,11 +111,12 @@ Replace the paragraph that beings "Implementations must guarantee that they use
 a unique `id` for every request" with:
 
 Each endpoint must guarantee that each request's `id` doesn't match the `id` of
-any other outstanding request from that endpoint, although the same `id` may be
-used for an inbound request and an outbound request. The host must similarly
-guarantee that a `CompileRequest`'s [compilation ID] doesn't match the
-compilation ID of any other outstanding `CompileRequest`. The compiler must
-ensure that all outbound requests' compilation IDs match that of the
+any other outstanding with the same [compilation ID] request from that endpoint.
+The same `id` may be used for an inbound request and an outbound request, and
+the same `id` may be used for two requests with different compilation IDs. The
+host must similarly guarantee that a `CompileRequest`'s compilation ID doesn't
+match the compilation ID of any other outstanding `CompileRequest`. The compiler
+must ensure that all outbound requests' compilation IDs match that of the
 `CompileRequest` that triggered its associated compilation.
 
 The compilation ID 0 is reserved for `VersionRequest` and `VersionResponse`,
--- a/proposal/embedded-protocol-2.md
+++ b/proposal/embedded-protocol-2.md
@@ -8,6 +8,8 @@
 
 * [Background](#background)
 * [Summary](#summary)
+  * [Design Decisions](#design-decisions)
+    * [Length Before Compilation ID](#length-before-compilation-id)
 * [Overview](#overview)
   * [Packet Structure](#packet-structure)
 * [RPCs](#rpcs)
--- a/proposal/embedded-protocol-2.md
+++ b/proposal/embedded-protocol-2.md
@@ -54,6 +54,8 @@ This proposal makes three breaking changes to the embedded Sass protocol:
 
 ## Overview
 
+### Packet Structure
+
 Replace the last paragraph of the [embedded protocol overview] with:
 
 [embedded protocol overview]: ../spec/embedded-protocol.md#overview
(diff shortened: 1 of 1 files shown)


ntkme:
I think it is more ideal if the length on the wire is just the length of the proto message. Otherwise the logic of parsing a stream will get overly complicated due to need to subtract the length of varint id (most stream varint read implementation only returns the decoded varint and do not know about the wire length of the varint). I also suggest switch the order of id and length, so that the length clearly means the remaining length to be read.

E.g.

Current:

```
length = readvarint
id = readvarint
varintlength = encodevarint(id).length
proto = read(length - varintlength)
```

Suggestion:

```
id = readvarint
length = readvarint
proto = read(length)
```

proposal/embedded-protocol-2.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+## RPCs
+
+### Type Definitions
+
+Replace the following RPC type definitions:
+
+* *Requests* usually include a mandatory `uint32 id` field so that the other
+  endpoint can respond, except for `CompilationRequest` which uses the
```
Goodwine: Is `CompilationRequest` supposed to be `CompileRequest`? (If so, same below with `CompilationResponse`)

Goodwine: The text on these both bullet points mention that `Compile{Request,Response}` use the compilation ID as its ID. However these fields will be removed from the proto message. I get that this ID is being bubbled up to the wire packet, but these paragraphs sound like they refer to the proto message.

The paragraphs also start saying that "usually" messages will include a mandatory `id`. I wonder if maybe these messages would be better worded as something like:

> - *Requests* must include a mandatory `uint32 id`, except for `CompileRequest` which uses the wire packet compilation ID as its own ID.

(or something better, but along those lines 🤔)

proposal/embedded-protocol-2.md
```diff
@@ -0,0 +1,168 @@
...
+[varint]: https://developers.google.com/protocol-buffers/docs/encoding#varints
+[the protocol buffer documentation]: https://developers.google.com/protocol-buffers/docs/techniques#streaming
+
+For a length-delimited stream, each packet has the following structure:
+
+```
+╔══════════╦══════════════════╗
+║ varint   ║ Length           ║
+╠══════════╬══════════════════╣
+║ varint   ║ Compilation ID   ║
```
Goodwine, line 128: +1 to @ntkme's comment about swapping the order of these 2 numbers

Goodwine (changes requested):

proposal/embedded-protocol-2.md
ntkme: For compiler side requests like FunctionCallRequest etc. In theory, it is possible to have more than 1 requests inflight for the same compilation therefore we need id to track unique requests inside a single compilation.

nex3:
The problem with swapping length and compilation ID is that the length isn't guaranteed to be present. The protocol is notionally transport-independent, and the length is only specified _if_ the transport itself doesn't have built-in message boundaries. If you were running the embedded protocol over websocket, for example, you wouldn't want a length. For now this is academic since the only compiler implementation only supports stdin/stdout, but it's very possible it'll become relevant once Dart's WASM support matures. In a case like that, you really want the host and compiler to be able to implement the logic of the protocol independent of the transport, and just get fed sliced up blobs that they can then decode.

proposal/embedded-protocol-2.md
nex3: The idea is that this is referring to "compilation ID" as a term-of-art which was defined in the overview. To make that clearer, I'll add a "Packet Structure" subheading and link it from the term.

ntkme:
Let’s say we just want a logically length varint delimited stream, the problem is on the efficiency of decoding:

The example I gave above is one way to implement it without buffer copy, the alternative implementation would be even more problematic:

```
length = readvarint
payload = read(length)
id = readvarint(payload)
varintlength = encodevarint(id).length
proto = payload.slice(varintlength)
```

The last operation can be a very expansive buffer copy if the language does not support virtual buffer view wrapper (or pointer offset manipulation like C language).

ntkme:
If we want to keep the order of [total_length, id, proto], I think we should use a fixed length type for id e.g. uint32 in big endian, a.k.a. network byte order, which will be easier to parse.

nex3:
Yes, this does lean on the ability to pass a lightweight slice of a binary buffer to protobuf decoding (or determining the length of a varint to determine the true protobuf length), but I'd argue that's already the case since we need to extract the protobuf after the length varint anyway. And the alternative would mean substantially complicating the interface for actually handling the compilation ID beyond the existing complexity inherent in pulling it out from the protobuf.

> If we want to keep the order of [total_length, id, proto], I think we should use a fixed length type for id e.g. uint32 in big endian, a.k.a. network byte order, and that will solve all the performance issues.

I think it would be pretty confusing to have two header ints that use totally different encodings. If the benefit of this is just to make it easy to determine the size of the ID on the wire, I don't think that's really a huge issue; it's pretty straightforward to just do a lookup for varint length:

| value                   | length in bytes |
|-------------------------|-----------------|
| [0, 128)                | 1               |
| [128, 16384)            | 2               |
| [16384, 2097152)        | 3               |
| [2097152, 268435456)    | 4               |
| [268435456, 4294967296) | 5               |

ntkme:
> it's pretty straightforward to just do a lookup for varint length

That's fair.

proposal/embedded-protocol-2.md
```diff
@@ -0,0 +1,173 @@
...
+Replace the paragraph that beings "Implementations must guarantee that they use
+a unique `id` for every request" with:
+
+Each endpoint must guarantee that each request's `id` doesn't match the `id` of
+any other outstanding request from that endpoint, although the same `id` may be
+used for an inbound request and an outbound request. The host must similarly
+guarantee that a `CompileRequest`'s [compilation ID] doesn't match the
+compilation ID of any other outstanding `CompileRequest`. The compiler must
+ensure that all outbound requests' compilation IDs match that of the
+`CompileRequest` that triggered its associated compilation.
```
ntkme: We should clarify the difference between compilation id and request id inside a compilation.

It is ok for compiler side requests to have overlapping id as long as they belongs to different compilations, and it would be unnecessarily difficult to synchronize different workers to make sure they don't user overlapping id. 

E.g.

Compilation 1 -> FunctionCallRequest 1 
Compilation 2 -> FunctionCallRequest 1

Even if FunctionCallRequest share the same id, we can route them correctly because the routing is based on compilation id.

ntkme: Of course, in a single threaded implementation like what we have today, making the request id under all compilations completely unique is not a problem at all. It's just that we should remove this constraint when we move to a multi-threaded model.

ntkme: In other worlds, in a single threaded implementation we have today, the outstanding request on compiler side can be tracked via tuple of `(compilation_id, request_id)` instead of just `request_id`, therefore the `request_id` does not need to be unique across different compilations, it just need to be unique within its own compilation.

Again, no need to change current implementation right now. This is just about relax a constraint in the protocol that is not necessary.

Goodwine (approved):

ntkme (approved):
LGTM. @nex3 Thanks!

proposal/embedded-protocol-2.md
```diff
@@ -0,0 +1,174 @@
...
+  which uses the compilation ID as its ID. All response message types begin with
+  the corresponding request name and end with `Response`.
+
+### ID Requirements
+
+Replace the paragraph that beings "Implementations must guarantee that they use
+a unique `id` for every request" with:
+
+Each endpoint must guarantee that each request's `id` doesn't match the `id` of
+any other outstanding with the same [compilation ID] request from that endpoint.
```
ntkme:
```suggestion
any other outstanding request with the same [compilation ID] from that endpoint.
```

ntkme (changes requested):

nex3:
@Goodwine I added a section documenting the packet order design decision, can you take a quick look to make sure it makes sense?

proposal/embedded-protocol-2.md
```diff
@@ -53,6 +53,46 @@ This proposal makes three breaking changes to the embedded Sass protocol:
 * Move `CompileSuccess.loaded_urls` to `CompileResult.loaded_urls` so it's
   available even when compilation fails.
 
+### Design Decisions
+
+#### Length Before Compilation ID
+
+This proposal includes the compilation ID for each request _after_ the length,
+and includes the size of the compilation ID in the total length.
```
Goodwine: Nit:  although it makes sense, I found the use of "includes" slightly confusing, perhaps it's my lack of familiarity with it in this context. I wonder if a different word such as "declares" or "defines" may make this clearer 🤔

Goodwine (approved):
only a minor comment, but otherwise LGTM!

ntkme:
I spent some time trying to implement this change on the host side and found that reserving id 0 for VersionRequest would make implementation way more complex, with an extra dedicated tracker for multiple requests sharing id 0.

My suggestion would be that allow any id to be used for VersionRequest or CompileRequest (or any kind of host request in the future) as long as they are unique when inflight. For example:

```
Host Thread 0 <-> CompileRequest id 0 <-> Compiler Thread 0
Host Thread 1 <-> VersionRequest id 1 <-> Compiler Thread 1
Host Thread 2 <-> CompileRequest id 2 <-> Compiler Thread 2
Host Thread 3 <-> VersionRequest id 3 <-> Compiler Thread 3
```

In other words, on the compiler side the logic would changes from the current proposal like:

```
// main thread
if (id == 0) {
  // handle version request
} else {
  // send to worker based on unique id
}

// worker thread
{
  // handle compile request
}
```

To:

```
// main thread
{
  // send to worker based on unique id
}

// worker thread
switch (request) {
  case CompileRequest:
    // handle compile request
  case VersionRequest:
    // handle version request
}
```

The benefit here is that it would greatly simplify the logic on host side that we don't need to do anything special for id 0 as all host requests are treated equally. 

@nex3 @Goodwine Any thoughts?

nex3:
My concern there is that we might want to add an incoming request in the future that _must_ be handled by the compiler's main thread. For example, we might want to add something relating to reporting information about or managing the rest of the workers. If we don't have a way to easily tell that a request is intended for the main thread, we'd need to parse into each request to determine if it's something that needs to be handled locally or something that can be handed off, which obviates most of the use of a wire ID in the first place.

ntkme:
I thought about the same about handling special request on main thread. I think we can say id 0 is reserved as a “hint” to compiler that this request is “preferred” (if compiler supports multi-threading) to be handled in main thread (some request type may only be valid when using request id 0 in the future). Only one request with id 0 can be sent at a time (as main thread will be blocked anyways), which would be the same as any other id.

With the model I stated above, we can also allow CompilerRequest with id 0 (indicating that it is preferred to be run on main thread instead of in worker thread, but we may facing the waitFor issue again...) - this would have a little bit performance benefit for avoid setting up isolate in the current implementation of node-embedded-host which create a new compiler process every single time.

nex3:
If ID zero is only a hint, then what happens if we _do_ get a request with a non-zero ID that has to be handled in the main thread?

ntkme:
> If ID zero is only a hint, then what happens if we do get a request with a non-zero ID that has to be handled in the main thread?

That would be a protocol error back to host in my opinion.

nex3:
If we make it a protocol error _sometimes_, I think that's effectively the same as requiring the host to always send wire ID 0 for the requests that actually must run on the main thread, which just puts us right back where we started.

ntkme:
I realized that the complexity can be worked around on the host side by using unique request ids for wire id 0 and then it just need two different trackers but both can use the same implementation.

I’m good with this proposal.

By the way it’s been a week that we can mark this as accepted.

PR #3605, opened 2023-06-07
Add JS type defs for calculations

jerivas:
https://github.com/sass/dart-sass/pull/1988
https://github.com/sass/sass-spec/pull/1918


js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * The set of possible operators in a Sass calculation.
+ */
+type CalculationOperator = '+' | '-' | '*' | '/';
+
+/**
+ * A binary operation that can appear in a {@link SassCalculation}.
+ */
+export class CalculationOperation implements ValueObject {
```
jerivas, line 94: @nex3 In the accepted spec, `CalculationOperation` and `CalculationInterpolation` are abstract classes, but it seems like we want to actually allow JS users to instantiate them. I removed the `abstract` modifier here but would like to get confirmation on that.

nex3: Yeah, they should be concrete. Can you update the proposal as well, and add a note to its changelog?

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,141 @@
+import {List, ValueObject} from 'immutable';
+import {Value, SassNumber, SassString} from './index';
+
+/** The type of values that can be arguments to a SassCalculation. */
+type CalculationValue =
```
nex3: This should be exported, so other libraries can refer to it as necessary. It should also (along with any other top-level types here) be in `@category Custom Function`.

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,141 @@
...
+  ): SassCalculation;
+
+  /**
+   * Creates a value that represents `clamp(value, min, max)`.
+   *
+   * @throws `Error` if any of `value`, `min`, or `max` are or transitively
+   * contain a quoted {@link SassString}.
+   * @throws `Error` if `value` is undefined and `max` is not undefined.
+   * @throws `Error` if `value` or `max` is undefined and neither `min` nor
+     `value` is a {@link SassString} that begins with `var(`
```
nex3: I think this is actually wrong in the spec. It should allow any string or interpolation to be considered as multiple arguments, because it _could_ contain a comma (e.g. `clamp(#{"1, 2, 3"})`)

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,141 @@
...
+  get name(): string;
+
+  /** Returns a list of internal's `arguments` */
+  get arguments(): List<CalculationValue>;
+}
+
+/**
+ * The set of possible operators in a Sass calculation.
+ */
+type CalculationOperator = '+' | '-' | '*' | '/';
```
nex3: This should also be exported

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,141 @@
...
+  /** Returns internal's `right` field. */
+  get right(): CalculationValue;
+
+  /** Whether internal is equal to `other.internal` in Sass. */
+  equals(other: CalculationOperation): boolean;
+
+  /** Returns the same number for any two `CalculationOperation`s that are equal
+   * according to `equals`.
+   */
+  hashCode(): number;
```
nex3: You don't actually need to document these separately; they're inherited from `ValueObject` anyway.

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,141 @@
...
+     @returns A calculation with the name `clamp` and `min`, `value`, and `max`
+     as it's arguments, excluding any arguments that are undefined.
+   */
+  static clamp(
+    min: CalculationValue,
+    value?: CalculationValue,
+    max?: CalculationValue
+  ): SassCalculation;
+
+  /** Returns internal's `name` field. */
```
nex3: The type definitions in `js-api-doc` are user-facing documentation, not specifications. Users don't know what "internal" is, they just think of this class as a representation of the calculation itself, and the documentation should reflect that.

spec/js-api/value/calculation.d.ts.md
```diff
@@ -1,7 +1,7 @@
-# JavaScript Calculation API: Draft 2
+# JavaScript Calculation API
```
nex3: The format of a specification document isn't the same as the format of a proposal. It doesn't need a tracking issue or changelog; it doesn't need a summary and background to explain what it's changing and why. It just needs to describe how the API should behave canonically. See the other value specifications in this directory for relevant templates.

spec/js-api/value/calculation.d.ts.md
```diff
@@ -75,18 +75,13 @@ function boundary rather than when a calculation is constructed.
 ```ts
 import {List, ValueObject} from 'immutable';
 
-import {Value, SassNumber, SassString} from '../spec/js-api/value';
+import {Value, SassNumber, SassString} from './index';
 ```
 
 ## Types
 
 ### `Value`
```
nex3: Similarly to the above: the proposal includes changes to _every_ type that gets updated, but the calculation specification should only define the `SassCalculation` itself and the types it uses.

jerivas:
@nex3 I think I addressed all your comments. This is ready for another round of review

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,138 @@
...
+export type CalculationOperator = '+' | '-' | '*' | '/';
+
+/**
+ * A binary operation that can appear in a {@link SassCalculation}.
+ * @category Custom Function
+ */
+export class CalculationOperation implements ValueObject {
+  /**
+   * Creates a Sass CalculationOperation by setting the fields to the arguments
+   * of the corresponding names, and returns it.
```
nex3: I think this still reads a bit spec-y. Maybe just "Creates a Sass CalculationOperation with the given [operator], [left], and [right] values."?

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * A string injected into a {@link SassCalculation} using interpolation.
+ * @category Custom Function
+ */
+export class CalculationInterpolation implements ValueObject {
+  /**
+   * Creates a Sass CalculationInterpolation by setting the value field to the
+   * value argument and returns it.
```
nex3: Similar comment to the constructor above.

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,138 @@
...
+  /** Returns the operation's `right` field. */
+  get right(): CalculationValue;
+
+  equals(other: CalculationOperation): boolean;
+
+  hashCode(): number;
+}
+
+/**
+ * A string injected into a {@link SassCalculation} using interpolation.
```
nex3: Might be worth mentioning here that interpolations are different from unquoted strings in that they're always surrounded in parentheses when they appear in `CalculationOperation`s.

nex3 (approved):

Issue #2739, opened 2019-08-30
Ship a built-in package importer


nex3:
- [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/package-importer.d.ts.md) (https://github.com/sass/sass/pull/3660)
- [x] Dart Sass (https://github.com/sass/dart-sass/pull/2130)
- [x] Embedded Host (https://github.com/sass/embedded-host-node/pull/260)
- [x] Tests (https://github.com/sass/sass-spec/pull/1952)
- [x] [Documentation](https://github.com/sass/sass-site/issues/938) (https://github.com/sass/sass-site/pull/944)
- [x] Update canonical spec (https://github.com/sass/sass/pull/3728)

---

There's currently no standard way for Sass users to import packages from dependencies. There are a number of domain-specific solutions, such as Webpack's [tilde-based imports](https://github.com/webpack-contrib/sass-loader#resolving-import-at-rules), `npm-sass`'s [load-path-like functionality](https://www.npmjs.com/package/npm-sass), or Dart Sass's support for Dart's `package:` URLs.

Most of these have the disadvantage of needing to be explicitly opted-into, and all of them make it difficult to share stylesheets across different contexts. For example, if package A depends on package B, how can A import B in a way that will work in both Webpack and in Dart?

This is the kind of situation where centralization is a boon. If we can build into the Sass language a notion of a "package import" that's flexible enough to work across contexts, we can make it usable by all stylesheets with confidence that even if they're used by different build runners or even ported to a different package manager their dependencies will continue to work.

## What does it look like?

The [Sass specification](https://github.com/sass/sass/blob/master/spec/at-rules/import.md#loading-an-import) talks about imports in terms of URLs. The current JS API deals with them as a combination of raw strings and filesystem paths, but I'd like to move away from that as part of #2509.

The current most popular solution for package imports is probably `sass-loader`'s, which passes any import that begins with `~` through Webpack's built-in resolution infrastructure. We could re-use this syntax, but it doesn't work well with URL semantics. A string beginning with a tilde is syntactically a relative URL, which means we'd need to check for the relative path `~package` first before passing the URL to the package importer.

We'd have a similar problem if we automatically added `node_modules` to the load path. Every instance of `@use "bootstrap"` would need to check the relative path first, as well as potentially every load path, before checking the package importer. It also makes package stylesheets less visually distinctive, which can be confusing for readers.

[...]

PR #3609, opened 2023-06-07
[Previous URL] Expand to embedded protocol and filesystem importers

nex3:
This updates the embedded protocol to support the proposal and adds
support for passing the previous URL to filesystem importers.

### DIFF
--- a/proposal/prev-url.d.ts.md
+++ b/proposal/prev-url.d.ts.md
@@ -17,10 +17,20 @@ import {PromiseOr} from '../spec/js-api/util/promise_or';
       * [Unavailable for Pre-Resolved Loads](#unavailable-for-pre-resolved-loads)
       * [Unavailable for Absolute Loads](#unavailable-for-absolute-loads)
 * [Types](#types)
+  * [`FileImporter`](#fileimporter)
+    * [`findFileUrl`](#findfileurl)
   * [`Importer`](#importer)
     * [`nonCanonicalScheme`](#noncanonicalscheme)
     * [`canonicalize`](#canonicalize)
 * [Embedded Protocol](#embedded-protocol)
+  * [`Importer`](#importer-1)
+    * [`non_canonical_scheme`](#non_canonical_scheme)
+  * [`CanonicalizeRequest`](#canonicalizerequest)
+    * [`containing_url`](#containing_url)
+  * [`CanonicalizeResponse`](#canonicalizeresponse)
+    * [`url`](#url)
+  * [`FileImportRequest`](#fileimportrequest)
+    * [`containing_url`](#containing_url-1)
 
 ## Background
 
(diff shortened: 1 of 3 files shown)


nex3:
@ntkme I'm interested in your feedback on this as well.

proposal/prev-url.d.ts.md
```diff
@@ -209,5 +253,67 @@ nonCanonicalScheme?: string | string[];
...
+## Embedded Protocol
+
+### `Importer`
+
+#### `non_canonical_scheme`
+
+The set of URL schemes that are considered *non-canonical* for this importer.
+This must be empty unless `importer.importer_id` is set.
+
+If any element of this contains a character other than a lowercase ASCII letter,
```
Goodwine, line 268: nitpicking: empty string should also throw an error

proposal/prev-url.d.ts.md
```diff
@@ -209,5 +253,67 @@ nonCanonicalScheme?: string | string[];
...
+```
+
+## Embedded Protocol
+
+### `Importer`
+
+#### `non_canonical_scheme`
+
+The set of URL schemes that are considered *non-canonical* for this importer.
+This must be empty unless `importer.importer_id` is set.
```
Goodwine, line 266: Can it be empty when `importer.importer_id` is set?

proposal/prev-url.d.ts.md
Goodwine, line 1: I realize it's not part of this PR, but I wonder if the TS type for `nonCanonicalScheme` should be:

```ts
  nonCanonicalScheme?: string | [string, ...string[]];
```

Just to capture that if an array is provided it will have at least 1 element

Goodwine (commented):
Is the idea for this proposal as a whole (ie not just this PR) that custom importers would be able to use custom schemas to emulate NodeJS behavior of using the closest `node_module`?

e.g. from the Issue, for a file: `a/node_modules/b/node_modules`, URLs loaded from `/a/` could have `schema1://` and files loaded from `/b/` could have `schema2://`? 

Otherwise, I'm slightly confused how the non-canonical schemes come into play after reading the summary 🤣😭 (the relative loads do make sense)

proposal/prev-url.d.ts.md
nex3: We do have well-defined semantics for an empty array (all schemes are considered canonical, just like `undefined`) so I don't think that adds much value. For example, consider a user who might want to assemble this array programmatically—it may indeed be more convenient for them to allow it to be empty.

proposal/prev-url.d.ts.md
nex3: Yes, it just means that no URL schemes are considered non-canonical.

nex3 (commented):
One use-case for non-canonical schemes is indeed to allow for https://github.com/sass/sass/issues/2739, where we have a `pkg:` URL scheme that can resolve differently based on the nearest `node_modules` location.

Goodwine (approved):

PR #3610, opened 2023-06-08
[Calc Functions] Ensure slash division works on min/max/abs/round

jathak:
This updates the definition of potentially slash-separated numbers to exclude operations involving a calculation that was previously a Sass global function.

In the process of updating this, I also changed the language of this section of the definition to refer to "each operand" rather than "both operands", as the definition as written implies that both operands have to be the same type (of numbers, calculations, and product expressions), rather than allowing each operand to be a different one of these three.


pamelalozano16 (approved):
lgtm :)

nex3 (approved):

PR #3611, opened 2023-06-09
[Deprecations API] Support embedded protocol

jathak:
This also fixes the actual spec text to make all invalid flag values warnings instead of errors (we updated this in the summary, but not the actual text)


accepted/deprecations-api.d.ts.md
```diff
@@ -514,3 +565,59 @@ export const deprecations: Deprecations;
...
+## Embedded Protocol
+
+### CompileRequest
+
+#### `fatal_deprecation`
+
+A set of deprecation IDs to treat as fatal.
+
+If a deprecation warning of any provided type is encountered during compilation,
+the compiler must respond with a `CompileFailure` instead.
```
nex3: Maybe "instead of a `CompileSuccess`"?

nex3 (approved):

PR #3619, opened 2023-06-20
Revert "Add JS type defs for calculations (#3605)"

nex3:
This reverts commit 39915044f65828173564751cfd93610a80284eaf. This
should only have been landed in parallel with the implementation and
tests.


nex3:
@jerivas Heads up, I merged your PR too early and had to revert because of course the types aren't implemented yet. Please re-open a PR to roll this forward and I'll be careful not to land it until the parallel PRs are ready as well :sweat_smile:.

jathak (approved):
Not sure if you want to do a partial revert or just fully revert and partially roll forward, but the proposal changes here should be fine; it's just the actual spec changes that need to wait

nex3:
Good call, un-reverted those.

jerivas:
> Please re-open a PR to roll this forward

@nex3 will do!

PR #3622, opened 2023-06-21
Add JS type definitions for calculations

jerivas:
A continuation of #3605 and #3619

Related to 
https://github.com/sass/dart-sass/pull/1988
https://github.com/sass/sass-spec/pull/1918


nex3 (approved):
Will hold off on merging until the implementation/test PRs are also in place :sweat_smile:

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,139 @@
...
+  /** Returns the operation's `operator` field. */
+  get operator(): CalculationOperator;
+
+  /** Returns the operation's `left` field. */
+  get left(): CalculationValue;
+
+  /** Returns the operation's `right` field. */
+  get right(): CalculationValue;
+
+  equals(other: CalculationOperation): boolean;
```
jerivas: @nex3 looking again at this, shouldn't this be `other: Value` or even `other: any`?

js-api-doc/value/calculation.d.ts
```diff
@@ -0,0 +1,139 @@
...
+  );
+
+  /** Returns the operation's `operator` field. */
+  get operator(): CalculationOperator;
+
+  /** Returns the operation's `left` field. */
+  get left(): CalculationValue;
+
+  /** Returns the operation's `right` field. */
+  get right(): CalculationValue;
```
jerivas, line 110: Curious as to why these are specified as getters instead of `readonly` properties?

nex3: We want to leave open the possibility for implementations to implement these using getters. Does TypeScript allow that for readonly properties? I don't actually know.

js-api-doc/value/calculation.d.ts
nex3: Yes, it probably should. It's pretty common for implementations to find small things like this that need to be updated in the proposal; feel free to lump those proposal fixes into this PR.

jerivas: Makes sense. This came up because the embedded host uses `other: Value`

js-api-doc/value/calculation.d.ts
jerivas: To test, I went with `readonly` in https://github.com/oddbird/embedded-host-node/blob/js-api-calculations/lib/src/value/calculations.ts#L113-L115 and tsc didn't complain, but I also see the value in making them getters. Will defer to you on this one

nex3: We use getters elsewhere so I think let's stick with them for consistency.

js-api-doc/value/calculation.d.ts
nex3: Actually, `other: Value` is wrong because `CalculationOperation` isn't a `Value`. I think `other: unknown` is what `Immutable.js` uses.

jerivas: Fixed in #3631

PR #3626, opened 2023-07-02
Proposal for First Class Mixins: Draft 1

### LINKED ISSUE #626: Dynamic mixin invocation

connorskees:
Builds on top of https://github.com/sass/sass/pull/3442 by addressing PR review and adding the function `meta.accepts-content`. 

Inspired in part by https://github.com/sass/dart-sass/issues/1958
Closes https://github.com/sass/sass/pull/3442
See #626


proposal/first-class-mixins.md
```diff
@@ -0,0 +1,199 @@
...
+}
+```
+
+> The internal representation of a mixin may also include function arguments, a body, a span, or other fields, but these properties are not meaningfully exposed to users.
+>
+> When a mixin value is created, it captures the current execution environment and a pointer to the mixin object being referenced. This is also not directly exposed to users.
+
+### Operations
+
+The only operation permitted for mixin objects is checking for equality.
```
nex3: This isn't entirely clear on what happens if you try to do an operation that's not permitted. If the intention is to throw an error for all other operations, say so explicitly.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,199 @@
...
+>
+> When a mixin value is created, it captures the current execution environment and a pointer to the mixin object being referenced. This is also not directly exposed to users.
+
+### Operations
+
+The only operation permitted for mixin objects is checking for equality. 
+
+#### Equality
+
+Two mixins are considered equal if they refer to the exact same mixin as declared in the source code.
```
nex3: What about mixins that aren't declared in source code, such as `meta.load-css()`?

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,199 @@
...
+  name: string;
+
+  /**
+   * Whether or not the body of this mixin contains the `@content` rule
+   */
+  hasContent: boolean;
+}
+```
+
+> The internal representation of a mixin may also include function arguments, a body, a span, or other fields, but these properties are not meaningfully exposed to users.
```
nex3: Style nit: please wrap lines at 80 characters.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,199 @@
...
+   * The name of this mixin
+   */
+  name: string;
+
+  /**
+   * Whether or not the body of this mixin contains the `@content` rule
+   */
+  hasContent: boolean;
+}
+```
```
nex3: Since the spec already has an (implicit) representation of mixins, and this proposal is worded so that "a mixin" as a module member is interchangeable with "a mixin" as a value object, it might make more sense to just say "the mixin value is now considered a Sass value type" rather than defining a separate type just for it.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Mixins
+
+### `meta.apply()`
+
+```
+meta.apply($mixin, $args...)
+```
+
+> Mixins must be invoked with `@include`, so `include` is guaranted to exist.
```
nex3: You need to indicate somewhere what `include` actually _refers to_. I'd say something like "If the current `@include` rule has a `ContentBlock`" in the second bullet point below, and move this note to that bullet point.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,199 @@
...
+
+* If `$mixin` is not a mixin, throw an error.
+
+* If `include` has a `ContentBlock`, it should be passed down to `$mixin`.
+
+* Assign $args to `$mixin`'s ArgumentDeclaration in `$mixin`'s scope.
```
nex3: Elsewhere we use the phrasing "execute an `ArgumentInvocation` with an `ArgumentDeclaration`". We don't have an `ArgumentInvocation` (since the args are already resolved by the time this is running), but you can create one and say "execute the `ArgumentInvocation` `(...$args)` with `$mixin`'s `ArgumentDeclaration` in `$mixin`'s scope."

(This line is also missing a few code quotes.)

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,199 @@
...
+
+* Execute each statement in `$mixin`.
+
+* Let `content` be the result of executing `$mixin` with `$args`.
+
+* Treat `content` as though it were the contents of this mixin.
```
nex3: I don't think you need these steps. Executing the mixin is already a mutating process.

nex3 (changes requested):
Thanks for taking a stab at this! This would be a very nice bread-and-butter feature to land.

js-api-doc/value/mixin.d.ts
```diff
@@ -0,0 +1,19 @@
...
+  /**
+   * Creates a new first-class mixin that can be invoked using
+   * [`meta.apply()`](https://sass-lang.com/documentation/modules/meta#apply).
+   *
+   * @param signature - The function signature, like you'd write for the
+   * [`@function rule`](https://sass-lang.com/documentation/at-rules/function).
+   * @param callback - The callback that's invoked when this function is called,
+   * just like for a {@link CustomFunction}.
+   */
+  constructor(signature: string, callback: (args: Value[]) => string);
```
connorskees: I'm not very familiar with the JS API, so this is more of a placeholder. I would argue that this function should return a Sass AST node, but I'm not sure if such functionality is currently exposed through the JS API. If it is not, I would propose that it be impossible for users to construct mixin objects through the JS API, at least as part of an initial implementation.

js-api-doc/value/index.d.ts
nex3, line 1: Proposals shouldn't directly update the existing JS API files or docs. They should open the modules and modify them from the proposal itself. You can see an example of this in action in the [calculation API proposal](https://github.com/sass/sass/blob/main/accepted/calculation-api.d.ts.md).

js-api-doc/value/mixin.d.ts
nex3: We don't want to make implementations beholden to exposing an entire Sass AST with a specific API. As far as the JS API and Embedded Protocol are concerned, a mixin should just be an opaque object that can't meaningfully be interacted with beyond returning it as-is. The JS constructor should be defined to just throw an error.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,197 @@
...
+
+The only operation permitted for mixin objects is checking for equality. All
+other operations throw an error.
+
+#### Equality
+
+Two user-declared mixins are considered equal if they refer to the exact same
+mixin as declared in the source code.
+
+Builtin mixins -- i.e. those declared by the Sass standard library, and not declared directly in Sass -- are equal to only themselves.
```
nex3: This is still potentially ambiguous. Does "equal only to itself" mean that separate calls to `meta.get-mixin()` return unequal values even if they refer to a mixin of the same name from the same module?

nex3 (changes requested):
This still needs updates for the embedded protocol.

proposal/first-class-mixins.d.ts.md
nex3, line 1: This should be in the same file as the main proposal. Since it's all Markdown anyway, you can just have the parts that need to define TypeScript do so and have the rest discuss the language feature.

spec/embedded-protocol.md
nex3, line 1: Again, this shouldn't directly update existing specs. Updating the specs comes _after_ the proposal has been accepted (see https://github.com/sass/sass/blob/main/CONTRIBUTING.md#process).

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,204 @@
...
+### `meta.apply()`
+
+```
+meta.apply($mixin, $args...)
+```
+
+* If `$mixin` is not a mixin, throw an error.
+
+* If the current `@include` rule has a `ContentBlock`, it should be passed down
+  to `$mixin`. Mixins must be invoked with `@include`, so `include` is
```
nex3: `include` isn't defined here. It's probably easiest to just say "so the current include rule" instead.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,204 @@
...
+  guaranteed to exist.
+
+* Create an `ArgumentInvocation` from `(...$args)`
+
+* Execute the `ArgumentInvocation` `(...$args)` with `$mixin`'s `ArgumentDeclaration` in `$mixin`'s scope
```
nex3: Nit: line length

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,204 @@
...
+meta.apply($mixin, $args...)
+```
+
+* If `$mixin` is not a mixin, throw an error.
+
+* If the current `@include` rule has a `ContentBlock`, it should be passed down
+  to `$mixin`. Mixins must be invoked with `@include`, so `include` is
+  guaranteed to exist.
+
+* Create an `ArgumentInvocation` from `(...$args)`
```
nex3: Nit: missing period here and below.

nex3: You probably don't need this line, since you're already saying "the `ArgumentInvocation` `(...$args)`" below.

nex3 (changes requested):

spec/embedded-protocol.md
connorskees: I did read this document, but misinterpreted some parts of it. I've added `JavaScript API` and `Embedded Protocol` sections to the main proposal document.

connorskees:
I will add that I am comfortable volunteering to drive implementation should this proposal be accepted.

proposal/first-class-mixins.md
nex3, line 1: Now that this contains TS as well, it should be `.d.ts.md` so that the TypeScript files will get generated and typechecked.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,292 @@
...
+
+* `meta.get-function()` => `meta.get-mixin()`
+
+* `meta.module-functions()` => `meta.module-mixins()`
+
+* `meta.call()` => `meta.apply()`
+
+## Types
+
+This proposal promotes the [mixin value][] to a Sass value type.
```
nex3: Style nit:

```suggestion
This proposal promotes the [mixin value] to a Sass value type.
```

Although it's not currently consistent in this repository, for new Markdown we omit the optional `[]` for references.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,292 @@
...
+
+This is a new function in the `sass:meta` module.
+
+```
+meta.accepts-content($mixin)
+```
+
+* If `$mixin` is not a mixin, throw an error.
+
+* Return a boolean which is true if the body of the mixin has an `@content`
```
nex3: Nit:

```suggestion
* Return a boolean which is true if the body of `$mixin` has an `@content`
```

Just a little clearer what the referent is.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,292 @@
...
+
+```
+meta.apply($mixin, $args...)
+```
+
+* If `$mixin` is not a mixin, throw an error.
+
+* If the current `@include` rule has a `ContentBlock`, it should be passed down
+  to `$mixin`. Mixins must be invoked with `@include`, so the current include
+  rule is guaranteed to exist.
```
nex3: "passed down to" is a little vague here... since the existing semantics for `@content` blocks are so intimately tied to a specific `@include` rule, maybe a better way to frame this is by saying in the "Execute each" line below something like "Treat the `@include` rule that invoked `meta.apply` as the `@include` rule that invoked `$mixin`" with a non-normative note indicating that this ensures that any `@content` rules in `$mixin` will use `meta.apply()`'s `ContentBlock`.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Mixins
+
+### `meta.apply()`
+
+```
+meta.apply($mixin, $args...)
+```
+
+* If `$mixin` is not a mixin, throw an error.
```
nex3: This should also throw an error if the `@include` rule that invoked `meta.apply()` has a `ContentBlock` but `$mixin`'s body doesn't contain `@content`.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,292 @@
...
+### Design Decisions
+
+Mixins differ from functions in that the result of their execution is a Sass AST
+node, and not a SassScript value. Sass today does not expose ways to create or
+manipulate AST nodes through the JavaScript API, nor does it intend to do so in
+the future.
+
+For this reason, it is not meaningful -- or even possible -- to construct or 
+execute a mixin through the JavaScript API. A mixin object shall be opaque, and
+the only operation available shall be to return the object as-is.
```
nex3: This should be a subheading of `## Summary`.

proposal/first-class-mixins.md
```diff
@@ -0,0 +1,292 @@
...
+
+### API
```
nex3: You don't need this header, this can just be directly under `## JavaScript API`.

nex3 (changes requested):
Some minor change requests, but this is generally looking really good!

proposal/first-class-mixins.md
connorskees: Implemented [here](https://github.com/sass/sass/pull/3626/commits/abbfecbce5b43608144ca8b37976f06fdec53763#diff-646077cb818e2f24ca0be3af54b487a8254fff79c51d33181a2e6c5483792f88R66). Since these are design decisions only about the JavaScript API, I updated the header and moved it under the toplevel `## Summary`.

I wasn't entirely sure if you meant that I should add a summary subheader to the JavaScript API header as well, or if the JavaScript API header should _also_ be moved under the main summary header. Please let me know if you had other thoughts here.

nex3 (approved):

PR #3630, opened 2023-07-06
[Calc Functions] Match the spec's behavior better for percentages

nex3:
More functions disallow percentages in CSS than our spec did, so this
updates the spec to match.

### DIFF
--- a/accepted/calc-functions.md
+++ b/accepted/calc-functions.md
@@ -300,8 +300,8 @@ This algorithm takes a calculation `calc` and returns a number or a calculation.
 
   [`sass:math`]: ../spec/built-in-modules/math.md
 
-  > Other than `round()`, all of these functions will throw errors if their
-  > arguments have any units.
+  > The `sass:math` functions will check units here for the functions that
+  > require specific or no units.
 
 * If `calc`'s name is `"abs"` and `arguments` contains exactly a single number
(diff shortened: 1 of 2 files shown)


nex3:
cc @pamelalozano16

jathak (approved):

accepted/calc-functions.md
```diff
@@ -294,16 +294,25 @@ This algorithm takes a calculation `calc` and returns a number or a calculation.
   > otherwise calculations' arguments must match the expected number.
 
 * If `calc`'s name is `"sin"`, `"cos"`, `"tan"`, `"asin"`, `"acos"`, `"atan"`,
```
pamelalozano16, line 296: I believe functions `sin`, `cos`, and `tan` should accept units (maybe only angle units)? E.g cos math function: 
```
final _cos = _function(
    "cos",
    r"$number",
    (arguments) => SassNumber(math.cos(arguments[0]
        .assertNumber("number")
        .coerceValueToUnit("rad", "number"))));
```
So `a {b: cos(1grad)}` would result in `b: 0.9998766325;`

nex3: Good point, I'll call those out in the note below.

jathak (approved):

PR #3660, opened 2023-08-18
Proposal: Package Importer and built-in Node implementation

jamesnw:
See #2739


proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+## Summary
+
+> This section is non-normative.
+
+Sass users often need to use styles from a dependency to customize an existing
+theme or access styling utilities.
+
+This proposal defines a `pkg:` URL scheme for usage with `@use` that directs an
+implementation to resolve a URL within a dependency. The implementation will
+resolve the dependency URL using the standard resolution for that environment.
```
nex3: From the conversation I had with @ntkme in https://github.com/sass/sass/issues/2739, I think this wording is a bit misleading. Maybe something more like "using a well-known or custom dependency resolution algorithm"?

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+Once resolved, this URL will be loaded in the same way as any other `file:` URL.
+
+This proposal also defines a built-in Node importer.
+
+For example, `@use "pkg:bootstrap";` would resolve to the path of a
+library-defined export within the `bootstrap` dependency. In Node, that would be
```
nex3:
```suggestion
library-defined export within the `bootstrap` dependency. In Node, that could be
```

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+
+### Node built-in importer
+
+The built-in Node importer resolves in the following order:
+
+1. `sass` condition in package.json `exports`.
+
+2. `style` condition in package.json `exports`.
+
+3. If no subpath, then find root export:
```
nex3: Nit: Can you make this a full sentence?

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+
+  1. `sass` key at package.json root.
+
+  2. `style` key at package.json root.
+
+  3. `index` file at package root, resolved for file extensions and partials.
+
+4. If there is a subpath, resolve that path relative to the package root, and
+   resolve for file extensions and partials
```
nex3: Nit: trailing period

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+
+For library creators, the recommended method is to add a `sass` conditional
+export key as the first key in `package.json`.
```
nex3: Do we need to recommend that this be the _first_ key?

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+fetches. If we instead require the browser version to have a fully resolved URL,
+we negate many of this spec's benefits. Users may write their own custom
+importers to fit their needs.
+
+[resolving a file: url]: ../spec/modules.md#resolving-a-file-url
+
+#### Available as an opt-in setting
+
+The `pkg:` import loader will be exposed through an opt-in setting as it adds the
+potential for file system interaction to `compileString` and
```
nex3: Nit: I'd say "unexpected file system interaction" here, since it's already possible to interact with the filesystem if you configure it that way.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+`compileStringAsync`. Specifically, we want people who invoke Sass compilation
+functions to have control over what files get accessed, and there's even a risk
+of leaking file contents in error messages.
+
+#### Available in legacy API
+
+The built-in Node Package Importer will be added to the legacy API in order to
+reduce the barrier to adoption.
```
nex3: It's probably also worth mentioning that part of the reasoning here is that we expect it to be relatively easy to implement.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+we will not apply that logic to the destination, and will expect library authors
+to map the export to the correct place. In other words, given a `package.json`
+with `exports` as below, The Node package importer will resolve a
+`@use "pkg:pkgName/variables";` to the destination of the `_variables.scss` export.
+
+```json
+{
+  "_variables.scss": {
+    "sass": "./src/sass/_variables.scss"
+  }
```
nex3: Shouldn't this be under an `"exports"` key?

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+```ts
+declare module '../spec/js-api/options' {
+  interface Options<sync extends 'sync' | 'async'> {
+    /**
+     * Specifies which, if any, built-in package importer to resolve any URL
+     * with the `pkg:` scheme. The `node` importer follows Node resolution logic.
+     *
+     * @defaultValue `undefined`
+     * @category Input
+     */
```
nex3: You don't need any documentation comments here. The APIs defined here are purely specification; we keep a separate copy of the APIs with doc comments in the [`js-api-doc` directory](https://github.com/sass/sass/tree/main/js-api-doc).

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+```
+
+## Semantics
+
+### Package Importers
+
+This proposal defines the requirements for Package Importers written by users or
+provided by implementations. It is a type of [Importer] and, in addition to the
+standard requirements for importers, it must handle only the following
+non-canonical URLs:
```
nex3: It's not entirely clear that _all_ of these requirements must be met (instead of _any_ of them). Consider wording it like

> it must handle only non-canonical URLs that:
>
> * have the scheme `pkg`, and

and so on.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+* A URL whose path begins with `/`.
+* A URL with non-empty/null username, password, host, port, query, or fragment.
+
+
+[importer]: ../spec/modules.md#importer
+
+### Node Package Importer
+
+The Node package importer is an [importer] with an associated absolute `pkg:`
+URL. When the Node package importer is invoked with a string named `string` and
```
nex3: What's the associated `pkg:` URL doing here?

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+a [previous URL] `previousURL`:
+
+* Let `url` be the result of [parsing `string` as a URL][parsing a URL]. If this
+  returns a failure, throw that failure.
```
nex3, line 325: Frustratingly, if you don't pass a base URL to the WHATWG URL parsing algorithm, it will throw an error for any relative URLs. (You can test this by running `new URL("foo")` in your browser console.) You can work around this by just saying "if `string` is a relative URL, return null".

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+
+* If `url`'s scheme is not `pkg:`, return null.
+
+* Let `resolved` be the result of [resolving a `pkg:` URL] with `url` and
```
nex3: Just saying "resolving a `pkg:` URL" here is kind of confusing, because it doesn't make it clear that it's using the Node algorithm specifically. Maybe add "as Node" on the end.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+* Let `syntax` be:
+
+  * "scss" if `resolved` ends in `.scss`.
+
+  * "indented" if `resolved` ends in `.sass`.
+
+  * "css" if `resolved` ends in `.css`.
+
+  > The algorithm for [resolving a `file:` URL](../spec/modules.md#resolving-a-file-url)
+  > guarantees that `resolved` will have one of these extensions.
```
nex3: I think it's really the Node algorithm that guarantees this, since it's the one that ultimately throws an error if the extension isn't `.scss`, `.sass`, or `.css`.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+[parsing a URL]: https://url.spec.whatwg.org/#concept-url-parser
+[resolving a `pkg:` URL]: #node-algorithm-for-resolving-a-pkg-url
+
+
+## Procedures
+
+### Node Algorithm for Resolving a `pkg:` URL
+
+This algorithm takes a URL with scheme `pkg:` named `url`, and an optional URL
+`previousURL`. It returns a canonical file path or null.
```
nex3:
```suggestion
`previousURL`. It returns a canonical `file:` URL or null.
```

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+* Let `packageRoot` be the result of [resolving the root directory for a
+  package] with `packageName` and `previousURL`.
+
+* If a `package.json` file does not exist at `packageRoot`, throw an error.
+
+* Let `packageManifest` be the result of parsing the `package.json` file at
+  `packageRoot` as [JSON].
+
+* Let `resolved` be the result of [resolving package exports] with `sass` as the
+  condition, `packageRoot`, `subpath`, and `packageManifest`.
```
nex3: Nit: be consistent about whether to camel-case `subPath`.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+* Otherwise, return the first URL path segment.
+
+### Resolving the root directory for a package
+
+This algorithm takes a string, `packageName`, and an optional absolute URL
+`previousURL`, and returns an absolute URL to the root directory for the most
+proximate installed `packageName`.
+
+* Let `parentURL` be the value of `previousURL` if it is defined, otherwise the
+  URL to the current directory.
```
nex3: Building in an unconfigurable notion of a global "current directory" worries me a bit. In Node, do we really want this to be the process's working directory, or do we want it to be `require.main.filename`? The latter is what Node actually uses to do its resolution.

I think a better option here is to define the Node.js package importer as taking a base `file:` URL as a parameter, and then define the JS API to pass `require.main.filename` as that parameter. (We can potentially add a way to explicitly configure it later if users request it.)

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+
+* Let `exports` be the value of `packageManifest.exports`.
+
+* If `exports` is undefined, return null.
+
+* Let `subpathVariants` be the result of [Export load paths] with `subpath`.
+
+* Let `resolvedPaths` be a list of the results of calling
+  `PACKAGE_EXPORTS_RESOLVE(packageRoot, subpathVariant, exports, [condition])`
+  as defined in the [Node resolution algorithm specification, with each
```
nex3: Nit:

```suggestion
  as defined in the [Node resolution algorithm specification], with each
```

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+  `subpathIndexVariants` as `subpathVariant`.
+
+* If `resolvedIndexPaths` contains more than one resolved URL, throw an error.
+
+* If `resolvedIndexPaths` contains exactly one resolved URL, return it.
+
+* Return null.
+
+> Where possible in Node, it can use [resolve.exports] which exposes the Node
```
nex3: "it" -> "implementations"

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,484 @@
...
+This algorithm takes a string `packagePath`, which is the root directory for a
+package, and `packageManifest`, which is the contents of that package's
+`package.json` file, and returns a file URL.
+
+* Let `sassValue` be the value of `sass` in `packageManifest`.
+
+* If `sassValue` is a relative path with an extension of `sass`, `scss` or
+  `css`:
+
+  * If `sassValue` starts with `./`, remove that substring.
```
nex3: Rather than explicitly removing this, I'd make the bullet point below "Return the canonicalized `file:` URL". This ensures that we canonicalize away any stray `/./`s or whatever that appear in `sassValue` itself as well.

nex3 (changes requested):
Looking pretty good! Mostly just a collection of nits, with only a few logic suggestions.

Also, I just added a Markdown linter to main, so I recommend merging that in and running `npm run fix`.

proposal/package-importer.d.ts.md
jamesnw: The main point of this sentence is to convey that each importer will do what makes most sense for that environment, and not necessarily to say how that decision will be made. You're correct that this is more expansive than what is currently conveyed. I went with- 

> The implementation will resolve the dependency URL using the resolution standards and conventions for that environment.

Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: When Node resolves conditional exports, it will always resolve with `default`, `import/require` and `node` conditions *in addition to* any custom conditions, and will resolve the first entry in the object with a condition that matches. Because of this, `sass` will always need to come before other conditions. However, my wording was not clear, so I've updated it-


> For library creators, the recommended method is to add a `sass` conditional
> export to `package.json`, with the `sass` key being the first listed condition for the
> exported file path.

Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Replaced with this, which I think helps reiterate some of the concepts and distinctions.

> The Node Package Importer is the built-in implementation of a [Package Importer]
using the standards and conventions of the Node ecosystem.

Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: I updated this to-

  > The algorithm for [resolving a `pkg:` URL as Node] guarantees that
  > `resolved` will have one of these extensions.

While Node itself will throw errors, it's the "Node Algorithm for Resolving a `pkg:` URL" that is mediating those errors and guaranteeing it. For example, given `pkg:package/foo`, Node will throw errors for all the non-matching partials and extensions. 

Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1-

> The built-in Node Package Importer will be added to the legacy API in order to
> reduce the barrier to adoption. While the legacy API is deprecated, we
> anticipate the implementation to be straightforward.

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Removed in 5014ab1. Should I also add the documented API to `js-api-doc` as part of this PR?

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: Addressed in 5014ab1

proposal/package-importer.d.ts.md
jamesnw: I agree, this seemed like a magic value appearing. I like your suggestion, and added this in d2eb4f9. I'd welcome your feedback on if I need to be more specific on using `require.main.filename`.

jamesnw (commented):
Thanks @nex3  for your feedback. I've applied the suggestions, and where I added new content, added it to the response so you can more easily track.

proposal/package-importer.d.ts.md
jamesnw: Confirmed in conversation that this documentation will come as part of a later step.

proposal/package-importer.d.ts.md
nex3: I think part of what's confusing about this is the term "The implementation", which implies that there will be one canonical implementation of `pkg:` imports for each Sass API. Maybe soften that to "Sass interfaces may provide one or more implementations that ..."?

proposal/package-importer.d.ts.md
nex3: Of those, only `"default"` should affect _our_ algorithm, right? Looking at [the Node resolution spec](https://nodejs.org/api/esm.html#resolution-algorithm-specification), it looks like the `"node"` and `"import"` conditions are only respected through the `ESM_RESOLVE` and `PACKAGE_SELF_RESOLVE` algorithms, which I _believe_ can't actually be reached from the way we invoke `PACKAGE_RESOLVE`. Maybe we can just make this "listed before `"default"`"?

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,463 @@
...
+
+For example, `@use "pkg:bootstrap";` would resolve to the path of a
+library-defined export within the `bootstrap` dependency. In Node, that could be
+resolved within `node_modules`, using the [Node resolution algorithm].
+
+[node resolution algorithm]: https://nodejs.org/api/packages.html
+
+### Node built-in importer
+
+The built-in Node importer resolves in the following order:
```
nex3, line 70: We should document that this _does_ use `default` (although also mention that relying on this is discouraged).

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,463 @@
...
+This proposal defines the requirements for Package Importers written by users or
+provided by implementations. It is a type of [Importer] and, in addition to the
+standard requirements for importers, it must handle only non-canonical URLs that:
+
+* have the scheme `pkg`, and
+* whose path begins with a package name, and
+* optionally followed by a path, with path segments separated with a forward
+  slash.
+
+The package name will often be the first path segment, but the importer should
```
nex3: Nit: I'd change "should" to "may" here.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,463 @@
...
+take into account any conventions in the environment. For instance, Node
+supports scoped package names, which start with `@` followed by 2 path segments.
```
nex3: Maybe add a note here that package names which aren't purely alphanumeric may be less portable across different package importers.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,463 @@
...
+
+Package Importers will reject the following patterns:
```
nex3: Nit: "will" -> "must"

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,463 @@
...
+
+* A URL whose path begins with `/`.
+* A URL with non-empty/null username, password, host, port, query, or fragment.
+
+[importer]: ../spec/modules.md#importer
+
+### Node Package Importer
+
+The Node Package Importer is the built-in implementation of a [Package Importer]
```
nex3: "the built-in" -> "an", just to make it a little clearer that this isn't mandated for all Sass implementations, only the ones that provide JS APIs.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,463 @@
...
+
+### Node Algorithm for Resolving a `pkg:` URL
+
+This algorithm takes a URL with scheme `pkg:` named `url`, and an optional URL
+`baseURL`. It returns a canonical `file:` URL or null.
+
+* Let `fullPath` be `url`'s path.
+
+* Let `packageName` be the result of [resolving a package name] with `fullPath`,
+  and `subpath` be the path without the `packageName`.
```
nex3: "the path" -> "`fullPath`"

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,463 @@
...
+* Let `packageManifest` be the result of parsing the `package.json` file at
+  `packageRoot` as [JSON].
+
+* Let `resolved` be the result of [resolving package exports] with `sass` as the
+  condition, `packageRoot`, `subpath`, and `packageManifest`.
+
+  * If `resolved` has the scheme `file:` and an extension of `sass`, `scss` or
+    `css`, return it.
+
+  * Otherwise, if `resolved` is not null, throw an error.
```
nex3: Nit: don't indent these, since they're not within a conditional block. Same below.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,463 @@
...
+[Package Importer]: #package-importers
+[parsing a URL]: https://url.spec.whatwg.org/#concept-url-parser
+[resolving a `pkg:` URL as Node]: #node-algorithm-for-resolving-a-pkg-url
+
+## Procedures
+
+### Node Algorithm for Resolving a `pkg:` URL
+
+This algorithm takes a URL with scheme `pkg:` named `url`, and an optional URL
+`baseURL`. It returns a canonical `file:` URL or null.
```
nex3: I don't think `baseURL` is optional here anymore.

proposal/package-importer.d.ts.md
nex3: I do think this needs to be more specific. The algorithm itself shouldn't specify what its argument will be; the point of an argument is that it's set by the callers that _invoke_ the algorithm. You also can't just add a new algorithm to invoking an importer, because "an importer" is already a well-defined interface [specified here](https://github.com/sass/sass/blob/main/spec/modules.md#importer).

Instead, you need an additional layer that _creates_ the Node resolution importer given an entrypoint URL. (Think of it like a function that returns a closure in JavaScript.) Then in the Node resolution algorithm itself, you can invoke resolving a `pkg:` URL as Node "with `baseURL` set to the canonical URL of the [current source file](https://github.com/sass/sass/blob/main/spec/spec.md#current-source-file) that contained the load if it's a `file:` URL, and `entrypointURL` otherwise". (I'm stealing the verbiage from [the prev-url spec](https://github.com/sass/sass/blob/main/accepted/prev-url.d.ts.md).)

You'll also want to be more specific in the section that defines `pkgImporter`. It's still written like user-facing documentation, but it needs to specify how to construct the Node resolution importer (specifically by passing in `require.main.filename`) _and_ where it goes in the list of importers (probably after user-defined importers but before load paths).

nex3 (changes requested):
One other thing I just realized: we should add a Node resolution option to the embedded protocol as well, or else embedded hosts will have to reimplement it by hand. It shouldn't be more complicated than an option with an enum value with a single element, though.

proposal/package-importer.d.ts.md
jamesnw: This is a great question. When possible, we'll be using `resolve.exports`, which always adds the 

There's a fair amount of conversation [here](https://github.com/lukeed/resolve.exports/issues/12), which boils down to wanting to adhere to the entirety of the node spec, which specifies:

- default
- import or require (The resolution spec we are referring to is specifically the ECMAScript Module, not CommonJS, so in our case it is import)
- node

I think we have a few options-

1. Explicitly pass `node`, `import`, and `default` in our algorithm to match the `resolve.exports` functionality 
2. Use `resolve.exports`'s `unsafe` mode which would only pass `default` in addition to our custom condition. This would break from default Node.js behavior
3. Adopt a simpler algorithm that would not follow Node.js behavior at all, but do a more direct lookup of `exports.[path].sass` (and many recursive variants). 

While I'm tempted by #3 in order to make things a bit simpler for users, I think the best decision is #1, as it best supports the ecosystem standards and conventions.

proposal/package-importer.d.ts.md
jamesnw: Resolved in fc72b46

proposal/package-importer.d.ts.md
jamesnw: Resolved in fc72b46

proposal/package-importer.d.ts.md
jamesnw: Resolved in fc72b46

proposal/package-importer.d.ts.md
jamesnw: Resolved in fc72b46

proposal/package-importer.d.ts.md
jamesnw: Resolved in fc72b46

jamesnw (commented):
I still have some work on Resolving the Package Root and specifying how to create the Node resolution importer, but I addressed other comments and added the embedded protocol.

proposal/package-importer.d.ts.md
jamesnw: Resolved in 7651924

proposal/package-importer.d.ts.md
jamesnw: Resolved in 7651924

proposal/package-importer.d.ts.md
jamesnw: Resolved in 7651924

proposal/package-importer.d.ts.md
nex3: As discussed offline, we can't use `resolve.exports` here directly because we'll need to implement the algorithm in pure Dart anyway (and all current implementations will be able to use that implementation directly). Also, since the Node spec itself doesn't require the use of `"node"` or `"import"` keys, we aren't bound by spec considerations to use them, so we'll only respect `"default"`, `"sass"`, and `"style"`.

proposal/package-importer.d.ts.md
jamesnw: Addressed in 7f8ec04

proposal/package-importer.d.ts.md
jamesnw: I updated this in 7f8ec04. Notably, I switched from a 2 step process that checked for `sass` and `style` separately to one that checks for both simultaneously. This addresses a case where the `style` and `default` conditions are set.

proposal/package-importer.d.ts.md
jamesnw: I updated this in 0c7ef34. I followed the pattern from loadPaths generating File Importers, adding the concept of an "associated" value. Instead of adding a separate instantiation procedure, I have the `pkgImporter` procedure creating a Node Package Importer with an associated `entryPointURL`. Let me know if this works.

I also added the ordering in the global imports list as part of the generic Package Imports requirements.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,502 @@
...
+
+After the first bullet points in [`compile`] and [`compileString`] in the
+Javascript Compile API, insert:
+
+* If `options.pkgImporter` equals `'node'`:
+
+  * Let `pkgImporter` be a [Node Package Importer] with an associated
+    `entryPointURL` of `require.main.filename`.
+
+  * Append `pkgImporter` to the `options.importers`.
```
jamesnw: @nex3 Here I'm mutating `options.importers` directly which feels incorrect from a programming standpoint, but sufficiently clear from a specification standpoint. Creating an intermediary combined list is possible, but would be a much more intrusive patch, but let me know if that would be preferable.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,507 @@
...
+
+[`compile`]: ../spec/js-api/compile.d.ts.md#compile
+[`compileString`]: ../spec/js-api/compile.d.ts.md#compilestring
+[Node Package Importer]: #node-package-importer
+
+```ts
+declare module '../spec/js-api/options' {
+  interface Options<sync extends 'sync' | 'async'> {
+    pkgImporter?: 'node';
+  }
```
ntkme: My concern is that this design is a limited to a single pkg importer. E.g. In ruby implementation it is likely I want to ship two `pkgImporter`s, one for `node` and one for `gem`. Also, I'd like to be able to enable multiple pkg importers with user defined order.

If we are going to mutate `options.importers` anyways, we can ship a `NodePkgImporter` stub singleton class, which mocks the importer interface, but does not contain actual implementation. This way users can directly put `NodePkgImporter` into `options.importers` array in the desired position. When we see `NodePkgImporter` as we process the `options.importers`, we can internally swap the stub class to an actual implementation, or for embedded-sass we can swap that with the embedded protocol constant.

For example in ruby:

``` ruby
Sass.compile_string(
  '@import "pkg:bootstrap";',
   importers: [Sass::NodePkgImporter, Sass::GemPkgImporter, ...other_importers]
)
```

@nex3 What do you think?

nex3: The main question I have here is, what's the use-case for having multiple package importers that coexist in the same compilation? Is it realistic that users would want to try loading from `node_modules` and then fall back to RubyGems?

ntkme: It is probably rare in the real world usage, but it is possible for a certain package to be only available in one package manager or another, that in some cases user will have to mix two pkg importers to load different packages.

nex3: All right, I think that's reasonable.

@jamesnw Can you update this to change the modern API from using a flag to providing an opaque `nodePkgImporter` object? [This stackoverflow issue](https://stackoverflow.com/questions/56737033/how-to-define-an-opaque-type-in-typescript) provides a suggestion for how to define such a typing. The actual object itself can just be a `new Object()` that we compare with `===`.

jamesnw: I've implemented that here- [6cdf76a](https://github.com/sass/sass/pull/3660/commits/6cdf76a3d2c268ef2c5be53790213af6bf5ed668)

It currently is not valid Typescript, as [Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces) doesn't work if non-function types are not unique or the same type. Here, we are needing to completely _replace_ the type of `Options.importers` to add the new NodePackageImporter type. 

The error it's showing is `Subsequent property declarations must have the same type.` I'll continue digging, but if there's an example of where we've replaced a type elsewhere, that would be helpful.

jamesnw: @nex3 After some investigation, I think our 2 best options are-

1. Introduce a new option called something better than `importers2` that allows for Importer, FileImporter, or NodePackageImporter. 
2. Decide to overwrite rather than merge the new definition. This would be a non-breaking, additive change, but still require mutating the original spec directly.

nex3: Ah, we've finally reached the limits on TypeScript's ability to do what we want. Let's call it `importers_new_` to distinguish it, I suppose.

nex3 (approved):

proposal/package-importer.d.ts.md
jamesnw: Updated in https://github.com/sass/sass/pull/3660/commits/827855a369e213d07d32978b4679dd199103537a

This should resolve the Static Analysis and TypeDoc build failures in the CI. The Link Referents failure is for a link to  https://runtime-keys.proposal.wintercg.org/#adding-a-key that is valid, but appears to be blocking requests from Github. Unless there's some workaround, it may be easiest to remove the link.

ntkme: I wonder for the JS API why cannot it just be an Importer type on the public API?

e.g.

``` js
export const nodePkgImporter = {
  canonicalize: function () {
    throw 'This is only a stub that is not meant to be invoked by user';
  },
  load: function () {
    throw 'This is only a stub that is not meant to be invoked by user';
  }
}
```

``` js
if (importer === sass.nodePkgImporter) {
  // enable the internal implementation
}
```

ntkme:
One thing that might be missing in the embedded protocol is the current working directory. I suppose we want to lookup `node_modules` starting from the directory containing the input file. However if the input is a string that does not have a URL, the lookup should start from the working directory. As for a persisted and reused compiler started previously, the host may have changed its directory and thus not having the same working directory. So I think we need to pass the working directory to compiler via embedded protocol as a context information for nodePkgImporter.

jamesnw:
> One thing that might be missing in the embedded protocol is the current working directory. I suppose we want to lookup `node_modules` starting from the directory containing the input file. However if the input is a string that does not have a URL, the lookup should start from the working directory. As for a persisted and reused compiler started previously, the host may have changed its directory and thus not having the same working directory. So I think we need to pass the working directory to compiler via embedded protocol as a context information for nodePkgImporter.

@ntkme Does the [Containing Url](https://github.com/sass/sass/blob/main/accepted/prev-url.d.ts.md) proposal address this need? I think has been accepted but not implemented.

proposal/package-importer.d.ts.md
jamesnw: I like this solution- it allows us to keep the simpler ‘importers’ option and doesn’t rely on the undocumented Typescript conventions for opaque objects. I implemented this in https://github.com/sass/sass/pull/3660/commits/943b0813de2ddf311979404a180665278593e4db - @nex3 does this work for you?

ntkme:
The containing url proposal was to provide the containing url of a import to custom importer functions (from compiler to host). Here we're considering providing the working directory for input that does not have a url (from host to compiler).

E.g. In node REPL, you can run `const sass = require('sass')` in a working directory that has `node_modules/sass`, because node would begin the lookup from the current working directory. So in sass, if we're running a string compilation that does not have a URL (filename) like `sass.compileString('@use "pkg:bootstrap";')`, it should to be treated similarly I suppose.

nex3:
I agree with @ntkme, we need to have some way to pass in an `entryPointURL` to the package importer.

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,527 @@
...
+* Return `paths`.
+
+[basename]: ../spec/modules.md#basename
+
+## Embedded Protocol
+
+```proto
+message CompileRequest {
+  message Importer {
+    PackageImporter packageImporter = 14;
```
nex3: This should be a new member of the `Importer.importer` oneof, since it's a new type of importer. Also, it doesn't need index 14; the index just needs to be unique within the `Importer` record.

jamesnw: Ah, right, I forgot to update to mirror the move in the JS API from a flag to a Importer object.

I updated this in https://github.com/sass/sass/pull/3660/commits/8658bc3e29f34bfbe206cac35b2f24c44ae90ec4. 

For the `entryPointUrl`, I added a NodePackageImporter message with a string `entry_point_url`. The other option I considered but rejected was flattening the object with a single key into-

```proto
    oneof importer {
      string node_package_importer_entry_point_url = 4;
    }
```

The nested message style I went with would allow for easier future additions, but I'm fine with either option (or something else).

@ntkme  Thanks for catching and explaining this- does this address the need?

ntkme:
Do we want to support loading from `NODE_PATH` or `GLOBAL_FOLDERS`? For example, loading a package installed globally like `npm install -g bootstrap`.

In node `NODE_PATH + GLOBAL_FOLDERS` are searched after local `node_modules`. A few implementation details in node:
- `NODE_PATH + GLOBAL_FOLDERS` can be retrieved from `require("module").globalPaths`.
- Mutating `NODE_PATH` during node runtime does not affect the load path, however mutating `require("module").globalPaths` does affect the load path.
    - This means if we want to be 100% compatible with node resolve we need to pass `require("module").globalPaths` in embedded protocol for each compilation to be accurate.

https://nodejs.org/api/modules.html#requireresolverequest-options
https://nodejs.org/api/modules.html#loading-from-the-global-folders
https://nodejs.org/api/modules.html#all-together
https://nodejs.org/api/esm.html#importmetaresolvespecifier

https://github.com/nodejs/node/blob/v20.0.0/lib/internal/modules/cjs/loader.js#L1397-L1425

@nex3 @jamesnw Any thoughts?

proposal/package-importer.d.ts.md
```diff
@@ -0,0 +1,529 @@
...
+#### Instantiating a `nodePackageImporter`
+
+After the first bullet points in [`compile`] and [`compileString`] in the
+Javascript Compile API, insert:
+
+* If any object in `options.importers` is exactly equal to the object
+  `nodePackageImporter`:
+
+  * Let `pkgImporter` be a [Node Package Importer] with an associated
+    `entryPointURL` of `require.main.filename`.
```
ntkme, line 260: When running node REPL `require.main` is `null`. Node has a hack internally to make the import work by adding the working directory as the local search path:

https://github.com/nodejs/node/blob/e7618fb5a5fc25d76b6474e2a6607f04fd6f10e0/lib/internal/modules/cjs/loader.js#L824-L832

I suppose we can set `entryPointURL` to `file://${process.cwd()}/` (node the slash at the end is important), which will give us enough information on the compiler side.

nex3:
My instinct is that trying to resolve from globally-installed packages will cause more headache than value for users. @alexander-akait what's the prior art here with sass-loader? Does it look for dependencies from `NODE_PATH` and `GLOBAL_FOLDERS`?

proposal/package-importer.d.ts.md
jamesnw: @ntkme I updated this for the Embedded Protocol in https://github.com/sass/sass/pull/3660/commits/3e3c5f3d9fe30035c6ee1d3bc4e3b2e40c762880. Thanks!

proposal/package-importer.d.ts.md
```diff
@@ -504,6 +504,16 @@ potential subpaths, resolving for partials and file extensions.
 
 ## Embedded Protocol
 
+An Importer that resolves `pkg:` URLs using the [node resolution algorithm]. It
+is instantiated with an associated `entry_point_url` set to the value of
+`file://${process.cwd()}/`.
```
nex3: This is confusing—if the `entry_point_url` is a field of the proto, isn't it up to the host to provide it rather than the spec?

jamesnw: Ah, yes. Does this need to be specified elsewhere for the host, or is that not part of the spec?

nex3: The JS host isn't part of the spec, except in that it has to follow the JS API spec.

jamesnw: Removed the host in [d14aba8](https://github.com/sass/sass/pull/3660/commits/d14aba830cef542511a58b8e1d492cc74b636bbb).

nex3:
Not hearing from @alexander-akait, I think the safest bet is to just not support global load paths for now. It'd be easier to add them later than it would be to remove them. @jamesnw can you add a design note documenting the rationale here?

jamesnw:
> Not hearing from @alexander-akait, I think the safest bet is to just not support global load paths for now. It'd be easier to add them later than it would be to remove them. @jamesnw can you add a design note documenting the rationale here?

I added this in https://github.com/sass/sass/pull/3660/commits/e733b7810ffcfd2f86565de147d3e86459e50a4f.

ECMAScript modules (used with `import`) do not support `NODE_PATH`s, but it is still supported (but discouraged) in CommonJS modules (used with `require()`). So I documented the decision to choose ECMAScript resolution, and noted no `NODE_PATH` as a result of that.

I suspect that Webpack would use CommonJS resolution for `sass-loader`, so this may be a surprise for some. But because Node docs discourage using `NODE_PATHS`  and there's a straightforward workaround with symlinks, I don't think we need to support it.

proposal/package-importer.d.ts.md
jamesnw: In talking with @nex3,  we decided to use the opaque option to make it more clear that it's not something a user can create. I also got clarification that the `importers_new_` was only for declaration merging in the spec, but on implementation, that would replace the existing `importers` option, and not introduce a new option to the API. Implemented in https://github.com/sass/sass/pull/3660/commits/c8fcd99b1d870041116b931618d996c9aafea0b3

proposal/package-importer.d.ts.md
jamesnw: I added a requirement to add it to a copy of `options.importers` in https://github.com/sass/sass/pull/3660/commits/c8fcd99b1d870041116b931618d996c9aafea0b3

jamesnw:
@nex3 I've made the changes we talked about yesterday, and this is ready for another review. Thanks!

PR #3661, opened 2023-08-18
[Calc Functions] Substantially refactor calculation parsing

nex3:
This allows calculation syntax to coexist with user-defined Sass
functions instead of needing every new calculation function to break
all existing Sass functions of the same name.

See #3504


accepted/calc-functions.md
```diff
@@ -276,6 +244,35 @@ Let `n1` and `n2` be two numbers. To determine `n1 % n2`:
 
 ## Procedures
 
+### Evaluating a `FunctionCall` as a Calculation
+
+This algorithm takes a [`FunctionCall`] `call` whose name is a plain identifier
```
Goodwine, line 248: The FunctionCall link in this line is broken

accepted/calc-functions.md
```diff
@@ -276,6 +244,35 @@ Let `n1` and `n2` be two numbers. To determine `n1 % n2`:
...
+and returns a number or a calculation.
+
+* If `call`'s `ArgumentInvocation` contains one or more `KeywordArg`s or one or
+  more `RestArg`s, throw an error.
```
Goodwine: Neither `KeywordArg` or `RestArg` seem to be defined 🤔 or at least I'm not able to find it on the spec nor here. I did find `KeywordArgument` in the `@use` spec, but I couldn't find the other anywhere.

Goodwine (changes requested):

accepted/calc-functions.md
```diff
@@ -511,93 +508,152 @@ This algorithm takes a calculation `calc` and returns a number or a calculation.
...
 
-To evaluate a `UnaryCalcExpression`, `BinaryCalcExpression`, `HypotExpression`,
-`CssMinMax`, `CssRound`, or `CssAbs`:
+[the semantics for `FunctionCall`]: ../spec/functions.md#functioncall
 
-* Let `name` be the lower-case value of the expression's first token without the
-  trailing parenthesis.
+* If `function` is null; `name` is case-insensitively equal to `"min"`, `"max"`,
+  `"round"`, or `"abs"`; and all arguments in `call`'s `ArgumentInvocation` are
+  [calculation-safe], return the result of evaluating `call` [as a calculation].
```
Goodwine, line 522: This `[as a calculation]` link is also broken. I also looked around and the `[compatible]` link is broken also, but that seems to have been an issue from a previous PR

Goodwine (changes requested):

accepted/calc-functions.md
nex3: The spec is created lazily, which means sometimes we refer to concepts that exist in theory but not in practice. That said, we should match the same name as the `@use` spec.

accepted/calc-functions.md
```diff
@@ -249,8 +248,10 @@ Let `n1` and `n2` be two numbers. To determine `n1 % n2`:
 This algorithm takes a [`FunctionCall`] `call` whose name is a plain identifier
 and returns a number or a calculation.
 
-* If `call`'s `ArgumentInvocation` contains one or more `KeywordArg`s or one or
-  more `RestArg`s, throw an error.
+[`FunctionCall`]: ../spec/functions.md#functioncall
+
+* If `call`'s `ArgumentInvocation` contains one or more `KeywordArgument`s or
+  one or more `RestArgument`s, throw an error.
```
Goodwine, line 254: my only nitpick is that this refers to `$foo...` as a `RestArgumant` where other places refer to this as an "argument list"

https://sass-lang.com/documentation/at-rules/mixin/#taking-arbitrary-arguments

That said, I did find a couple other places in the spec mentioning "rest arguments", so if this is OK then LGTM :)

Goodwine (approved):

accepted/calc-functions.md
stof: The argument list is a type of value passed to the function. The RestArgument is about describing the structure of the ArgumentInvocation

Issue #2831, opened 2020-03-05
Support CSS Color Level 4 hwb/lab/lch/gray functions (and related adjustments)


mirisuzanne:
- [x] [Proposal (Language)](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md)
- [x] [Proposal (JS API)](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces-js.d.ts.md)
- [x] [Embedded Protocol](https://github.com/sass/sass/issues/3576) (https://github.com/sass/sass/pull/3718)
- [x] Request for Comments (https://github.com/sass/sass-site/pull/671)
- **Implementation**
  - [x] [Dart Sass](https://github.com/sass/dart-sass/issues/1805) (https://github.com/sass/dart-sass/pull/2341)
  - [x] [Node.js Embedded Host](https://github.com/sass/embedded-host-node/issues/170) (https://github.com/sass/embedded-host-node/pull/330)
  - [x] [Tests](https://github.com/sass/sass-spec/issues/1828) (https://github.com/sass/sass-spec/pull/2013)
  - [x] [Documentation](https://github.com/sass/sass-site/issues/672) (https://github.com/sass/sass-site/pull/1115)
  - [x] Specification (https://github.com/sass/sass/pull/3930, ...)
  - [x] [Migrator](https://github.com/sass/migrator/issues/233) (https://github.com/sass/migrator/pull/260)
- **Breaking Changes**
  - [ ] [Dart Sass](https://github.com/sass/dart-sass/issues/1805)
  - [ ] [Tests](https://github.com/sass/sass-spec/issues/1828)
  - [ ] [Documentation](https://github.com/sass/sass-site/issues/672)

---

See [Color Level 4](https://www.w3.org/TR/css-color-4/)

Having access to CIE colors/adjustments would be particularly useful for design systems, and it seems like we could provide the syntax in `sass:color` without waiting on browsers to support the CSS variant.

When I say adjustments, I'm not referring to the new CSS color functions in Level 5. I think we need to wait for those to solidify. I'm thinking of something similar to, or extending, `color.adjust()`.

nex3:
I'd definitely like to expand the set of built-in color functions, like we did for math functions in #851. Miriam, are you interested in writing up a proposal?

mirisuzanne:
Yeah, I'm happy to. I see a few potential issues right away, and I'd love your thoughts:

[...]

PR #3665, opened 2023-08-21
Proposal: Add New Color Spaces to JS API

jamesnw:
See #2831


proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+```ts
+import {Value} from '../spec/js-api/value';
+```
+
+## Types
+
+### Color Space Definitions
+
+```ts
+type ColorSpaceRGB =
```
nex3: I think these types should probably be exported, so external TypeScript users can reference them when e.g. wrapping Sass's APIs.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+  | ColorSpaceHSL
+  | ColorSpaceXYZ
+  | ColorSpaceLCH
+  | ColorSpaceLAB;
+
+type PolarColorSpace =
+  | ColorSpaceHWB
+  | ColorSpaceHSL
+  | ColorSpaceLCH
+  | ColorSpaceLAB;
```
nex3: Lab is rectangular, not polar (every polar space has a "hue" channel).

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+get space(): KnownColorSpace;
+```
+
+#### `channels`
+
+Returns an array of channel values for [`internal`], with missing channels
+converted to `0`.
+
+```ts
+get channels(): string | number;
```
nex3: Shouldn't this be `: number[]`?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+
+type RectangularColorSpace = Omit<KnownColorSpace, PolarColorSpace>;
+
+type HueInterpolationMethod =
+  | 'shorter'
+  | 'longer'
+  | 'decreasing'
+  | 'increasing';
+
+type ChannelValue = string | number | null;
```
nex3: Why can channels be strings?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+get channels(): string | number;
+```
+
+#### `channelsOrNull`
+
+Returns an array of channel values for [`internal`], with missing channels
+converted to `null`.
+
+```ts
+get channelsOrNull(): ChannelValue;
```
nex3: Shouldn't this be an array as well?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+
+#### `channel`
+
+Returns the value of the given `channel` in [`internal`], after converting it to
+`space` if necessary. It should be used instead of the old channel-specific
+functions such as `color.red()` and `color.hue()`.
+
+```ts
+channel(options: {
+  channel: ChannelName;
```
nex3: Generally, mandatory parameters should be positional unless there's a substantial risk of confusion between multiple mandatory parameters.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+  space?: KnownColorSpace;
+}): ChannelValue;
```
nex3: If you use overloads here, you can actually type this so that it statically verifies that the channel name is valid for the color space:

```ts
channel(channel: ChannelNameRGB, {space: ColorSpaceRGB});
channel(channel: ChannelNameHSL, {space: ColorSpaceHSL});
// etc
```

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+Returns an array of channel values for [`internal`], with missing channels
+converted to `null`.
+
+```ts
+get channelsOrNull(): ChannelValue;
+```
+
+#### `isLegacy`
+
+Returns whether [`internal`] is in a legacy color space (`rgb`, `hsl`, or
```
nex3: Nit: link to "legacy color" in the color spec.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+`hwb`).
+
+```ts
+get isLegacy(): boolean;
+```
+
+#### `isInGamut`
+
+Returns whether [`internal`] is in-gamut for its color space (as opposed to
+having one or more of its channels out of bounds, like `rgb(300 0 0)`).
```
nex3: Remember, this is a spec, not just documentation, so we need to specify what it means for a color to be "in-gamut". Fortunately, you can just link to the definition that's already present in the color spec.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+
+```ts
+get isInGamut(): boolean;
+```
+
+#### `channel`
+
+Returns the value of the given `channel` in [`internal`], after converting it to
+`space` if necessary. It should be used instead of the old channel-specific
+functions such as `color.red()` and `color.hue()`.
```
nex3: These usage notes don't belong in the spec, since they're for Sass author consumption rather than implementation author consumption.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+having one or more of its channels out of bounds, like `rgb(300 0 0)`).
+
+```ts
+get isInGamut(): boolean;
+```
+
+#### `channel`
+
+Returns the value of the given `channel` in [`internal`], after converting it to
+`space` if necessary. It should be used instead of the old channel-specific
```
nex3: As above, you need to be explicit about what it means to convert a color to a space (and again you can just link to the color spec). In this case, "if necessary" is unnecessary, since the "Converting a Color" procedure already short-circuits if the color is in the target space.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+combination of channels in that space may be changed. Throws an error if any
+`channel` is not present in `space`.
+
+* If `space` is not defined, let `space` be the value of [`space(internal)`].
+
+* Let `arguments` be the key value pairs in `channels` as keyword arguments.
+
+* Return the value of the result of [`change(internal, ...arguments)`].
+
+[`space(internal)`]: #space
```
nex3: Is this supposed to link to the `color.space()` function?

As a stylistic convention, when we reference built-in functions in the spec, we use an explicit namespace.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+#### `changeChannels`
+
+Returns a new SassColor as the result of changing some of [`internal`]'s
+channels. The `space` value defaults to the `space` of [`internal`], and any
+combination of channels in that space may be changed. Throws an error if any
+`channel` is not present in `space`.
+
+* If `space` is not defined, let `space` be the value of [`space(internal)`].
+
+* Let `arguments` be the key value pairs in `channels` as keyword arguments.
```
nex3: "as keyword arguments" isn't clear here, because "keyword arguments" aren't a Sass value type. If you're passing this to a Sass rest argument, it should be a map, but then you'd also need to be explicit about how to convert from JS Strings, Numbers, and `undefined`s to Sass values.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+}): SassColor;
+```
+
+[`internal`]: ../spec/js-api/value/color.d.ts.md#internal
+
+### New Constructors
+
+Because the value of each channel may be a string, number, or `null`, this
+algorithm checks if an option with a key exists, instead of checking if it is
+set.
```
nex3: What's the difference between "existing" and being "set"?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+  space?: PolarColorSpace;
+  method?: HueInterpolationMethod;
+}): SassColor;
+```
+
+[`internal`]: ../spec/js-api/value/color.d.ts.md#internal
+
+### New Constructors
+
+Because the value of each channel may be a string, number, or `null`, this
```
nex3: If you're going to define this in terms of Sass color functions, you'll need to have some logic to explicitly convert the JS `null` to the unquoted Sass string `none`.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+
+* If `options.alpha` is set:
+
+  * Let `alpha` be a Sass number with a value of `options.alpha`
+
+  * Set [`internal`] to the result of
+    [`color(channel1 channel2 channel3 / alpha)`]
+
+* Otherwise, set [`internal`] to the result of
+  [`color(channel1 channel2 channel3 )`]
```
nex3: This seems to throw away the `space` value. Kust passing channel numbers to the `color()` function (e.g. `color(1 2 3)`) isn't valid. In addition, the CSS `color()` function _can't_ create a color with every space—colors that have their own functions, like `rgb()` or `lab()` _must_ be created with those functions.

It's also not clear to me why this procedure is defined here with a big conditional instead of having each constructor overload below define its own individual procedure.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+
+```ts
+constructor(
+  options: {
+    red: ChannelValue;
+    green: ChannelValue;
+    blue: ChannelValue;
+    alpha?: number;
+  },
+  space?: ColorSpaceRGB
```
nex3: Why isn't this a named argument in `options`?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+[`space(internal)`]: #space
+
+[`change(internal, ...arguments)`]: ./color-4-new-spaces.md#colorchange
+
+```ts
+changeChannels(
+  channels: {
+    [key in ChannelNameRGB]?: ChannelValue;
+  },
+  space?: ColorSpaceRGB
```
nex3: It seems like space should maybe be another key in `options`.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+* `green`
+* `blue`
+* `hue`
+* `saturation`
+* `lightness`
+* `whiteness`
+* `blackness`
+* `alpha`
+
+In addition, `change` is deprecated in favor of `changeChannels`.
```
nex3: Does this need to be renamed? Isn't `change()` broadly forward-compatible with `changeChannels()`?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+  // The color's alpha channel. Mandatory. Must be between 0 and 1,
+  // inclusive.
+  double alpha = 5;
+}
+```
+
+### Deprecations
+
+The `RgbColor`, `HslColor` and `HwbColor` SassScript values will be marked as
+deprecated in the Embedded Protocol.
```
nex3: I'm not sure it's worth continuing to support these. We'll need to make this a breaking protocol change anyway, since the embedded compiler is going to be sending colors as `SassColor` instead of these other types, so we might as well just get rid of them entirely.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,540 @@
...
+* `hue`
+* `saturation`
+* `lightness`
+* `whiteness`
+* `blackness`
+* `alpha`
+
+In addition, `change` is deprecated in favor of `changeChannels`.
+
+## Embedded Protocol
```
nex3, line 901: Add a note that this will be a breaking change for the embedded protocol.

nex3 (changes requested):

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Good point, resolved in 50e403c

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 50e403c

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: This was to address the falsy case where `options.red = null` but is no longer needed with the constructor changes, so it was removed in 171c1d9

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 18cc44d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 18cc44d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 18cc44d.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 18cc44d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Yes, resolved in 18cc44d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved  in 18cc44d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 18cc44d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: I moved it inside, using the same pattern as `channel()`. You can't mix mapped types with static keys in the same type, so I did this with an intersection type. The type is a bit less legible, but it works as expected.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: I was somewhat following the Color.js pattern of defining the color as its own argument, but, as was the case in the other places, there wasn't a compelling reason to do so here.  I moved it inside options in 86f780d.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 86f780d

Overload order here matters somewhat, as if multiple overloads match on the channel, Typescript will only autocomplete the space of the first overload. So, for instance, if you type `c.channel("hue", {space: "`, only `hsl` will be provided as an autocomplete. I decided alphabetical order would be the least arbitrary.

I also put a space-less overload first (`channel(channel: ChannelName): ChannelValue;`), as I assume that will be the most common use case. 

I also wanted the types to communicate clearly that space is optional, so I made the `options` value optional, even though it is functionally identical to having the space-less overload for all spaces.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 86f780d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in [18cc44d](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: They can't- I was considering handling `none` as the string `"none"` but decided not to, and forgot to remove. Removed in 05a1b4d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 05a1b4d

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Addressed in 67c78b9 with a new procedure

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: You're absolutely right, not sure what I was thinking here. I've made these adjustments in 67c78b9, and by taking into account the space argument, we don't need to be as concerned with constructor evaluation order anymore.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: I ended up replacing this with an alternate method that doesn't pass it to a Sass rest argument.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Addressed in c54902d

jamesnw:
@nex3  This is ready for re-review. Notably, I made some significant updates around handling alpha.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -127,61 +127,107 @@ get space(): KnownColorSpace;
...
 get isLegacy(): boolean;
 ```
 
+[legacy color space]: ./color-4-new-spaces.md#legacy-color
+
 #### `isInGamut`
 
-Returns whether [`internal`] is in-gamut for its color space (as opposed to
+Returns whether [`internal`] is [in-gamut] for its color space (as opposed to
 having one or more of its channels out of bounds, like `rgb(300 0 0)`).
```
nex3: Nit: it's fine to have a non-normative clarifying note here, but for consistency with other specs it should be formatted as a block quote.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -48,67 +51,64 @@ import {Value} from '../spec/js-api/value';
...
+
+export type HueInterpolationMethod =
   | 'decreasing'
-  | 'increasing';
+  | 'increasing'
+  | 'longer'
+  | 'shorter';
 
-type ChannelValue = string | number | null;
+export type ChannelValue = number | null;
```
nex3: I'm not sure this is worth defining as its own type rather than just declaring inline.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -127,61 +127,107 @@ get space(): KnownColorSpace;
...
-functions such as `color.red()` and `color.hue()`.
+Returns the value of the given `channel` in [`internal`]. If `space` is set, it
+will first [convert] to that `space`.
 
 ```ts
-channel(options: {
-  channel: ChannelName;
-  space?: KnownColorSpace;
-}): ChannelValue;
+channel(channel: ChannelName): ChannelValue;
```
nex3: I think this shouldn't ever return `null`, since most of the time users won't want to have to explicitly think about missing channels.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -127,61 +127,107 @@ get space(): KnownColorSpace;
...
+channel(
+  channel: ChannelNameHSL,
+  options?: {space: ColorSpaceHSL}
```
nex3: I think these should be mandatory, because what we're trying to express in these types is (for example) _if_ you pass an HSL color space _then_ you have to pass an HSL channel name. If the color space isn't passed, it'll just use the first overload.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -127,61 +127,107 @@ get space(): KnownColorSpace;
...
-automatically when [toSpace()] returns a color with a powerless channel.
+Returns whether the given `channel` of [`internal`] is missing. [Missing
+channels][missing components] can be explicitly specified using the special
+value `none` and can appear automatically when [toSpace()] returns a color with
+a powerless channel. Throws an error if `channel` is not a channel in
+[`internal`]'s `space`.
 
 ```ts
-isChannelMissing(options: {channel: ChannelName}): boolean;
+isChannelMissing(channel: ChannelName): boolean;
```
nex3: This and `isAlphaMissing` should be getters.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -127,61 +127,107 @@ get space(): KnownColorSpace;
...
 
+[in-gamut]: ./color-4-new-spaces.md#coloris-in-gamut
+
 #### `channel`
 
-Returns the value of the given `channel` in [`internal`], after converting it to
-`space` if necessary. It should be used instead of the old channel-specific
-functions such as `color.red()` and `color.hue()`.
+Returns the value of the given `channel` in [`internal`]. If `space` is set, it
+will first [convert] to that `space`.
```
nex3: This, and some of the members below, are still written as documentation rather than a specification. As a rule of thumb, constructs like "it will" don't belong in a specification—they exist to communicate to users what is going to happen when the function is invoked, but we want to communicate to implementers what _must_ happen when the function is invoked (very precisely with an explicit ordering).

If it helps, try writing it as a bulleted list of steps rather than just a paragraph.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
-* Return the value of the result of [`change(internal, ...arguments)`].
-
-[`space(internal)`]: #space
-
-[`change(internal, ...arguments)`]: ./color-4-new-spaces.md#colorchange
+This algorithm takes a JavaScript object `options` and returns a new SassColor
+as the result of changing some of [`internal`]'s components.
+
+The `space` value defaults to the `space` of [`internal`], and any combination
+of channels and alpha in that space may be changed.
```
nex3: "any combination ... may be changed" in a spec suggests that it's up to the implementer what gets changed and when, which isn't accurate.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+
+If `space` is not a [legacy color space], a channel value of `null` will result
+in a [missing component][missing components] value for that channel.
```
nex3: You don't need to write this as part of the specification, because it's already implicit in the way you're invoking the constructor below. If this is intended to be purely documentation to allow a human to more easily understand what's going on, put it in a block quote to clarify that it's non-normative.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+
+* Let `initialSpace` be the value of [`internal.space()`].
+
+* If `options.space` is not defined, let `space` be the value of `initialSpace`.
```
nex3: This is backwards-incompatible with the existing behavior: currently the user can change the hue of a color that was originally specified as RGB, and there's no problem. We'll need to add some logic for determining `space` based on `options` if `initialSpace` is legacy and `options.space` isn't passed.

We should also emit a deprecation warning if this behavior gets triggered.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+
+* Let `keys` be a list of the keys in `options` without `space`.
+
+* Let `components` be `"alpha"` and the channels in `space`.
```
nex3: Should this be "the names of the channels in `space`"?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+
+* If any key in `keys` is not the name of a channel in `components`, throw an
+  error.
+
+* If `space` is not equal to `initialSpace`, let `convertedColor` be the result
```
nex3: You don't refer to `convertedColor` again after this, which means that when you reference the channels it's implicitly referencing it from the _original_ color.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+This algorithm takes a JavaScript object `options` and returns a new SassColor
+as the result of changing some of [`internal`]'s components.
+
+The `space` value defaults to the `space` of [`internal`], and any combination
+of channels and alpha in that space may be changed.
+
+If `space` is not a [legacy color space], a channel value of `null` will result
+in a [missing component][missing components] value for that channel.
+
+* Let `initialSpace` be the value of [`internal.space()`].
```
nex3: `internal` is specifically the _Sass_ value as distinguished from the _JS_ value, so calling methods on it doesn't really make sense. I think you want this to be `this.space()`.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+    space: space
+  })
+  ```
+
+* If space equals `lab` or `oklab`, let `changedColor` be the result of:
+
+  ```js
+  SassColor({
+    lightness: keys.includes('lightness') ? options.lightness : internal.channel('lightness'),
+    a: keys.includes('a') ? options.a : internal.channel('a'),
```
nex3: Since we're changing `channel()` not to return `null`, this will need a little more logic to handle already-missing channel values. You could potentially factor that out into a procedure.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+
+* If any key in `keys` is not the name of a channel in `components`, throw an
+  error.
+
+* If `space` is not equal to `initialSpace`, let `convertedColor` be the result
+  of [`internal.toSpace(space)`].
+
+* If `space` equals `hsl`, let `changedColor` be the result of:
+
+```js
```
nex3: Nit: misaligned code block (also below).

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+  SassColor({
+    hue: options.hue ?? internal.channel('hue'),
```
nex3: We should probably also emit a deprecation warning if the user explicitly passes `null` for a legacy color. This was historically disallowed by the TypeScript types, but if users were using pure JS it worked the same as `undefined`.

Actually, we could also say that if users pass `space` explicitly, even legacy colors treat `null` as missing.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -218,68 +285,190 @@ colors as best they can, but it may be necessary in some cases.
...
+    [key in ChannelNameXYZ]?: ChannelValue;
+  } & {
+    alpha?: ChannelValue;
+    space: ColorSpaceXYZ;
+  }
 ): SassColor;
 ```
 
-### `interpolate()`
+#### `interpolate`
```
nex3: This also needs to be rewritten in spec language

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -312,181 +501,169 @@ interpolate(options: {
...
 
-Because the value of each channel may be a string, number, or `null`, this
-algorithm checks if an option with a key exists, instead of checking if it is
-set.
-
-* If `options.space` is not set, follow the legacy procedure for [construction].
-
-* Let `space` be a string with the value of `options.space`.
+* If `options.space` is not set, or `space` is a [legacy color space], follow
+  the legacy procedure for [construction].
```
nex3: Given that "legacy" is a pretty overloaded term already, I'd say "the previous procedure" here instead.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -495,8 +672,8 @@ constructor(
 
 ### Deprecations
 
-A number of SassColor getters only make sense for legacy colors, and so are
-being deprecated for `channel`.
+A number of SassColor getters only make sense for [legacy color space], and so
+are being deprecated for `channel`.
```
nex3: We've started adding explicit deprecation names, since users can specify deprecations to mute or upgrade to errors in the API. How about `color-4-api` for these?

nex3 (changes requested):

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Would you recommend this mirroring `channels` and `channelsOrNull`, where this returns 0 if null, and a new `channelOrNull` method would potentially return null?

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in a0f37a0

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Did you mean `alpha` should be a getter? `isChannelMissing` can't be a getter because it takes arguments.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: In a0f37a0, I moved this into a non-normative block quote, and changed to

> The `space` value defaults to the `space` of [`internal`], and the caller may
> specify any combination of channels and alpha in that space to be changed.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in a0f37a0

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in a0f37a0

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in a0f37a0

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in a0f37a0

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: I'm considering having the specification for `channel` simply refer to the spec for [color.channel](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md#colorchannel-1). However, that handles alpha- should `channel` here also allow for the alpha component? 

The alpha question also applies to `isChannelMissing` -> `color.is-missing` . I could rename `isChannelMissing` to `isMissing` and then refer to the spec for [color.is-missing](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md#coloris-missing-1) 

I think I can have the specs here also point `isChannelPowerless` to `color.is-powerless` and `interpolate` to `color.mix`. I think it works to point to the specs there, because either we will be able to use the logic in dart-sass, or we will want to re-implement that exact behavior.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Replaced in ab9cb0a. It's used in ~30 places, but it somewhat obscures the types in error messages, etc.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Specified in d71b038

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 1c66e56

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 1c66e56

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 9c3853a.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: In ab8864f, I emitted the `alpha-null` deprecation.

I'm not sure that we can treat `null` as missing if the change is happening in a legacy space. If a color has an `initialSpace` of `hsl` with an `space` of `hsl` specified, and `saturation: null`, we would end with a color that could not be represented. 

If `initialSpace` is `lch`, but set the saturation to missing in the `hsl` space, the `hsl` intermediary color would not be able to be represented, and before conversion back to `lch`, the `saturation` would be set to `0`.

Because the CSS Color Conversion spec specifies that converting colors with missing channels sets that value to 0, I updated the specification to only convert to `initialSpace` at the end if `space` is not equal to `initialSpace`.

jamesnw (commented):
@nex3 Thanks! I have several questions below.

proposal/color-4-new-spaces-js.d.ts.md
nex3: I think it's fine to have `isChannelMissing` and `channelsOrNull` be the two ways to access missing status, I don't think we need `channelOrNull` as well.

proposal/color-4-new-spaces-js.d.ts.md
nex3: Ah, yes, that is what I meant.

proposal/color-4-new-spaces-js.d.ts.md
nex3: > I'm considering having the specification for `channel` simply refer to the spec for [color.channel](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md#colorchannel-1). However, that handles alpha- should `channel` here also allow for the alpha component? 

Yeah, I think any API that takes a channel name should be able to handle alpha as well, even if there are also explicit alpha APIs (because every color has an alpha).

> The alpha question also applies to `isChannelMissing` -> `color.is-missing` . I could rename `isChannelMissing` to `isMissing` and then refer to the spec for [color.is-missing](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md#coloris-missing-1) 

I think the name `isChannelMissing` is clearer here, but I think referring to the Sass function is fine.

proposal/color-4-new-spaces-js.d.ts.md
nex3: I'm not sure what you mean by "can't be represented"... `hsl(0deg none 50%)` is a valid CSS color that's displayed identically to `hsl(0deg 0% 50%)` but handles interpolation differently.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: I've updated these. My main concern was conflating channels and components, but I think it's matching the Sass Color 4 proposal fairly closely now.

Relevant commits- 422af09eb7ec627a10178e07df803f68b6130e80, b42c3233343c48736f5a2a77e7dfaab7d1f64037, 6a80a53877246aeba052bdc828d50ca24a47cdbe

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Updated in [b42c323](https://github.com/sass/sass/pull/3665/commits/b42c3233343c48736f5a2a77e7dfaab7d1f64037)

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: This was removed in a refactor in 422af09eb7ec627a10178e07df803f68b6130e80

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Updated in 6a80a53877246aeba052bdc828d50ca24a47cdbe

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: I refactored this out into a procedure in 4d94520cc6925c432515ee886625ed57294486d9

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: You're correct, I was conflating some unrelated ideas here. I added the logic to allow legacy colors to treat null as missing if space is set in 1e1dae7a819ed0c7d279efc520050edeaf9c32f4. Combined with some refactoring, this diff may be the most helpful- https://github.com/oddbird/sass/compare/e3b8ef726f75679a10ab28ba4713b17e4aa4a7d5...4ada2835df5a714a3901b1bbe140216d20a28eca

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,864 @@
...
+
+```ts
+get isLegacy(): boolean;
+```
+
+[legacy color space]: ./color-4-new-spaces.md#legacy-color
+
+#### `isInGamut`
+
+Returns the result of [`color.is-in-gamut(internal, space)`].
```
nex3: This is a bit (more) pedantic (than usual), but it's probably a good idea to add "as a JS boolean" here just to clarify that we aren't returning the Sass boolean value.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,864 @@
...
+  * Let `value` be the channel value in `color` with name of `component`.
+
+  * If `value` is `none`, let `value` be `null`.
+
+  * Append `value` to `channels`.
+
+* Return `channels`.
+
+```ts
+get channelsOrNull(): [
```
nex3: I think this should only include the three channels that are different between color spaces.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,864 @@
...
+
+[`internal`]: ../spec/js-api/value/color.d.ts.md#internal
+
+```ts
+get space(): KnownColorSpace;
+```
+
+#### `toSpace`
+
+Returns the result of [`color.to-space(internal, space)`] as a new SassColor.
```
nex3: I don't think you need to explicitly specify "as a new SassColor" here, particularly since it's likely some implementations will want to optimize the "same as existing space" case to return the same object.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,864 @@
...
+
+* Let `initialSpace` be the value of [`this.space()`].
+
+* Let `space` be `options.space` if it is defined, and the value of
+  `initialSpace` otherwise.
+
+* If `channel` is not "alpha" or a channel in `space`, throw an error.
+
+* If `space` is not equal to `initialSpace`, let `color` be the result
+  of [`this.toSpace(space)`]. Otherwise let `color` be `this`.
```
nex3: I think you can just assign this unconditionally, since `toSpace()` should be a no-op for the same color space anyway.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,864 @@
...
+```
+
+#### `alpha`
+
+Returns the result of calling [`this.channel('alpha')`].
+
+[`this.channel('alpha')`]: #channel
+
+```ts
+get alpha(): number | null;
```
nex3:
```suggestion
get alpha(): number;
```

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,864 @@
...
+
+* Let `initialSpace` be the value of [`this.space()`].
+
+* Let `spaceSetExplicitly` be `true` if `options.space` is defined, and `false`
+  otherwise.
+
+* Let `space` be `options.space` if `spaceSetExplicitly` is true, and the value of
+  `initialSpace` otherwise.
+
+* If `initialSpace` is a [legacy color space] and `options.space` is not set:
```
nex3: Style nit: I'd either refer to `spaceSetExplicitly` here, or just refer directly to `options.space` being defined everywhere instead of using the intermediate variable.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,864 @@
...
+
+* Otherwise, use the constructor that matches the value of `options.space`.
+
+[construction]: ../spec/js-api/value/color.d.ts.md#constructor
+
+#### LAB Channel Constructor
+
+Create a new SassColor in a color space with LAB channels -- `lab` and `oklab`.
+
+* Let `lightness` be the result of [parsing a channel value] with value
```
nex3, line 633: Looks like this link reference isn't defined.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,864 @@
...
+While the [legacy color space] constructors do not require a space, replace the
+[existing types] with these constuctor overloads for forward compatibility.
+
+[existing types]: ../spec/js-api/value/color.d.ts.md#constructor
+
+```ts
+constructor(options: {
+  red: number;
+  green: number;
+  blue: number;
```
nex3: Shouldn't we allow `null` here so users can set missing channels for these spaces?

nex3 (changes requested):
I think this is pretty close to ready to go!

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Fixed in 3599c5bdd932eef6b28a1e9a2a969656694fc510

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Resolved in 3599c5bdd932eef6b28a1e9a2a969656694fc510

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Added to `isInGamut`, `isChannelMissing`, `isAlphaMissing`, and `isChannelPowerless` in b40cca105e34095a241c612ca7e169c536cd4194

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: In light of your comment on `channel` (and to avoid a potentially non-insignificant difference between implementations),  I went further and specified that it should be a no-op if the space is not changing.  Changed in be40def4295de5511b6d2870742d877c68e982d9,

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Updated in be40def4295de5511b6d2870742d877c68e982d9. I also updated this in `change`.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: I think I misinterpreted an earlier comment about including alpha as a channel, but I think this makes more sense, given the channel vs component distinction. I removed alpha from `channelsOrNull` and `channels` in 03254fa2ff5902a4ae3fba6a854d06e680d0a379.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Good question- I think we need to do some more explicit handling here. The [New Color Spaces Sass Proposal](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md#rgb-and-rgba) explicitly says for RGB and HSL `Missing channels are not allowed in legacy syntax.`. I think this is where I was getting confused elsewhere with legacy colors not supporting missing channels, but it's actually that the legacy Sass constructors for RGB and HSL don't allow missing colors, but they also have new Sass constructors that do.

I think I have a plan, but want to double check-

1. Replace the [JS API color constructor](https://github.com/sass/sass/blob/main/spec/js-api/value/color.d.ts.md#constructor) instead of referring to it
2. For RGB, if space is not set, round channels to an integer to preserve past functionality (and emit deprecation warning? Maybe only if a non-integer is provided?)
3. Set `internal` to the result of `rgb(red green blue / alpha)` instead of `rgb(red, green, blue, alpha)` (and similar for hsl and hwb.) With these constructors, `none` is allowed.
4. Because we're switching all construction to use the modern space-separated syntax, we can allow `null` even if `space` is not set.

@nex3  Is that all correct?

jamesnw: I implemented this in eac2876322245614b5e3d2e7bd8dceb9a9a483dd, and decided not to emit a deprecation in step 2.

nex3: This all sounds good, but I wouldn't round in step 2.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,938 @@
...
+  space: ColorSpaceLCH;
+});
+```
+
+#### Predefined RGB Channel Constructor
+
+Create a new SassColor in a color space with RGB channels -- `srgb`,
+`srgb-linear`, `display-p3`, `a98-rgb`, and `prophoto-rgb`. `rgb` is not
+supported with this constructor, as the legacy `rgb` color does not support
+[missing components].
```
nex3: This last sentence is no longer true, right?

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: I removed the rounding in https://github.com/sass/sass/pull/3665/commits/00b80f55b1b2af0157716a81c6a1cbbe02fd2769. 

I also realized that I could emit the null-alpha deprecation if alpha is null and space is not defined, so I did that in that commit for RGB, and in HSL and HWB in https://github.com/sass/sass/pull/3665/commits/d0fb9e0f0b3174c7907a57f42eff2f43a20ded6f.

nex3 (changes requested):

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Updated in 5df5b7502c6c9931cfe76348274a6d15694ef836. 

While the Typescript definition could trivially be adjusted to include `rgb`, It's a separate constructor so that it explicitly overrides the legacy constructor, because the `rgb` constructor has an additional deprecation step, and so that it's clear that space is not optional for the predefined rgb spaces.

nex3 (approved):

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,931 @@
...
+### Modified Legacy Color Constructors
+
+These will replace the [existing constructors] for legacy colors.
+
+#### HSL Constructor
+
+Create a new SassColor in the `hsl` color space.
+
+* If `options.alpha` is `null` and `options.space` is not set, emit a
+  deprecation warning named `null-alpha`.
```
jgerigmeyer, line 764: @jamesnw I want to be sure I understand the intent behind the "and `options.space` is not set" clauses in these modifications to the legacy color constructors. Does that mean if `alpha: null` and `space: 'hsl'` are used, then `alpha` will be interpreted as `"none"` as per [Parsing a Channel Value](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces-js.d.ts.md#parsing-a-channel-value)?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -0,0 +1,931 @@
...
+
+* `red`
+* `green`
+* `blue`
+* `hue`
+* `saturation`
+* `lightness`
+* `whiteness`
+* `blackness`
+* `alpha`
```
jgerigmeyer, line 862: @jamesnw Is `alpha` being deprecated? It looks like it's identical to the "new" `alpha()` fn [described above](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces-js.d.ts.md#alpha).

proposal/color-4-new-spaces-js.d.ts.md
nex3: That's right. If you use the new constructor format, you get the new semantics.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Good catch, `alpha` is not deprecated. Updated in https://github.com/sass/sass/pull/3704/commits/b9484d122369877c39e3edc6a3219011afbb6c5f

PR #3667, opened 2023-08-22
Add a Markdown linter

nex3:
I had hoped that this would catch missing link reference issues like
those in sass/sass#3661, but unfortunately those aren't supported
yet (see DavidAnson/markdownlint#915).


Goodwine (changes requested):
I don't think this linter is catching the missing references, the `accepted/calc-functions.md` file still has broken links for `[compatible]` and the linter seems to be fine with it 🤔

nex3:
Yes, that's what I mentioned above when I linked https://github.com/DavidAnson/markdownlint/issues/915 :slightly_smiling_face:

Goodwine (approved):

PR #3670, opened 2023-08-25
Use buf instead of protoc to compile protobufs

### LINKED ISSUE #3666: Migrate CI from arduino/setup-protoc to buf
Context https://github.com/sass/sass/pull/3649#issuecomment-1679450953:

- protoc is not shipping JS since 3.21.x
- Coincidentally, protoc dropped the major version from its releases, so it was released as `v21.0`
- arduino/setup-protoc@v2 no longer searches for the full version, only for the "patch" version.
- We need to get the `protoc-gen-js` release separately from https://github.com/protocolbuffers/protobuf-javascript
- The npm package `protoc-gen-js` is available, but there's no guarantee it'll keep being updated. https://github.com/protocolbuffers/protobuf-javascript
- protoc version is now `3.24.0` but JS stayed behind at `3.21.2`
- I think we got lucky and the only reason `setup-protoc@v1` doesn't break is because the breaking version and all subsequent releases from `v3.21.0` dropped the `3.`, so the github action `setup-protoc@v1` cannot download any future releases.

Buf works, we have migrated other repos to buf, and we should do the same here

Goodwine:
This is a more consistent way of testing the validation of protobufs. This now avoids relying on protoc which dropped built-in support for generating JS. Fixes #3666

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,10 +72,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: arduino/setup-protoc@v1
-        with: { version: "3.x", repo-token: "${{ github.token }}" }
+      - uses: bufbuild/buf-setup-action@v1.26.1
+        with: {github_token: "${{ github.token }}"}
       - name: Generate protobuf code
-        run: protoc --js_out=/tmp spec/embedded_sass.proto
+        run: buf generate --output=/tmp --path=spec/embedded_sass.proto
 
   embedded_protocol_versions:
     name: "Validate Embedded Protocol versions"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
       - uses: bufbuild/buf-setup-action@v1.26.1
         with: {github_token: "${{ github.token }}"}
       - name: Generate protobuf code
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -72,10 +72,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: arduino/setup-protoc@v1
-        with: { version: "3.x", repo-token: "${{ github.token }}" }
+      - uses: bufbuild/buf-setup-action@v1.26.1
+        with: {github_token: "${{ github.token }}"}
       - name: Generate protobuf code
-        run: protoc --js_out=/tmp spec/embedded_sass.proto
+        run: buf generate --output=/tmp --path=spec/embedded_sass.proto
```
nex3: We should specify the input in `buf.work.yaml`, and I don't think there's a particular reason to generate to `/tmp` as opposed to the default location.

nex3 (changes requested):

.github/workflows/ci.yml
Goodwine: done, and good point. For some reason I thought the CI would complain if there was uncommitted code 😅. That said, I added a `.gitignore` to prevent accidentally committing the generated code if we also accidentally run `buf generate` :)

nex3 (approved):

PR #3674, opened 2023-08-29
add first class mixins to embedded sass protobuf and JS API

connorskees:
See https://github.com/sass/sass/pull/3626 and https://github.com/sass/dart-sass/pull/2073/


nex3 (approved):

ntkme:
We should bump the version number in https://github.com/sass/sass/blob/main/spec/EMBEDDED_PROTOCOL_VERSION

stof:
As this creates a new kind of value that can be sent to the host, is it actually a minor version or a major one ?

ntkme:
Historically this kind of change has been considered as minor. The host can upgrade to new protocol version without supporting added type, and all previous feature would still work.

nex3 (approved):

ntkme:
This looks good! Only thing is that we should probably edit the title of this PR as the scope of this PR has increased.

nex3 (approved):

PR #3675, opened 2023-08-30
[Calc Functions] Update based on implementation feedback

nex3:
See #3504


stof:
With this new merged syntax, how is it expected to interact with the /-as-division deprecation ? In particular with the expected future state where `/` should be parsed as division in calculation but as a list separator outside them ?

nex3:
Good question. The same syntax would still parse, so the main issue is one of precedence—a `SlashListExpression` has lower precedence than _all_ math functions, so `a * b / c * d` would parse as `(a * b) / (c * d)` rather than `((a * b) / c) * d`. We'll need to have an algorithm for redoing the precedence when evaluating an expression as a calculation. I'll add that to the spec.

accepted/calc-functions.d.ts.md
Goodwine, line 1: Are both of these files `calc-functions.d.ts.md` and `calc-functions.md` supposed to co-exist? I thought the former generated the latter, and I can't tell the differences very well 😅 

I'll review the other file first and then try to see if I can figure if there's any change from that other file to this file.

accepted/calc-functions.md
```diff
@@ -518,8 +569,10 @@ global function:
 [the semantics for `FunctionCall`]: ../spec/functions.md#functioncall
 
 * If `function` is null; `name` is case-insensitively equal to `"min"`, `"max"`,
-  `"round"`, or `"abs"`; and all arguments in `call`'s `ArgumentInvocation` are
-  [calculation-safe], return the result of evaluating `call` [as a calculation].
+  `"round"`, or `"abs"`; `call`'s `ArgumentInvocation` doesn't have any
+  `KeywordArgument`s or `RestArgument`s; and all arguments in `call`'s
```
Goodwine, line 573: Should this keyword/rest condition apply to the other calc functions too? or why is it that only these four get this special behavior?

accepted/calc-functions.md
```diff
@@ -575,21 +629,18 @@ To evaluate a `SumExpresssion` or a `ProductExpression` as a calculation value:
...
+* Let `result` be the result of evaluating `expression` as a calculation value.
 
-  * If `result` is a number or a calculation, return it.
+* If `result` is a number or a calculation, return it.
 
-    > This could happen if the user defines a `var` function in Sass.
+  > Otherwise, it must be an unquoted string.
 
-  * If `result` is not an unquoted string, throw an error.
+* If `result` begins case-insensitively with `"var("`, or if `expression` is an
```
Goodwine, line 643: Currently `calc(1 / #{2 / 5})` results in `calc(1/2/5)` which evaluates to `0.1`.
IIUC this change means that it'll now produce `calc(1/(2/5))` which evaluates to `2.5`.

Is that correct?

accepted/calc-functions.md
```diff
@@ -611,13 +662,183 @@ To evaluate an `InterpolatedIdentifier` `ident` as a calculation value:
...
+      * Let `last-left` be the last operand of `left` and `first-right` the
+        first operand of `right`.
+
+      * Set `left` to a `SumExpression` that begins with all operands and
+        operators of `left` except `last-left`, followed by a
+        `SlashListExpression` with elements `last-left` and `first-right`,
+        followed by all operators and operands of `right` except `first-right`.
+
+        > For example, if we insert virtual parentheses to explicitly indicate
+        > precedence groupings, `(1 + 2) / (3 + 4)` becomes `1 + (2 / 3) + 4`.
```
Goodwine: IIUC the user input here is `1 + 2 / 3 + 4`, but the parser reads it like `slash_list(1+2, 3+4)`, and this process changes it to `1 + slash_list(2,3) + 4`. right?

If so, I think that even though you mention "virtual parentheses" (TIL `-is` is singular and `-es` is plural!) it's confusing to see _"`(1 + 2) / (3 + 4)` becomes `1 + (2 / 3) + 4`"_. Like I can understand it, but it takes me a couple reading passes.

accepted/calc-functions.d.ts.md
nex3: Oops, that's a Git mishap on my part. I've fixed it now. (The only difference was that the more recent one didn't explicitly re-specify the TypeScript API, which prevented a TypeScript error.)

accepted/calc-functions.md
nex3: These four get special behavior because they'll fall back to calling global Sass functions. For the others, we want keywords and rest arguments to be errors instead—if we exempted calls that included them, rest arguments would just be passed to plain CSS functions, so something like `clamp(1 2 3...)` would return the unquoted string `"clamp(1, 2, 3)"` instead of producing an error. I'll add a note to explain this.

accepted/calc-functions.md
nex3: No. These are the semantics for evaluating a `ParenthesizedExpression`, and `calc(1 / #{2 / 5})` doesn't contain a parenthesized expression. It would only produce `calc(1 / (2 / 5))` if you wrote `calc(1 / (#{2 / 5}))`.

accepted/calc-functions.md
nex3: Updated to use `slash-list` notation.

accepted/calc-functions.md
```diff
@@ -577,10 +577,15 @@ global function:
   [calculation-safe]: #calculation-safe-expression
   [as a calculation]: #evaluating-a-functioncall-as-a-calculation
 
+  > For calculation functions that overlap with global Sass function names, we
+  > want anything Sass-specific like this to end up calling the Sass function.
+  > For all other calculation functions, we want those constructs to throw an
+  > error (which they do when evaluating `call` [as a calculation]).
```
Goodwine, line 583: Very helpful, thanks!

Goodwine (approved):

Issue #3673, opened 2023-08-29
Support the relative color syntax of CSS Color 5


stof:
CSS Color level 5 introduces the relative color syntax which allows to compute a color based on another one: https://www.w3.org/TR/css-color-5/#relative-colors

This feature is already available in Safari 16.4+ and Chrome will [ship it soon](https://groups.google.com/a/chromium.org/g/blink-dev/c/KeKBTjKEdaU/m/NCHAo8n-AgAJ).

Currently, this syntax throws an error when using it in a Sass file with a color function known by Sass (using it in `oklab` works for now due to the lack of support for `oklab` in Sass and so not treating it like a color at all, but this would change once https://github.com/sass/sass/issues/2831 will be implemented): https://sass-lang.com/playground/#MTFhJTIwJTdCJTBBJTIwJTIwYiUzQSUyMHJnYihmcm9tJTIwbWFnZW50YSUyMGNhbGMociUyMCUyQiUyMDNkZWcpJTIwZyUyMGIpJTNCJTBBJTdE

Following the design principles of Sass, this syntax should be accepted.

stof:
Idea for a proposal:

For all global color factory functions, add a check in their single-argument syntax. If the first list item (accounting properly for the space separate list being on the left of an optional `/` separator in case the alpha channel is provided) is an unquoted string being equal to  `from` (case insensitively), emit the function as a CSS function instead of producing a Sass color.

A future proposal might want to add full support for processing relative colors in Sass itself but I don't think this is necessary for the initial support. The common use case for relative colors in the browser is anyway to use a CSS custom variable to provide the base color, which would not be resolvable in Sass (and so would still have to output it as a CSS function). Projects wanting to do color computation with known colors in Sass will probably stick to using the `sass:color` functions instead.

nex3:
We can start thinking about handling this properly once we're closer to done with support for Color Level 4 :sweat_smile:. For now, I wouldn't be opposed to a stop-gap measure where color function invocations beginning with `from` just return strings.

stof:
My proposal is not dependent on support for Color Level 4 (the "future proposal" part would need it, but I'm not even sure this is something that should be proposed).

stof:
@nex3 does this initial proposal require a formal proposal or can it go through the fast track ?

nex3:
I think it's fine to fast-track.

stof:
This is now implemented. I don't think this needs any dedicated documentation as it is about CSS compatibility so I'm closing this issue as completed.

PR #3676, opened 2023-09-02
Add support for the relative color syntax from CSS Color 5

stof:
When using the relative color syntax, the color function is emitted as a CSS function (similar to cases using a CSS variable in arguments).

This fast track proposal updates the definition of the `rgb` and `hsl` global functions. The proposal for color spaces is also updated to reflect the new parsing.

#3673

### DIFF
--- a/proposal/color-4-new-spaces.changes.md
+++ b/proposal/color-4-new-spaces.changes.md
@@ -1,3 +1,8 @@
+## Draft 1.11
+
+* Add support for the relative color syntax in the algorithm parsing color
+  arguments, for CSS compatibility.
+
 ## Draft 1.10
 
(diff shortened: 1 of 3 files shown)


nex3 (approved):

nex3:
Spec changes look good, but per the fast track process this should be landed _along with_ patches to Dart Sass and the tests.

stof:
@nex3 I have the dart-sass change in progress locally. I will open the PR soon.

stof:
@nex3 PRs for sass-spec and dart-sass are now ready

stof:
@nex3 as you merged the implementation in dart-sass, I think this should be merged as well.

PR #3693, opened 2023-09-18
[Calc Functions] Apply to main specification

nex3:
See #3504


spec/js-api/value/calculation.d.ts.md
```diff
@@ -237,41 +237,48 @@ hashCode(): number;
...
 
-A private property like [`Value.internal`] that refers to a Sass
-[`CalculationInterpolation`].
+A private property like [`Value.internal`] that refers to a Sass string.
 
 #### Constructor
 
-Creates a Sass [`CalculationInterpolation`] by setting the `value` field to the
-`value` argument and returns it.
+Creates a `CalculationInterpolation` with its `internal` set to an unquoted Sass
```
Goodwine: Should this line read without `its`? or maybe it's missing a word after `internal`?

Goodwine (approved):

PR #3699, opened 2023-09-26
Package Importer updates

### LINKED ISSUE #3701: The proposed ordering of importers is inconsistent for the package importer proposal
The section `Updated importers options` in `types` says:

> If any object in `options.importers` is exactly equal to the object `nodePackageImporter`:
> 
>    - Let `pkgImporter` be a [Node Package Importer](https://github.com/sass/sass/blob/main/proposal/package-importer.d.ts.md#node-package-importer) with an associated `entryPointURL` of `require.main.filename`.
> 
>    - Replace `nodePackageImporter` with `pkgImporter` in a copy of `options.importers`.

The `Semantics` section after that says

> Package Importers must be added to the [global importer list](https://github.com/sass/sass/blob/main/spec/modules.md#global-importer-list) immediately after any user-provided importers.

This is inconsistent as the importers option allows to put them anywhere among the user-provided importers.
Is this sentence in `semantics` meant to apply only to the new `pkgImporter` option of the legacy API ? If yes, it might make sense to move it to the section describing that option instead.

jamesnw:
Changes: 

* Throw an error if `nodePackageImporter` is used in the browser.

* Remove specified order in the global import list, as users can specify the
  order within the `importers` option.

Closes https://github.com/sass/sass/issues/3701

### DIFF
--- a/proposal/package-importer.d.ts.md
+++ b/proposal/package-importer.d.ts.md
@@ -273,8 +273,12 @@ Javascript Compile API, insert:
 ### Legacy API `pkgImporter`
 
 If set, the compiler will use the specified built-in package importer to resolve
-any URL with the `pkg:` scheme. Currently, the only available package importer
-is `node`, which follows Node resolution logic to locate Sass files.
+any URL with the `pkg:` scheme. This step will be inserted immediately before
+the existing legacy importer logic, and if the package importer returns `null`,
+the legacy importer logic will be invoked.
+
+Currently, the only available package importer is `node`, which follows Node
+resolution logic to locate Sass files.
 
 Defaults to undefined.
 
--- a/proposal/package-importer.d.ts.md
+++ b/proposal/package-importer.d.ts.md
@@ -256,7 +256,10 @@ Javascript Compile API, insert:
 * If any object in `options.importers` is exactly equal to the object
   `nodePackageImporter`:
 
-  * If the execution environment is the browser, throw an error.
+  * If no filesystem is available, throw an error.
+
+  > This primarily refers to a browser environment, but applies to other
+  > sandboxed JavaScript environments as well.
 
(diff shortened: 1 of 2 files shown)


proposal/package-importer.d.ts.md
```diff
@@ -256,6 +256,8 @@ Javascript Compile API, insert:
 * If any object in `options.importers` is exactly equal to the object
   `nodePackageImporter`:
 
+  * If the execution environment is the browser, throw an error.
```
jamesnw: @nex3 Is a more formal way needed to specify the execution environment?

nex3: I think this is fine, although it might be slightly more accurate to say "if no filesystem is available" in case this is run in a different kind of sandboxed JS environment.

jamesnw: Updated in https://github.com/sass/sass/pull/3699/commits/85a27d87d6cc5fc29ce878592a7edfb3a9042e35

proposal/package-importer.d.ts.md
```diff
@@ -256,6 +256,11 @@ Javascript Compile API, insert:
 * If any object in `options.importers` is exactly equal to the object
   `nodePackageImporter`:
 
+  * If no filesystem is available, throw an error.
+
+  > This primarily refers to a browser environment, but applies to other
+  > sandboxed JavaScript environments as well.
```
nex3: Nit: +2 indent here so this is inside the previous bullet point.

nex3 (approved):

proposal/package-importer.d.ts.md
jamesnw: Fixed in https://github.com/sass/sass/pull/3699/commits/adcf4faa586c82e6ff438a700078cdfeb83060e4

PR #3704, opened 2023-09-28
Draft 1.1 changes for Color spaces JS API 

jamesnw:
Updates to https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces-js.d.ts.md

### DIFF
--- a/proposal/color-4-new-spaces-js.changes.md
+++ b/proposal/color-4-new-spaces-js.changes.md
@@ -11,6 +11,14 @@
 
 * Use `Exclude<>` instead of `Omit<>` for union types.
 
+* Update `change` for legacy colors:
+
+  * Make procedure for determining space backwards compatible
+  
+  * If channels from multiple spaces are specified and no space is specified,
+    emit a deprecation warning instead of throwing an error to prevent a
+    breaking change.
+
 ## Draft 1
 
 * Initial draft
--- a/proposal/color-4-new-spaces-js.changes.md
+++ b/proposal/color-4-new-spaces-js.changes.md
@@ -19,6 +19,8 @@
     emit a deprecation warning instead of throwing an error to prevent a
     breaking change.
 
+* Fix channel names for `change` with `oklch` and `lch`.
+
 ## Draft 1
 
 * Initial draft
--- a/proposal/color-4-new-spaces-js.changes.md
+++ b/proposal/color-4-new-spaces-js.changes.md
@@ -9,6 +9,8 @@
 * Rename types: `ColorSpaceLAB` to `ColorSpaceLab`, `ChannelNameLAB` to
   `ChannelNameLab`.
 
+* Use `Exclude<>` instead of `Omit<>` for union types.
+
 ## Draft 1
 
 * Initial draft
--- a/proposal/color-4-new-spaces-js.changes.md
+++ b/proposal/color-4-new-spaces-js.changes.md
@@ -1,0 +1,9 @@
+## Draft 1.1
+
+* Clarify values in `channels` and `channelsOrNull`.
+
+* Throw an error if construction space can not be determined.
+
+## Draft 1
+
+* Initial draft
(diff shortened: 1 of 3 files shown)


jgerigmeyer:
@nex3 Of course there may be more changes along the way, but this first set is probably ready for review/merge?

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -401,14 +409,18 @@ as the result of changing some of [`internal`]'s components.
...
 * Let `components` be `"alpha"` and the names of the channels in `space`.
 
-* If any key in `keys` is not the name of a channel in `components`, throw an
-  error.
+* If any key in `keys` is not the name of a channel in `components`:
+
+  * If `initialSpace` is a [legacy color space] and `spaceSetExplicitly` is
+    false, emit a deprecation warning named `color-4-api`.
+
+  * Otherwise, throw an error.
```
nex3: I think we can comfortably start throwing an error without updating the spec here and call it a potentially-breaking bug fix. We have a general rule that if users violate the TypeScript types, the behavior is undefined. We'll try to deprecate in cases where it's very likely that the types were being violated in good faith (like passing `null` instead of `undefined` to optional parameters), but I think it's pretty unlikely that anyone is actually running `new SassColor({hue: ..., red: ...})` and expecting it to work.

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -512,13 +524,13 @@ as the result of changing some of [`internal`]'s components.
   })
   ```
 
-* If space equals `lch` or `oklch`, let `changedColor` be the result of:
+* If `space` equals `lch` or `oklch`, let `changedColor` be the result of:
 
   ```js
   SassColor({
```
nex3: It occurs to me that these should probably be

```suggestion
  new SassColor({
```

nex3 (changes requested):

proposal/color-4-new-spaces-js.d.ts.md
jgerigmeyer: Addressed in https://github.com/sass/sass/pull/3704/commits/7308783af5646e553c35b4944216bcc78b781a44

proposal/color-4-new-spaces-js.d.ts.md
jgerigmeyer: Addressed in https://github.com/sass/sass/pull/3704/commits/7308783af5646e553c35b4944216bcc78b781a44

jgerigmeyer:
@nex3 Ready for another round of review.

nex3 (approved):

proposal/color-4-new-spaces-js.changes.md
```diff
@@ -0,0 +1,21 @@
...
+
+* Remove `alpha` from list of deprecated getters.
+
+* Rename types: `ColorSpaceLAB` to `ColorSpaceLab`, `ChannelNameLAB` to
+  `ChannelNameLab`.
+
+* Use `Exclude<>` instead of `Omit<>` for union types.
+
+* Make procedure for determining space backwards compatible when using `change`
+  for legacy colors.
```
nex3, line 15: I think this is no longer relevant.

jgerigmeyer: @nex3 I think this is still relevant for the [changes below](https://github.com/sass/sass/pull/3704/files#diff-c478516801a088f8c5bb0f50f2949f93259f54932decf8367c5d6a733ee6e17aR394-R400), updating how `space` is determined based on the presence of `whiteness`/`blackness`, `hue`/`saturation`/`lightness` etc.

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: @nex3 This brings a related point up, then. Currently the proposed types for `change` do allow for `myColor.change({hue: ..., red: ..., x: ...})` through the overload without a space defined. We could limit the types by removing that overload, and then setting the `space` to optional in all the other types. I think this also communicates the interface a bit more directly as well.

PR #3711, opened 2023-10-07
[First-Class Mixins] Flush to spec

### LINKED ISSUE #626: Dynamic mixin invocation

nex3:
Closes #626

### DIFF
--- a/spec/at-rules/function.md
+++ b/spec/at-rules/function.md
@@ -55,7 +55,7 @@ To execute a `@function` rule `rule`:
 
     > This overrides the previous definition, if one exists.
 
-  * Set [the current import context][]'s function `name` to `function`.
+  * Set [the current import context]'s function `name` to `function`.
 
     > This happens regardless of whether or not it begins with `-` or `_`.
 
--- a/spec/at-rules/function.md
+++ b/spec/at-rules/function.md
@@ -28,20 +28,38 @@ To execute a `@function` rule `rule`:
 
   [vendor prefix]: ../syntax.md#vendor-prefix
 
+* Let `parent` be the [current scope].
+
+  [current scope]: ../spec.md#scope
+
+* Let `function` be a [function] named `name` which does the following when
+  executed with `args`:
+
+  [function]: ../types/functions.md
+
+  * With the current scope set to an empty [scope] with `parent` as its parent:
+
+    * Evaluate `args` with `rule`'s `ArgumentDeclaration`.
+
+    * Execute each statement in `rule`.
+
+    * Return the value from the `@return` rule if one was executed, or throw an
+      error if no `@return` rule was executed.
+
+  [scope]: ../spec.md#scope
+
 * If `rule` is outside of any block of statements:
 
   * If `name` *doesn't* begin with `-` or `_`, set [the current module][]'s
-    function `name` to `rule`.
+    function `name` to `function`.
 
     > This overrides the previous definition, if one exists.
 
-  * Set [the current import context][]'s function `name` to `rule`.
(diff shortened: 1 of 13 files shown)


spec/at-rules/function.md
```diff
@@ -28,20 +28,38 @@ To execute a `@function` rule `rule`:
...
 * If `rule` is outside of any block of statements:
 
   * If `name` *doesn't* begin with `-` or `_`, set [the current module][]'s
-    function `name` to `rule`.
+    function `name` to `function`.
 
     > This overrides the previous definition, if one exists.
 
-  * Set [the current import context][]'s function `name` to `rule`.
+  * Set [the current import context][]'s function `name` to `function`.
```
Goodwine: Nit: since you're changing this line, would you mind removing the `[]` part from the linkified text?

spec/built-in-modules/meta.md
```diff
@@ -144,6 +162,31 @@ This function is also available as a global function named `get-function()`.
...
+
+```
+get-mixin($name, $module: null)
+```
+
+* If `$name` is not a string, throw an error.
+
+* If `$module` is null:
+
+  * Return the result of [resolving a mixin] named `$name`. If this returns
```
Goodwine, line 183: the `resolving a mixin` method seems to make it like `get-mixin('foo.bar')` should work and return the `bar` mixin from the namespace `foo`, but that doesn't seem to be the case to how it's implemented. 

Should this clarify that it should be the name for a mixin declared in the current namespace?

spec/built-in-modules/meta.md
```diff
@@ -198,16 +241,14 @@ This function is also available as a global function named `mixin-exists()`.
 
 * If `$module` is null:
 
-  * Return whether [resolving a mixin][] named `$name` returns null.
-  
-  [resolving a mixin]: ../modules.md#resolving-a-member
+  * Return whether [resolving a mixin] named `$name` returns null.
```
Goodwine, line 258: ditto

spec/types/functions.md
```diff
@@ -0,0 +1,62 @@
...
+> As an example, if we declare two functions:
+>
+> ```scss
+> @function foo() {
+>   @return red;
+> }
+>
+> $a: meta.get-function(foo);
+>
+> @mixin foo {
```
Goodwine:
```suggestion
> @function foo {
```

spec/types/functions.md
```diff
@@ -0,0 +1,62 @@
...
+>   @return red;
+> }
+>
+> $b: meta.get-mixin(foo);
```
Goodwine:
```suggestion
> $b: meta.get-function(foo);
```

Goodwine (changes requested):

Goodwine (approved):

PR #3716, opened 2023-10-10
Clarifying performance expectations about sass and sass-embedded

ntkme:


js-api-doc/compile.d.ts
```diff
@@ -44,6 +44,11 @@ export interface CompileResult {
  *
  * This only allows synchronous {@link Importer}s and {@link CustomFunction}s.
  *
+ * **Heads up!**  When using `sass-embedded` npm package, **{@link compileAsync}
```
nex3: Nit: "the `sass-embedded` npm package" (also below)

nex3 (changes requested):

nex3 (approved):
Thanks!

Issue #3296, opened 2022-04-23
Provide a way to share resources across multiple compilations


sapphi-red:
* [Proposal](https://github.com/sass/sass/blob/main/proposal/shared-resources.d.ts.md)

---

It seems currently sass-embedded starts one process for each compile request and stops it after compilation.

When using Vue components with sass, Vite requests one compile request for one file. This makes many compile requests and it slows down by process creation overhead.

[esbuild reuses process](https://github.com/evanw/esbuild/blob/e8768569a042be09be0b01445fbe399ed043793b/lib/npm/node.ts#L245-L356) to avoid this situation. I suppose the performance can be improved in the same way.

ntkme:
A quick benchmark with sass-embedded-host-ruby as it is capable of reusing compiler process:

``` ruby
require 'benchmark'
require 'sass-embedded'

n = 1000
Benchmark.bm do |x|
  x.report('reuse') do
    n.times do
      Sass.compile_string('')
    end
  end
  x.report('  new') do
    n.times do
      sass = Sass::Compiler.new
      sass.compile_string('')
      sass.close
    end
  end
end
```

```
       user     system      total        real
reuse  0.148874   0.068218   0.217092 (  0.446515)
  new  0.694413   1.849822  12.046062 ( 10.330498)
```

You can see the overhead of starting a new compiler process in my environment is about 12 milliseconds. For a project with tons of files to compile, the overhead can add up to something noticeable.

nex3:
This is something we'd like to add, but it's non-trivial from a design standpoint. We'd need a way for the user to represent when a shared process should be spun up and when it should be disabled, and we'd want to make sure that API will work well for integration with various frameworks that tend to do many different Sass compilations over time. Ideally it would also "rhyme" with the API for sass/sass#2745, since that represents a more intense version of the same underlying "shared resources across compilations" logic.

I'm going to move this to the language repo, since it will require a JS API proposal, even if that proposal will only be (usefully) implemented in the embedded host.

ntkme:
@nex3 I think there are two stages of optimization here. The first one would be simply reuse the same dart-sass-embedded process but do not share resources across requests, which is possible today without the need of changing the API. I think that is what @sapphi-red and I are talking about in terms of process creation overhead. The shared resources, e.g. shared cache, would be another optimization that would need discussion in language repo.

nex3:
[...]

PR #3719, opened 2023-10-13
Proposal: Compiler interface for shared resources

jamesnw:
https://github.com/sass/sass/issues/3296

### DIFF
--- a/proposal/shared-resources.d.ts.md
+++ b/proposal/shared-resources.d.ts.md
@@ -2,10 +2,10 @@
 
 *([Issue](https://github.com/sass/sass/issues/3296))*
 
-This proposal adds an API design for sharing resources across multiple
-invocations of the Sass compiler's JavaScript API. This will provide Sass's
-users with a more efficient way of running Sass compilations across multiple
-files.
+This proposal adds an API design that allows for sharing resources across
+multiple invocations of the Sass compiler's JavaScript API. This will provide
+Sass's users with a more efficient way of running Sass compilations across
+multiple files.
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


jamesnw:
[I want to share resources across multiple invocations of the Sass compiler's JS API.](https://trello.com/c/BT4VgchI/4-i-want-to-share-resources-across-multiple-invocations-of-the-sass-compilers-js-api)

proposal/shared-resources.d.ts.md
```diff
@@ -0,0 +1,247 @@
...
+* Returns `true` when disposal is complete.
+
+### Disposing an Async Compiler
+
+When `dispose` is invoked on a Compiler or Async Compiler:
+
+* Any subsequent invocations of `compileAsync` and `compileStringAsync` must
+  throw an error.
+
+* Resolves a Promise with `true` when disposal is complete.
```
nex3: Why pull these into separate sections instead of just including them directly in the docs for their respective `dispose()` methods?

proposal/shared-resources.d.ts.md
```diff
@@ -0,0 +1,247 @@
...
+### Disposing a Compiler
+
+When `dispose` is invoked on a Compiler:
+
+* Any subsequent invocations of `compile` and `compileString` must throw an
+  error.
+
+* Returns `true` when disposal is complete.
+
+### Disposing an Async Compiler
```
nex3: What happens if a compilation is still running when an async compiler is disposed?

proposal/shared-resources.d.ts.md
```diff
@@ -0,0 +1,247 @@
...
+}
+```
+
+## Semantics
+
+### Compiler
+
+A Compiler must:
+
+* Have a lifetime, which starts upon construction.
```
nex3: You're using the term "lifetime" but you haven't defined it or described how it interacts with any other semantics. If all a "lifetime" is is just "after `dispose()` is called these other methods throw errors", I'm not sure you need to define it separately at all.

proposal/shared-resources.d.ts.md
```diff
@@ -0,0 +1,247 @@
...
+
+* `compile` and `compileString` must have identical semantics to the [Sass
+  interface].
```
nex3: This should be part of the specifications for these functions.

nex3 (changes requested):

proposal/shared-resources.d.ts.md
jamesnw: Removed this in ff34f2d8e6039df5b0adeeabefdee67edddfbdd5

proposal/shared-resources.d.ts.md
jamesnw: I moved this into the Types section in ff34f2d8e6039df5b0adeeabefdee67edddfbdd5.

proposal/shared-resources.d.ts.md
jamesnw: In ff34f2d8e6039df5b0adeeabefdee67edddfbdd5, I specified-

```
* Any compilations that have not yet been settled must be allowed to settle, and
  not be cancelled.

* Resolves a Promise with `true` when all compilations have been settled, and
  disposal is complete.```

proposal/shared-resources.d.ts.md
jamesnw: Combined in https://github.com/sass/sass/commit/ff34f2d8e6039df5b0adeeabefdee67edddfbdd5

jamesnw (commented):
@nex3 Thanks! I think this is ready for another review.

nex3 (approved):

ntkme:
Looks good to me. While splitting sync and async compilers is not the best in terms of being user friendly, I do understand this is a limitation due to how the implementation work in embedded-host-node. So I think it an acceptable compromise.

PR #3723, opened 2023-10-19
[Color 4] Draft 1.4 changes for Color Spaces JS API

jamesnw:


proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -752,11 +779,12 @@ Create a new SassColor in the `hsl` color space.
 * Let `saturation` be the result of [parsing a channel value] with value
   `options.saturation`.
 
-* Let `lightness` be the result of [parsing a channel value] with value
-  `options.lightness`.
+* Let `lightness` be the result of [parsing a clamped channel value] with
+  `value` of `options.lightness`, `minimum` of `0`, and `maximum` of `1`.
```
jgerigmeyer: @jamesnw For HSL, isn't the max lightness `100`?
```suggestion
  `value` of `options.lightness`, `minimum` of `0`, and `maximum` of `100`.
```

proposal/color-4-new-spaces-js.d.ts.md
```diff
@@ -865,6 +899,18 @@ This procedure takes a channel value `value`, and returns the special value
 * If `value` is the Javascript value `null`, return the unquoted Sass string
   `none`.
 
+### Parsing a Clamped Channel Value
+
+This procedure takes a channel value `value` and an inclusive range of `minimum`
+and `maximum`. It asserts the value is in the range, and returns the special
+value `none` if the value is `null`.
+
+* If `value` is less than `minimum`, throw an error.
```
nex3: These should be fuzzy comparisons. If the user passes `lightness: 100.000000000001`, that should be treated as `lightness: 100`.

nex3 (changes requested):

proposal/color-4-new-spaces-js.d.ts.md
jamesnw: Updated in https://github.com/sass/sass/pull/3723/commits/d7bdd2ec33c588d208ea5235a82829f7660f8960

In addition to lightness and alpha, the dart-sass implementation is also throwing out of range errors for-

- `hwb.whiteness`
- `hwb.blackness`
- `hsl.saturation`

Should I add this behavior to the proposal as well?
What are the overarching rules that this should follow? I think it is "If a channel is clamped, throw out of range errors", but to check-

1. Does bounded vs un-bounded make a difference?
2. Does legacy vs non-legacy make a difference?
3. Does legacy with space specified vs legacy without space specified make a difference?

nex3: Don't treat the Dart Sass implementation as authoritative—it's lagging behind the spec in a few crucial ways. Currently, the _only_ clamped channels are `lightness` and `alpha`, and so those should be the only ones that throw out-of-range errors (although that may change as the result of https://github.com/w3c/csswg-drafts/issues/9484).

nex3 (approved):

PR #3728, opened 2023-10-23
[Package Importer] Add types and specs

jamesnw:
Continuation of https://github.com/sass/sass/pull/3660

[Issue](https://github.com/sass/sass/issues/2739)
 
Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/package-importer.d.ts.md) is accepted.

- https://github.com/sass/dart-sass/pull/2130
- https://github.com/sass/embedded-host-node/pull/260
- https://github.com/sass/sass-spec/pull/1952


js-api-doc/index.d.ts
```diff
@@ -15,6 +15,8 @@ export {
   FileImporter,
   Importer,
   ImporterResult,
+  nodePackageImporter,
+  NodePackageImporter,
```
jamesnw, line 22: @nex3 Are there examples of other classes we want to document but not export? If I don't export NodePackageImporter here, `typedoc` can't include it in the docs, and throws a warning- `[warning] NodePackageImporter, defined in ./js-api-doc/importer.d.ts, is referenced by nodePackageImporter but not included in the documentation.`

One option would be to add it to `[intentionallyNotExported](https://typedoc.org/options/validation/#intentionallynotexported)`, which would silence the warning. We would need to document what the `NodePackageImporter` type is within the `nodePackageImporter` and `Options.importers` documentation.

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,37 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
 
+/**
+ * An opaque branded type, representing the built-in Node Package Importer. This
+ * is used for type checking. Users should use the exposed `nodePackageImporter`
+ * and not instantiate their own.
+ *
+ * @category Importer
+ */
+type NodePackageImporter = {
+  _NodePackageImporterBrand: any;
```
jamesnw: @nex3 This has a `no explicit any` eslint warning. Should I change this to `never` instead?

nex3: I think `never` means that you're declaring that the object doesn't include the property. Does `unknown` work?

js-api-doc/index.d.ts
nex3: Let's just export the type. It makes sense that users might want to reference it in a union type or something at some point.

js-api-doc/importer.d.ts
jamesnw: `unknown` does work. I was suggesting `never` because the `_NodePackageImporterBrand` is a build-time only property, and never has to actually receive a value, and `never` is a stronger deterrent for users who may want to create their own. 

In  experimenting, `never` is more "the object requires the property is included, but it never has a value". Setting an object property to `never` doesn't seem to be documented by Typescript, as it is somewhat illogical, so I'm going to go with `unknown`.

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,37 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ * An opaque branded type, representing the built-in Node Package Importer. This
+ * is used for type checking. Users should use the exposed `nodePackageImporter`
+ * and not instantiate their own.
+ *
+ * @category Importer
+ */
+export type NodePackageImporter = {
+  _NodePackageImporterBrand: unknown;
+};
+/**
```
nex3: Style nit: Empty lines between declarations.

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,37 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
   nonCanonicalScheme?: string | string[];
 }
 
+/**
+ * An opaque branded type, representing the built-in Node Package Importer. This
+ * is used for type checking. Users should use the exposed `nodePackageImporter`
```
nex3: Use `{@link nodePackageImporter}` to link to the member rather than just referring to it by name.

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,37 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ * @category Importer
+ */
+export type NodePackageImporter = {
+  _NodePackageImporterBrand: unknown;
+};
+/**
+ * The built-in Node Package Importer, which is a Package Importer that resolves
+ * using the standards and conventions of the Node ecosystem. It enables a
+ * `pkg:` URL scheme for usage with `@use` that directs an implementation to
+ * resolve a URL within a dependency.
```
nex3: This will be the primary documentation that people will look to for information on this importer, so it's worth being particularly clear and thorough. A few specifics:

* "Node" should probably be "Node.js" throughout, since that's the full name of the platform.
* I don't think "Package Importer" should be capitalized—it's not a proper noun. That said, we _should_ find a place to document what it means to be a "package importer" and link to it here.
* "resolves using" isn't really proper grammar.
* "Enables a `pkg:` URL scheme" is pretty vague.
* "directs an implementation" is very confusing. Documentation inherently describes the behavior of an implementation—in what sense does directing an implementation to do something differ from just doing that thing?

I'd rewrite it as something like:

> The built-in Node.js [package importer](https://example.org). This loads `pkg:` URLs from `node_modules` according to the standard Node.js resolution algorithm.

You should also explicitly describe how to make a package compatible with it, all the ways it handles exports and style declarations in `package.json`, and so on. Remember, this is _the only place_ that a reader will be able to see this information—no one is going to go look up the proposal once this is released.

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,37 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ * A Node Package Importer is exposed as a constant that can be added to the
+ * `importers` option.
+ *
+ * @example
+ *
+ *```js
+ *const sass = require('sass');
+ *sass.compile('style.scss', {
+ *  importers: [sass.nodePackageImporter]
+ *});
```
nex3: This example should probably use `compileString()` so it can show an actual package import in use (ideally from a package that's already naturally compatible with it).

js-api-doc/legacy/options.d.ts
```diff
@@ -508,6 +508,13 @@ export interface LegacySharedOptions<sync extends 'sync' | 'async'> {
    * @compatibility dart: "1.43.0", node: false
    */
   logger?: Logger;
+
+  /**
+   * If this option is set to `'node'`, Sass will use the built-in Node Package
+   * Importer to resolve Sass files with a `pkg:` URL scheme.
+   */
+
```
nex3: No empty line between doc comments and members.

js-api-doc/legacy/options.d.ts
```diff
@@ -508,6 +508,13 @@ export interface LegacySharedOptions<sync extends 'sync' | 'async'> {
    * @compatibility dart: "1.43.0", node: false
    */
   logger?: Logger;
+
+  /**
+   * If this option is set to `'node'`, Sass will use the built-in Node Package
+   * Importer to resolve Sass files with a `pkg:` URL scheme.
+   */
```
nex3: Similar comments to `nodePackageImporter`. This doesn't need to be as detailed, but it should link to `nodePackageImporter` for full information.

js-api-doc/options.d.ts
```diff
@@ -208,8 +208,8 @@ export interface Options<sync extends 'sync' | 'async'> {
    * - The importer that was used to load the current stylesheet, with the
    *   loaded URL resolved relative to the current stylesheet's canonical URL.
    *
-   * - Each {@link Importer} or {@link FileImporter} in {@link importers}, in
-   *   order.
+   * - Each {@link Importer} or {@link FileImporter} or
+   *   {@link NodePackageImporter} in {@link importers}, in order.
```
nex3: "Importer, FileImporter, or NodePackageImporter"

spec/embedded_sass.proto
```diff
@@ -98,6 +98,11 @@ message InboundMessage {
         // generating this ID and ensuring that it's unique across all importers
         // registered for this compilation.
         uint32 file_importer_id = 3;
+
+        // A built-in Package Importer, with an associated `entry_point_url`,
+        // that resolves `pkg:` URLs using the standards and conventions of the
+        // Node ecosystem.
```
nex3: This should link to the specification for a Node.js package importer.

spec/embedded_sass.proto
```diff
@@ -1060,3 +1065,11 @@ enum CalculationOperator {
   // The division operator.
   DIVIDE = 3;
 }
+
+// The built-in Node Package Importer, which is a Package Importer that resolves
+//using the standards and conventions of the Node ecosystem. It enables a `pkg:`
```
nex3: Nit: missing space

spec/embedded_sass.proto
```diff
@@ -1060,3 +1065,11 @@ enum CalculationOperator {
...
+// URL scheme for usage with `@use` that directs an implementation to resolve a
+// URL within a dependency. It must be instantiated with an`entry_point_url`.
```
nex3: Describe what `entry_point_url` means.

nex3 (changes requested):

js-api-doc/importer.d.ts
jamesnw: Fixed.

js-api-doc/importer.d.ts
jamesnw: Fixed

js-api-doc/legacy/options.d.ts
jamesnw: Fixed

js-api-doc/options.d.ts
jamesnw: Fixed

spec/embedded_sass.proto
jamesnw: Added

spec/embedded_sass.proto
jamesnw: Fixed

js-api-doc/importer.d.ts
jamesnw: Updated to use '@angular/material'

spec/embedded_sass.proto
jamesnw: Updated with-

 It must be instantiated with an`entry_point_url`, which is the path to a file within the NPM package that contains the NPM packages that will be imported using the `pkg:` scheme.

js-api-doc/importer.d.ts
jamesnw: Updated in 9d1d3d2c4d9f04f9a3d4021cf06ee93fe089e531

js-api-doc/legacy/options.d.ts
jamesnw: Updated in 9d1d3d2c4d9f04f9a3d4021cf06ee93fe089e531

js-api-doc/importer.d.ts
```diff
@@ -309,32 +309,120 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
- * An opaque branded type, representing the built-in Node Package Importer. This
- * is used for type checking. Users should use the exposed `nodePackageImporter`
- * and not instantiate their own.
+ * An opaque branded type, representing the built-in Node.js Package Importer.
+ * This is used for type checking. Users should use the exposed
+ * {@link nodePackageImporter} and not instantiate their own.
  *
  * @category Importer
  */
 export type NodePackageImporter = {
```
nex3: Should this be a symbol type now?

js-api-doc/importer.d.ts
```diff
@@ -309,32 +309,120 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
- *
  *```js
  *const sass = require('sass');
- *sass.compile('style.scss', {
+ *sass.compileString('@use "pkg:@angular/material', {
  *  importers: [sass.nodePackageImporter]
  *});
  *```
+ *
+ * Package authors can control what is exposed to their users through their
```
nex3, line 329: The way this is currently structured, I believe this text is considered part of the `@example` block created by the `@example` tag above. That is, I think an `@example` covers everything beneath it until the next tag, rather than just being a way of annotating a specific block of code. Given that, I think you should move the example usage to the bottom of the documentation and remove the `@example` tag from other code blocks.

js-api-doc/importer.d.ts
```diff
@@ -309,32 +309,120 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ * - `@use "pkg:uicomponents/colors";` to import the colors partial.
+ * - `@use "pkg:uicomponents/theme/purple";` to import a purple theme.
+ *
+ * Note that while library users can rely on the importer to resolve
+ * [partials](https://sass-lang.com/documentation/at-rules/use#partials), [index
+ * files](https://sass-lang.com/documentation/at-rules/use#index-files), and
+ * extensions, library authors must specify the entire file path.
+ *
+ * In addition to the `sass` condition, the `style` condition is also
+ * acceptable, but authors should avoid using the `default` condition for Sass
```
nex3: I'd say "and Sass will match the `default` condition if it's a relevant file type but authors are discouraged from relying on this". The current phrasing makes it ambiguous about whether `default` works _in practice_ or not.

js-api-doc/importer.d.ts
```diff
@@ -309,32 +309,120 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ *
+ * This allows a user to write `@use "pkg:uicomponents";` to import the `index.scss` file.
+ *
+ * Finally, the Node.js package importer will look for an `index` file at the
+ * package root, resolving partials and extensions. For example, if the file
+ * `_index.scss` exists in the package root of `uicomponents`, a user can import
+ * that with `@use "pkg:uicomponents";`.
+ *
+ * If a `pkg:` URL includes a subpath that doesn't have a match in package entry
+ * points, the Node.js importer will attempt to find that file relative to the
```
nex3, line 412: "Relative to the package root"

js-api-doc/legacy/options.d.ts
```diff
@@ -510,10 +510,21 @@ export interface LegacySharedOptions<sync extends 'sync' | 'async'> {
...
-   * Importer to resolve Sass files with a `pkg:` URL scheme.
+   * If this option is set to `'node'`, Sass will use the built-in Node.js
+   * package importer to resolve Sass files with a `pkg:` URL scheme. Details
+   * for library authors and users can be found in the
+   * {@link nodePackageImporter} documentation.
+   *
+   * @example
+   * ```js
+   * sass.renderSync({
+   *  data: '@use "pkg:@angular/material";',
```
nex3: Nit: two spaces indentation.

spec/embedded_sass.proto
```diff
@@ -1066,10 +1068,12 @@ enum CalculationOperator {
...
-// The built-in Node Package Importer, which is a Package Importer that resolves
-//using the standards and conventions of the Node ecosystem. It enables a `pkg:`
-// URL scheme for usage with `@use` that directs an implementation to resolve a
-// URL within a dependency. It must be instantiated with an`entry_point_url`.
+// The built-in Node.js Package Importer, which is a Package Importer that
+// resolves using the standards and conventions of the Node.js ecosystem. It
+// enables a `pkg:` URL scheme for usage with `@use` that directs an
+// implementation to resolve a URL within a dependency. It must be instantiated
+// with an`entry_point_url`, which is the path to a file within the NPM package
+// that contains the NPM packages that will be imported using the `pkg:` scheme.
```
nex3: Since this is effectively spec text, rather than sort of vaguely describing `entry_point_file`'s general meaning I think it's better to specifically say "which is passed as <argument name> to the Node.js resolution algorithm <link>".

nex3 (changes requested):

js-api-doc/importer.d.ts
jamesnw: Updated in 06e5f4bbc94e7554231de8d29647a3d887b0fad5

js-api-doc/importer.d.ts
jamesnw: Updated in 7a75284a8431c990281aa89cea06f9cbc52ea976

js-api-doc/importer.d.ts
jamesnw: Updated in 7a75284a8431c990281aa89cea06f9cbc52ea976

js-api-doc/legacy/options.d.ts
jamesnw: Fixed in 7a75284a8431c990281aa89cea06f9cbc52ea976

spec/embedded_sass.proto
jamesnw: Updated in 7a75284a8431c990281aa89cea06f9cbc52ea976

js-api-doc/importer.d.ts
jamesnw: Ok, I updated this in 62f28c77578daf85d6b0e5b7bbc500f7df3212fe. I don't love that it puts the code for package users after the examples for package authors- do you have any recommended alternatives? Perhaps putting the package user example and info on the const `nodePackageImporter` and the package author info on the type `NodePackageImporter`?

nex3: Maybe a better solution is to forego the `@example` tag altogether and just use Markdown headers to discuss different subjects. That way you can have a usage example first with just a normal code block, then `## Writing Sass packages` afterwards.

spec/js-api/compile.d.ts.md
```diff
@@ -40,6 +40,22 @@ export interface CompileResult {
...
+* If any object in `options.importers` is exactly equal to the object
+  `nodePackageImporter`:
+
+  * If no filesystem is available, throw an error.
+
+    > This primarily refers to a browser environment, but applies to other
+    > sandboxed JavaScript environments as well.
+
+  * Let `pkgImporter` be a [Node Package Importer] with an associated
+    `entryPointURL` of `path`.
```
nex3: See discussion on the other PR about `entryPointURL`.

nex3 (changes requested):

spec/embedded_sass.proto
```diff
@@ -1060,3 +1067,13 @@ enum CalculationOperator {
...
 }
+
+// The built-in Node.js Package Importer, which is a Package Importer that
+// resolves using the standards and conventions of the Node.js ecosystem. It
+// enables a `pkg:` URL scheme for usage with `@use` that directs an
+// implementation to resolve a URL within a dependency. It must be instantiated
+// with an`entry_point_url`, which is passed as the `parentURL` to the Node.js
+// resolution algorithm.
+message NodePackageImporter {
+  string entry_point_url = 1;
```
nex3: Is there any case where this isn't going to be a `file:` URL? If not, should we make it a path instead of a URL?

spec/embedded_sass.proto
```diff
@@ -1060,3 +1067,13 @@ enum CalculationOperator {
   // The division operator.
   DIVIDE = 3;
 }
+
+// The built-in Node.js Package Importer, which is a Package Importer that
+// resolves using the standards and conventions of the Node.js ecosystem. It
+// enables a `pkg:` URL scheme for usage with `@use` that directs an
+// implementation to resolve a URL within a dependency. It must be instantiated
+// with an`entry_point_url`, which is passed as the `parentURL` to the Node.js
+// resolution algorithm.
```
nex3: The last sentence should probably be moved to the documentation of `entry_point_url`. Also, we should more clearly specify what a valid URL is (does it have to be absolute? does it have to be `file:`? if it is relative, to what?)

nex3 (changes requested):

js-api-doc/importer.d.ts
jamesnw: Updated in d78c4a502235ebea2c9b3aa49e58bd0cbe28b824

spec/embedded_sass.proto
jamesnw: Moved and documented in daf424181f903bfadeb37fcce494ea82623d298d

spec/embedded_sass.proto
jamesnw: This can be a path. I've updated that in daf424181f903bfadeb37fcce494ea82623d298d

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,133 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ * This allows a package user to write:
+ *
+ * - `@use "pkg:uicomponents";` to import the root export.
+ * - `@use "pkg:uicomponents/colors";` to import the colors partial.
+ * - `@use "pkg:uicomponents/theme/purple";` to import a purple theme.
+ *
+ * Note that while library users can rely on the importer to resolve
+ * [partials](https://sass-lang.com/documentation/at-rules/use#partials), [index
+ * files](https://sass-lang.com/documentation/at-rules/use#index-files), and
+ * extensions, library authors must specify the entire file path.
```
nex3: I'd add "in exports" to the end of this sentence just to be maximally clear.

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,133 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ *
+ * In addition to the `sass` condition, the `style` condition is also
+ * acceptable. Sass will match the default condition if it's a relevant file
+ * type, but authors are discouraged from relying on this. Notably, the key
+ * order matters, and the importer will resolve to the
```
nex3: Nit: looks like this needs a reflow.

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,133 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ * - `@use "pkg:uicomponents/colors";` to import the colors partial.
+ * - `@use "pkg:uicomponents/theme/purple";` to import a purple theme.
+ *
+ * Note that while library users can rely on the importer to resolve
+ * [partials](https://sass-lang.com/documentation/at-rules/use#partials), [index
+ * files](https://sass-lang.com/documentation/at-rules/use#index-files), and
+ * extensions, library authors must specify the entire file path.
+ *
+ * In addition to the `sass` condition, the `style` condition is also
+ * acceptable. Sass will match the default condition if it's a relevant file
```
nex3: Code-quote `default` here to match `sass` and `style` above.

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,133 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ * type, but authors are discouraged from relying on this. Notably, the key
+ * order matters, and the importer will resolve to the
+ * first value with a key that is `sass`, `style`, or `default`.
```
nex3: Maybe add "so you should always put `default` last."

js-api-doc/importer.d.ts
```diff
@@ -308,6 +308,133 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
...
+ * "pkg:uicomponents/src/sass/colors";`.
+ *
+ * @compatibility dart: "2.0", node: false
+ * @category Importer
+ */
+export class NodePackageImporter {
+  /**
+   * The NodePackageImporter has an optional `entryPointPath` option, which is
+   * the path where the Node Package Importer should start when resolving `pkg:`
+   * URLs. This will be used as the `parentURL` in the [Node Module
```
nex3: "when resolving `pkg:` URLs in sources other than files on disk."

js-api-doc/legacy/options.d.ts
```diff
@@ -508,6 +508,27 @@ export interface LegacySharedOptions<sync extends 'sync' | 'async'> {
...
+   *   pkgImporter: {type: 'node'}
+   * });
+   * ```
+   * @category Plugins
+   * @compatibility dart: "2.0", node: false
+   */
+  pkgImporter?: {
+    type: 'node';
+    entryPointPath?: string;
+  };
```
nex3: As per my comment on the other PR, I think this should just take an instance of `NodePackageImporter`.

spec/embedded_sass.proto
```diff
@@ -98,6 +98,13 @@ message InboundMessage {
         // generating this ID and ensuring that it's unique across all importers
         // registered for this compilation.
         uint32 file_importer_id = 3;
+
+        // The [Node.js package importer], which is a built-in Package Importer,
+        // with an associated `entry_point_url`, that resolves `pkg:` URLs using
```
nex3: Nit: remove the commas around "with an associated `entry_point_url`".

spec/embedded_sass.proto
```diff
@@ -1060,3 +1067,14 @@ enum CalculationOperator {
...
 }
+
+// The built-in Node.js Package Importer, which is a Package Importer that
+// resolves using the standards and conventions of the Node.js ecosystem. It
+// enables a `pkg:` URL scheme for usage with `@use` that directs an
+// implementation to resolve a URL within a dependency. 
+message NodePackageImporter {
+  // The path to associate with the Node Package Importer. It can either be an
+  // absolute path, or a relative path that will be resolved relative to the
+  // current working directory.
```
nex3: This should either explain or point to an explanation of the meaning and use of this path.

spec/js-api/compile.d.ts.md
```diff
@@ -91,6 +110,21 @@ export function compileAsync(
...
+
+    > This primarily refers to a browser environment, but applies to other
+    > sandboxed JavaScript environments as well.
+
+  * Let `pkgImporter` be a [Node Package Importer] with an associated
+    `entryPointURL` of `options.url` if it is a `file:` URL or the current
+    working directory otherwise.
+
+  * Replace `nodePackageImporter` with `pkgImporter` in a copy of
+    `options.importers`.
```
nex3: I don't think this is accurate any more.

nex3 (changes requested):

js-api-doc/importer.d.ts
jamesnw: Updated in ac235148468a3778500a37d757bef8236616b0de

js-api-doc/importer.d.ts
jamesnw: Updated in ac235148468a3778500a37d757bef8236616b0de

js-api-doc/importer.d.ts
jamesnw: Updated in ac235148468a3778500a37d757bef8236616b0de

js-api-doc/importer.d.ts
jamesnw: Updated in ac235148468a3778500a37d757bef8236616b0de

js-api-doc/importer.d.ts
jamesnw: Updated in ac235148468a3778500a37d757bef8236616b0de

js-api-doc/legacy/options.d.ts
jamesnw: Updated in ac235148468a3778500a37d757bef8236616b0de and dbf2e85ea705a040b9dc506b84d71cb709180b20

spec/embedded_sass.proto
jamesnw: Updated in https://github.com/sass/sass/commit/ac235148468a3778500a37d757bef8236616b0de

spec/embedded_sass.proto
```diff
@@ -98,6 +98,13 @@ message InboundMessage {
         // generating this ID and ensuring that it's unique across all importers
         // registered for this compilation.
         uint32 file_importer_id = 3;
+
+        // The [Node.js package importer], which is a built-in Package Importer
+        // with an associated `entry_point_url` that resolves `pkg:` URLs using
+        // the standards and conventions of the Node ecosystem.
+        //
+        // [Node.js package importer]: https://github.com/sass/sass/tree/main/spec/modules.md#node-package-importer
+        NodePackageImporter node_package_importer = 5;
```
ntkme, line 107: Any reason this is `5`, I see `4` is not used?

```suggestion
        NodePackageImporter node_package_importer = 4;
```

ntkme (changes requested):

spec/embedded_sass.proto
jamesnw, line 107: This conflicts with `non_canonical_scheme`, see https://github.com/sass/sass/actions/runs/6628127415/job/18004541179

spec/embedded_sass.proto
jamesnw: Pointed to JS docs in ec40399d98f7c378125f207f5ec375e9f83e42ec

spec/js-api/compile.d.ts.md
jamesnw: Applied draft changes in ec40399d98f7c378125f207f5ec375e9f83e42ec

js-api-doc/importer.d.ts
```diff
@@ -375,13 +376,14 @@ export interface Importer<sync extends 'sync' | 'async' = 'sync' | 'async'> {
  * Note that while library users can rely on the importer to resolve
  * [partials](https://sass-lang.com/documentation/at-rules/use#partials), [index
  * files](https://sass-lang.com/documentation/at-rules/use#index-files), and
- * extensions, library authors must specify the entire file path.
+ * extensions, library authors must specify the entire file path in exports.
```
nex3: Nit: `exports` should be code font, since it refers to a field in the package.json.

nex3 (approved):

js-api-doc/importer.d.ts
jamesnw: Updated in 41fc3132671d87f75175f02076e76299b318a7b7

spec/embedded_sass.proto
```diff
@@ -1060,3 +1067,17 @@ enum CalculationOperator {
...
+
+// The built-in Node.js Package Importer, which is a Package Importer that
+// resolves using the standards and conventions of the Node.js ecosystem. It
+// enables a `pkg:` URL scheme for usage with `@use` that directs an
+// implementation to resolve a URL within a dependency. 
+message NodePackageImporter {
+  // The path to associate with the Node Package Importer, with semantics
+  // identical to the [entryPointPath option] in the JavaScript API. It can
+  // either be an absolute path, or a relative path that will be resolved
+  // relative to the current working directory.
```
ntkme: We should enforce this to be only absolute path inside protocol. Relative path should be resolve by the embedded host.

See this comment for background: https://github.com/sass/embedded-host-node/pull/261#issuecomment-1877725010

ntkme (changes requested):

spec/embedded_sass.proto
jamesnw: Good catch, updated in https://github.com/sass/sass/pull/3728/commits/6f7bf999ceb65edb0f36923b1a32e19a178b984d

spec/EMBEDDED_PROTOCOL_VERSION
```diff
@@ -1 +1 @@
-2.3.0
+3.0.0-dev
```
ntkme: As we don't have any other planned protocol changes for next version right now, I suggest just change this to a stable version, so that it will be tagged automatically as we merge this PR.

jamesnw: Thanks- updated in https://github.com/sass/sass/pull/3728/commits/63c45082729730eb0c74cb83c0536e13fc7d9b1c

jgerigmeyer (approved):

ntkme (approved):

EMBEDDED_PROTOCOL_CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.4.0
```
ntkme: We now need a rebase and another version bump here.

jamesnw: Done in https://github.com/sass/sass/pull/3728/commits/1c36ec5577d75558f6f18e95c9f77358167b826b

ntkme: Please also update `spec/EMBEDDED_PROTOCOL_VERSION`. Thanks.

jamesnw: Apologies, done in https://github.com/sass/sass/pull/3728/commits/7ec5cb7b87a9e3b1eb495935953fc73ec0d9156f

nex3 (approved):

nex3 (changes requested):
Since `baseURL` is now a directory rather than a file path, and the Node.js algorithms expect a file path, we should explicitly add a basename before passing it into those algorithms.

jamesnw:
> Since `baseURL` is now a directory rather than a file path, and the Node.js algorithms expect a file path, we should explicitly add a basename before passing it into those algorithms.

@nex3 Based on my reading of the [algorithm](https://nodejs.org/api/esm.html#resolution-algorithm), the `parentURL`
doesn't make any assertions on whether it is a file or directory. Passing a file in a directory is identical to passing in the directory, as `parentURL` is only used for relative resolution. 

It's also unclear to me what we would use as a basename when the user only provides a directory.

ntkme:
I think what @nex3 means is that we can should pass `/path/to/dir` as `/path/to/dir/.`. Many times algorithm does not check if a path is a file or a directory, and this can potentially be dangerous. E.g. `dirname('/path/to/dir')` is different from `dirname('/path/to/dir/.')`, so that if an algorithm expect a file, it might be better to pass `/path/to/dir/.` to met the expectation.

accepted/package-importer.d.ts.md
```diff
@@ -290,19 +296,22 @@ any URL with the `pkg:` scheme. This step will be inserted immediately before
...
 
 Defaults to undefined.
 
 ```ts
+import {NodePackageImporter as BaseNodePackageImporter} from '../spec/js-api/importer';
+
 declare module '../spec/js-api/legacy/options' {
   export interface LegacySharedOptions<sync extends 'sync' | 'async'> {
-    pkgImporter?: NodePackageImporter;
+    pkgImporter?: BaseNodePackageImporter;
```
jgerigmeyer, line 314: @nex3 This workaround was required by the new symbol key, so that the two definitions of `pkgImporter` are now actually using the same definition of `NodePackageImporter`.

jamesnw:
> I think what @nex3 means is that we can should pass `/path/to/dir` as `/path/to/dir/.`. Many times algorithm does not check if a path is a file or a directory, and this can potentially be dangerous. E.g. `dirname('/path/to/dir')` is different from `dirname('/path/to/dir/.')`, so that if an algorithm expect a file, it might be better to pass `/path/to/dir/.` to met the expectation.

@nex3  Does this match your expectations- adding `.` as a basename, and updating the spec to read-

```md
  * Let `pkgImporter` be a [Node Package Importer] with an associated
    `entryPointURL` of the absolute file URL for `entryPointDirectory`
    appended with a [single-dot URL path segment](https://url.spec.whatwg.org/#single-dot-path-segment).
```

My reading of this is that this is only a change to the spec, and we don't need to update the implementation as well, correct?

nex3:
The `parentURL` parameter that's passed into `PACKAGE_RESOLVE` is ultimately used in two places:

1. In `PACKAGE_RESOLVE` itself, `node_modules/${packageSpecifier}` is resolved relative to `parentURL` and then `parentURL` is set to `dirname(parentURL)`.

2. It's passed to `LOOKUP_PACKAGE_SCOPE`, which immediately sets it to its own directory name at the beginning of the loop.

None of these operations actually looks at the basename. Even for resolving the relative path, `node_modules/foo` relative to `a/b/c` is `a/b/node_modules/foo`—the fact that the basename was `c` doesn't matter. This means, as @ntkme points out, that you need to pass in something like `a/b/c/.` if you want to actually check _within_ `a/b/c`. I don't know why the Node.js algorithm is written this way... it seems silly to me to pass in information that's just going to be thrown out, which is why I wanted to avoid doing the same thing in the Sass API.

All that said, I think the best place to compensate for a weird API like this is immediately before invoking it. Store the information "at rest" in a way that makes the most sense semantically—as the directory that we first check for `node_modules`—and only add `/.` immediately before invoking the Node.js algorithm.

> My reading of this is that this is only a change to the spec, and we don't need to update the implementation as well, correct?

That's right. The implementation now avoids the issue entirely by always operating in terms of directories in the first place.

jamesnw:
@nex3 Good point about appending `/.` immediately before invoking the algorithm. I added it in https://github.com/sass/sass/pull/3728/commits/e2517ca1d9646ecc7a1fe1b395fc5039d6f523c1.

nex3 (approved):

PR #3741, opened 2023-11-11
Only clamp channels on color creation

### LINKED ISSUE #3737: Color Spaces: Representing clamped out-of-gamut colors
Per https://github.com/w3c/csswg-drafts/issues/9484, it's possible for color space conversion to produce a color that's outside the gamut of the space it's represented in _even when that space is clamped at parsed-value time_. This means that `color.to-space(color(xyz 1 1 1), lab)` represents a valid and meaningful CSS color value in the Lab color space that cannot be represented with the `lab()` syntax because its lightness is `100.12`.

To handle edge cases like this, I think we should add a special case to the color serialization spec that checks whether a non-legacy color has a channel value outside its clamped range and, if so, emits an XYZ color (which has no range gamut limitations whatsoever) that's converted to the given color space via `color-mix()`. For example, `color.to-space(color(xyz 1 1 1), lab)` would be emitted as `color-mix(in lab, color(xyz-d65 1 1 1) 100%, black)`. This is a little silly in that Sass doesn't support the native CSS `color-mix()` function, but it's the only way I know of to ensure that the colors we emit represent the same colors as Sass's notion.

@jamesnw This probably means we should also allow out-of-range lightness values for Lab/OKLab/LCH/OKLCH in the JS API and the embedded protocol as well.

nex3:
Closes #3737


proposal/color-4-new-spaces.md
```diff
@@ -420,21 +427,66 @@ being deprecated in favor of color-space-friendly functions like
...
+allow the internal representation of these colors to go out-of-gamut.
+
+[but *not* in interpolation]: https://github.com/w3c/csswg-drafts/issues/9484
+
+The question remains as to how to handle cases that don't directly correspond to
+CSS, such as the `color.change()` function. Because out-of-gamut clamped
+channels are meaningful in CSS, we've chosen the design principle of preserving
+them in all situations where clamping isn't specifically mandated by CSS. In
+addition, to ensure that that the colors represent the same values in CSS that
+they do in Sass, they're be serialized to special formats that preserves their
```
stof:
```suggestion
they do in Sass, they'll be serialized to special formats that preserves their
```

or

```suggestion
they do in Sass, they're serialized to special formats that preserves their
```

proposal/color-4-new-spaces.md
```diff
@@ -420,21 +427,66 @@ being deprecated in favor of color-space-friendly functions like
...
+out-of-gamut values:
+
+* Out-of-gamut RGB colors with are serialized to `hsl()`, which per spec is not
```
stof:
```suggestion
* Out-of-gamut RGB colors are serialized to `hsl()`, which per spec is not
```

proposal/color-4-new-spaces.md
```diff
@@ -786,6 +845,18 @@ To serialize a non-legacy color `color`:
 
 * Let `space-name` be an unquoted lowercase string of `color`'s space name.
 
+* If `color` has a clamped channel whose value is out-of-bounds:
+
+  * Let `xyz` be the result of [converting] `color` into the `xyz` color space.
+
+  * Emit "color-mix(in ", followed by `space-name`, ", ", the result of
+    serializing `xyz`, and then " 100%, black)".
```
mirisuzanne, line 853: I get why we need something like this, but it also feels like a bit of a hack. Does it seem like a reliable solution long term, if we're stuck with it?

mirisuzanne (approved):

proposal/color-4-new-spaces.md
nex3: This is sort of a question for the CSS working group. Maybe the actual questions are twofold:

1. Is there an observable difference in a CSS file between representations of the same color in different spaces? That is, could we get away with just serializing this color as XYZ without the `color-mix()` translation layer?

2. If there _is_ an observable difference, is this translation layer likely to be stable going forward?

mirisuzanne: The other impact to consider here is: authors spend a lot of time in browser dev tools, especially when debugging. When they see this syntax, will it make any sense?

nex3: Probably not, but if it is observably different and there's no other way to represent out-of-gamut colors in plain CSS I don't see that we have much choice.

mirisuzanne: I guess I'm not totally convinced that it's our responsibility to help authors get-and-stay out-of-gamut in ways that CSS doesn't allow (and browsers handle quite poorly at the moment). Is that really an essential feature?

nex3: We could _potentially_ flex on this point, but it would involve compromising one of our other principles. Specifically:

* We could say that out-of-gamut colors just get serialized in a way that clamps in-browser, but this violates our principle that Sass representations of values should serialize to equivalent CSS representations.

* We could say that Sass _never generates_ out-of-gamut colors and just specify `color.to-space()` as clamping or gamut-mapping, but this violates the goal that colors should losslessly round-trip between space conversions which is kind of a foundational assumption for our design of a lot of color utilities.

Both of these would also mean that `color.mix()` wouldn't follow the same semantics as CSS, which in turn means that we'd never be able to add a Sass implementation of CSS's `color-mix()` without violating our core design principle of CSS compatibility.

Given that, I think generating messy output in these edge cases is a lower cost than these alternatives.

PR #3744, opened 2023-11-13
[Shared Resources] Add types and specs

jamesnw:
Continuation of [3719](https://github.com/sass/sass/pull/3719)

[Issue](https://github.com/sass/sass/issues/3296)

Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/shared-resources.d.ts.md) is accepted.

- https://github.com/sass/dart-sass/pull/2134
- https://github.com/sass/embedded-host-node/pull/261
- https://github.com/sass/sass-spec/pull/1954


jamesnw:
[I want to share resources across multiple invocations of the Sass compiler's JS API.](https://trello.com/c/BT4VgchI/4-i-want-to-share-resources-across-multiple-invocations-of-the-sass-compilers-js-api)

js-api-doc/compile.d.ts
```diff
@@ -37,6 +37,92 @@ export interface CompileResult {
...
+ *
+ * @category Compile
+ */
+export interface Compiler {
+  /**
+   * The {@link compile} method exposed through a Compiler instance while it is
+   * active. If this is called after {@link dispose} on the Compiler
+   * instance, an error will be thrown.
+   *
+   * During the Compiler instance's lifespace, given the same input, this will
```
nex3: "lifespace" -> "lifespan"

js-api-doc/compile.d.ts
```diff
@@ -37,6 +37,92 @@ export interface CompileResult {
...
+   * return an identical result to the {@link compileString} method exposed at
+   * the module root.
+   */
+  compileString(source: string, options?: StringOptions<'sync'>): CompileResult;
+
+  /**
+   * Ends the lifespan of this Compiler instance. After this is invoked, all
+   * calls to the Compiler instance's `compile` or `compileString` methods will
```
nex3: Link the method names.

js-api-doc/compile.d.ts
```diff
@@ -162,3 +248,61 @@ export function compileStringAsync(
   source: string,
   options?: StringOptions<'async'>
 ): Promise<CompileResult>;
+
+/**
+ * Creates an instance of a synchronous {@link Compiler}. Each compiler instance
+ * exposes the {@link compile} and {@link compileString} methods within the
+ * lifespan of the Compiler. To use asynchronous compilation, use
+ * {@link initAsyncCompiler};
```
nex3: ";" -> "."

js-api-doc/compile.d.ts
```diff
@@ -162,3 +248,61 @@ export function compileStringAsync(
   source: string,
   options?: StringOptions<'async'>
 ): Promise<CompileResult>;
+
+/**
```
nex3, line 276: Even though it's also explained in the individual `Compiler` methods, I think it's also worth mentioning here that these methods return the same values as the top-level `compile()` and `compileAsync()` methods since it's easy to think they might not.

js-api-doc/compile.d.ts
```diff
@@ -162,3 +248,61 @@ export function compileStringAsync(
...
+ * Creates an instance of a synchronous {@link Compiler}. Each compiler instance
+ * exposes the {@link compile} and {@link compileString} methods within the
+ * lifespan of the Compiler. To use asynchronous compilation, use
+ * {@link initAsyncCompiler};
+ *
+ * When using the `sass-embedded` npm package, this allows Sass to reuse a
```
nex3: Link to the package on npm.

nex3 (changes requested):

js-api-doc/compile.d.ts
jamesnw: 🤦🏻‍♂️ Updated in a32c8aabbe656fa869327b76140d539a60ea1535

js-api-doc/compile.d.ts
jamesnw: Updated in a32c8aabbe656fa869327b76140d539a60ea1535

js-api-doc/compile.d.ts
jamesnw: Updated in a32c8aabbe656fa869327b76140d539a60ea1535

js-api-doc/compile.d.ts
jamesnw: Updated in a32c8aabbe656fa869327b76140d539a60ea1535

js-api-doc/compile.d.ts
jamesnw: Updated in a32c8aabbe656fa869327b76140d539a60ea1535. Also added links other places the package was mentioned.

js-api-doc/compile.d.ts
```diff
@@ -99,12 +187,14 @@ export function compileAsync(
  *
  * This only allows synchronous {@link Importer}s and {@link CustomFunction}s.
  *
- * **Heads up!** When using the `sass-embedded` npm package,
+ * **Heads up!** When using the [sass-embedded] npm package,
```
jamesnw: @nex3 We may want to see if this is still true once the implementation is complete. The [initial benchmarking](https://github.com/sass/embedded-host-node/pull/261#issuecomment-1811187248) from @jerivas showed the difference between sync and async may now be negligible.

nex3: We'll need to test something that's more chatty... it makes sense that the difference would be very small when most of each compilation time is being spent in the compiler, but once a bunch of messages are being sent for each compilation that may change the equation.

js-api-doc/compile.d.ts
```diff
@@ -162,3 +252,67 @@ export function compileStringAsync(
...
+/**
+ * Creates an instance of a synchronous {@link Compiler}. Each compiler instance
+ * exposes the {@link compile} and {@link compileString} methods within the
+ * lifespan of the Compiler. Given identical input, these methods will return
+ * results identical to their counterparts exposed at the module root. To use
+ * asynchronous compilation, use {@link initAsyncCompiler}.
+ *
+ * When using the [sass-embedded] npm package, this allows Sass to reuse a
+ * process across multiple compilations, reducing the amount of time needed to
+ * create and destroy each process.
```
nex3: The second half of this sentence isn't totally accurate—you're not creating and destroying "each process" at all with a `Compiler` object, there's only one process the entire time. It might be better to just say something like "This is much faster than calling the top-level compile methods multiple times when using the sass-embedded npm package". The most important thing is the speed, the reason why is kind of an implementation detail.

jamesnw: Because there are multiple variables (number of compilations, implementation, compiler/top level), I went with the following in https://github.com/sass/sass/pull/3744/commits/5d4329bd5ff43007697a400245835e3951704a65 -

```
 * When calling the compile functions multiple times, using a compiler instance
 * with the [sass-embedded] npm package is much faster than using the top-level
 * compilation methods or the [sass] npm package.
 ```

File: js-api-doc/compile.d.ts (hunk 6 above)
Comment (jerivas) [reply js-api-doc/compile.d.ts]: We are now testing 1000 concurrent compilations on https://github.com/sass/sass-spec/pull/1954

Comment (nex3) [reply js-api-doc/compile.d.ts]: What's the result performance-wise?

**Review by nex3 — APPROVED** [review]

File: js-api-doc/compile.d.ts (hunk 6 above)
Comment (ntkme) [reply js-api-doc/compile.d.ts]: Async is much slower in dart version mainly because heavy use of `Future` and `FutureOr`. If you test with extremely simple and small input it won't show much difference but if you try to benchmark something large, e.g. bootstrap, you will still find async to be way slower.

Comment (nex3) [reply js-api-doc/compile.d.ts]: I'm specifically asking about the embedded host's performance in async versus sync mode. Either way, Dart Sass is running in sync mode.

Comment (jerivas) [reply js-api-doc/compile.d.ts]: I ran a new set of benchmarks here https://github.com/sass/embedded-host-node/pull/261#issuecomment-1857084145

File: js-api-doc/compile.d.ts [hunk 8]
```diff
@@ -37,18 +37,113 @@ export interface CompileResult {
...
 
+/**
+ * The result of creating a synchronous compiler. Returned by
+ * {@link initCompiler}.
+ *
+ * @category Compile
+ */
+export class Compiler {
+  /**Throws an error if constructed directly, instead of via
+   * {@link initCompiler}. */
```
nex3: Properly format this comment

js-api-doc/compile.d.ts
```diff
@@ -37,18 +37,113 @@ export interface CompileResult {
...
+
+/**
+ * The result of creating an asynchronous compiler. Returned by
+ * {@link initAsyncCompiler}.
+ *
+ * @category Compile
+ */
+export class AsyncCompiler {
+  /**Throws an error if constructed directly, instead of via
+   * {@link initAsyncCompiler}. */
```
nex3: This one too

js-api-doc/compile.d.ts
```diff
@@ -37,18 +37,113 @@ export interface CompileResult {
...
+  private constructor();
+  /**
```
nex3, line 96: Empty lines between members

nex3 (changes requested):

js-api-doc/compile.d.ts
```diff
@@ -140,10 +140,15 @@ export class AsyncCompiler {
...
- * concurrent compilations being blocked on main thread.
+ * **Heads up!** When using the [sass-embedded] npm package for single
+ * compilations, **{@link compileAsync} is almost always faster than
+ * {@link compile}**, due to the overhead of emulating synchronous messaging
+ * with worker threads and concurrent compilations being blocked on main thread.
+ *
+ * If you are running multiple compilations with the [sass-embedded] npm
+ * package, using an {@link AsyncCompiler} or a {@link Compiler} will be much
+ * faster on average, and the difference between synchronous and asynchronous
+ * compilation will likely not be significant.
```
ntkme, line 151: Is this true? If we use `Compiler` doesn't it means it has to compile files sequentially, while `AsyncCompiler` can initialize multiple Promises concurrently.

ntkme: Ok, I saw the benchmark code and apparently it did not consider the case where you can start multiple promises concurrently on `AsyncCompiler ` but you cannot start multiple compilations Compiler concurrently.

E.g. The following code shows the true benefit of concurrent compilations with async compiler, which sync compiler will not be able to compete.

``` js
const asyncCompiler = await initAsyncCompiler();
await Promise.all([
  asyncCompiler.compileAsync(path),
  asyncCompiler.compileAsync(path),
  asyncCompiler.compileAsync(path),
  asyncCompiler.compileAsync(path)
]);
```

js-api-doc/compile.d.ts
```diff
@@ -44,8 +44,9 @@ export interface CompileResult {
  * @category Compile
  */
 export class Compiler {
-  /**Throws an error if constructed directly, instead of via
-   * {@link initCompiler}. */
+  /** Throws an error if constructed directly, instead of via
```
nex3: Standard JSDoc style (as used by other comments here) has the first `/**` on its own line.

nex3 (changes requested):

js-api-doc/compile.d.ts
```diff
@@ -37,17 +37,120 @@ export interface CompileResult {
...
- * concurrent compilations being blocked on main thread.
+ * **Heads up!** When using the [sass-embedded] npm package for single
+ * compilations, **{@link compileAsync} is almost always faster than
+ * {@link compile}**, due to the overhead of emulating synchronous messaging
+ * with worker threads and concurrent compilations being blocked on main thread.
+ *
+ * If you are running multiple compilations with the [sass-embedded] npm
+ * package, using an {@link AsyncCompiler} or a {@link Compiler} will be much
+ * faster on average, and the difference between synchronous and asynchronous
+ * compilation will likely not be significant.
```
ntkme: https://github.com/sass/embedded-host-node/pull/261#issuecomment-1862093449

It turns out this conclusion was from testing of dart kernel-snapshot (JIT mode) without proper warmup and therefore is misleading.

ntkme (changes requested):

js-api-doc/compile.d.ts
jamesnw: Fixed in https://github.com/sass/sass/pull/3744/commits/11f1af3748befbd7e21e7a6e5cce55caf2b4ff8a

js-api-doc/compile.d.ts
jamesnw: Fixed in https://github.com/sass/sass/pull/3744/commits/11f1af3748befbd7e21e7a6e5cce55caf2b4ff8a

js-api-doc/compile.d.ts
jamesnw: Fixed in https://github.com/sass/sass/pull/3744/commits/11f1af3748befbd7e21e7a6e5cce55caf2b4ff8a

js-api-doc/compile.d.ts
jamesnw: I removed the second part of the sentence in https://github.com/sass/sass/pull/3744/commits/43be47316ad454f815d93424736f1db6c140bc0c.

js-api-doc/compile.d.ts
```diff
@@ -37,17 +37,121 @@ export interface CompileResult {
...
+   * During the Compiler instance's lifespan, given the same input, this will
+   * return an identical result to the {@link compileString} method exposed at
+   * the module root.
+   */
+  compileString(source: string, options?: StringOptions<'sync'>): CompileResult;
+
+  /**
+   * Ends the lifespan of this Compiler instance. After this is invoked, all
+   * calls to the Compiler instance's {@link compile} or {@link compileString} methods will
+   * result in an error.
```
nex3: Nit: looks like this paragraph needs to be reflowed.

js-api-doc/compile.d.ts
```diff
@@ -37,17 +37,121 @@ export interface CompileResult {
...
- * the overhead of emulating synchronous messaging with worker threads and
- * concurrent compilations being blocked on main thread.
+ * **Heads up!** When using the [sass-embedded] npm package for single
+ * compilations, **{@link compileAsync} is almost always faster than
+ * {@link compile}**, due to the overhead of emulating synchronous messaging
+ * with worker threads and concurrent compilations being blocked on main thread.
+ *
+ * If you are running multiple compilations with the [sass-embedded] npm
+ * package, using either an {@link AsyncCompiler} or a {@link Compiler} will be
+ * much faster on average.
```
nex3: Update this (and similar notes) with the results of the benchmarks in the other thread.

js-api-doc/compile.d.ts
```diff
@@ -162,3 +272,71 @@ export function compileStringAsync(
   source: string,
   options?: StringOptions<'async'>
 ): Promise<CompileResult>;
+
+/**
+ * Creates an instance of a synchronous {@link Compiler}. Each compiler instance
```
nex3: Nit: "an instance of a" is superfluous here and can be removed

nex3 (approved):

js-api-doc/compile.d.ts
jamesnw: Updated in 8a0acb29b9ea218c799156b3d6626b0af75938ff

js-api-doc/compile.d.ts
jamesnw: Updated in 8a0acb29b9ea218c799156b3d6626b0af75938ff

js-api-doc/compile.d.ts
jamesnw: I updated this in ab30a00c1950b27329817ca605d3a4a94b224926.

jamesnw (commented):
In documenting the benchmarks, I decided to also add a top level section to the README about speed, to pull it together into a single place. I also realized the README only mentioned `sass`, so I added info on `sass-embedded` there as well.

js-api-doc/README.md
```diff
@@ -1,13 +1,21 @@
-The [`sass` package] on npm is a pure-JavaScript package built from the [Dart
-Sass] implementation. In addition to Dart Sass's [command-line interface], it
-provides a JavaScript API that can be used to drive Sass compilations from
-JavaScript. It even allows an application to control {@link Options.importers |
+The Sass JavaScript API can be used to to drive Sass Compilations from JavaScript. It even allows an application to control {@link Options.importers |
```
nex3: Nit: Reflow this to match the 80-character line length.

js-api-doc/README.md
```diff
@@ -1,13 +1,21 @@
...
 how stylesheets are loaded} and {@link Options.functions | define custom
 functions}.
 
+The [`sass` package] on npm is a pure-JavaScript package built from the [Dart
+Sass] implementation, and includes Dart Sass's [command-line interface].
+
+The [`sass-embedded` package] on npm is a JavaScript wrapper around a Native
```
nex3: Nit:
```suggestion
The [`sass-embedded` package] on npm is a JavaScript wrapper around a native
```

js-api-doc/README.md
```diff
@@ -112,3 +120,38 @@ of Sass code by passing in a {@link LegacyStringOptions}.
...
+
+With the `sass` package, the synchronous calls will be faster than asynchronous
+calls due to the overhead of making the entire evaluation process asynchronous.
+While the {@link Compiler} and {@link AsyncCompiler} class are available, they
+aren't faster than than the module level compilation methods when using `sass`.
+
+### With the `sass-embedded` package
+
+The `sass-embedded` package provides significant speed improvements in certain
+situations, and is generally faster than `sass` for large compilations. When
```
nex3: I'd say "large or frequent compilations", since the shared resources API now makes it much faster to compile many small Sass files with the embedded host.

js-api-doc/README.md
```diff
@@ -112,3 +120,38 @@ of Sass code by passing in a {@link LegacyStringOptions}.
...
+
+While multiple factors go into how long Sass compilations take, there are
+general speed trends that can help you minimize your compilation time.
+
+### With the `sass` package
+
+With the `sass` package, the synchronous calls will be faster than asynchronous
+calls due to the overhead of making the entire evaluation process asynchronous.
+While the {@link Compiler} and {@link AsyncCompiler} class are available, they
+aren't faster than than the module level compilation methods when using `sass`.
```
nex3: Nit:
```suggestion
aren't faster than than the module-level compilation methods when using `sass`.
```

js-api-doc/README.md
```diff
@@ -112,3 +120,38 @@ of Sass code by passing in a {@link LegacyStringOptions}.
...
+
+### With the `sass-embedded` package
+
+The `sass-embedded` package provides significant speed improvements in certain
+situations, and is generally faster than `sass` for large compilations. When
+using the module level compilation methods, asynchronous calls are generally
```
nex3: Nit:
```suggestion
using the module-level compilation methods, asynchronous calls are generally
```

js-api-doc/README.md
```diff
@@ -112,3 +120,38 @@ of Sass code by passing in a {@link LegacyStringOptions}.
...
+faster than synchronous ones due to the overhead of emulating synchronous
+messaging with worker threads and concurrent compilations being blocked on the
+main thread.
+
+The {@link Compiler} and {@link AsyncCompiler} classes provide significant
+improvements when using the `sass-embedded` package. We are able to persist and
+reuse a process across multiple compilations, avoiding the need to repeatedly
+start up and tear down a process.
```
nex3: I'd frame this as the package doing the work rather than an ambiguous "we". Something like "These classes persist and reuse a single Dart process across multiple compilations"

js-api-doc/README.md
nex3, line 1: These are very nice additions, thanks for thinking of this!

js-api-doc/README.md
```diff
@@ -112,3 +120,38 @@ of Sass code by passing in a {@link LegacyStringOptions}.
...
+
+When compiling a single file using `sass-embedded`, there is not much difference
+between the synchronous and asynchronous methods. When compiling multiple times,
```
nex3: I think the crucial difference here is that `AsyncCompiler` is substantially faster when running multiple compilations _at the same time_. If users are running them sequentially, they won't see as much benefit.

nex3 (approved):

js-api-doc/README.md
jamesnw: Fixed in c8b1198eeb33347f8c5df28e6a5ff798d5375c09

js-api-doc/README.md
jamesnw: Fixed in c8b1198eeb33347f8c5df28e6a5ff798d5375c09

js-api-doc/README.md
jamesnw: Fixed in c8b1198eeb33347f8c5df28e6a5ff798d5375c09

js-api-doc/README.md
jamesnw: Fixed in c8b1198eeb33347f8c5df28e6a5ff798d5375c09

js-api-doc/README.md
jamesnw: Fixed in c8b1198eeb33347f8c5df28e6a5ff798d5375c09

js-api-doc/README.md
jamesnw: Fixed in c8b1198eeb33347f8c5df28e6a5ff798d5375c09

js-api-doc/README.md
jamesnw: Fixed in c8b1198eeb33347f8c5df28e6a5ff798d5375c09

js-api-doc/compile.d.ts
```diff
@@ -37,17 +37,121 @@ export interface CompileResult {
...
+ * {@link initCompiler}.
+ *
+ * @category Compile
+ */
+export class Compiler {
+  /**
+   * Throws an error if constructed directly, instead of via
+   * {@link initCompiler}.
+   */
+  private constructor();
```
ntkme, line 51: If we don't want to expose this `constructor`, I think we can just define `Compiler` as an `interface` instead of a `class`, and then have an internal `class CompilerImpl implements Compiler` that is not exported so that it cannot be invoked by the user. `initCompiler` basically just calls `new CompilerImpl()`.

jerivas: I think @nex3 mentioned that we want to expose the actual class so `instanceof` checks work as expected.

ntkme (approved):

jgerigmeyer (approved):

PR #3748, opened 2023-11-17
[Package Importer] Draft 1.4 changes

jamesnw:


proposal/package-importer.d.ts.md
```diff
@@ -225,9 +225,7 @@ import {FileImporter, Importer} from '../spec/js-api/importer';
 ### `nodePackageImporter`
 
 ```ts
-export type NodePackageImporter = {
-  _NodePackageImporterBrand: unknown;
-};
+export type NodePackageImporter = Symbol;
```
stof: You should not alias `NodePackageImporter` as the `Symbol` type IMO. This would mean that any Symbol matches this type.

What we need is the `unique symbol` type of typescript. This way, the type checker enforces using that symbol exactly.

```ts
export declare const nodePackageImporter: unique symbol;
export type NodePackageImporter: typeof nodePackageImporter;
```

jamesnw: Good point, I updated this.

proposal/package-importer.d.ts.md
```diff
@@ -261,10 +259,20 @@ Javascript Compile API, insert:
     > This primarily refers to a browser environment, but applies to other
     > sandboxed JavaScript environments as well.
 
-  * Let `pkgImporter` be a [Node Package Importer] with an associated
-    `entryPointURL` of `require.main.filename`.
+For [`compile`], insert:
 
-  * Replace `nodePackageImporter` with `pkgImporter` in a copy of
+* Let `pkgImporter` be a [Node Package Importer] with an associated
+    `entryPointURL` of `path`.
```
nex3: Nit: extra indentation here and below.

nex3: I think `entryPointURL` should still be `require.main.filename`. In a Node.js context, we want package resolution to match that of the host process.

nex3 (changes requested):

proposal/package-importer.d.ts.md
jamesnw: There's a few reasons I wanted to drop `require.main.filename`-

- It's only available in CommonJS
- The ECMAScript equivalent of `import.meta.filename` won't be available until [Node v21.2.0](https://nodejs.org/api/esm.html#importmetafilename)
- `require.main` appeared to be undefined via JS Interop in Dart

If we are using `require.main.filename` to mean the concept rather than the value, then I think there are questions about what that means as a concept. 

In the case of a user running `npm run dev` in a Vite project with Vue, is it the package.json that contains the `script`, the Vite entry point, the Vite file that contains the `sass.compile` call, or the Vue template file that contains the Sass? In almost all cases, any of these will work interchangeably.

As @ntkme [points out](https://github.com/sass/dart-sass/pull/2130#discussion_r1389612551), using path, url, or CWD matches the behavior of Node, and starts resolution from the main entry point to Sass.

jamesnw: Fixed

nex3: (We discussed this offline, but writing here as well for posterity)

The critical thing is that we need to match the package resolution behavior of the host Node.js process. The point of Node.js-based `pkg:` resolution is that the user can install Sass packages _alongside_ their normal npm dependencies, so we need to ensure that we use the same resolution. If they try to compile a Sass file that happens to exist in a different package tree, it's bad news if we use that package tree's dependencies instead of the ones of the running process. (Critically, I'm pretty sure this also matches the behavior of the current WebPack resolver which will make migration easier.)

jamesnw: Looking specifically at Webpack's algorithm, the package tree used is not the tree of the running process- it is the package tree of the `entry` defined in its config. The value of `require.main.filename` is `[package]/node_modules/webpack/bin/webpack.js`, but if the default `entry`  of `[package]/src/index.js` is used, `[package]/src/node_modules` is examined, but not `[package]/node_modules/webpack/bin/node_modules`. In both instances, the most likely case of `[package]/node_modules` is considered. It could give different results for nested packages or a globally installed `webpack` package.

Given this, I think the proposed path/url/CWD strategy most closely follows the Webpack expectations. If a Sass file is found by Webpack in a sibling dependency tree, I do think we would want to start with the node_modules most proximate to that file, and not in a subdirectory of the Webpack package.

nex3: That seems to me like more of an argument for making the entrypoint configurable than anything—as far as I know, there's not a _strong_ guarantee that all Sass entrypoints WebPack compiles will be children of the package that installed webpack (although it's certainly likely in most cases). Maybe `NodePackageImporter` should be a class that takes a path parameter rather than just a symbol—which also more directly represents the API we provide for embedded hosts.

I don't think the "path/url/CWD" strategy makes sense here. If we're resolving a `pkg:` import within a `file:` URL, we _always_ use the `file:` URL as `parentURL` for Node's `PACKAGE_RESOLVE` algorithm. This question is only relevant in the first place if we're handling `pkg:` URLs in non-`file:` sources, like a user writing `sass.compileString("@use 'pkg:foo'", ...)`. So in almost all cases, "path/url/CWD" just boils down to "use the current working directory" which I don't think is sufficiently robust.

jamesnw: In https://github.com/sass/sass/pull/3748/commits/ea86702f235b785599dbc3507c5101f92d4f3a5f, I made the entry point configurable. This removes the opaque nodePackageImporter symbol, and moves it to the same format as the other importers that take options. Now that it is configurable, it seemed less crucial that it is a specific symbol, but I'm open to other thoughts on that.

I also plan on typing `findFileUrl` and `canonicalize` as @hidden `never`, and adding `pkgImporter` as never to `Importer` and `FileImporter`- does that need to happen in the proposal, or just when applying it to the spec?

proposal/package-importer.d.ts.md
```diff
@@ -216,17 +216,27 @@ using symlinks if this behavior is desired.
 
 [ECMAScript modules]: https://nodejs.org/api/esm.html#no-node_path
 
+The [Node resolution algorithm] requires a `parentURL`, used for determining
+where in the file system to start searching for a module. When compiling a
```
nex3: "...to start searching for a module **if a `pkg:` URL is being resolved somewhere other than a file on disk**."

proposal/package-importer.d.ts.md
```diff
@@ -216,17 +216,27 @@ using symlinks if this behavior is desired.
...
-### `nodePackageImporter`
+### `NodePackageImporter`
 
 ```ts
-export declare const nodePackageImporter: unique symbol;
-export type NodePackageImporter = typeof nodePackageImporter;
+export interface NodePackageImporter {
+  pkgImporter: 'node';
+  entryPointPath?: string;
+}
```
nex3: I think we should make this a proper class, so that users call it as:

```js
sass.compileString(stylesheet, {
  importers: [new sass.NodePackageImporter(entrypoint)]
});
```

nex3 (changes requested):

proposal/package-importer.d.ts.md
jamesnw: Updated in https://github.com/sass/sass/pull/3748/commits/3afb76cc2bf5df211040bfb51323f4e5a0fd074e

proposal/package-importer.d.ts.md
jamesnw: In https://github.com/sass/sass/pull/3748/commits/3afb76cc2bf5df211040bfb51323f4e5a0fd074e, I changed this, with the addition of "...is
being resolved **in a source** somewhere other than..." to clarify we're talking about resolved **from** as opposed to the resolution destination.

proposal/package-importer.d.ts.md
```diff
@@ -216,19 +216,28 @@ using symlinks if this behavior is desired.
...
-### `nodePackageImporter`
+### `NodePackageImporter`
 
 ```ts
-export type NodePackageImporter = {
-  _NodePackageImporterBrand: unknown;
-};
-export declare const nodePackageImporter: NodePackageImporter;
+export class NodePackageImporter {
+  entryPointPath?: string;
```
nex3: I don't think we need to expose this publicly once the class is constructed. (And if we did, we wouldn't want it to be `undefined`, since it does have a well-defined default value.)

proposal/package-importer.d.ts.md
```diff
@@ -281,12 +293,19 @@ the legacy importer logic will be invoked.
...
 Defaults to undefined.
 
 ```ts
 declare module '../spec/js-api/legacy/options' {
   export interface LegacySharedOptions<sync extends 'sync' | 'async'> {
-    pkgImporter?: 'node';
+    pkgImporter?: {
+      type: 'node';
+      entryPointPath?: string;
+    };
```
nex3: I think this should just take a `NodePackageImporter` instance.

nex3 (changes requested):

proposal/package-importer.d.ts.md
jamesnw: Updated in 64dc0a12696821062abfae2c1d33c9247b1fe1b6

proposal/package-importer.d.ts.md
jamesnw: Fixed in 64dc0a12696821062abfae2c1d33c9247b1fe1b6

nex3 (approved):

PR #3773, opened 2024-01-10
Add CompileRequest.silent to embedded protocol

ntkme:
This is a performance enhancement that if the special js-api `Logger.silent` is requested by the user, we can set `CompileRequest.silent` to completely bypass the unnecessary serialization, sending, and deserialization of `LogEvent`s, thus making the compilation more efficient in the case where lots of log messages are supposed to be suppressed.


nex3:
Treating this as a fast-track feature.

nex3 (approved):

PR #3777, opened 2024-01-19
[Package Importer] Draft 1.5 

jamesnw:
- Specify no throws for package importer misses

See https://github.com/sass/dart-sass/pull/2130/files/e04b10982423d44237cb4d976c460e59478d731d#r1458161774

### DIFF
--- a/proposal/package-importer.changes.md
+++ b/proposal/package-importer.changes.md
@@ -3,6 +3,9 @@
 * Specify that Package Importers must not reject URL patterns that other Package
   Importers may be able to canonicalize.
 
+* Specify that the Node Package Importer throws if `entryPointPath` is not
+    passed and `require.main.filename` is not available.
+
 ## Draft 1.4
 
 * Allow NodePackageImporter to accept an entry point path.
--- a/proposal/package-importer.changes.md
+++ b/proposal/package-importer.changes.md
@@ -1,3 +1,8 @@
+## Draft 1.5
+
+* Specify that Package Importers must not reject URL patterns that other Package
+  Importers may be able to canonicalize.
+
 ## Draft 1.4
 
 * Allow NodePackageImporter to accept an entry point path.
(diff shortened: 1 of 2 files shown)


proposal/package-importer.d.ts.md
```diff
@@ -330,6 +330,10 @@ Package Importers must reject the following patterns:
 * A URL whose path begins with `/`.
 * A URL with non-empty/null username, password, host, port, query, or fragment.
 
+If the conventions or specifications for an environment disallow any other URL
+patterns, the Package Importer must not reject, and instead return `null`. This
```
nex3: "reject" isn't a well-defined term here. I'd say "must return `null` rather than throwing an error."

nex3 (changes requested):
It's probably also a good idea to specify that `new NodePackageImporter()` throws an error if a path isn't passed and the default can't be loaded.

proposal/package-importer.d.ts.md
jamesnw: Updated in https://github.com/sass/sass/pull/3777/commits/26d30d7b62448ee60074eb6e8d6a5cbbd727cb0d

jamesnw:
> It's probably also a good idea to specify that `new NodePackageImporter()` throws an error if a path isn't passed and the default can't be loaded.

Added in 517b1e2c16e25d5b7a3728dbb4f525993564cd35

proposal/package-importer.changes.md
```diff
@@ -1,3 +1,11 @@
+## Draft 1.5
+
+* Specify that Package Importers must not reject URL patterns that other Package
+  Importers may be able to canonicalize.
+
+* Specify that the Node Package Importer throws if `entryPointPath` is not
+    passed and `require.main.filename` is not available.
```
nex3: Nit: fix indentation

nex3 (approved):

proposal/package-importer.changes.md
jamesnw: Fixed in 3ac835a42cf5a7f691d7a92d613cbf81eaf29531

PR #3793, opened 2024-02-22
[Hotfix Node Package Importer]- Handle subpath without extensions

jamesnw:
https://github.com/sass/dart-sass/issues/2183

Adds `subpath` to the list of `export load paths`.

dart-sass- https://github.com/sass/dart-sass/pull/2184
sass-spec- https://github.com/sass/sass-spec/pull/1960

### DIFF
--- a/accepted/package-importer.d.ts.md
+++ b/accepted/package-importer.d.ts.md
@@ -554,8 +554,8 @@ potential subpaths, resolving for partials and file extensions.
 
   * Add `subpath` to `paths`.
 
-* Otherwise, add `subpath` + `.scss`, `subpath` + `.sass`, and `subpath` +
-  `.css` to `paths`.
+* Otherwise, add `subpath`, `subpath` + `.scss`, `subpath` + `.sass`, and
+  `subpath` + `.css` to `paths`.
 
 * If `subpath`'s [basename] does not start with `_`, for each `item` in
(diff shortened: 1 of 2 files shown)


jamesnw:
@nex3 Let me know if there is more process (proposal, draft, etc) needed, or if this qualifies as a fast track fix.

accepted/package-importer.d.ts.md
```diff
@@ -554,8 +554,8 @@ potential subpaths, resolving for partials and file extensions.
 
   * Add `subpath` to `paths`.
 
-* Otherwise, add `subpath` + `.scss`, `subpath` + `.sass`, and `subpath` +
-  `.css` to `paths`.
+* Otherwise, add `subpath`, `subpath` + `.scss`, `subpath` + `.sass`, and
```
ntkme, line 557: I think we should make it explicit to only add the extensions when it's not specified. If it already has an extension (.scss, .sass, or .css), we should not try add another.

jamesnw: This is cut off in this snippet, but the preceding lines contain an "if it has an extension", and this is the else.

```
* If `subpath` ends in `.scss`, `.sass`, or `.css`:

  * Add `subpath` to `paths`.

* Otherwise...
```

nex3 (approved):

PR #3826, opened 2024-04-01
Integrate the deprecations API into the spec/docs/protocol

jathak:
In the process of implementing this, I made two minor changes to the embedded protocol part of the proposal: (1) updating the field IDs since a new field got added to `CompileRequest` since the proposal was written and (2) adding a `deprecation_type` field to `LogEvent` so that the embedded host can pass that information to the JS API. Let me know if these changes are fine to just do here, or if I should factor them out into a separate PR.

[skip dart-sass]
[skip sass-embedded]


js-api-doc/deprecations.d.ts
```diff
@@ -0,0 +1,207 @@
...
+   * Deprecation for `@import` rules.
+   *
+   * This deprecation is not yet active, but will be soon.
+   */
+  import: Deprecation<'import'>;
+
+  /**
+   * Used for any user-emitted deprecation warnings.
+   */
+  'user-authored': Deprecation<'user-authored', 'user'>;
```
nex3, line 119: Do we have any testing in place that will catch if we add a deprecation to Dart Sass's enum that isn't listed here? If not, we should—I'm confident I'll make that mistake in the future :sweat_smile:

js-api-doc/logger/index.d.ts
```diff
@@ -43,6 +44,8 @@ export interface Logger {
    *
    * @param message - The warning message.
    * @param options.deprecation - Whether this is a deprecation warning.
+   * @param options.dperecationType - The type of deprecation this warning is
```
nex3:
```suggestion
   * @param options.deprecationType - The type of deprecation this warning is
```

js-api-doc/options.d.ts
```diff
@@ -276,6 +302,20 @@ export interface Options<sync extends 'sync' | 'async'> {
...
   quietDeps?: boolean;
 
+  /**
+   * A set of active deprecations to ignore.
+   *
+   * If a deprecation warning of any provided type is encountered during
+   * compilation, the compiler will ignore it instead.
+   *
+   * **Heads up!** The deprecated functionality you're depending on will
+   * eventually break
```
nex3: Nit: period

js-api-doc/deprecations.d.ts
```diff
@@ -0,0 +1,207 @@
...
+   * @param minor - The minor version.
+   * @param patch - The patch version.
+   */
+  constructor(major: number, minor: number, patch: number);
+  readonly major: number;
+  readonly minor: number;
+  readonly patch: number;
+
+  /**
+   * Parses a version from a string.
```
nex3, line 194: This should probably document that it'll throw an error if the version can't be parsed.

spec/js-api/deprecations.d.ts.md
```diff
@@ -0,0 +1,338 @@
+# Deprecations API
+
+> Interfaces for user-declared importers that customize how Sass loads
+> stylesheet dependencies.
```
nex3: Copy-paste error

spec/js-api/deprecations.d.ts.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Types
+
+### `Deprecations`
+
+```ts
+export interface Deprecations {
+```
+
+#### `call-string`
```
nex3: I think we can get away without summaries here, since the individual specs say "emit a deprecation named ...".

spec/js-api/deprecations.d.ts.md
```diff
@@ -0,0 +1,338 @@
...
+originating from user-authored code, this is null.
+
+```ts
+deprecatedIn: status extends 'future' | 'user' ? null : Version;
+```
+
+#### `obsoleteIn`
+
+The compiler version this feature was fully removed in, making the deprecation
+obsolete.
```
nex3, line 127: Probably worth mentioning that this is also implementation-dependent.

spec/js-api/deprecations.d.ts.md
```diff
@@ -0,0 +1,338 @@
...
+
+```ts
+constructor(major: number, minor: number, patch: number);
+```
+
+#### `major`
+
+The major version.
+
+This must be a non-negative integer.
```
nex3: Since the spec is describing how an implementation should _behave_, it should define what happens if the number is negative rather than just declaring that it must not be.

spec/js-api/deprecations.d.ts.md
```diff
@@ -0,0 +1,338 @@
...
+
+```ts
+readonly patch: number;
+```
+
+#### `parse`
+
+Parses a string in the form "major.minor.patch" into a `Version`.
```
nex3, line 195: Similarly, this should define what happens if parsing fails.

nex3 (changes requested):

nex3:
We should also explicitly specify what happens if the host refers to a deprecation ID that the compiler doesn't recognize or vice-versa. In the former case, I think the compiler should just throw an error, because the host should generally be responsible for ensuring it's speaking a compatible protocol to the compiler. In the latter case, I think it's fine to say that the warning is treated as a non-deprecation.

js-api-doc/logger/index.d.ts
```diff
@@ -51,6 +54,7 @@ export interface Logger {
     message: string,
     options: {
       deprecation: boolean;
+      deprecationType?: Deprecation;
```
nex3: We should probably type this explicitly so that `deprecationType` is set if and only if `deprecation` is true.

js-api-doc/deprecations.d.ts
jathak: Done

js-api-doc/deprecations.d.ts
jathak: I've added a test to https://github.com/sass/sass-spec/pull/1967 to ensure that Dart Sass and the embedded host don't expose any extra deprecations, which I think should be sufficient until we switch the deprecations list to a single source-of-truth (tracked in https://github.com/sass/sass/issues/3827)

js-api-doc/logger/index.d.ts
jathak: Done

js-api-doc/logger/index.d.ts
jathak: Done. Is this the cleanest way to write this type, or is there some fancy Typescript feature to make `deprecationType` directly dependent on `deprecation`'s value?

js-api-doc/options.d.ts
jathak: Done

spec/js-api/deprecations.d.ts.md
jathak: Done.

spec/js-api/deprecations.d.ts.md
jathak: Done (updated the constructor spec to say the compiler should throw an error)

spec/js-api/deprecations.d.ts.md
jathak: Done

spec/js-api/deprecations.d.ts.md
jathak: Removed

spec/js-api/deprecations.d.ts.md
jathak: Fixed

js-api-doc/logger/index.d.ts
nex3: I think this is the best option

spec/js-api/deprecations.d.ts.md
nex3: You could even get rid of the headers and have this all be a single code block.

spec/js-api/deprecations.d.ts.md
```diff
@@ -0,0 +1,314 @@
...
+
+#### `id`
+
+A kebab-case ID for this deprecation.
+
+```ts
+id: id;
+```
+
+The status of this deprecation.
```
nex3, line 89: Missing a header.

nex3 (approved):

Issue #3702, opened 2023-09-27
Deprecate and remove meta.feature-exists()


nex3:
- [x] [Proposal](https://github.com/sass/sass/blob/main/accepted/remove-feature-exists.md) (https://github.com/sass/sass/pull/3749)
- **Deprecation**
  - [x] Update official spec (https://github.com/sass/sass/pull/3860)
  - [x] Dart Sass (https://github.com/sass/dart-sass/pull/2235/)
  - [x] Tests (https://github.com/sass/sass-spec/pull/1986)
  - [x] Website (https://github.com/sass/sass-site/pull/1148)
- **Removal**
  - [x] Update official spec (https://github.com/sass/sass/pull/4230)
  - [x] Dart Sass (https://github.com/sass/dart-sass/pull/2781)
  - [x] Tests (https://github.com/sass/sass-spec/pull/2141)
  - [ ] Website

---

From https://github.com/sass/sass-spec/issues/1939#issuecomment-1738040431: 

> In practice, essentially all the new features we add fall into one of three categories:
>
> 1. New built-in functions or mixins, which are easy to detect using other `sass:meta` functions.
> 2. Language-level features which are relatively easy to detect on their own (for example, first-class calc can be detected with `calc(1 + 1) == 2`).
> 3. New syntax which can't even be parsed in implementations that don't support it, for which feature detection isn't particularly useful anyway.

This, combined with the fact that we haven't added a new feature to `meta.feature-exists()` in years, suggests that it's not useful anymore and should be removed.

nex3:
https://github.com/sass/sass/pull/3749 has landed and the proposal is out for review. Since it's a breaking change, I'll give it a month for public comment before we can mark it as accepted and implement.

nex3:
@stof We just released 1.77.0, so these will need to be 1.78.0.

stof:
@nex3 I updated all PRs to mention 1.78.0, and I also opened the missing PR for the website to document the deprecation.

stof:
@nex3 how should all the PRs related to this be handled ? As they haven't been merged back then, they conflict again. I'd rather avoid doing more rebases if they get stuck in waiting time again.

nex3:
[...]

PR #3828, opened 2024-04-06
Mark the "Remove feature-exists" proposal as accepted

stof:
The proposal has been out since almost 3 months now.

Refs https://github.com/sass/sass/issues/3702


nex3:
Ah, thanks!

PR #3835, opened 2024-04-12
Implement access tracking for containingUrl

ntkme:
https://github.com/sass/dart-sass/issues/2208


spec/embedded_sass.proto
```diff
@@ -192,6 +192,9 @@ message InboundMessage {
       // `null` instead.
       string error = 3;
     }
+
+    // Whether `containing_url` in `CanonicalizeRequest` is unused.
+    bool containing_url_unused = 4;
```
ntkme, line 203: In protobuf, `bool` default to `false` when it's unset.

The default `unused = false` is effectively `accessed = true`, which means `cacheable = false`. This is the ideal default, that it works correctly even if the host does not implement the tracking, making the protocol backward compatible.

nex3: Can you mention this reasoning in the documentation?

spec/embedded_sass.proto
```diff
@@ -257,6 +260,9 @@ message InboundMessage {
       // An error message explaining why the URL could not be loaded.
       string error = 3;
     }
+
+    // Whether `containing_url` in `FileImportRequest` is unused.
+    bool containing_url_unused= 4;
```
nex3: Style nit:
```suggestion
    bool containing_url_unused = 4;
```

nex3 (changes requested):

spec/embedded_sass.proto
ntkme: Added. Please feel free to edit this PR if you would like to rephrase the language.

nex3 (approved):

PR #3836, opened 2024-04-13
[Color 4] Require an explicit gamut mapping method

nex3:
[skip dart-sass]
[skip sass-embedded]


mirisuzanne (approved):

accepted/color-4-new-spaces.md
```diff
@@ -325,20 +327,20 @@ one or more of its channels out of bounds, like `rgb(300 0 0)`).
 
 #### `color.to-gamut()`
 
-This function returns a color that is in the given gamut, using the recommended
-[CSS Gamut Mapping Algorithm][css-mapping] to 'map' out-of-gamut colors into
-the desired gamut with as little perceptual change as possible. In many cases
-this can be more reliable for generating fallback values, rather than the
-'channel clipping' approach used by current browsers.
+This function returns a color that is in the given gamut, using the current CSS
+Color 4's `local-minde` algorithm to 'map' out-of-gamut colors into the desired
```
Goodwine: Nit: I find "using the current CSS Color 4's `local-minde` algorithm" slightly confusing because I'm not sure what exactly "current" is supposed to mean. Is that referring to date/version, or is it referring to choose one of the (hypothetically) many "local-minde" algorithms. I assume it's the former since the latter doesn't make much sense. But because I'm not familiar with the "local-minde" algorithm, I read it as a generic term such as "hashing" or "sorting". And I end up confused when swapping-in one of those words for "local-minde".

I wonder if something like this would be less ambiguous:

"using the `local-minde` algorithm defined in the current CSS Color 4 spec" or something like that 🤔

accepted/color-4-new-spaces.md
```diff
@@ -1016,28 +1031,48 @@ The individual conversion algorithms are:
...
+destination of `destination`.
+
+[css-mapping]: https://www.w3.org/TR/2024/CRD-css-color-4-20240213/#css-gamut-mapping-algorithm
+
 > This algorithm implements a relative colorimetric intent, and colors inside
-> the destination gamut are unchanged. Since the process is lossy, authors
-> should be encouraged to let the browser handle gamut mapping when possible.
+> the destination gamut are unchanged.
+
+#### `clip`
```
Goodwine, line 1064: (non actionable)
Interesting, I thought this method would have been called `clamp`, I guess both work 😮

js-api-doc/value/color.d.ts
```diff
@@ -269,14 +269,26 @@ export class SassColor extends Value {
...
+   *   candidate recommendation]. This maps in the Oklch color space, using the
+   *   [deltaEOK] color difference formula and the [local-MINDE] improvement.
+   *
+   * * `clip`: Clamp each color channel that's outside the gamut to the minimum
+   *   or maximum value for that channel. This algorithm will produce poor
+   *   visual results, but it may be useful to match the behavior of other
+   *   situations in which a color can be clipped.
+   *
+   * [the original Color Level 4 candidate recommendation]: https://www.w3.org/TR/2024/CRD-css-color-4-20240213/#css-gamut-mapping
+   * [the original Color Level 4 candidate recommendation]: https://www.w3.org/TR/2024/CRD-css-color-4-20240213/#color-difference-OK
```
Goodwine: Is this link supposed to be `deltaEOK`?

Goodwine (changes requested):

accepted/color-4-new-spaces.md
nex3: Rewritten.

accepted/color-4-new-spaces.md
nex3: "clamp" is usually used for individual colors, while "clip" is used for the strategy as a whole.

js-api-doc/value/color.d.ts
```diff
@@ -79,6 +79,25 @@ export type HueInterpolationMethod =
...
+ *   candidate recommendation]. This maps in the Oklch color space, using the
+ *   [deltaEOK] color difference formula and the [local-MINDE] improvement.
+ *
+ * * `clip`: Clamp each color channel that's outside the gamut to the minimum or
+ *   maximum value for that channel. This algorithm will produce poor visual
+ *   results, but it may be useful to match the behavior of other situations in
+ *   which a color can be clipped.
+ *
+ * [the original Color Level 4 candidate recommendation]: https://www.w3.org/TR/2024/CRD-css-color-4-20240213/#css-gamut-mapping
+ * [deltaEok]: https://www.w3.org/TR/2024/CRD-css-color-4-20240213/#color-difference-OK
```
Goodwine: Nit: `deltaEOK` above, but `deltaEok` here

spec/js-api/value/color.d.ts.md
```diff
@@ -224,6 +225,14 @@ export type HueInterpolationMethod =
   | 'shorter';
 ```
 
+### `GamutMapMethod`
+
+Methods by which colors in bounded spaces can be mapped to within their gamut.
+
+```ts
+export type GamutMapMethod = 'clip' | 'local-minde';
```
Goodwine, line 233: I like this!

accepted/color-4-new-spaces.md
Goodwine: I see, that makes sense

Goodwine (approved):

PR #3843, opened 2024-04-19
Pull childless `@layer` rules to the top of the output stylesheet

### LINKED ISSUE #3842: SCSS @use vs CSS @layer Problems.... plz fixed.. this issues..
While using CSS's @layer in scss, 
I found that it did not work in files that had to be compiled at the top level.

The contents of my compiled file are as follows.

//style.scss
@layer reset, components, utility;
@use 'components';
@use 'pages';

The problem in that code is
@use must be at the top of the code
This problem seems to have arisen because it is similar to the @layer concept in CSS.

Is there a way to fix this problem or can it be fixed in a future update?

---

I wrote this issue through a translator.
Some expressions may be strange.

nex3:
These rules are only allowed at the beginning of stylesheets. Allowing
them anywhere in Sass and moving them to the top of the output matches
the behavior of plain-CSS `@import` rules.

This is fast-track proposal. Although this does add an error when
multiple childless `@layer` rules are specified in the same
stylesheet, per sass/dart-sass#2225 that's not considered a breaking
change because it was already invalid CSS.

Closes #3842

### DIFF
--- a/spec/at-rules/extend.md
+++ b/spec/at-rules/extend.md
@@ -86,6 +86,11 @@ shorthands:
 * `extend(extendee, extensions)` for iteratively running `extendee =
   extend(extendee, extension)` for each `extension` in `extensions`.
 
+### Initial Rule
+
+An "initial rule" is a CSS at-rule that is either an `@import` or a `@layer`
+rule with no children.
+
 ## Semantics
 
 The `@extend` rule means that all elements matching the [extender](#extender)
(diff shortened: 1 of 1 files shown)


Yedfree (approved):

stof:
@nex3 what is the status of this fast-track proposal ?

nex3:
I realized there were some issues with the original semantics I had been considering, but ran out of time to pursue it further. It's on my list of things to circle back to.

PR #3844, opened 2024-04-20
[Ordered Comments] Flush to the spec

nex3:
See #2848

### DIFF
--- a/spec/at-rules/extend.md
+++ b/spec/at-rules/extend.md
@@ -203,7 +203,13 @@ that includes CSS for *all* modules transitively used or forwarded by
 
   * If `domestic` has already been traversed, do nothing.
 
-  * Otherwise, traverse every module in `domestic`'s dependencies.
+  * For each module `upstream` in `domestic`'s dependencies:
+
+    * For each unmarked comment in `domestic`'s CSS, if that comment originally
+      appeared before the `@use` or `@forward` rule that loaded `upstream`, add
+      a copy of that comment to `css` and then mark it.
+
+    * Traverse `upstream`.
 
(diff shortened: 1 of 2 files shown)


spec/at-rules/extend.md
```diff
@@ -212,15 +218,17 @@ that includes CSS for *all* modules transitively used or forwarded by
...
+    > subsequence is empty and `initial-imports` is inserted at the beginning of
+    > `css`.
 
   * For each top-level statement `statement` in `domestic`'s CSS tree after
     `initial-imports`:
 
-    * If `statement` is an `@import` rule, insert a copy of `statement` in `css`
-      after the last `@import` rule, or at the beginning of `css` if it doesn't
-      contain any `@import` rules.
+    * If `statement` is a marked comment, ignore it.
```
Goodwine, line 231: I don't see where `statement` gets marked, I think "them mark it" is missing in the bullet point below 🤔

Goodwine (changes requested):

spec/at-rules/extend.md
nex3: It's marked on line 210, above.

Goodwine: That bullet marks a "comment" rather than a `statement`

nex3: Right—this bullet checks if `statement` is a comment that is marked.

Goodwine: 🤦🏽‍♂️ for some reason I read this as "If `statement` is marked, ignore it" lol

Goodwine (approved):

PR #3847, opened 2024-04-22
Fix some bugs in the style rule and extend specs

nex3:
* Adds style rules to the stylesheet when they're encountered rather
  than after all their children have run, so that nested rules appear
  after their parents.

* Removes private placeholder selectors from style rules *after*
  they've been extended by local extensions.

* Explicitly handle omitting placeholder-only selectors from the CSS
  generated by Resolving Extensions.

### DIFF
--- a/spec/style-rules.md
+++ b/spec/style-rules.md
@@ -4,6 +4,7 @@
 
 * [Definitions](#definitions)
   * [Current Style Rule](#current-style-rule)
+  * [Current Keyframe Block](#current-keyframe-block)
 * [Semantics](#semantics)
 
 ## Definitions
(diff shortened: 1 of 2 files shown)


spec/style-rules.md
```diff
@@ -4,28 +4,50 @@
...
+  prefixes is "keyframes":
+
+  * Let `selector` be the result of parsing `selector-text` as a keyframe
+    selector.
+
+  * Append a keyframe block with selector `selector` to `parent`.
+
+  * Evaluate each child of `rule`.
+
+  * Cease evaluating `rule`.
```
Goodwine, line 47: I'm kinda confused by the last bullet. What does it mean by "cease evaluating" and why is it not implied that `rule` will stop being evaluated when all children are evaluated? 

Or does this line means that the next bullet point in line 49 should be ignored?

nex3: Yeah, it's intended to be the spec equivalent of `return;`. I'm not thrilled with the wording either... if you have a better suggestion I'm very interested.

Goodwine: Oh I see, I was confused partly because lines 75 and 77 look similar (two "otherwise" bullets chained together) but they didn't have the "cease evaluating" text. I see now that that's because in _that_ case the evaluation keeps going into line 81, and in _this_ case the evaluation stops completely.

I think this is fine, thanks for the clarification :)

Goodwine (approved):

Issue #3846, opened 2024-04-22
Make interleaved declaration semantics CSS-compatible


nex3:
- [x] Proposal (https://github.com/sass/sass/pull/3848)
- [x] Update full spec (https://github.com/sass/sass/pull/3885)
- **Deprecation**
  - [x] Dart Sass (https://github.com/sass/dart-sass/pull/2267)
  - [x] Tests (https://github.com/sass/sass-spec/pull/2002)
  - [x] Website (https://github.com/sass/sass-site/pull/1101)
- **Breaking change**
  - [x] Dart Sass (https://github.com/sass/dart-sass/pull/2629)
  - [x] Tests (https://github.com/sass/sass-spec/pull/2075)
  - [x] Website (https://github.com/sass/sass-site/pull/1452)

---

https://github.com/w3c/csswg-drafts/issues/8738 changes the behavior of declarations interleaved with nested rules away from behavior that matches Sass's historic behavior (hoisting all properties to the original nested rule), and instead treats those behaviors as logically appearing after the nested rule. For example, the source:

```scss
a {
  color: red;
  @media screen {color: blue}
  color: green;
}
```

currently compiles to:

```scss
a {
  color: red;
  color: green;
}

@media screen {
  a {
    color: blue;
  }
}
```

but the CSS semantics are now equivalent to:

```scss
a {
  color: red;
}

@media screen {
  a {
    color: blue;
  }
}

a {
  color: green;
}
```

...which is meaningfully different. Reportedly, Chrome will be implementing the new semantics soon.

Although our approach to full compatibility with plain CSS nesting semantics is generally fairly cautious due to Sass's strong existing semantic grounding (see https://github.com/sass/sass/issues/3030), in this case the set of styles that will need to change is fairly narrow and a workaround is easy (wrapping the properties in `& { ... }`), so it's worth changing this behavior eagerly.

nex3:
The proposal has landed. Since it does involve a deprecation, I'm going to give it a month for public comment.

PR #3848, opened 2024-04-22
[Interleaved Declarations] Add a proposal

nex3:
See #3846

### DIFF
--- a/proposal/interleaved-declarations.md
+++ b/proposal/interleaved-declarations.md
@@ -80,18 +80,47 @@ Change the definition of the [current style rule][old] to:
 > Differences are highlighted in bold.
 
 The *current style rule* is the CSS style rule that was created by the innermost
-[execution of a style rule], `@media` rule, `@supports` rule, unknown at-rule,
-**or [declaration]**.
+[execution of a style rule], `@media` rule, `@supports` rule, unknown at-rule.
+**This may be overridden by the [execution of a declaration].**
 
 [execution of a style rule]: ../spec/style-rules.md#semantics
-[declaration]: #semantics
+[execution of a declaration]: #semantics
+
+### Current Style Rule
+
+Change the definition of the [current style rule][old style] to:
+
+[old style]: ../spec/style-rules.md#current-style-rule
+
+> Differences are highlighted in bold.
+
+The *current style rule* is the CSS style rule that was created by the innermost
+[execution of a style rule], `@media` rule, `@supports` rule, unknown at-rule.
+**This may be overridden by the [execution of a declaration].**
+
+[execution of a style rule]: ../spec/style-rules.md#semantics
+[execution of a declaration]: #semantics
+
+### Current Keyframe Block
+
+Change the definition of the [current keyframe block][old keyframe] to:
+
+[old keyframe]: ../spec/style-rules.md#current-keyframe-block
+
+> Differences are highlighted in bold.
+
+The *current keyframe block* is the CSS keyframe block that was created by the
+innermost [execution of a style rule]. **This may be overridden by the
+[execution of a declaration].**
 
 ## Declarations
 
 ### Semantics
 
-Add the following to the semantics for executing a declaration `declaration`
+Add the following to the semantics for [executing a declaration] `declaration`
 before "Append `css` to `parent`":
+
+[executing a declaration]: ../spec/declarations.md#semantics
 
 * If `parent` isn't the last statement in its parent:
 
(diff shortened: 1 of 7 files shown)


proposal/interleaved-declarations.md
```diff
@@ -0,0 +1,125 @@
...
+the original style rule. Although this doesn't match the logical ordering of the
+source document, it *does* avoid duplicating the style rule (and any other
+nested rules) for each set of interleaved properties.
+
+Many years after this decision was made, the [CSS Nesting Module] was published
+by the W3C, adding native support for nesting rules within style rules using
+essentially the same syntax as Sass. Although initially its semantics for
+interleaved declarations were the same as Sass's, [a later update] (four days
+ago at time of writing) changed it to preserve the logical ordering of
+interleaved declarations.
```
Goodwine, line 33: Out of curiosity, given that a decision was made before, but it changed now... how "final" is this decision? As in, if Sass were to make changes in response to that CSS spec, what are the odds of it changing back?

proposal/interleaved-declarations.md
```diff
@@ -0,0 +1,125 @@
...
+**or [declaration]**.
+
+[execution of a style rule]: ../spec/style-rules.md#semantics
+[declaration]: #semantics
+
+## Declarations
+
+### Semantics
+
+Add the following to the semantics for executing a declaration `declaration`
```
Goodwine: This means that the changes below will be made on this file, right?
[spec/declarations.md](https://github.com/sass/sass/blob/e01cbeb8fd6b3b796ba223042c4f305392ac3856/spec/declarations.md)

If so, I think it would be helpful to have a link to that

proposal/interleaved-declarations.md
```diff
@@ -0,0 +1,125 @@
...
+before "Append `css` to `parent`":
+
+* If `parent` isn't the last statement in its parent:
+
+  * Let `copy` by a copy of `parent` without any children.
+
+  * Append `copy` to `parent`'s parent.
+
+  * Set `parent` to `copy`.
```
Goodwine: I'm slightly confused by this part. From the example above:
```scss
.a {
  color: red;
  @media screen {color: blue}
  color: green;
}
```

1. `copy` = `@media screen{}`
2. `copy` (`@media screen{}`) is appended to `parent`'s parent (`.a{color: red}`)
3. `parent` = `copy`

This would mean (IIUC) that it would produce the same(?) as the input:

```css
.a {
  color: red;
  @media screen {color: blue}
  color: green;
}
```

Or at the very least, with this wording I don't get that `@media` will wrap `.a`. 
I feel like it's missing "make an empty copy of all ancestors, and insert them inside `copy`, then make the innermost element inside `copy` the new `parent`" or something like that.

Or is that implied given how the other rules are executed?

proposal/interleaved-declarations.md
nex3: I suspect this is fairly final. One way to think about it it that it went from "the CSSWG hadn't talked about this in depth and the editor just defaulted to the Sass behavior because it's definitely feasible if not ideal" to "the CSSWG has talked about this and found a way around the difficulties and now it's resolved".

proposal/interleaved-declarations.md
nex3: Good call. When I was first writing this `declarations.md` didn't exist yet, but now it does!

proposal/interleaved-declarations.md
nex3: When you're executing `color: blue`, the structure of the generated CSS will be:

```css
a {color: red}
@media screen {a {}}
```

and `parent` will be `a {}`. This _is_ the last statement in its parent, so the new condition doesn't trigger.

Where it matters is when you're executing `color: green`. At this point, the CSS will be:

```css
a {color: red}
@media screen {a {color: blue}}
```

and `parent` is `a {color: red}`. This is _not_ the last statement in its parent, because the `@media` rule comes after it. So:

1. `copy` = `a {}`
2. `copy` (`a {}`) is appended to `parent`'s parent (the root stylesheet)
3. `parent` = `copy`

Then `color: green` is appended to `parent` and you get:

```css
a {color: red}
@media screen {a {color: blue}}
a {color: green}
```

This _does_ make me realize though that the current spec implies that the spec as written implies that the parent selector is duplicated anew for _each_ interleaved declaration, which isn't right.

Goodwine: gotcha, that makes sense

Goodwine (approved):

PR #3849, opened 2024-04-22
Flesh out some specs in preparation for modifying them

nex3:


spec/at-rules/unknown.md
```diff
@@ -14,11 +16,51 @@ structures it allows. It uses the following grammar:
...
+    * If `rule`'s name is "font-face", or if its [unprefixed] name is
+      "keyframes", append `css` to [the current module]'s CSS.
+
+    * Otherwise:
+
+      * Let `copy` by a copy of `parent` without any children.
+
+      * Append `copy` to `parent`'s parent.
+
+      * Append `css` to `copy`.
```
Goodwine: The way I read these lines would mean that:
```scss
.foo {
  a: b;
  @c {
    d: e;
  }
}
```
becomes
```scss
.foo {
  a: b;
}
.foo { // `copy` appended to `parent`
  @c { // `css` appended to `copy`
    d: e; // `rule`'s Statements appended to `css`
  }
}
```
but in practice I get:
```scss
.foo {
  a: b;
}
@c { // `css` appended to `parent`
  .foo { // `copy` appended to `css`
    d: e; // `rule`'s Statements appended to `copy`
  }
}
```

Goodwine: I think this needs to be something like...
```suggestion
      * Let `copy` by a copy of `parent` without any children.

      * Append `css` to `parent`'s parent.

      * Append `copy` to `css`.
      
      * Assign `copy` to `css`.
```

Goodwine (changes requested):

spec/at-rules/unknown.md
nex3: Ah, you're right, thanks!

Goodwine (approved):

PR #3858, opened 2024-05-02
[Color 4] Various fixes

nex3:
* Zero missing channels before conversion in `color.same()`

  Otherwise, you get weird behavior when conversion to the XYZ space preserves missing channels, which can cause colors to be considered `same()` even when they're visually distinct.

* Treat missing channels as distinct from 0 for `==`

  Now that `color.same()` explicitly treats missing channels as 0, it makes more sense for `==` to fill the stricter role and treat them as different. This is especially true because it already requires (non-legacy) colors to be in the same space, so missing channels are much more consistently meaningful than they would be in cross-space comparisons.

* Improve specifications for missing alpha values

### DIFF
--- a/accepted/color-4-new-spaces.md
+++ b/accepted/color-4-new-spaces.md
@@ -1557,15 +1557,15 @@ in [CSS Color Level 4][color-4].
 
 ### Scaling a Number
 
-This algorithm takes a number `number`, a value `factor`, a number `max`, and
-an optional number `min`. It's written "scale `<number>` by `<factor>` with a
-`max` of `<max>` and a `min` of `<min>`". It returns a number with a value
-between `min` (or 0) and `max` and the same units as `number`.
+This algorithm takes a number `number`, a value `factor`, a number `max`, and an
+optional number `min`. It's written "scale `<number>` by `<factor>` with a `max`
+of `<max>` and a `min` of `<min>`". It returns a number with the same units as
+`number`.
 
-> Note that this no longer assumes the original `number` is in a range of
-> 0 to `max`. We now allow scaling up negative numbers, and scaling down
-> numbers above the `max` value. The inverse operations return the `number`
-> unchanged, since that's the asymptotic scale behavior approaching boundaries.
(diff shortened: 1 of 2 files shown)


Goodwine (approved):

accepted/color-4-new-spaces.md
```diff
@@ -631,20 +631,18 @@ Legacy colors that have [missing] components are
...
+For determining *equality* between two colors `color1` and `color2`:
 
-* If both colors are [legacy colors](#legacy-color):
+* If both colors are [legacy colors](#legacy-color) in different spaces, return
+  `color.to-space(color1, rgb) == color.to-space(color2, rgb)`.
 
-  * Set each color to the result of [converting] the color into `rgb` space.
+* Otherwise, return whether the colors are in the same color space and their
+  channel and alpha values are fuzzy-equal, with [missing] channels treated as
+  0.
```
Goodwine: Did this mean to say that missing channels are treated distinct than zero instead?

Goodwine (approved):

Goodwine (approved):

mirisuzanne (approved):

accepted/color-4-new-spaces.md
```diff
@@ -2300,7 +2306,8 @@ ie-hex-str($color)
 
 * If `$color` is not a color, throw an error.
 
-* Let `rgb` be the result of [converting] and [gamut mapping] `$color` to `rgb`.
+* Let `rgb` be the result of [converting] and [gamut mapping] `$color` to `rgb`
```
Goodwine, line 2309: Nit: it feels a little odd that `rgb` here stands for a variable and for a channel

Goodwine (approved):

accepted/color-4-new-spaces.md
nex3: It's pretty common here to use the name of a color space to represent "the current color in that space". It is a bit odd but I think not so much that a reader can't get used to it, and it's substantially terser than using something like `rgbColor` everywhere.

PR #3749, opened 2023-11-17
Proposal: deprecate and remove `feature-exists`

stof:
https://github.com/sass/sass/issues/3702

PR #3860, opened 2024-05-06
Update the spec for the deprecation of feature-exists

stof:
See https://github.com/sass/sass/issues/3702


nex3:
Since this introduces a deprecation and will ultimately produce a breaking change, it's ineligible for a fast track proposal. It'll need a full proposal document, and we'll have to have at least some amount of a public response period before we start landing implementation PRs.

stof:
@nex3 this is not a fast-track proposal. The proposal has been submitted in https://github.com/sass/sass/pull/3749, which has been marked as accepted in April, but without updating the spec (this PR fixes that mistake done when marking the proposal as affected)

js-api-doc/deprecations.d.ts
```diff
@@ -113,6 +113,13 @@ export interface Deprecations {
    */
   'css-function-mixin': Deprecation<'css-function-mixin'>;
 
+  /**
+   * Deprecation for `meta.feature-exists()`.
+   *
+   * This deprecation became active in Dart Sass 1.78.0.
+   */
+  'feature-exists': Deprecation<'feature-exists'>;
```
chrss3697, line 128: Que es loque tengo q hacer

chrss3697 (approved):

nex3 (approved):

PR #3862, opened 2024-05-10
[Color 4] Be stricter about slash-separated strings

nex3:


accepted/color-4-new-spaces.md
```diff
@@ -1141,26 +1141,17 @@ The procedure is:
     * Let `split-last` be the result calling `string.split()` with the last
       element of `input` as the string to split, and `/` as the separator.
 
-    * If `split-last` has two items, and one or both items are an unquoted
-      string that's case-insensitively equal to 'none':
+    * If `split-last` doesn't have exactly two items, throw an error.
 
-      > Special handling for `none/none`, `none/<number>`, and `<number>/none`.
+    > This solves for a legacy handling of `/` in Sass that would produce an
```
Goodwine: Should this note be moved to in between lines 1139 and 1141? It seems like this entire section is meant for that rather than just the bullet points below 🤔

accepted/color-4-new-spaces.md
```diff
@@ -1141,26 +1141,17 @@ The procedure is:
     * Let `split-last` be the result calling `string.split()` with the last
       element of `input` as the string to split, and `/` as the separator.
 
-    * If `split-last` has two items, and one or both items are an unquoted
-      string that's case-insensitively equal to 'none':
+    * If `split-last` doesn't have exactly two items, throw an error.
```
Goodwine: Is this meant to deal with the following case?
```css
rgba(1 2 calc(var(--x)/var(--x))/var(--x))
```
IIUC by this point `components == [1, 2]` and `split-last == [calc(var(--x), var(--x)), var(--x)]` which would fail, however I think it should technically be permitted 🤔

nex3: Hmm, that's an interesting point. I think you're right that allowing that is more important than disallowing `rgb(1 2 3 a/b/c/)`.

accepted/color-4-new-spaces.md
```diff
@@ -1138,14 +1138,19 @@ The procedure is:
...
       element of `input` as the string to split, and `/` as the separator.
 
-    * If `split-last` doesn't have exactly two items, throw an error.
+    * If `split-last` doesn't have exactly two items, return an unquoted string
+      with the value of `input`.
 
-    > This solves for a legacy handling of `/` in Sass that would produce an
-    > unquoted string when the alpha value is a CSS function such as `var()`,
-    > when either value is `none`, or when using relative color syntax.
+      > This ensures that `rgb(1 2 3 calc(var(--a) / var(--b)) / var(--c))` is
```
Goodwine: I think I messed up my example and the `3` is extra

```suggestion
      > This ensures that `rgb(1 2 calc(var(--a) / var(--b)) / var(--c))` is
```

Goodwine (approved):

PR #3869, opened 2024-05-18
[Color 4] Do a full re-read and revision

nex3:


accepted/color-4-new-spaces.md
```diff
@@ -1055,28 +1059,27 @@ The individual conversion algorithms are:
...
+* If `color.is-in-gamut(color)`, return `origin`.
 
 * Let `mapped` be the result of running the method defined below whose name
-  matches `method`. If no such method matches, throw an error.
+  matches `method`.
 
-* Return the result of [converting] `mapped` into `origin-space`.
+* Return the result of [converting] `mapped` into `origin`'s space.
 
 #### `local-minde`
```
stof, line 1085: should the spec in `local-minde` specifically refer to the 13 February 2024 draft of CSS Color Module Level 4 ? What happens if the algorithm in the CSS spec is updated between that draft and the CR of Color Level 4 ?

accepted/color-4-new-spaces.md
```diff
@@ -1258,88 +1258,79 @@ a number relative to the range `[0,max]` without clamping.
...
 
 ### Validating a Color Channel
 
 [validating]: #validating-a-color-channel
 
-This process accepts a SassScript value `channel` to validate, a [known color
-space] `space` to validate against, and the `key` name of the channel. It
+This process accepts a SassScript value `value` to validate, a [known color
+space] `space` to validate against, and `channel` a channel in `space`. It
 throws an error if the channel is invalid for the color space, or returns a
```
stof, line 1272: this error is not described in the algorithm. As the order of errors is observable (by seeing which error you get when passing 2 invalid values), I think it should.

accepted/color-4-new-spaces.md
```diff
@@ -987,16 +984,18 @@ matched.
...
-> the process of 'carrying over' missing values on analogous channels, we have
+> the process of "carrying over" missing values on analogous channels, we have
 > to handle that here.
 
-* Let `origin-space` be `origin-color`'s color space.
+* If `target-space` is a SassScript value, set it to the result of [looking up a
+  known color space] named `target-space`.
 
-* If `origin-space == target-space` return `origin-color`.
+* Let `origin-color`'s space is already `target-space`, return `origin-color`.
```
jathak:
```suggestion
* If `origin-color`'s space is already `target-space`, return `origin-color`.
```

accepted/color-4-new-spaces.md
```diff
@@ -1233,14 +1233,14 @@ The procedure is:
     > Doing this late in the process allows us to throw any obvious syntax
     > errors, even for colors that can't be fully resolved during compilation.
 
-* If the length of `channels` is not equal to `expected`, throw an error.
+* If the length of `channels` is not equal 3, throw an error.
```
jathak:
```suggestion
* If the length of `channels` is not 3, throw an error.
```

accepted/color-4-new-spaces.md
```diff
@@ -1671,16 +1625,11 @@ is-in-gamut($color, $space: null)
...
-  `space-name`.
+* Let `color` be `$color` if `$space` is null, and the result of [converting]
+  `$color` to `$space` otherwise.
 
-* Let `color` be the result of calling `color.to-space($color, space)`.
-
-* For all bounded channels in `space`, if the associated channel value in
-  `$color` is fuzzy greater-than the bounded maximum, or fuzzy less-than the
+* For all bounded channels in `color`'s space, if the associated channel value
+  in `$color` fuzzy greater-than the bounded maximum, or fuzzy less-than the
```
jathak:
```suggestion
  in `$color` is fuzzy greater-than the bounded maximum or fuzzy less-than the
```

jathak (approved):

accepted/color-4-new-spaces.md
nex3: It throws an error if the number's value is `NaN`. There aren't any other circumstances where a number is fully invalid for a channel value. Although come to think of it, this should include infinite numbers as well.

nex3: Actually, rereading the Color 4 spec, it looks like it actually _preserves_ degenerate channel values and just lets them do whatever float stuff they're going to do, so we should do the same and avoid ever throwing an error here.

PR #3885, opened 2024-06-19
[Interleaved Declarations] Mark as accepted

nex3:
See #3846


jathak (approved):

nex3:
Major issue: CSS has removed `@nest` (https://github.com/w3c/csswg-drafts/issues/10234), so we'll need to pull that out of this proposal and implementation as well.

jathak (approved):

PR #3907, opened 2024-07-29
Backport deprecation API to legacy JS API

jathak:
This is a fast-track proposal to backport the deprecation options to work in the legacy JS API. Some bundlers don't support the modern API, so adding these options to the legacy API allows users of those bundlers to still control their deprecation warnings.

The behavior of each option in the legacy API matches its behavior in the modern API (the text here is directly copied from the modern API spec).

[skip dart-sass]
[skip sass-embedded]


nex3 (approved):

spec/js-api/legacy/options.d.ts.md
nex3, line 1: For the specs, I think it's fine to just link to the non-legacy API definitions rather than writing them out again.

nex3 (approved):

PR #3912, opened 2024-08-12
Merge main into feature.color-4

nex3:
[skip dart-sass]
[skip sass-embedded]


spec/at-rules/nest.md
```diff
@@ -0,0 +1,33 @@
+# `@nest` Rule
```
jathak, line 1: Wasn't this removed?

jathak (approved):

spec/at-rules/nest.md
nex3: Good catch. Looks like this is an issue in main, so I've put out https://github.com/sass/sass/pull/3913 to fix it.

PR #3969, opened 2024-10-24
Document existing indented syntax

jamesnw:
In preparation for proposing changes to the indented syntax, this documents existing behavior of the indented syntax, specifically around Statements.

In addition: 

- Explicit curly brace-wrapped Statements are replaced with a new `Block` production that supports both syntaxes.
- This also documents `InterpolatedDeclarationValue` production, which is somewhat orthogonal to the syntax, and those changes can be removed if it's easier.

### DIFF
--- a/spec/at-rules/mixin.md
+++ b/spec/at-rules/mixin.md
@@ -17,11 +17,10 @@
 ### Syntax
 
 <x><pre>
-**MixinRule** ::= '@mixin' [\<ident-token>] [ArgumentDeclaration]? [Block]
+**MixinRule** ::= '@mixin' [\<ident-token>] ArgumentDeclaration? [Block]
 </pre></x>
 
 [\<ident-token>]: https://drafts.csswg.org/css-syntax-3/#ident-token-diagram
-[ArgumentDeclaration]: ../syntax.md#argumentdeclaration
 [Block]: ../statement.md#block
 
 No whitespace is allowed between the `Identifier` and the `ArgumentDeclaration`
--- a/spec/at-rules/mixin.md
+++ b/spec/at-rules/mixin.md
@@ -17,10 +17,12 @@
 ### Syntax
 
 <x><pre>
-**MixinRule** ::= '@mixin' [\<ident-token>] ArgumentDeclaration? '{' Statements '}'
+**MixinRule** ::= '@mixin' [\<ident-token>] [ArgumentDeclaration]? [Block]
 </pre></x>
 
 [\<ident-token>]: https://drafts.csswg.org/css-syntax-3/#ident-token-diagram
+[ArgumentDeclaration]: ../syntax.md#argumentdeclaration
+[Block]: ../statement.md#block
 
 No whitespace is allowed between the `Identifier` and the `ArgumentDeclaration`
 in `MixinRule`.
@@ -75,7 +77,7 @@ To execute a `@mixin` rule `rule`:
 <x><pre>
 **IncludeRule**      ::= '@include' [NamespacedIdentifier] ArgumentInvocation?
 &#32;                    ContentBlock?
-**ContentBlock**     ::= UsingDeclaration? '{' Statements '}'
+**ContentBlock**     ::= UsingDeclaration? Block
 **UsingDeclaration** ::= 'using' ArgumentDeclaration
 </pre></x>
 
--- a/spec/at-rules/mixin.md
+++ b/spec/at-rules/mixin.md
@@ -76,7 +76,7 @@ To execute a `@mixin` rule `rule`:
 <x><pre>
 **IncludeRule**      ::= '@include' [NamespacedIdentifier] ArgumentInvocation?
 &#32;                    ContentBlock?
-**ContentBlock**     ::= UsingDeclaration? Block
+**ContentBlock**     ::= UsingDeclaration? [Block]
 **UsingDeclaration** ::= 'using' ArgumentDeclaration
 </pre></x>
 
--- a/spec/declarations.md
+++ b/spec/declarations.md
@@ -11,9 +11,11 @@
 
 <x><pre>
 **Declaration**         ::= StandardDeclaration | CustomDeclaration
-**StandardDeclaration** ::= [InterpolatedIdentifier]¹ ':' (Value | Value? Block )
+**StandardDeclaration** ::= [InterpolatedIdentifier]¹ ':' (Value | Value? [Block] )
 **CustomDeclaration**   ::= [InterpolatedIdentifier]² ':' InterpolatedDeclarationValue
 </pre></x>
+
+[Block]: statement.md#block
 
 1. This may not begin with "--".
 2. This *must* begin with "--".
--- a/spec/declarations.md
+++ b/spec/declarations.md
@@ -10,10 +10,9 @@
 ## Syntax
 
 <x><pre>
-**Declaration**              ::= StandardDeclaration | CustomDeclaration
-**StandardDeclaration**      ::= [InterpolatedIdentifier]¹ ':' (Value | Value? Block )
-**CustomDeclaration**        ::= [InterpolatedIdentifier]² ':' InterpolatedDeclarationValue
-InterpolatedDeclarationValue ::= (Interpolation | String)+
+**Declaration**         ::= StandardDeclaration | CustomDeclaration
+**StandardDeclaration** ::= [InterpolatedIdentifier]¹ ':' (Value | Value? Block )
+**CustomDeclaration**   ::= [InterpolatedIdentifier]² ':' InterpolatedDeclarationValue
 </pre></x>
 
 1. This may not begin with "--".
--- a/spec/declarations.md
+++ b/spec/declarations.md
@@ -10,9 +10,10 @@
 ## Syntax
 
 <x><pre>
-**Declaration**         ::= StandardDeclaration | CustomDeclaration
-**StandardDeclaration** ::= [InterpolatedIdentifier]¹ ':' (Value | Value? '{' Statements '}')
-**CustomDeclaration**   ::= [InterpolatedIdentifier]² ':' InterpolatedDeclarationValue
+**Declaration**              ::= StandardDeclaration | CustomDeclaration
+**StandardDeclaration**      ::= [InterpolatedIdentifier]¹ ':' (Value | Value? Block )
+**CustomDeclaration**        ::= [InterpolatedIdentifier]² ':' InterpolatedDeclarationValue
+InterpolatedDeclarationValue ::= (Interpolation | String)+
 </pre></x>
 
 1. This may not begin with "--".
--- a/spec/statement.md
+++ b/spec/statement.md
@@ -10,6 +10,10 @@
   * [IndentedStatements](#indentedstatements)
   * [Statements](#statements)
   * [Block](#block)
+  * [Comments](#comments)
+    * [LoudComment](#loudcomment)
+    * [SilentComment](#silentcomment)
+    * [WhitespaceComment](#whitespacecomment)
   * [Whitespace](#whitespace)
   * [Indentation](#indentation)
 
@@ -60,8 +64,10 @@ If a `WhitespaceComment` would be ambiguous with a `Statement` in the
 ### IndentedStatements
 
 <x><pre>
-**IndentedStatements**  ::= (Statement IndentedSame)* Statement
+**IndentedStatements**  ::= (Statement [IndentSame])* Statement
 </pre></x>
+
+[IndentSame]: #indentation
 
 ### Statements
 
@@ -78,12 +84,11 @@ If a `WhitespaceComment` would be ambiguous with a `Statement` in the
 
 <x><pre>
 **ScssBlock**      ::= '{' [ScssStatements] '}'
-**IndentedBlock**  ::= [IndentMore] [IndentedStatements] [IndentLess]
+**IndentedBlock**  ::= [IndentMore] [IndentedStatements]
 **Block**          ::= (ScssBlock | IndentedBlock)¹
 </pre></x>
 
 [IndentMore]: #indentation
-[IndentLess]: #indentation
 
 1: Only the production for the current syntax is valid.
 
@@ -92,8 +97,8 @@ If a `WhitespaceComment` would be ambiguous with a `Statement` in the
 #### LoudComment
 
 <x><pre>
-**ScssLoudComment**          ::= '/*' '!'? (.* | Interpolation)* '*/'
-**IndentedLoudComment**      ::= '/*' '!'? ([^IndentedLess] | Interpolation)*
+**ScssLoudComment**          ::= '/*' '!'? (.* | Interpolation)*'*/'
+**IndentedLoudComment**      ::= '/*' '!'? (.* | Interpolation)*
 **LoudComment**              ::= (ScssLoudComment | IndentedLoudComment)¹
 </pre></x>
 
@@ -104,8 +109,8 @@ If a `WhitespaceComment` would be ambiguous with a `Statement` in the
 #### SilentComment
 
 <x><pre>
-**ScssSilentComment**          ::= '//' [^LineBreak]*
-**IndentedSilentComment**      ::= '//' [^IndentedLess]*
+**ScssSilentComment**          ::= '//' .*
+**IndentedSilentComment**      ::= '//' .*
 **SilentComment**              ::= (ScssSilentComment | IndentedSilentComment)¹
 </pre></x>
 
@@ -114,9 +119,9 @@ If a `WhitespaceComment` would be ambiguous with a `Statement` in the
 #### WhitespaceComment
 
 <x><pre>
-**ScssWhitespaceComment**          ::= '//' .* '//' | '/*' .* '*/'
+**ScssWhitespaceComment**          ::= '//' .*'//' | '/*' .*'*/'
 **IndentedWhitespaceComment**      ::= '/*' .* '*/'
-**WhitespaceComment*¹              ::= ScssWhitespaceComment 
+**WhitespaceComment*¹              ::= ScssWhitespaceComment
 &#32;                                | IndentedWhitespaceComment
 </pre></x>
 
@@ -124,14 +129,11 @@ If a `WhitespaceComment` would be ambiguous with a `Statement` in the
 
 ### Whitespace
 
-> Whitespace separates productions inside a statement when disambiguation is
-> necessary. The Whitespace productions do not separate Statements.
-
 <x><pre>
 **LineBreak**               ::= CarriageReturn | LineFeed | FormFeed
 **ScssWhitespace**          ::= LineBreak | Space | Tab
 **IndentedWhitespace**      ::= Space | Tab
-**Whitespace**              ::= (ScssWhitespace | IndentedWhitespace)¹ 
+**Whitespace**              ::= (ScssWhitespace | IndentedWhitespace)¹
 &#32;                         | [WhitespaceComment]
 </pre></x>
 
@@ -142,10 +144,12 @@ If a `WhitespaceComment` would be ambiguous with a `Statement` in the
 ### Indentation
 
 <x><pre>
-**IndentCharacter**         ::= Space | Tab
-**IndentSame**              ::= [LineBreak] [IndentCharacter]{ Current }
-**IndentLess**              ::= [LineBreak] [IndentCharacter]{ 0, Current - 1 }
-**IndentMore**              ::= [LineBreak] [IndentCharacter]{ Current + 1, }
+**WhitespaceOnlyLine**          ::= IndentSame (Whitespace)*
+**IndentCharacter**             ::= Space | Tab
+**IndentSame**                  ::= [WhitespaceOnlyLine]* [LineBreak]
+&#32;                               [IndentCharacter]{ Current }
+**IndentMore**                  ::= [WhitespaceOnlyLine]* [LineBreak]
+&#32;                               [IndentCharacter]{ Current + 1, }
 </pre></x>
 
 [LineBreak]: #whitespace
@@ -156,8 +160,7 @@ The [IndentCharacter] must be the [document indentation character].
 [document indentation character]: #document-indentation-character
 
 `Current` is the [current indentation level] for a document. After consuming an
-`IndentSame`, `IndentLess`, or `IndentMore` production, the [current indentation
-level] is set to the number of [IndentCharacter]s found. At the end of a
-document, the [current indentation level] is set to 0.
+`IndentSame` or `IndentMore` production, the [current indentation level] is set
+to the number of [IndentCharacter]s found.
 
 [current indentation level]: #current-indentation-level
--- a/spec/statement.md
+++ b/spec/statement.md
@@ -55,7 +55,7 @@ character used in an [`IndentMore`] production in a document.
 **ScssStatements**      ::= (Statement ';'?¹)* Statement?
 </pre></x>
 
-1. This production is mandatory unless the previous `Statement` is a
+1: This production is mandatory unless the previous `Statement` is a
 `LoudComment`, `SilentComment`, or ends in a `Block`.
 
 If a `WhitespaceComment` would be ambiguous with a `Statement` in the
@@ -81,7 +81,7 @@ productions.
 [ScssStatements]: #scssstatements
 [IndentedStatements]: #indentedstatements
 
-1. Only the production for the current syntax is valid.
+1: Only the production for the current syntax is valid.
 
 ### Block
 
@@ -93,7 +93,7 @@ productions.
 
 [IndentMore]: #indentation
 
-1. Only the production for the current syntax is valid.
+1: Only the production for the current syntax is valid.
 
 ### Comments
 
@@ -101,10 +101,11 @@ productions.
 
 <x><pre>
 **ScssLoudComment**          ::= '/\*' (.\*¹ | Interpolation)\* '\*/'
-**IndentedLoudCommentNest**  ::= [IndentMore] ((.\*² | Interpolation)\*
-&#32;                            [IndentSame])\* (.\*² | Interpolation)\*
-**IndentedLoudComment**      ::= '/\*' (.\*² | Interpolation)\*
-&#32;                            IndentedLoudCommentNest?
+**InterpolatedCommentText**  ::= (.\*² | Interpolation)\*
+**IndentedLoudChildren**     ::= (InterpolatedCommentText [IndentSame])\*
+&#32;                            InterpolatedCommentText
+**IndentedLoudComment**      ::= '/\*' InterpolatedCommentText
+&#32;                            ([IndentMore] IndentedLoudChildren)?
 **LoudComment**              ::= (ScssLoudComment | IndentedLoudComment)³
 </pre></x>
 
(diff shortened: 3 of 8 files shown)


spec/declarations.md
```diff
@@ -10,9 +10,10 @@
 ## Syntax
 
 <x><pre>
-**Declaration**         ::= StandardDeclaration | CustomDeclaration
-**StandardDeclaration** ::= [InterpolatedIdentifier]¹ ':' (Value | Value? '{' Statements '}')
-**CustomDeclaration**   ::= [InterpolatedIdentifier]² ':' InterpolatedDeclarationValue
+**Declaration**              ::= StandardDeclaration | CustomDeclaration
+**StandardDeclaration**      ::= [InterpolatedIdentifier]¹ ':' (Value | Value? Block )
+**CustomDeclaration**        ::= [InterpolatedIdentifier]² ':' InterpolatedDeclarationValue
+InterpolatedDeclarationValue ::= (Interpolation | String)+
```
nex3: I don't think this definition is right—specifically, "String" doesn't really encapsulate the fact that _almost any_ token is allowed here. It's certainly not just things that parse as unquoted or quoted strings in Sass. If this definition will be relevant for the proposal, it's probably necessary to specifically enumerate the tokens it can contain.

Also, style nit: the left-hand side of syntax definitions should always be bold.

```suggestion
**InterpolatedDeclarationValue** ::= (Interpolation | String)+
```

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+  * [Whitespace](#whitespace)
+  * [Indentation](#indentation)
+
+## Definitions
+
+### Current indentation level
+
+The `current indentation level` is the count of the [`documentation indentation
+character`] between a new line and any non-whitespace characters of the last
+consumed line.
```
nex3: This should explicitly exclude lines that are only whitespace from consideration.

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+[`documentation indentation character`]: #document-indentation-character
+
+### Document indentation character
+
+The `document indentation character` is the character used for calculating the
+[`current indentation level`], and is either a space or tab character.
+
+[`current indentation level`]: #current-indentation-level
+
+It is defined as the first tab or space character used as an [IndentMore]
```
nex3:
```suggestion
It is defined as the first tab or space character used as an [`IndentMore`]
```

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+
+The `current indentation level` is the count of the [`documentation indentation
+character`] between a new line and any non-whitespace characters of the last
+consumed line.
+
+A document parsed in the indented syntax may not begin with whitespace, and the
+initial `current indentation level` for any document is `0`.
+
+Changes in the indentation level are used by the indented syntax to start and
+end blocks of statements.
```
nex3: This should be in a block quote to indicate that it's descriptive rather than normative.

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+
+[`documentation indentation character`]: #document-indentation-character
+
+### Document indentation character
+
+The `document indentation character` is the character used for calculating the
+[`current indentation level`], and is either a space or tab character.
```
nex3: The definition should come before the non-normative description that explains why the definition is useful, and the description should be in a block quote.

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+### ScssStatements
+
+<x><pre>
+**ScssStatements**      ::= (Statement ';'?¹)*Statement?
+</pre></x>
+
+1: This production is mandatory unless the previous `Statement` is a
+`LoudComment` or `SilentComment`, or after the final production in a
+`ScssStatements` production, which is either at the end of a `Block` or the end
+of the document.
```
nex3: You don't need to explicitly exempt the final statement—that's why the BNF ends in `Statement?` without a semicolon. You *do* need to explicitly exempt `';` after a `Statement` that ends in a `Block`, or else `a {b {} c: d}` would be invalid.

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+
+In the indented syntax, no character other than the `document indentation
+character` may be used for indentation.
+
+## Syntax
+
+### ScssStatements
+
+<x><pre>
+**ScssStatements**      ::= (Statement ';'?¹)*Statement?
```
nex3:
```suggestion
**ScssStatements**      ::= (Statement ';'?¹)* Statement?
```

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+### IndentedStatements
+
+<x><pre>
+**IndentedStatements**  ::= (Statement IndentedSame?¹)* Statement?
+</pre></x>
+
+1: This production is mandatory unless the previous `Statement` is a
+`LoudComment` or `SilentComment`, or after the final production in a
+`IndentedStatements` production, which is either at the end of a `Block` or the
+end of the document.
```
nex3: This is _not_ true for the indented syntax. If you could leave out `IndentedSame`, then `/**/ a: b` would parse as two statements in sequence rather than an error as it currently does. It's also not true that `Statement IndentedSame` is a valid `IndentedStatements`—that would consume the indentation without consuming a statement afterwards, leaving the parser in a confusing and likely broken state.

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+
+If a `WhitespaceComment` would be ambiguous with a `Statement` in the
+`IndentedStatements` rule, parse it preferentially as a `Statement`.
```
nex3: How would this be ambiguous in the indented syntax?

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Statements
+
+<x><pre>
+**Statements**          ::= [ScssStatements] | [IndentedStatements]¹
```
nex3: Where you had the footnote, it looks like it's just associated with `IndentedStatements` and not `ScssStatements`. Another alternative would be to add parentheses to make the grouping explicit.

```suggestion
**Statements**¹          ::= [ScssStatements] | [IndentedStatements]
```

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+**IndentedBlock**  ::= [IndentMore] [IndentedStatements] [IndentLess]
+**Block**          ::= ScssBlock | IndentedBlock¹
+</pre></x>
+
+[IndentMore]: #indentation
+[IndentLess]: #indentation
+
+1: Only the production for the current syntax is valid.
+
+### Whitespace
```
nex3: What about comments? What about a `Whitespace` production?

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+
+> Whitespace separates productions inside a statement when disambiguation is
+> necessary. The Whitespace productions do not separate Statements.
```
nex3: Several issues here:

* This seems like it's supposed to be normative, but it's in a block quote which indicates descriptive text.
* What does "inside a statement" mean? Block productions and thus the `Statements` production are also inside statements.
* When is disambiguation necessary? Does this mean that whitespace isn't allowed any time it wouldn't strictly disambiguate something?
* Statements certainly _are_ separated by whitespace in SCSS. `@foo; @bar` is a valid stylesheet with whitespace between the statements.

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+**IndentCharacter**         ::= Space | Tab
+**IndentSame**              ::= [LineBreak] [IndentCharacter]{ Current }
+**IndentLess**              ::= [LineBreak] [IndentCharacter]{ 0, Current - 1 }
+**IndentMore**              ::= [LineBreak] [IndentCharacter]{ Current + 1, }
+</pre></x>
+
+[LineBreak]: #whitespace
+[IndentCharacter]: #whitespace
+
+The [IndentCharacter] must be the [Document indentation character].
```
nex3:
```suggestion
The [IndentCharacter] must be the [document indentation character].
```

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+**ScssWhitespace**          ::= LineBreak | Space | Tab
+**IndentedWhitespace**      ::= Space | Tab
+</pre></x>
+
+### Indentation
+
+<x><pre>
+**IndentCharacter**         ::= Space | Tab
+**IndentSame**              ::= [LineBreak] [IndentCharacter]{ Current }
+**IndentLess**              ::= [LineBreak] [IndentCharacter]{ 0, Current - 1 }
```
nex3: If this consumes the indentation, then there won't be anything for the next statement to consume as `IndentSame`. The end of the document should also be considered an `IndentLess`.

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+**IndentMore**              ::= [LineBreak] [IndentCharacter]{ Current + 1, }
```
nex3:
```suggestion
**IndentMore**              ::= [LineBreak] [IndentCharacter]{ Current + 1 }
```

spec/statement.md
```diff
@@ -0,0 +1,131 @@
...
+[IndentCharacter]: #whitespace
+
+The [IndentCharacter] must be the [Document indentation character].
+
+[Document indentation character]: #document-indentation-character
+
+`Current` is the [current indentation level] for a document. After consuming an
+`IndentSame`, `IndentLess`, or `IndentMore` production, the [current indentation
+level] is set to the number of [IndentCharacter]s found. At the end of a
+document, the [current indentation level] is set to 0.
```
nex3: Why?

spec/statement.md
nex3: These should all allow arbitrarily many whitespace-only lines.

nex3 (changes requested):

spec/declarations.md
jamesnw: I think I will remove this to keep scope narrow (unless it does become relevant).

spec/statement.md
jamesnw: I updated this by adding a second sentence. Alternatively, I considered-

```md
The `current indentation level` is the count of the [`documentation indentation
character`] immediately preceding the start of the last consumed statement and 
after the closest linebreak.
```

I also excluded lines that do not start a statement. The only case I can think where this currently applies is with multiline comments, but it will be useful to have already defined.

spec/statement.md
jamesnw: Added in 70b12049d7bc462463d05af4dca8419ad98a4100

spec/statement.md
jamesnw: This was indeed meant to be a non-normative comment to remind myself to consider the difference between subtraction and unary negation, for example, and the difference between Whitespace and Indentation. It's not needed anymore.

spec/statement.md
jamesnw: It was an incorrect attempt to trigger an `IndentLess` production.

spec/statement.md
jamesnw: Definitely check my logic on this, but I think I can remove the `IndentLess` production. It was only used in the Comments and `IndentedBlock` productions, and I can remove it because it was essentially operating as a look ahead, and wasn't necessary. 

In `IndentedBlock`, the `IndentedStatements` production will consume until there is no longer an  `IndentSame`, so we don't need to explicitly end it with `IndentLess`. I'm less confident on the Indented comment productions ending with a wildcard.

spec/statement.md
jamesnw: I have the trailing comma to denote that there can be 1 or more `IndentCharacter`.

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+
+#### LoudComment
+
+<x><pre>
+**ScssLoudComment**          ::= '/*' '!'? (.* | Interpolation)*'*/'
+**IndentedLoudComment**      ::= '/*' '!'? (.* | Interpolation)*
+**LoudComment**              ::= (ScssLoudComment | IndentedLoudComment)¹
+</pre></x>
+
+> TODO: Wildcard and character set regex syntax conflict with markdown.
```
jamesnw: If the `.*` syntax is what's needed here, it is messing with markdown, and I wasn't sure if it should be correct BNF or correctly rendered markdown in that case.

I also considered using `'/*' '!'? ([^IndentationLess] | Interpolation)*` to help narrow when the comment production ends without consuming the indentation. The `[]` also conflicted with markdown syntax.

jamesnw (commented):
@nex3 This is ready for another round of review. Thanks!

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+    * [WhitespaceComment](#whitespacecomment)
+  * [Whitespace](#whitespace)
+  * [Indentation](#indentation)
+
+## Definitions
+
+### Current indentation level
+
+The `current indentation level` is the count of the [`documentation indentation
+character`] between a new line and any non-whitespace characters of the last
```
nex3: "the last consumed statement" is a good way to indicate this :+1:.

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+
+> Changes in the indentation level are used by the indented syntax to start and
+> end blocks of statements.
+
+[`documentation indentation character`]: #document-indentation-character
+
+### Document indentation character
+
+The `document indentation character` is defined as the first tab or space
+character used as an [`IndentMore`] production in a document.
```
nex3:
```suggestion
character used in an [`IndentMore`] production in a document.
```

since indentation is likely to use multiple spaces.

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+**ScssStatements**      ::= (Statement ';'?¹)* Statement?
+</pre></x>
+
+1: This production is mandatory unless the previous `Statement` is a
+`LoudComment`, `SilentComment`, or ends in a `Block`.
+
+If a `WhitespaceComment` would be ambiguous with a `Statement` in the
+`ScssStatements` rule, parse it preferentially as a `Statement`.
+
+### IndentedStatements
```
nex3, line 66: This should include a requirement that the statements not include newlines outside of `IndentSame` productions, or else it's halfway to the more flexible syntax already.

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+<x><pre>
+**IndentedStatements**  ::= (Statement [IndentSame])* Statement
+</pre></x>
+
+[IndentSame]: #indentation
+
+### Statements
+
+<x><pre>
+**Statements**          ::= ([ScssStatements] | [IndentedStatements])¹
```
nex3: Let's call this `Stylesheet` to emphasize that it's the root production of the document. It should also allow an optional U+FEFF at the beginning as a [byte-order mark](https://en.wikipedia.org/wiki/Byte_order_mark).

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+[IndentMore]: #indentation
+
+1: Only the production for the current syntax is valid.
+
+### Comments
+
+#### LoudComment
+
+<x><pre>
+**ScssLoudComment**          ::= '/*' '!'? (.* | Interpolation)*'*/'
```
nex3: You don't need explicit `'!'?` here, since it's covered by the general production. You _do_ need to make sure the `.*` isn't allowed to consume `#{` or `*/`, though. (Similarly below.)

Also, this is a style nit but it affects readability pretty strongly, make sure to have spaces between your productions:

```suggestion
**ScssLoudComment**          ::= '/*' '!'? (.* | Interpolation)* '*/'
```

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+**IndentedLoudComment**      ::= '/*' '!'? (.* | Interpolation)*
```
nex3: Similarly, this `.*` needs to indicate that it can't contain newlines. (Similarly below.)

Also: statement-level comments in the indented syntax allow nested content. (It's weird, I know.)

spec/statement.md
nex3: You can avoid markdown parsing by putting a backslash before the character. For example, `.\* .\*` renders as ".\* .\*" and `\[foo\]` renders as \[foo\].

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+**IndentedSilentComment**      ::= '//' .*
+**SilentComment**              ::= (ScssSilentComment | IndentedSilentComment)¹
+</pre></x>
+
+1: Only the production for the current syntax is valid.
+
+#### WhitespaceComment
+
+<x><pre>
+**ScssWhitespaceComment**          ::= '//' .*'//' | '/*' .*'*/'
```
nex3: Why the trailing `'//'`?

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+</pre></x>
+
+1: Only the production for the current syntax is valid.
+
+[WhitespaceComment]: #whitespacecomment
+
+### Indentation
+
+<x><pre>
+**WhitespaceOnlyLine**          ::= IndentSame (Whitespace)*
```
nex3: This shouldn't require `IndentSame`. It's valid to have empty zero-indent lines between indented lines.

It should also require a `LineBreak` at the end.

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+**IndentCharacter**             ::= Space | Tab
+**IndentSame**                  ::= WhitespaceOnlyLine* [LineBreak]
```
nex3: `WhitespaceOnlyLine*` should come after `LineBreak`; otherwise you're parsing trailing whitespace after the last statement as "a line".

spec/statement.md
```diff
@@ -0,0 +1,166 @@
...
+&#32;                               [IndentCharacter]{ Current }
+**IndentMore**                  ::= WhitespaceOnlyLine* [LineBreak]
+&#32;                               [IndentCharacter]{ Current + 1, }
```
nex3: It might be clearer to write this

```suggestion
&#32;                               [IndentCharacter]{ ≥ Current + 1 }
```

nex3 (changes requested):
This is getting there!

jamesnw:
@nex3 Updates made. Specifying the `IndentedLoudCommentNest` production is quite verbose- I'd welcome ideas on if there are simpler or clearer ways of doing that.

spec/at-rules/mixin.md
```diff
@@ -75,7 +76,7 @@ To execute a `@mixin` rule `rule`:
 <x><pre>
 **IncludeRule**      ::= '@include' [NamespacedIdentifier] ArgumentInvocation?
 &#32;                    ContentBlock?
-**ContentBlock**     ::= UsingDeclaration? '{' Statements '}'
+**ContentBlock**     ::= UsingDeclaration? Block
```
nex3:
```suggestion
**ContentBlock**     ::= UsingDeclaration? [Block]
```

spec/declarations.md
```diff
@@ -11,7 +11,7 @@
 
 <x><pre>
 **Declaration**         ::= StandardDeclaration | CustomDeclaration
-**StandardDeclaration** ::= [InterpolatedIdentifier]¹ ':' (Value | Value? '{' Statements '}')
+**StandardDeclaration** ::= [InterpolatedIdentifier]¹ ':' (Value | Value? Block )
```
nex3: Link this `Block`

spec/statement.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Syntax
+
+### ScssStatements
+
+<x><pre>
+**ScssStatements**      ::= (Statement ';'?¹)* Statement?
+</pre></x>
+
+1. This production is mandatory unless the previous `Statement` is a
```
nex3: Use `1:` rather than `1.` for annotation references. Otherwise Markdown will render them as an ordered list.

spec/statement.md
```diff
@@ -0,0 +1,176 @@
...
+### Comments
+
+#### LoudComment
+
+<x><pre>
+**ScssLoudComment**          ::= '/\*' (.\*¹ | Interpolation)\* '\*/'
+**IndentedLoudCommentNest**  ::= [IndentMore] ((.\*² | Interpolation)\*
+&#32;                            [IndentSame])\* (.\*² | Interpolation)\*
+**IndentedLoudComment**      ::= '/\*' (.\*² | Interpolation)\*
+&#32;                            IndentedLoudCommentNest?
```
nex3: I would write this as:

```suggestion
**InterpolatedCommentText**  ::= (.\*² | Interpolation)\*
**IndentedLoudChildren**     ::= (InterpolatedCommentText [IndentSame])\*
&#32;                            InterpolatedCommentText
**IndentedLoudComment**      ::= '/\*' InterpolatedCommentText
&#32;                            ([IndentMore] IndentedLoudChildren)?
```

This way you avoid repeating the interpolation construction over and over again, and `IndentedLoudChildren` production matches the structure of `IndentedStatements`.

spec/statement.md
```diff
@@ -0,0 +1,176 @@
...
+
+1. This may not contain `#{` or `*/`.
+2. This may not contain `#{` or newlines.
+3. Only the production for the current syntax is valid.
+
+#### SilentComment
+
+<x><pre>
+**ScssSilentComment**          ::= '//' .\*¹
+**IndentedSilentComment**      ::= '//' .\*²
```
nex3: This can also have text nested within it.

spec/statement.md
```diff
@@ -0,0 +1,176 @@
...
+
+1. This may not contain newlines.
+2. This may not contain newlines outside of [IndentSame] productions.
+3. Only the production for the current syntax is valid.
+
+#### WhitespaceComment
+
+<x><pre>
+**ScssWhitespaceComment**          ::= '//' .\*¹ | '/\*' .\*² '\*/'
+**IndentedWhitespaceComment**      ::= '/\*' .\*² '\*/'
```
nex3: The indented syntax also allows `//` as whitespace comments [at the end of a line](https://sass-lang.com/playground/#eJwzMHRISU0qTVdIy89X0NdXSM7PzU3NKwEAVhAHng==). Also, this should not allow newlines.

spec/statement.md
```diff
@@ -0,0 +1,176 @@
...
+
+<x><pre>
+**LineBreak**               ::= CarriageReturn | LineFeed | FormFeed
+**ScssWhitespace**          ::= LineBreak | Space | Tab
+**IndentedWhitespace**      ::= Space | Tab
+**Whitespace**              ::= (ScssWhitespace | IndentedWhitespace)¹
+&#32;                         | [WhitespaceComment]
+</pre></x>
+
+1. Only the production for the current syntax is valid.
```
nex3: Potentially a simpler way to define this would be:

```suggestion
**LineBreak**  ::= CarriageReturn | LineFeed | FormFeed
**Whitespace** ::= LineBreak¹ | Space | Tab | [WhitespaceComment]
</pre></x>

1: This is not allowed in the indented syntax.
```

nex3 (changes requested):

spec/statement.md
jamesnw: In sections with multiple annotation references, I kept the ordered list to avoid rendering it as a single paragraph. If there is a single reference, I changed it to `1:`.

spec/statement.md
jamesnw: Added in 22c1607cb6494b98dd6aeccf8da9b877234439f3

spec/statement.md
jamesnw: Added in 22c1607cb6494b98dd6aeccf8da9b877234439f3

spec/statement.md
nex3: Instead of this, please separate out the separate references into their own paragraphs, as [in the existing spec](https://github.com/sass/sass/blob/1c9ec00bbb061fa1be92c2e8c81144e6b8d8159b/spec/functions.md?plain=1#L66-L73)

spec/statement.md
```diff
@@ -0,0 +1,174 @@
...
+**ScssSilentComment**          ::= '//' .\*¹
+**CommentText**                ::= .\*²
+**IndentedSilentChildren**     ::= (CommentText [IndentSame])\* CommentText
+**IndentedSilentComment**      ::= '//' CommentText ([IndentMore]
+&#32;                              IndentedSilentChildren)?
+**SilentComment**              ::= (ScssSilentComment | IndentedSilentComment)³
+</pre></x>
+
+1. This may not contain newlines.
+2. This may not contain newlines outside of [IndentSame] productions.
```
nex3: How would `CommentText` contain an `IndentSame` production?

spec/statement.md
```diff
@@ -0,0 +1,174 @@
...
+3. Only the production for the current syntax is valid.
+
+#### SilentComment
+
+<x><pre>
+**ScssSilentComment**          ::= '//' .\*¹
+**CommentText**                ::= .\*²
+**IndentedSilentChildren**     ::= (CommentText [IndentSame])\* CommentText
+**IndentedSilentComment**      ::= '//' CommentText ([IndentMore]
+&#32;                              IndentedSilentChildren)?
```
nex3: Don't split parentheses across lines like this.

```suggestion
**IndentedSilentComment**      ::= '//' CommentText
&#32;                              ([IndentMore] IndentedSilentChildren)?
```

nex3 (changes requested):
Just a couple more changes and I think this is good to go!

spec/statement.md
jamesnw: Good point. I changed it to `[WhitespaceOnlyLine]` because [silent comments consume](https://sass-lang.com/playground/#eJwVizEOwjAQBPt7xZZQBdr8xlwWfJJzjuyLIL/HLqYYaebxXBYRIA2A1wqdFpnQuu/0gCaHuZZzI77Zgv1ISlQvF4o5++hvc+iXR/oh2yeXQbDB+lxra9S4i8g2Wq54/wEWeCSu) WhitespaceOnlyLines as well.

spec/statement.md
jamesnw: Updated in https://github.com/sass/sass/pull/3969/commits/63f009e0bd754eb37423a240e7b97dab6753d011

jamesnw: Updated in https://github.com/sass/sass/pull/3969/commits/63f009e0bd754eb37423a240e7b97dab6753d011

spec/statement.md
```diff
@@ -0,0 +1,181 @@
...
+</pre></x>
+
+1: This is not allowed in the indented syntax.
+
+[WhitespaceComment]: #whitespacecomment
+
+### Indentation
+
+<x><pre>
+**WhitespaceOnlyLine**          ::= IndentSame? [Whitespace]\* [LineBreak]
```
nex3: There's not really a reason to have `IndentSame?` here. The point is that empty lines can have any whitespace and have no bearing on the current indentation level.

spec/statement.md
nex3: The issue is that this doesn't contain *any* other productions. It's just matching pure characters. It should forbid newlines entirely.

nex3 (changes requested):

spec/statement.md
jamesnw: Ok, that makes sense. I forbade newlines in https://github.com/sass/sass/pull/3969/commits/1c19f21c9e0a9c5736dfc78d52839c10480af847.

nex3 (approved):

spec/statement.md
```diff
@@ -0,0 +1,177 @@
...
+    * [SilentComment](#silentcomment)
+    * [WhitespaceComment](#whitespacecomment)
+  * [Whitespace](#whitespace)
+  * [Indentation](#indentation)
+
+## Definitions
+
+### Current indentation level
+
+The *current indentation level* is the number of [documentation indentation
```
stof, line 24: shouldn't this be `document` instead of `documentation` ?

jamesnw: Opened #3992 to fix. Thanks!

PR #3973, opened 2024-11-01
Proposal: Improvements to indented syntax

jamesnw:
[Issue](https://github.com/sass/sass/issues/216)

This proposal improves the indented syntax format, allowing multiline expressions, semicolons, and curly braces.

It alters the existing syntax as documented in https://github.com/sass/sass/pull/3969.

[Rendered proposal](https://github.com/sass/sass/blob/281b6f234ba0e2e8feb58166a1ecbac3acc2a7a3/proposal/indented-syntax-improvements.md)

### DIFF
--- a/proposal/indented-syntax-improvements.md
+++ b/proposal/indented-syntax-improvements.md
@@ -208,7 +208,8 @@ In the indented syntax, this may not contain newlines.
 Replace footnote 1 with:
 
 1. In the indented syntax, `LineBreak` is not whitespace in the `IncludeAtRule`,
-   `SupportsAtRule`, [`MediaAtRule`], `KeyframesAtRule` or [`UnknownAtRule`].
+   `SupportsAtRule`, [`MediaAtRule`], `KeyframesAtRule`, [`UnknownAtRule`], or
+   in any position where the semantics of a `Statement` define a possible end.
 
 [`MediaAtRule`]: ../spec/at-rules/media.md
 [`UnknownAtRule`]: ../spec/at-rules/unknown.md
--- a/proposal/indented-syntax-improvements.md
+++ b/proposal/indented-syntax-improvements.md
@@ -42,8 +42,8 @@ line break is treated as continuing white space.
 In addition, this proposal adds semicolons to the indented syntax as explicit
 statement ends, and allows curly braces to wrap blocks.
 
-This proposal will make it possible to include line breaks by using
-an order of operations operator, essentially wrapping the expression in parentheses.
+This proposal will make it possible to include line breaks by wrapping the
+expression in parentheses.
 
 ```sass
 @font-face
@@ -56,7 +56,7 @@ an order of operations operator, essentially wrapping the expression in parenthe
   )
 ```
 
-Alternatively, authors can use Scss block syntax with braces and semicolons
+Alternatively, authors can use SCSS block syntax with braces and semicolons
 within the indented document.
 
 ```sass
@@ -78,7 +78,7 @@ a statement can not end, an error will be thrown.
 
 ### Places where a line break must create a statement break
 
-By design, line breaks are ignored as meaningless white space, except in
+By design, line breaks are ignored as meaningless white space except in
 contexts where the semantics define that a statement may end. When an author
 inserts a line break they intend to be meaningless in a position where a
 statement may end, the statement will end, and parsing will most likely fail on
@@ -92,8 +92,8 @@ to end.
 
 This may be surprising in more complex situations, for instance, with binary
 operators. `$foo: 3\n+ 4\n` ends the statement after `3`, but `$foo: 3 +\n4\n`
-ends the statement after `4`. Wrapping with the order of operations operator
-`()` allows authors more flexibility with `$foo: (3\n+ 4)`.
+ends the statement after `4`. Wrapping with parentheses allows authors more
+flexibility with `$foo: (3\n+ 4)`.
 
 This also applies to flow control at-rules. `@if $a \n and $b` would end the
 statement after `$a`, but `@if ($a \n and $b)` can be parsed.
@@ -107,7 +107,8 @@ A line break in a list that is not in a `BracketedListExpression` or enclosed in
 statement has ended. Alternates `$var: (1 2\n3)`, `$var: [1 2\n3]`, and `$var:
 (1, 2,\n 3)` can be parsed.
 
-Comma separated lists can not use a trailing comma to signify that a list will
+Existing syntax allows for trailing commas in Sass lists. However, comma
+separated lists can not use a trailing comma to signify that a list will
 continue after the line break, as this would break existing stylesheets with
 trailing commas.
 
@@ -118,11 +119,13 @@ wrapped in `#{}` so line breaks do not need to end statements.
 #### At-rules
 
 For any at-rule that is supported by native CSS, line breaks after the `@` and
-before a block or statement are not supported. This includes `@include`,
-(which overlaps with Sass), `@supports`, `@media`, `@keyframes` and any unknown
-at-rule.
+before a block or statement are not supported. An exception is that line breaks
+are supported inside parentheses. This includes
+`@import`, (which overlaps with Sass), `@supports`, `@media`, `@keyframes` and
+any unknown at-rule.
 
-These rules should be emitted as is, with no special handling from Sass.
+These rules should be emitted as is, with no special handling from Sass,
+including all parentheses.
 
 ### Design Decisions
 
@@ -130,41 +133,44 @@ While some CSS at-rules may have contexts where a line break would not be
 meaningful, custom handling of line breaks is outside of the scope of this
 proposal. For instance, `@media (hover: hover) and \n (color)` is not supported,
 even though line breaks do not end statements after boolean operators in general
-SassScript.
+SassScript. This avoids a requirement for Sass to add custom support for new CSS
+at-rules.
 
 We considered borrowing alternate syntax from other languages, such as a leading
 `>` or `|` from YAML or a trailing `|` from HAML. These introduce syntax that is
-novel to Sass, and we instead opted to borrow syntax from the Scss format.
+novel to Sass, and we instead opted to borrow syntax from the SCSS format. It
+also could introduce incompatibilities with future CSS features.
 
 These changes also allow authors using the indented syntax to add more explicit
 blocks with curly braces and line breaks with semicolons. While this introduces
 the changes for all authors, authors will still be able choose to limit the
 syntax with linters.
 
-In the indented syntax, authors can use a semicolon to explicitly end a
-statement. Subsequent lines in the same block still must have the same level of
-indentation. This means that the indented format won't support multiple
-statements on a single line, even if they are separated by a semicolon.
-
-```sass
-$font-stack: Helvetica, sans-serif;
-// or
-$primary-color: #333
-```
-
 In the indented syntax, authors can use curly braces to explicitly wrap a block
 of statements, but must use semicolons to separate statements inside that block.
-The braces essentially let authors opt in to Scss format for part of the
-document. We opted to require semicolons inside of curly braces to prevent
-issues with multiple nested formats.
+The braces essentially let authors opt in to SCSS format for part of the
+document. This means indentation rules do not apply and multiple statements can
+be on a single line, separated by semicolons. We opted to require semicolons
+inside of curly braces to prevent issues with multiple nested formats. 
 
 ```sass
 $font-stack: Helvetica, sans-serif
 $primary-color: #333
 body {
   font: 100% $font-stack;
-  color: $primary-color;
+  color: $primary-color; background: blue;
 }
+```
+
+In the indented syntax outside of curly braces, authors can use a semicolon to
+explicitly end a statement. Subsequent lines in the same block still must have
+the same level of indentation. This means that the indented format won't support
+multiple statements on a single line, even if they are separated by a semicolon.
+
+```sass
+$font-stack: Helvetica, sans-serif;
+// or
+$primary-color: #333
 ```
 
 When line breaks do not end statements, the line breaks are treated as white
@@ -195,7 +201,7 @@ productions.
 
 Replace footnote 1 with:
 
-1: In the Scss syntax, only the `ScssBlock` production is valid.
+1: In the SCSS syntax, only the `ScssBlock` production is valid.
 
 ### WhitespaceComment
 
@@ -207,9 +213,16 @@ In the indented syntax, this may not contain newlines.
 
 Replace footnote 1 with:
 
-1. In the indented syntax, `LineBreak` is not whitespace in the `IncludeAtRule`,
-   `SupportsAtRule`, [`MediaAtRule`], `KeyframesAtRule`, [`UnknownAtRule`], or
-   in any position where the semantics of a `Statement` define a possible end.
+1. In the indented syntax, [`LineBreak`] is not whitespace in the
+   `ImportAtRule`, `SupportsAtRule`, [`MediaAtRule`], `KeyframesAtRule`, or
+   [`UnknownAtRule`], except inside of parentheses or square brackets as defined
+   by CSS syntax. If a [`LineBreak`] in [`Whitespace`] would cause it to be
+   ambiguous with an [`IndentSame`] or [`IndentMore`] production, parse it
+   preferentially as [`IndentSame`] or [`IndentMore`].
 
 [`MediaAtRule`]: ../spec/at-rules/media.md
 [`UnknownAtRule`]: ../spec/at-rules/unknown.md
+[`LineBreak`]: ../spec/statement.md#whitespace
+[`Whitespace`]: ../spec/statement.md#whitespace
+[`IndentSame`]: ../spec/statement.md#indentation
+[`IndentMore`]: ../spec/statement.md#indentation
--- a/proposal/indented-syntax-improvements.md
+++ b/proposal/indented-syntax-improvements.md
@@ -9,8 +9,10 @@ expressions, semicolons, and curly braces.
 
 * [Background](#background)
 * [Summary](#summary)
+  * [Line breaks](#line-breaks)
+  * [Semicolons](#semicolons)
+  * [SCSS-in-Sass](#scss-in-sass)
   * [Ending statements](#ending-statements)
-  * [Places where a line break must create a statement break](#places-where-a-line-break-must-create-a-statement-break)
     * [After a SassScript value](#after-a-sassscript-value)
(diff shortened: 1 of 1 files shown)


jamesnw:
@nex3 Given the following code-

```sass
a
b
  c: d
```
Sass [currently](https://sass-lang.com/playground/#eJwzMEzkSuJSUEi2UkgBABAfApo=) outputs a warning about the `a` selector having no properties, and the css `b: {c: d}`.

If we allow newlines in selectors (without trailing commas), this will meaningfully change the output of the css to `a b {c: d}`. 

Should we continue to disallow newlines as whitespace in selectors?

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+
+In the indented syntax, this may not contain newlines.
+
+### Whitespace
+
+Replace footnote 1 with:
+
+1. In the indented syntax, `LineBreak` is not whitespace in the `IncludeAtRule`,
+   `SupportsAtRule`, [`MediaAtRule`], `KeyframesAtRule`, [`UnknownAtRule`], or
+   in any position where the semantics of a `Statement` define a possible end.
```
jamesnw: @nex3 I'd welcome your feedback on if there is a better way to say "where the semantics of a `Statement` define a possible end."

I would like to say "When a linebreak is ambiguous as Whitespace or a PossibleStatementEnd, parse it preferentially as a PossibleStatementEnd", but the PossibleStatementEnd concept doesn't (and shouldn't) exist.  

Would it make sense to add syntax for Expression, roughly something like-

```
**Expression**  ::= Whitespace?¹  SingleExpression (WhiteSpace² SingleExpression)*
```
Existing: 
1 & 2: This may not include newlines in the indented syntax

Proposal:

1: This may include newlines in the indented syntax
2: This may not include newlines in the indented syntax.

`SingleExpression` could still contain newlines, so I'm unsure if this helps specify it, or if it just pushes the question of where newlines cause statement breaks another layer deeper.

nex3:
> Given the following code-
> 
> ```sass
> a
> b
>   c: d
> ```
> 
> Sass [currently](https://sass-lang.com/playground/#eJwzMEzkSuJSUEi2UkgBABAfApo=) outputs a warning about the `a` selector having no properties, and the css `b: {c: d}`.
> 
> If we allow newlines in selectors (without trailing commas), this will meaningfully change the output of the css to `a b {c: d}`.
> 
> Should we continue to disallow newlines as whitespace in selectors?

I think we should continue to disallow *top-level* newlines except after a comma. This fits with the expectation that a newline is treated as the end of the current parsing context if that's valid, and only treated as whitespace if the parsing context contains some unterminated production.

proposal/indented-syntax-improvements.md
nex3: I'd lean on the same wording we're already using in [`ScssStatements`](java/com/google/assistant/aog/console/web/modules/languages/) and say something like "If a `LineBreak` in `Whitespace` would cause it to be ambiguous with an `IndentSame` or `IndentMore` production, parse it preferentially as `IndentSame` or `IndentMore`."

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+how authors can format their code. However, the parser can tell from context
+whether a statement can end at a given point. This document proposes that line
+breaks only end statements when a statement can end, and in any other case, a
+line break is treated as continuing white space.
+
+In addition, this proposal adds semicolons to the indented syntax as explicit
+statement ends, and allows curly braces to wrap blocks.
+
+This proposal will make it possible to include line breaks by using
+an order of operations operator, essentially wrapping the expression in parentheses.
```
nex3: Technically in Sass an "operator" only refers to infix or prefix operators like `+` or `not`. Parentheses are just called "parentheses".

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+  font-family: "Leatherman"
+  src: (
+    local("Leatherman"),
+    url("leatherman-COLRv1.otf") format("opentype") tech(color-COLRv1),
+    url("leatherman-outline.otf") format("opentype"),
+    url("leatherman-outline.woff") format("woff")
+  )
+```
+
+Alternatively, authors can use Scss block syntax with braces and semicolons
```
nex3:
```suggestion
Alternatively, authors can use SCSS block syntax with braces and semicolons
```

Also below. Unlike "Sass", "SCSS" *is* a proper acronym.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+### Ending statements
+
+Semicolons will always cause a statement break. If one occurs in a context where
+a statement can not end, an error will be thrown.
+
+### Places where a line break must create a statement break
+
+By design, line breaks are ignored as meaningless white space, except in
```
nex3:
```suggestion
By design, line breaks are ignored as meaningless white space except in
```

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+
+A line break in a list that is not in a `BracketedListExpression` or enclosed in
+`()` must cause a statement break.
+
+`$var: 1 2\n3` and `$var: 1, 2\n, 3` can not be parsed to determine when the
+statement has ended. Alternates `$var: (1 2\n3)`, `$var: [1 2\n3]`, and `$var:
+(1, 2,\n 3)` can be parsed.
+
+Comma separated lists can not use a trailing comma to signify that a list will
+continue after the line break, as this would break existing stylesheets with
```
nex3: I'd mention explicitly that trailing commas are allowed in Sass lists.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+Because arguments to functions and mixins are already wrapped in `()`, line
+breaks in arguments do not need to cause a statement break. Interpolations are
+wrapped in `#{}` so line breaks do not need to end statements.
+
+#### At-rules
+
+For any at-rule that is supported by native CSS, line breaks after the `@` and
+before a block or statement are not supported. This includes `@include`,
+(which overlaps with Sass), `@supports`, `@media`, `@keyframes` and any unknown
+at-rule.
```
nex3: An important exception here is that line breaks *are* allowed within parentheses (or square brackets but I don't think any at-rules use those currently). `@media screen and\n  (width >= 10px)` isn't allowed, but `@media screen and (\n  width >= 10px\n)` is.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+trailing commas.
+
+Because arguments to functions and mixins are already wrapped in `()`, line
+breaks in arguments do not need to cause a statement break. Interpolations are
+wrapped in `#{}` so line breaks do not need to end statements.
+
+#### At-rules
+
+For any at-rule that is supported by native CSS, line breaks after the `@` and
+before a block or statement are not supported. This includes `@include`,
```
nex3: Is `@include` here supposed to be `@import`? There is no plain CSS `@include` rule.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+
+These rules should be emitted as is, with no special handling from Sass.
+
+### Design Decisions
+
+While some CSS at-rules may have contexts where a line break would not be
+meaningful, custom handling of line breaks is outside of the scope of this
+proposal. For instance, `@media (hover: hover) and \n (color)` is not supported,
+even though line breaks do not end statements after boolean operators in general
+SassScript.
```
nex3: I think it's worth explaining that this is because we don't want Sass's behavior for at-rules to have to depend on us adding support for newer CSS syntax.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+
+We considered borrowing alternate syntax from other languages, such as a leading
+`>` or `|` from YAML or a trailing `|` from HAML. These introduce syntax that is
+novel to Sass, and we instead opted to borrow syntax from the Scss format.
```
nex3: Another downside of this syntax (and any new syntax) is that it could conflict with future CSS features.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+
+These changes also allow authors using the indented syntax to add more explicit
+blocks with curly braces and line breaks with semicolons. While this introduces
+the changes for all authors, authors will still be able choose to limit the
+syntax with linters.
+
+In the indented syntax, authors can use a semicolon to explicitly end a
+statement. Subsequent lines in the same block still must have the same level of
+indentation. This means that the indented format won't support multiple
+statements on a single line, even if they are separated by a semicolon.
```
nex3: except in the context of a curly brace block, right?

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,215 @@
...
+body {
+  font: 100% $font-stack;
+  color: $primary-color;
+}
+```
+
+When line breaks do not end statements, the line breaks are treated as white
+space, and do not have any requirements about the amount of indentation. The
+next statement's indentation is compared to the indentation at the start of the
+current statement to determine nesting or block closure.
```
nex3: These feel like they belong outside of the Design Decisions section, since they're more documentation of the format than specifically descriptions of why we chose one option over another.

nex3 (changes requested):

proposal/indented-syntax-improvements.md
jamesnw: Correct- thanks!

proposal/indented-syntax-improvements.md
jamesnw: Good point. I added a clarification that parentheses are emitted as part of the at-rule, as we don't want to track where parentheses match CSS syntax, and where they could be added solely for arbitrary newlines. For example we would emit both sets of parentheses in `@media (screen \n and) (width >= 10px)`, which wouldn't be valid CSS.

proposal/indented-syntax-improvements.md
jamesnw: Good point. I put this clarification with the section on curly braces, and put that before the section on semicolons.

jamesnw:
> > Given the following code-
> > ```sass
> > a
> > b
> >   c: d
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Sass [currently](https://sass-lang.com/playground/#eJwzMEzkSuJSUEi2UkgBABAfApo=) outputs a warning about the `a` selector having no properties, and the css `b: {c: d}`.
> > If we allow newlines in selectors (without trailing commas), this will meaningfully change the output of the css to `a b {c: d}`.
> > Should we continue to disallow newlines as whitespace in selectors?
> 
> I think we should continue to disallow _top-level_ newlines except after a comma. This fits with the expectation that a newline is treated as the end of the current parsing context if that's valid, and only treated as whitespace if the parsing context contains some unterminated production.

To clarify- are you specifying _top-level_ newlines here to distinguish from similar syntax newlines inside of curly braces? The same issue would occur in a nested selector, with `b` not having any properties.

```sass
a
  b
  c
    d: e
```

jamesnw (commented):
@nex3 I've incorporated the feedback, and substantially reworked the Summary and Design Decision sections, and this is ready for another round.

nex3:
> To clarify- are you specifying _top-level_ newlines here to distinguish from similar syntax newlines inside of curly braces?

I mean "top-level" in the expression-level sense: outside of any matching parentheses or square/curly braces. So `a\nb` is parsed as two style rules but `a:is(\n  b\n)` is parsed as one.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,244 @@
...
+
+The indented `.sass` syntax uses whitespace to end statements (linebreaks) and
+indicate nesting (indentation). While this syntax is preferred by some authors,
+it presents authoring challenges, specifically around long lists.
+
+## Summary
+
+> This section is non-normative.
+
+This proposal changes where line breaks cause statement breaks, allows the use
```
nex3: Instead of "changes where line breaks cause statement breaks", I'd say "adds support for multiline statements", which I think is the aspect of this that users will care the most about.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,244 @@
...
+of semicolons, and adds a method to opt-in to SCSS formatting to the indented
+syntax.
+
+### Line breaks
+
+In the Sass indented syntax, line breaks always end a statement, which limits
+how authors can format their code. However, the parser can tell from context
+whether a statement can end at a given point. This document proposes that line
+breaks only end statements when a statement can end, and in any other case, a
+line break is treated as continuing white space.
```
nex3: I'm not sure how clear "continuing white space" is here. Maybe "non-meaningful whitespace"?

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,244 @@
...
+statement.
+
+```sass
+@each $item in 
+  1 2 3
+    .item-#{
+      $item
+    }
+      content: $item *
+        10
```
nex3: I'd prefer to recommend a style where statement indentation is consistently 2 spaces and continued-expression indentation is 4 spaces to contrast.

```suggestion
@each $item in 
    1 2 3
  .item-#{
    $item
  }
    content: $item *
        10
```

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,244 @@
...
+    url("leatherman-outline.woff") format("woff")
+  )
+```
+
+### Semicolons
+
+In the indented syntax outside of curly braces, authors can use a semicolon to
+explicitly end a statement. Subsequent lines in the same block still must have
+the same level of indentation. This means that the indented format won't support
+multiple statements on a single line, even if they are separated by a semicolon.
```
nex3: I'd put "outside of curly braces" here instead of in the first sentence.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,244 @@
...
+operators. `$foo: 3\n+ 4\n` ends the statement after `3`, but `$foo: 3 +\n4\n`
+ends the statement after `4`. Wrapping with parentheses allows authors more
+flexibility with `$foo: (3\n+ 4)`.
+
+This also applies to flow control at-rules. `@if $a \n and $b` would end the
+statement after `$a`, but `@if ($a \n and $b)` can be parsed.
+
+#### After a non-enclosed list begins
+
+A line break in a list that is not in a `BracketedListExpression` or enclosed in
```
nex3: Since this is still the summary, we shouldn't assume that readers will have familiarity with spec production names. I'd just say "not enclosed in `[]` or `()`".

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,244 @@
...
+`()` must cause a statement break.
+
+`$var: 1 2\n3` and `$var: 1, 2\n, 3` can not be parsed to determine when the
+statement has ended. Alternates `$var: (1 2\n3)`, `$var: [1 2\n3]`, and `$var:
+(1, 2,\n 3)` can be parsed.
+
+Existing syntax allows for trailing commas in Sass lists. However, comma
```
nex3: "However" suggests that this goes against the fact that the existing syntax allows trailing commas, when in fact this is *because* the existing syntax allows trailing commas.

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,244 @@
...
+
+In the indented syntax, this may not contain newlines.
+
+### Whitespace
+
+Replace footnote 1 with:
+
+1. In the indented syntax, [`LineBreak`] is not whitespace in the
+   `ImportAtRule`, `SupportsAtRule`, [`MediaAtRule`], `KeyframesAtRule`, or
+   [`UnknownAtRule`], except inside of parentheses or square brackets as defined
```
nex3: You don't need to special-case `UnknownAtRule` here because statement breaks are allowed essentially anywhere outside of matched braces anyway, so the existing clause will already work. You don't need to special-case `KeyframesAtRule` because it's actually parsed as an `UnknownAtRule` and only recognized as `@keyframes` during evaluation.

nex3 (changes requested):

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,243 @@
...
+The indented `.sass` syntax uses whitespace to end statements (linebreaks) and
+indicate nesting (indentation). While this syntax is preferred by some authors,
+it presents authoring challenges, specifically around long lists.
+
+## Summary
+
+> This section is non-normative.
+
+This proposal adds support for multiline statements, allows the use of
+semicolons, and adds a method to opt-in to SCSS formatting to the indented
```
nex3: Nit:
```suggestion
semicolons, and adds a method to opt in to SCSS formatting to the indented
```

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,243 @@
...
+This also applies to flow control at-rules. `@if $a \n and $b` would end the
+statement after `$a`, but `@if ($a \n and $b)` can be parsed.
+
+#### After a non-enclosed list begins
+
+A line break in a list that is not enclosed in `[]` or `()` must cause a
+statement break.
+
+`$var: 1 2\n3` and `$var: 1, 2\n, 3` can not be parsed to determine when the
+statement has ended. Alternates `$var: (1 2\n3)`, `$var: [1 2\n3]`, and `$var:
```
nex3:
```suggestion
statement has ended. Alternatives `$var: (1 2\n3)`, `$var: [1 2\n3]`, and `$var:
```

proposal/indented-syntax-improvements.md
```diff
@@ -0,0 +1,243 @@
...
+Because arguments to functions and mixins are already wrapped in `()`, line
+breaks in arguments do not need to cause a statement break. Interpolations are
+wrapped in `#{}` so line breaks do not need to end statements.
+
+#### At-rules
+
+For any at-rule that is supported by native CSS, line breaks after the `@` and
+before a block or statement are not supported. An exception is that line breaks
+are supported inside parentheses. This includes
+`@import`, (which overlaps with Sass), `@supports`, `@media`, `@keyframes` and
```
nex3: Nit: reflow

nex3 (approved):
Just a few style changes and I think this is good to go!

jamesnw:
@nex3 I've made the updates. Thanks!

PR #3990, opened 2024-11-27
Add `calc-size()` as a calculation function

### LINKED ISSUE #3989: Add support for `calc-size()` as a calculation function
- [ ] Specification (https://github.com/sass/sass/pull/3990)
- [ ] Tests (https://github.com/sass/sass-spec/pull/2039)
- [ ] Dart Sass (https://github.com/sass/dart-sass/pull/2446)

---

1.81.0 compiled with dart2js 3.5.4

When doing:
```
height: calc-size(auto, size - 20px);
```
The output is:
```
height: calc-size(auto, size-20px);
```
Which breaks css

Also 
```
height: calc-size(auto, size + 20px);
```
The output is:
```
height: calc-size(auto, size20px);
```

When parsing calc-size() the second parameter should be formatted as is with symbols and spaces.
One workaround I have right now is to add calc().
So 
```
height: calc-size(auto, size - 20px);
```
Works if you do:
```
height: calc-size(auto, calc(size - 20px));
```
But this is not needed in the final output.

nex3:
Since this can't ever be evaluated at build time, there's no need to
add any additional semantics for it.

Closes #3989

### DIFF
--- a/spec/types/calculation.md
+++ b/spec/types/calculation.md
@@ -243,7 +243,7 @@ This algorithm takes a calculation `calc` and returns a number or a calculation.
 > to the input.
 
 * If `calc` was parsed from an expression within a `SupportsDeclaration`'s
-  `Expression`, but outside any interpolation, return a `calc` as-is.
+  `Expression`, but outside any interpolation, return `calc` as-is.
 
 * Let `arguments` be the result of [simplifying] each of `calc`'s arguments.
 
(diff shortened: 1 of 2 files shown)


spec/types/calculation.md
```diff
@@ -252,12 +252,22 @@ This algorithm takes a calculation `calc` and returns a number or a calculation.
...
+  `"atan`", `"sqrt"`, `"log"`, `"exp"`, `"sign"` and `arguments` has more than
+  one element, throw an error.
+
+* If `calc`'s name is `"mod"`, `"rem"`, `"atan2"`, `"pow"`, or `"calc-size"`:
+
+  * If `arguments` has fewer than two elements; and none of those are unquoted
+    strings, throw an error.
+
+    > It's valid to write `pow(var(--two-args))` or `pow(#{"2, 3"})`, but
+    > otherwise calculations' arguments must match the expected number.
```
pamelalozano16, line 265: Why is this only valid for calculations with two arguments? We could also have `round(var(--three-args))` or `round(#{"2, 3, 4"})` right?

pamelalozano16 (approved):

spec/types/calculation.md
nex3: `round()` and `clamp()` handle their fewer-than-three-arg cases below.

PR #4001, opened 2024-12-11
Allow trailing commas after rest parameters

### LINKED ISSUE #3994: Allow trailing commas after rest parameters
Currently, all parameter and argument lists allow trailing commas *except* after a rest parameter. So `foo($args...,)` is a valid way to call `foo()` but not to declare it. We should make this consistent by allowing the comma everywhere.

- [x] Specification (https://github.com/sass/sass/pull/4001)
- [x] Tests (https://github.com/sass/sass-spec/pull/2040)
- [x] Dart Sass (https://github.com/sass/dart-sass/pull/2464)

nex3:

Closes #3994

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -107,8 +107,8 @@ No whitespace is allowed between components of an `InterpolatedUnquotedUrlConten
 
 <x><pre>
 **ParameterList** ::= '(' Parameter (',' Parameter)\* ','? ')'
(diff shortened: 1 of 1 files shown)


jathak (approved):

nex3:
Worth noting: I realized while testing this that currently a trailing comma is also not allowed after a keyword rest argument, so this is effectively proposing a fix there as well.

PR #4002, opened 2024-12-12
[Indented Syntax Improvements] Draft 1.1 - Remove SCSS-in-Sass

### LINKED ISSUE #3991: Proposal feedback: Indented Syntax Improvements
Hello! I couldn't find an issue template or examples of other RFC feedback issues at first glance. Hope this works OK for you 😄 

Related to [this proposal](https://github.com/sass/sass/blob/main/proposal/indented-syntax-improvements.md) and [blog post](https://sass-lang.com/blog/).

SCSS in Sass will complicate things for language servers, and I wonder if the added complexity is worth it.

Today in the language server we can use the file extension or language ID of the document to determine which syntax is in use.
Were Sass to allow SCSS syntax inside of intented documents, language servers will have to adapt features to
figure out which regions of the document have which syntax.

For instance:

- [Diagnostics to highlight inconsistent indentation](https://github.com/wkillerud/some-sass/blob/588e19ee62477c2d02af59d8ec91c5aa4f67f1a4/packages/language-services/src/features/do-diagnostics.ts#L29) would need to know whether a line is part of an SCSS block to not report false positives.
- Code actions like [extract function](https://github.com/wkillerud/some-sass/blob/588e19ee62477c2d02af59d8ec91c5aa4f67f1a4/packages/language-services/src/features/code-actions.ts#L70) would need to know whether the target area is inside an SCSS block to have correct syntax.
- Code completions/IntelliSense would also need awareness of the current syntax to provide correct snippets in some cases (with/without semicolon, whether or not linebreaks are allowed).

Language servers can adapt of course, but I wonder if the added complexity is worth it to inline SCSS code instead of using a separate file? From the proposal this seems like an alternative to the suggested changes to line breaks, not strictly required.

jamesnw:
This removes the SCSS-in-Sass syntax, based on feedback in #3991 as well as comments in #216.

### DIFF
--- a/proposal/indented-syntax-improvements.md
+++ b/proposal/indented-syntax-improvements.md
@@ -1,6 +1,7 @@
 # Indented Syntax Improvements: Draft 1.1
 
 *([Issue](https://github.com/sass/sass/issues/216))*
+*([Changelog](./indented-syntax-improvements.changes.md))
 
 This proposal improves the indented syntax format, allowing multiline
 expressions and semicolons.
--- a/proposal/indented-syntax-improvements.md
+++ b/proposal/indented-syntax-improvements.md
@@ -1,7 +1,6 @@
 # Indented Syntax Improvements: Draft 1.1
 
-*([Issue](https://github.com/sass/sass/issues/216))*
-*([Changelog](./indented-syntax-improvements.changes.md))
+*([Issue](https://github.com/sass/sass/issues/216), [Changelog](./indented-syntax-improvements.changes.md))*
 
 This proposal improves the indented syntax format, allowing multiline
 expressions and semicolons.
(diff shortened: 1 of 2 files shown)


proposal/indented-syntax-improvements.md
```diff
@@ -1,25 +1,23 @@
-# Indented Syntax Improvements
+# Indented Syntax Improvements: Draft 1.1
 
 *([Issue](https://github.com/sass/sass/issues/216))*
```
nex3: Please link to the changelog here as well.

nex3 (approved):

proposal/indented-syntax-improvements.md
jamesnw: Added!

nex3 (approved):

Issue #4030, opened 2025-02-12
Support the <attr-type> production


Goodwine:
- [x] Proposal (https://github.com/sass/sass/pull/4033)

**Deprecation**

- [x] [Tests](https://github.com/sass/sass-spec/issues/2050) (https://github.com/sass/sass-spec/pull/2051)
- [x] [Implementation](https://github.com/sass/dart-sass/issues/2539) (https://github.com/sass/dart-sass/pull/2540)
- [x] [Documentation](https://github.com/sass/sass-site/issues/1333) (https://github.com/sass/sass-site/pull/1340)
- [x] Update canonical spec (https://github.com/sass/sass/pull/4045)

**Final Changes**

- [x] [Tests](https://github.com/sass/sass-spec/issues/2050) (https://github.com/sass/sass-spec/pull/2076)
- [x] [Implementation](https://github.com/sass/dart-sass/issues/2539) (https://github.com/sass/dart-sass/pull/2630)
- [x] [Documentation](https://github.com/sass/sass-site/issues/1333) (https://github.com/sass/sass-site/pull/1453)

---

CSS has an `attr()` function which can have a nested `type()` function call for an argument. The `type()` function accepts a "syntax argument" which is passed as an angle bracketed string, as in `<number>`.

https://developer.mozilla.org/en-US/docs/Web/CSS/attr

Example of valid CSS (I think):
```scss
.a {
  content: attr(id type(<custom-ident>))
}
```

This throws an error when passed to Sass as a `.scss` file ([playground]):

```
expected ")".
  ╷
2 │   content: attr(id type(<custom-ident>))
  │                         ^
  ╵
  - 2:25  root stylesheet
```

[playground]: https://sass-lang.com/playground/#eJwzNNRLVKjmUlBIzs8rSc0rsVJILCkp0shMUSipLEjVsEkuLS7Jz9XNTAHK2WlqctUCAIPXEBU=

---

### Workaround

One workaround that can be used is quoting the "syntax" value, and using interpolation or `string.unquote()` to unquote it. For example:

```scss
.a {
  content: attr(id type(#{'<custom-ident>'}))
}
```

nex3:
Although MDN reports this as unsupported on all browsers, [this codepen](https://codepen.io/nex3/pen/dPyPXjZ) demonstrates that it does seem to work on Chrome at least as of v133, so we should support it (probably by declaring `type()` a special function).

nex3:
[...]

PR #4031, opened 2025-02-13
[Attr Types] Add proposal

nex3:
See https://github.com/sass/sass/issues/4030

### DIFF
--- a/accepted/attr-type.md
+++ b/accepted/attr-type.md
@@ -1,0 +1,100 @@
+# Attr Type: Draft 1
+
+*([Issue](https://github.com/sass/sass/issues/4030))*
+
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+* [Syntax](#syntax)
+  * [`SpecialFunctionExpression`](#specialfunctionexpression)
+  * [`SingleExpression`](#singleexpression)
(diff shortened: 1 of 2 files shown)


accepted/attr-type.md
```diff
@@ -0,0 +1,100 @@
...
+
+* [The `<attr-unit>` production], which is mostly just identifiers but also
+  includes `%` as a value in its own right. This character is already used in
+  SassScript as the modulo operator.
+
+[The `<syntax>` production]: https://drafts.csswg.org/css-values-5/#typedef-syntax
+[The `<attr-unit>` production]: https://drafts.csswg.org/css-values-5/#typedef-attr-unit
+
+This is further complicated by the fact that the `attr()` function in which the
+`<attr-type>` production appears has been supported in browsers for a number of
```
Goodwine: I find reading "appears has been supported" a little bit odd because it kinda reads very similar to "appears to have been supported", but IIUC the "appears" verb is for the "production" subject 🤔 

Can you reword this slightly to avoid that possible confusion?

Goodwine (approved):

accepted/attr-type.md
nex3: Added parentheses to clarify

PR #4040, opened 2025-03-01
Mark bogus combinators and moz-document as obsolete

nex3:
See #3340 

[skip dart-sass]
[skip sass-embedded]


jathak (approved):

jathak (changes requested):
This needs `npm run sync-deprecations` to ensure the JS API is updated.

Not sure why the CI isn't failing, since it should be comparing the checksums

jathak (approved):

spec/deprecations.yaml
```diff
@@ -64,6 +64,7 @@ bogus-combinators:
   dart-sass:
     status: active
```
jathak: This needs to be changed to `status: obsolete`

PR #4045, opened 2025-03-08
[Attr Type] Flush to spec

nex3:
See #4030


Goodwine (approved):
lgtm sans broken links

Issue #4034, opened 2025-02-17
Configuration for private variables of modules


stof:
- [x] Proposal (https://github.com/sass/sass/pull/4049)

### Phase 1

- [x] Flush to spec (https://github.com/sass/sass/pull/4125)
- [x] [Implementation](https://github.com/sass/dart-sass/issues/2632) (https://github.com/sass/dart-sass/pull/2631)
- [x] Tests (https://github.com/sass/sass-spec/pull/2077)
- [x] Documentation (https://github.com/sass/sass-site/pull/1454)

### Phase 2

- [ ] Deprecation metadata
- [ ] [Implementation](https://github.com/sass/dart-sass/issues/2632)
- [ ] Tests
- [ ] Documentation

---

While reviewing a PR from a colleague, I discovered that configuring a private variable of a module works, which quite surprised me (I first wanted to post review comments saying this was broken, but our CI proved me wrong as it managed to build the CSS).

```scss
// _a.scss
$_separator-size: 8px !default;

a {
  gap: $_separator-size;
}
```

```scss
// input.scss
@use "a" with { $_separator-size: 16px; }
```

Looking at sass-spec, this is not covered by any tests, so I'm not sure it is intended to allow private members to be configurable. If it is intended, I would suggest adding tests in sass-spec (I could contribute them btw). Otherwise, this might deserve a deprecation.

stof:
Btw, regarding https://github.com/sass/sass/issues/3171, this feature might even be useful as it means that those private members can be configured but cannot be written later.

nex3:
I think this should be an error, since it violates the principle that within a module `$-foo` is a strictly local identifier. We should deprecate both passing it and declaring `!default` on a private variable name.

PR #4047, opened 2025-03-10
Deprecate configuring private variables when loading Sass modules

Goodwine:
- Deprecate `!default` in private sass variables
- Deprecate configuration map keys for private variables (e.g. `$-foo`)

https://github.com/sass/sass/issues/4034

### DIFF
--- a/spec/deprecations.yaml
+++ b/spec/deprecations.yaml
@@ -147,3 +147,9 @@ global-builtin:
   dart-sass:
     status: active
     deprecated: 1.80.0
+
+private-identifier-module-config:
+  description: Private identifiers as configuration keys in `@use`, `@forward`, or `load-css()` are deprecated.
+  dart-sass:
+    status: active
+    deprecated: 1.85.2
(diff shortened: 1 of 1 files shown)


nex3:
This needs a proper [language proposal](https://github.com/sass/sass/blob/main/CONTRIBUTING.md#process) following the full process.

Goodwine:
oh ok, I thought the language spec already disallowed it and it was a matter of implementing.

I will send a proposal then, FWIW I learned through the implementation that failing on `with($-a: b)` isn't necessary if we disallowed `!default` on `$-a` in the first place.

stof:
The language proposal was merged at https://github.com/sass/sass/pull/4049

stof:
> I will send a proposal then, FWIW I learned through the implementation that failing on `with($-a: b)` isn't necessary if we disallowed `!default` on `$-a` in the first place.

to me, it is still good to report it, because it reports deprecation in different places (and so if you use ``quietDeps`, one of them might be reported while the other is silenced)

Goodwine:
per the updated proposal, `with($-foo:bar)` will be an issue, but declaring `$-foo: bar !default` won't be. However, the merged proposal is still not marked as "accepted".

Once the proposal is marked as accepted I'll revisit this PR to implement that.

Issue #4048, opened 2025-03-10
Improve checks for invalid CSS function names


nex3:
Currently, it's possible for users to define functions whose names differ from reserved names only by case, such as `EXPRESSION()` or `URL()`. These functions can't be invoked normally, because the logic for function _calls_ is case-insensitive, but unlike their lowercase equivalents they're not forbidden at definition-time. We should make the invalid function name check case-insensitive to deprecate and eventually forbid these names.

**Edit:** I've updated the title to broaden the scope here a bit. There are a number of improvements to be made to this check, mostly in the form of _loosening_ it for cases that aren't (or are no longer) of real concern.

---

- [x] Proposal (https://github.com/sass/sass/pull/4053)

**Deprecation**

- [x] [Tests](https://github.com/sass/sass-spec/issues/2052) (https://github.com/sass/sass-spec/pull/2109)
- [x] [Implementation](https://github.com/sass/dart-sass/issues/2543) (https://github.com/sass/dart-sass/pull/2731)
- [x] [Documentation](https://github.com/sass/sass-site/issues/1339) (https://github.com/sass/sass-site/pull/1527)
- [x] Update canonical spec (https://github.com/sass/sass/pull/4186)

**Final Changes**

- [ ] [Tests](https://github.com/sass/sass-spec/issues/2052)
- [ ] [Implementation](https://github.com/sass/dart-sass/issues/2543)
- [ ] [Documentation](https://github.com/sass/sass-site/issues/1339)

ghost:
Hi, I'm interested in contributing. Since this issue is pending a deprecation cycle, I was wondering if there's a related task or a smaller issue you'd recommend for a first time contributor. I'd love to get involved. Thanks

nex3:
Hi @ugwudev, and thanks for reaching out! Implementing a deprecation isn't too difficult, but there are a few steps:

1. In this repo, add the new deprecation to [`spec/deprecations.yaml`](https://github.com/sass/sass/blob/main/spec/deprecations.yaml). This is the shared source of truth for which deprecations exist, their names, and their documentation.

2. In the https://github.com/sass/dart-sass repo, symlink this repo into `build/sass`, then run `dart run grinder deprecations`. This will add a deprecation enum entry for the new deprecation you just added. Then find the place where the deprecation should be emitted and emit it.

[...]

PR #4053, opened 2025-03-17
[Function Name] Add proposal

nex3:
See https://github.com/sass/sass/issues/4048

### DIFF
--- a/proposal/function-name.md
+++ b/proposal/function-name.md
@@ -1,0 +1,101 @@
+# Function Name: Draft 1.0
+
+*([Issue](https://github.com/sass/sass/issues/4048))*
+
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+* [Semantics](#semantics)
+  * [`@function`](#function)
+* [Deprecation Process](#deprecation-process)
+  * [Phase 1](#phase-1)
+  * [Phase 2](#phase-2)
+
+## Background
+
+> This section is non-normative.
+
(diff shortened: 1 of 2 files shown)


proposal/function-name.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Deprecation Process
+
+The deprecation process will be divided into two phases:
+
+### Phase 1
+
+> This phase adds no breaking changes. Its purpose is to notify users of the
+> upcoming changes to behavior and give them a chance to move towards passing
+> future-proof units.
```
jathak:
```suggestion
> future-proof names.
```

jathak (approved):

nex3:
The CI seems to be failing to verify w3c pages that are up for me. I'm going to assume it's a transitory error and try again another day, but if it keeps happening we may need to exclude them from CI entirely.

proposal/function-name.md
```diff
@@ -0,0 +1,101 @@
...
+## Semantics
+
+### `@function`
+
+Remove the second bullet point and replace the fourth bullet point of [the
+semantics for `@function`] with:
+
+[the semantics for `@function`]: ../spec/at-rules/function.md#semantics
+
+* If `name` is `and`, `or`, or `not`, throw an error.
```
stof, line 73: should this be case insensitive as well ?

nex3: No, Sass identifiers are always case-sensitive.

PR #4056, opened 2025-03-26
Update proposal: Disallow private identifiers in a `with` clause

Goodwine:


proposal/default-on-private-variables.md
```diff
@@ -64,4 +65,5 @@ emit a deprecation warning named `private-variable-default`.
 
 Phase 2 implements the full changes described above. Per the Dart Sass
 compatibility policy, it won't be released until at least three months after the
-first release with the deprecation warning.
+first release with the deprecation warning. However, the implementation of Phase
+2 is contingent upon the complete removal of `@import`.
```
nex3: Probably worth explaining why here.

nex3 (changes requested):

Goodwine:
We discussed it offline. We want to still allow `!default` on private variables because it allows setting default values. It's kind of like sugar syntax for `@if not $-foo { $-foo = mydefaultvalue }`

So, we'll instead disallow the use of private identifiers in a with clause.

proposal/default-on-private-variables.md
```diff
@@ -15,36 +18,67 @@
...
 
-  > Sass treats `_` and `-` as equivalent in variable names.
+  * If the `argument`'s identifier name begins with `-`, throw an error.
+
+    > Sass treats `_` and `-` as equivalent in identifiers.
+
+  * Otherwise, let `value` be the result of evaluating `argument`'s expression.
+
+  * Add a variable to `rule-config` with the same name as `argument`'s
+    identifier and with `value` as its value.
```
nex3: You don't necessarily need to repeat this here; you could just say "Add the following line before evaluating a `KeywordArgument`'s expression:" and then add the error-throwing clause.

proposal/default-on-private-variables.md
```diff
@@ -56,9 +90,9 @@ The deprecation process will be divided into two phases:
 > upcoming changes to behavior and give them a chance to move towards passing
 > future-proof units.
 
-Phase 1 does not change the semantics of variable declarations. However, if a
-private variable (one whose name starts with `_` or `-`) has a `!default` flag,
-emit a deprecation warning named `private-variable-default`.
+Phase 1 does not change the semantics of `with` clauses. However, we will emit a
+deprecation warning named `with-clause-private-variable` when a `with` clause
```
nex3: I suggest making this deprecation name terser, like `with-private`.

proposal/default-on-private-variables.md
```diff
@@ -56,9 +74,9 @@ The deprecation process will be divided into two phases:
 > upcoming changes to behavior and give them a chance to move towards passing
 > future-proof units.
 
-Phase 1 does not change the semantics of variable declarations. However, if a
-private variable (one whose name starts with `_` or `-`) has a `!default` flag,
-emit a deprecation warning named `private-variable-default`.
+Phase 1 does not change the semantics of `with` clauses. However, we will emit a
+deprecation warning named `with-private` when a `with` clause privacy of Sass
+modules.
```
nex3: "when a `with` clause privacy of Sass modules" is missing a verb.

Goodwine: oh no, it doesn't make any sense lol

nex3 (approved):

Issue #3886, opened 2024-06-19
Compatibility with the proposed CSS if function


stof:
* Phase 1
  - [x] Specification (https://github.com/sass/sass/pull/4155)
  - [x] Tests (https://github.com/sass/sass-spec/pull/2092)
  - [x] Dart Sass (https://github.com/sass/dart-sass/pull/2693)
  - [x] Website (https://github.com/sass/sass-site/pull/1504)
  - [x] [Migrator](https://github.com/sass/migrator/issues/283) (https://github.com/sass/migrator/pull/286)
* Phase 2
  - [ ] Specification
  - [ ] Tests
  - [ ] Dart Sass
  - [ ] Website

---

The CSSWG has resolved working on inline conditionals for CSS. For now, the proposed syntax uses a `if()` function.
See https://lea.verou.me/blog/2024/css-conditionals/ and https://github.com/w3c/csswg-drafts/issues/10064

Even though the syntax is not yet final (the proposal has just started), the current syntax will require adjustments in Sass:
- the special [`if`](https://sass-lang.com/documentation/modules/#if) Sass function (which has a special parsing and evaluation) will need to be distinguished from the CSS conditional
- the current proposal uses `if(condition ? then : else)` as syntax, which fails to parse in Sass (I'm testing that using `css-if` as the function name to avoid issues with the first point):
    - the ternary syntax fails to parse: https://sass-lang.com/playground/#eJwzNExUqOZSUEiyUkguLtbNTNMoSyzS0NVNzs9LySzJzM/TVLBXSMopTVWwUihKTdG05qoFAIT+D1Q=
    - the style condition fails to parse inside the condition of the function: https://sass-lang.com/playground/#eJwzNExUqOZSUEiyUkguLtbNTNMoLqnMSdXQ1S1LLLJSSMvP11SwV0jKKU1VsFIoSk3RtOaqBQCRIQ9c (note that this syntax works when storing the condition in a custom property thanks to the fact that custom properties accept any content in Sass, requiring to use interpolation to use SassScript expressions: https://sass-lang.com/playground/#eJwzNExUqOZSUNDVTc7PS8ksyczPs1IoLqnMSS3W0NUtSyyyUkjLz9fkqgUAA7AM4Q==)

The work on this proposal should be tracked to figure out how the compatibility layer should look like.

stof:
the spec has been updated and is available at https://drafts.csswg.org/css-values-5/#if-notation

I updated examples:

[...]

PR #4126, opened 2025-08-29
[Plain CSS If] Add a proposal

nex3:
See #3886

### DIFF
--- a/proposal/plain-css-if.md
+++ b/proposal/plain-css-if.md
@@ -49,7 +49,7 @@ functions remained in the global scope. This was for a few reasons:
 
 2. `if()` is part of the core language infrastructure (like a ternary expression
    in other languages) and so requiring a module to be loaded to use it would
-   was more onerous than we wanted to impose on users.
+   have been more onerous than we wanted to impose on users.
 
 3. It didn't fit cleanly into any of the core modules. `sass:meta` was the
    closest as sort of a grab-bag of functions that did things with odd
--- a/proposal/plain-css-if.md
+++ b/proposal/plain-css-if.md
@@ -1,0 +1,463 @@
+# Plain CSS If: Draft 1.0
+
+*([Issue](https://github.com/sass/sass/issues/3886))*
+
+## Table of Contents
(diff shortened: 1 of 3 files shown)


proposal/plain-css-if.md
```diff
@@ -0,0 +1,463 @@
...
+1. `if()`'s semantics aren't strictly the same as other functions. It only ever
+   evaluates its second *or* third argument based on which one is selected by
+   the first; the other one is ignored, which is detectable if it would have
+   produced an error when evaluated. This historically made it difficult to
+   implement as a module-based function, although the infrastructure laid down
+   for [calculation functions] makes this much more feasible today.
+
+2. `if()` is part of the core language infrastructure (like a ternary expression
+   in other languages) and so requiring a module to be loaded to use it would
+   was more onerous than we wanted to impose on users.
```
stof:
```suggestion
   be more onerous than we wanted to impose on users.
```

Alternatively, remove `would` instead

proposal/plain-css-if.md
```diff
@@ -0,0 +1,463 @@
...
+top-level `;` which is never legal in the old syntax *or* a single top-level `:`
+that's not preceded by any top-level commas. Since the old syntax takes three
+comma-separated arguments, if it contains a top-level `:` it will either have a
+comma before it or a `...` after it, neither of which is valid in the new
+syntax.
+
+In Phase 1, we will also retain the `if()` function as defined at the top level
+(which is distinct from the expression we handle at parse-time). However,
+evaluating either the top-level function or the expression will produce a
+deprecation error named `if-function`.
```
stof, line 455: This means that `meta.call(meta.get-function('if))` is deprecated, right ?
This is good news to me, given that this usage never had the semantic of a true conditional (as it evaluates all arguments eagerly).

nex3: That's correct. I expect the migrator will migrate these to a local function that uses the new `@if` syntax.

jathak (approved):

PR #4128, opened 2025-09-02
Merge origin/main into feature.v2

nex3:


accepted/interleaved-declarations.md
```diff
@@ -98,16 +100,24 @@ The *current keyframe block* is the CSS keyframe block that was created by the
...
+* Let `parent` be the [current style rule], [keyframe block], or at-rule if one
+  exists; or the innermost if multiple exist.
+
+  [current style rule]: ../spec/style-rules.md#current-style-rule
+  [keyframe block]: ../spec/style-rules.md#current-style-rule
 
-* If `parent` isn't the last statement in its parent:
+* If `parent` is null, return null.
+
+* Otherwise, if `parent` is the last statement in its parent, return `parent`.
```
pamelalozano16, line 118: Something like `If parent is the last statement within its container, return parent` would be more clear, or if we want to refer to the grandparent something like `If parent is the last child of its parent, return parent.`

accepted/interleaved-declarations.md
```diff
@@ -98,16 +100,24 @@ The *current keyframe block* is the CSS keyframe block that was created by the
...
+
+* Otherwise:
```
pamelalozano16, line 120: `Otherwise, if parent is followed by other statements`:

accepted/interleaved-declarations.md
nex3: This is a merge PR, so it shouldn't make changes on top of the commits being merged in except as necessary to resolve conflicts. If you notice any issues in the underlying commits, please send a PR to the branch being merged.

pamelalozano16 (approved):

PR #4132, opened 2025-09-04
Add deprecation entry for case-insensitive function names

ghost:
I added the deprecation for the function names URL() and EXPRESSION ()


spec/deprecations.yaml
```diff
@@ -167,9 +167,17 @@ misplaced-rest:
   dart-sass:
     status: active
     deprecated: 1.91.0
+    
+case-insensitive-function-names:
+  description: 'Functions whose names differ from reserved names only by case are now deprecated. These cannot be invoked normally and will be disallowed in a future release.'
```
nex3, line 172: This should be a brief sentence fragment describing what's deprecated—its main purpose is showing up in the [Sass documentation](https://sass-lang.com/documentation/cli/dart-sass/#fatal-deprecation) where it should fit on a single line. See other deprecation descriptions for an idea of what this should look like.

I think something like "User-defined functions with capitalized reserved names." would be better.

spec/deprecations.yaml
```diff
@@ -167,9 +167,17 @@ misplaced-rest:
   dart-sass:
     status: active
     deprecated: 1.91.0
+    
+case-insensitive-function-names:
```
nex3, line 171: The spec says that the deprecation name should be `function-case`.

nex3 (changes requested):
This PR should also move the spec and its changelog from `proposal/` to `accepted/`.

spec/deprecations.yaml
ghost: I will make the changes

spec/deprecations.yaml
ghost: Am trying to make the correction to the PR but am running into this error while trying to run the language test 
> 
> Error: Dart Sass process exited unexpectedly with code 254.
> ../dart-sass/lib/src/visitor/evaluate.dart:603:29: Error: Member not found: 'withPrivate'.
>                 Deprecation.withPrivate,
>                             ^^^^^^^^^^^
> ../dart-sass/lib/src/parse/stylesheet.dart:1657:36: Error: Member not found: 'withPrivate'.
>           deprecation: Deprecation.withPrivate,
>                                    ^^^^^^^^^^^
> ../dart-sass/lib/src/visitor/async_evaluate.dart:595:29: Error: Member not found: 'withPrivate'.
>                 Deprecation.withPrivate,
>                             ^^^^^^^^^^^
> 
> ~/css/sass-spec $

nex3: You may need to pull in the latest changes from the base repo.

PR #4170, opened 2025-12-27
Enhance sourceMapIncludeSources option

ntkme:
https://github.com/sass/dart-sass/issues/2712

This PR add new supported values for `sourceMapIncludeSources`:
- `'auto'` - same as `undefined`
-  `'always'` - same as `true`
-  `'never'` - same as `false`

The old boolean values `true` and `false` are deprecated and will be removed.


EMBEDDED_PROTOCOL_CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 4.0.0
```
nex3: According to the embedded protocol versioning policy, this isn't a breaking change because the wire protocol is backwards-compatible from the host's perspective. It should be 3.3.0.

js-api-doc/options.d.ts
```diff
@@ -352,10 +367,10 @@ export interface Options<sync extends 'sync' | 'async'> {
    *
    * This option has no effect if {@link sourceMap} is `false`.
```
nex3, line 368: The docs here should now say "When" rather than "Whether". We should also indicate that passing a boolean is deprecated.

nex3 (changes requested):
I'd like to deprecate passing `true` and `false` to `sourceMapIncludeSources`, which means this needs to be a full proposal document with an open comment period.

ntkme:
@nex3 Should we add a deprecation option for this? In my opinion this is not a language depreciation, and also not a major API deprecate like the legacy API itself. I'm afraid having a deprecation option for such a subtle change would introduce even more noisy. I'm thinking to only mention the deprecation in documentation, and at dart-sass 2.0 we remove the support for boolean as a breaking change. What do you think?

spec/deprecations.yaml
```diff
@@ -180,3 +180,9 @@ if-function:
   dart-sass:
     status: active
     deprecated: 1.95.0
+
+source-map-include-sources-boolean:
+  description: Passing a boolean value to as Options.sourceMapIncludeSources.
+  dart-sass:
+    status: active
+    deprecated: 1.99.0
```
nex3: This should probably be 1.98, since it hasn't yet been released.

EMBEDDED_PROTOCOL_CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 3.3.0
+
+* Add SourceMapIncludeSources options as an enum.
+
+  Note: although this is a breaking change for protobuf generated code, the
+  encoded protobuf payload is backward compatible, meaning it does not affect
+  previously built host on newly built compiler.
```
nex3: Nit:
```suggestion
  a previously built host on a newly built compiler.
```

nex3 (approved):

spec/js-api/options.d.ts.md
```diff
@@ -292,13 +304,19 @@ sourceMap?: boolean;
...
+text in [`CompileResult.sourceMap`] for each source.
 
-Defaults to false.
+If `'always'` (or true), the compiler must include the full Sass source text in
+[`CompileResult.sourceMap`] for all sources.
+
+If `'never'`, the compiler must not include the full Sass source text in
+[`CompileResult.sourceMap`] for any sources.
+
+Defaults to 'auto'.
```
nex3: This should match the text in the proposal.

nex3 (changes requested):
This also needs to include the proposal's changes for `importer.d.ts.md`.

PR #4175, opened 2026-01-01
Add deprecation entry for case-insensitive function names

kantnero:
Part of case-insensitive function deprecation:

- Dart Sass emits the warning: https://github.com/sass/dart-sass/pull/2716
- Tests in sass-spec: https://github.com/sass/sass-spec/pull/2100

### DIFF
--- a/spec/deprecations.yaml
+++ b/spec/deprecations.yaml
@@ -180,3 +180,9 @@ if-function:
   dart-sass:
     status: active
     deprecated: 1.95.0
+
+case-insensitive-function:
+  description: Function names whose name differ from reserved names.
+  dart-sass:
+    status: active
+    deprecated: 1.96.0


kantnero:
Part of case-insensitive function deprecation:

- Dart Sass emits the warning: https://github.com/sass/dart-sass/pull/2716
- Tests in sass-spec: https://github.com/sass/sass-spec/pull/2100

spec/deprecations.yaml
```diff
@@ -180,3 +180,9 @@ if-function:
   dart-sass:
     status: active
     deprecated: 1.95.0
+
+case-insensitive-function:
```
nex3, line 184: The proposal says that this deprecation should be called `function-case`.

spec/deprecations.yaml
```diff
@@ -180,3 +180,9 @@ if-function:
...
+  description: Function names whose name differ from reserved names.
```
nex3, line 185:
```suggestion
  description: Uppercase reserved function names.
```

nex3 (changes requested):
Thanks for looking into this!

nex3:
This should also move the function name proposal into the `accepted` directory.

PR #4181, opened 2026-01-21
Proposal for enhancing sourceMapIncludeSources option

ntkme:
https://github.com/sass/dart-sass/issues/2712


nex3:
Can you grant me push permissions for this PR?

ntkme:
> Can you grant me push permissions for this PR?

I'd like to but looks like because PR is from a fork in an org account, it's not possible: https://github.com/orgs/community/discussions/5634

@nex3 As a workaround, I send you an invite to the sass-contrib org.

nex3:
I accepted the invite, but it's still not letting me push:

```
128 git … push -v sass-contrib refs/heads/spec-source-map-include-sources\:refs/heads/spec-source-map-include-sources
Pushing to github.com:sass-contrib/sass
ERROR: Permission to sass-contrib/sass.git denied to nex3.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

ntkme:
Ok, I'm lazy and I just made you the owner :-)

I'm not going to ask you to do anything as an owner in that org, but I think you should have access to those community projects if you want to do any changes.

proposal/source-map-include-sources.md
```diff
@@ -0,0 +1,97 @@
...
+* If the value is `'never'`, no stylesheets' contents are included in the source
+  map's `sourcesContent`.
+
+* If the value is `'always'`, source contents are included in the source map's
+  `sourcesContent` for all stylesheets.
+
+### `ImporterResult.sourceMapUrl`
+
+* If the `sourceMapUrl` is not explictly defined in `ImportResult`, the
+  canonical URL is used as fallback.
```
nex3: Making this part of the spec for the `sourceMapUrl` field itself implies that we'd somehow inject the canonical URL when the API constructs an `ImporterResult` object, which isn't feasible. This should instead modify the way the `ImporterResult` is handled in the `Importer` section of the spec.

proposal/source-map-include-sources.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Deprecation Process
+
+The deprecation process will be divided into two phases:
+
+### Phase 1
```
nex3, line 85: This should define a named deprecation and emit it when a boolean is passed to `sourceMapIncludeSources`.

nex3 (changes requested):

proposal/source-map-include-sources.md
ntkme: What about this?

``` md
When constructing source map, the canonical URL is used for `sources` field if `ImporterResult.sourceMapUrl` is not explicitly defined
```

proposal/source-map-include-sources.md
ntkme: I made a comment regarding this earlier in the other PR: https://github.com/sass/sass/pull/4170#issuecomment-3780664378

Also paste it here for visibility:

> In my opinion this is not a language depreciation, and also not a major API deprecate like the legacy API itself. I'm afraid having a deprecation option for such a subtle change would introduce even more noises. I'm thinking to only mention the deprecation in documentation, and at dart-sass 2.0 we remove the support for boolean as a breaking change - that users would only need to make an one time change when they upgrade to 2.0.

proposal/source-map-include-sources.md
nex3: The way we usually do this is to define it in terms of changes to the existing spec text. Something like

```md
### `Importer`

Replace the bullet point "..." with "...".
```

proposal/source-map-include-sources.md
nex3: I really don't like the idea of deprecating something without providing any advance feedback to users that it's coming. Ideally, a user on the last version of the 1.x line with zero deprecation warnings should be able to upgrade to 2.0 without any changes at all. If it were possible to mark certain union members as deprecated in TypeScript that _might_ be sufficient, but I don't believe it is. And once we're emitting a message, the only standard way we have of doing that is to emit it as a specific named deprecation.

ntkme: > Ideally, a user on the last version of the 1.x line with zero deprecation warnings should be able to upgrade to 2.0 without any changes at all.

I disagree with this idea because it’s against semantic versioning. If user can smoothly upgrade to 2.0 with zero incompatibility, it should not even be a 2.0 release, it should just be 1.x. I understand that for a lot of sass language changes we don’t follow semantic versioning, due to changes in upstream CSS spec, which is why emit depreciations are necessary. However, I personally think this policy shouldn’t really be expanded to API changes, as they are supposed to follow semantic versioning.

Lots of libraries make breaking changes with major semantic versions without any prior notice. The important thing is to provide clear migration guidance for major breaking changes.

I’m open to add a depreciation if you think otherwise, in that case what should be the name? For this one I couldn’t come up a short yet descriptive name.

nex3: Specifically I want a user to be able to smoothly upgrade from 1.x to 2.0 _if they have no warnings_. Semantic versioning is still relevant, because it describes when a deprecation warning becomes an error (or a behavior change), but to whatever degree possible I don't want users to have to scour the changelog and their codebases to try to manually discern what they need to fix. If we can tell them directly, we should, and in this case we definitely can.

I'd name this `source-map-include-sources-bool`. It's not the shortest, but neither is the option itself.

proposal/source-map-include-sources.md
ntkme: The current spec only says:

> * If `result.sourceMapUrl` is defined and the implementation generates a source
  map, the implementation must use this URL in the source map to refer to source
  spans in `result.contents`.

Currently what to do with undefined `result.sourceMapUrl` is an undefined behavior. I'm not sure what's the best language here. Please feel free to edit the PR.

nex3 (approved):

PR #4206, opened 2026-03-27
Add support for top-level &

### LINKED ISSUE #4204: Consider supporting `&` as a top level selector (at least within `@scope`)
* [x] Specification (https://github.com/sass/sass/pull/4206)
* [x] Tests (https://github.com/sass/sass-spec/pull/2129)
* [x] Dart Sass (https://github.com/sass/dart-sass/pull/2758)
* [x] Documentation (https://github.com/sass/sass-site/pull/1568)

---

Take the following SCSS:

```scss
// https://sass-lang.com/playground/#eJwzNFRTqOZSUEjOz8kvslIoSk2x5qoFADrJBeQ=
& {
  color: red;
}
```

This will result in the error "Top-level selectors may not contain the parent selector &". In vanilla CSS [this is supported](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Nesting_selector#using_outside_nested_rule), where this `&` is an alias for `:scope`, which (outside of `@scope`) is the same as `:root`.

Within an `@scope` block, `&` similarly refers to `:scope` but also has a specificity of 0, so it behaves the same as `:where(:scope)`. (Alternatively, you can say that is equivalent to not having the `& {}` at all.) In Sass this is similarly not supported:

```scss
// https://sass-lang.com/playground/#eJwzNHQoTs4vSFXQ0EvLz9dUqOZSUFADkwoKyfk5+UVWCkWpKdZAfi1XLQAZEwwN
@scope (.foo) {
  & {
    color: red;
  }
}
```

I think that at minimum the `&` within `@scope` is worth supporting. It could be transpiled to CSS `:where(:scope)`, or maybe just straight up `&` since there are no browsers that support `@scope` but don't support `&`.

As an example use case:

```css
@scope (.foo) {
  &, & * {
    color: red;
  }
}
```

Without support in Sass, I need to write this as:

```css
@scope (.foo) {
  :where(:scope), :where(:scope) * {
    color: red;
  }
}
```

nex3:
This is a fast-track change.

Closes #4204

### DIFF
--- a/spec/style-rules.md
+++ b/spec/style-rules.md
@@ -74,8 +74,8 @@ To execute a style rule `rule`:
 * Otherwise, if `selector` contains one or more parent selectors with suffixes,
   throw an error.
 
-  > Suffixes on parent selectors are a Sass extension and so *only* allowed in
-  > Sass nesting contexts.
+  > Suffixes on parent selectors are a Sass extension and so are *only* allowed
+  > in Sass nesting contexts.
 
 * Let `css` be a CSS style rule with selector `selector`.
 
--- a/spec/style-rules.md
+++ b/spec/style-rules.md
@@ -71,24 +71,35 @@ To execute a style rule `rule`:
 
   [descendant combinator]: https://www.w3.org/TR/selectors-3/#descendant-combinators
 
-* Otherwise, if `selector` contains one or more parent selectors and `rule`'s
-  stylesheet wasn't [parsed as CSS], throw an error.
+* Otherwise, if `selector` contains one or more parent selectors with suffixes,
+  throw an error.
+
+  > Suffixes on parent selectors are a Sass extension and so *only* allowed in
+  > Sass nesting contexts.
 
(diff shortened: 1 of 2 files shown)


spec/style-rules.md
```diff
@@ -71,25 +71,36 @@ To execute a style rule `rule`:
 
   [descendant combinator]: https://www.w3.org/TR/selectors-3/#descendant-combinators
 
-* Otherwise, if `selector` contains one or more parent selectors and `rule`'s
-  stylesheet wasn't [parsed as CSS], throw an error.
+* Otherwise, if `selector` contains one or more parent selectors with suffixes,
+  throw an error.
+
+  > Suffixes on parent selectors are a Sass extension and so *only* allowed in
```
Goodwine: Nit: this sentence reads a bit awkward. Is it missing "are" before "only"?

Goodwine (approved):

Issue #4217, opened 2026-04-22
Support analogous sets of missing components in color conversions


nex3:
This tracks adding support for the changes made in https://github.com/w3c/csswg-drafts/issues/10210 to Sass.

Although this hasn't yet landed in browsers, I think it's worth implementing eagerly. Browser support for missing components is already fairly inconsistent, especially in terms of plain color conversions where this will be most visible to Sass users. In addition, this is unavoidably a breaking change, and one that we can't easily provide a deprecation cycle for, so landing it sooner will reduce the surface area of existing stylesheets that may be affected negatively by the breakage (although I expect this surface area to be small regardless, since missing channels are already a niche feature).

- [x] Proposal (https://github.com/sass/sass/pull/4218)
- [x] Integrate into spec (https://github.com/sass/sass/pull/4242)
- [x] Dart Sass (https://github.com/sass/dart-sass/pull/2810)
- [x] Tests (https://github.com/sass/sass-spec/pull/2159)

nex3:
I guess I never set an explicit timeline for this, but it's been three months, so I'm going to say it's done cooking.

PR #4218, opened 2026-04-22
[Analogous Sets] Add a proposal

nex3:
See https://github.com/sass/sass/issues/4217

### DIFF
--- a/proposal/analogous-sets.md
+++ b/proposal/analogous-sets.md
@@ -20,12 +20,14 @@
 > This section is non-normative.
 
 When Sass originally implemented support for the expanded color spaces
-introduced by [CSS Color 4], it contained the concept of "analogous components"
-across two color spaces whose missing status would be maintained during (some)
-conversions. At the time, analogous components were limited to individual
-component pairs: `red`/`green`/`blue` across the various RGB-style spaces,
-`lightness` between HSL/Lab/LCH, and so on. Sass implemented these by retaining
-missing channels during all color conversions.
+introduced by [CSS Color 4], it contained the concept of "[analogous
+components]" across two color spaces whose missing status would be maintained
+during (some) conversions. At the time, analogous components were limited to
+individual component pairs: `red`/`green`/`blue` across the various RGB-style
+spaces, `lightness` between HSL/Lab/LCH, and so on. Sass implemented these by
+retaining missing channels during all color conversions.
+
+[analogous components]: https://drafts.csswg.org/css-color-4/#analogous-components
 
 Since then, the CSS spec has expanded the notion of analogous components to
 include "analogous sets" as well, defined as the complement of any subset of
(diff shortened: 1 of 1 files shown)


proposal/analogous-sets.md
```diff
@@ -0,0 +1,126 @@
...
+  * [Analogous Mappings](#analogous-mappings)
+* [Procedures](#procedures)
+  * [Converting a Color](#converting-a-color)
+
+## Background
+
+> This section is non-normative.
+
+When Sass originally implemented support for the expanded color spaces
+introduced by [CSS Color 4], it contained the concept of "analogous components"
```
pamelalozano16: Maybe link for context "[analogous components](https://drafts.csswg.org/css-color-4/#analogous-components)" ?

proposal/analogous-sets.md
```diff
@@ -0,0 +1,126 @@
...
+[analogous components].
+
+[analogous sets]: https://drafts.csswg.org/css-color-4/#analogous-set
+[analogous components]: https://drafts.csswg.org/css-color-4/#analogous-components
+
+> Each pair of color spaces has a finite, well-defined set of analogous
+> mappings. Per the CSS spec, the set of all components of each space always
+> represents an analogous mapping.
+>
+> For example, the mappings between Lab and LCH are:
```
pamelalozano16, line 111: I would also add an example like this for context:

For example:
```css
.example {
  background: color-mix(in lch, red 50%, blue 50%);
}
```
> Here red and blue are typically defined in the RGB color model, but you told CSS to mix them in LCH
> So what happens?
> 1. Convert both colors into the LCH color space
> 2. Match *analogous components*:
> - Lightness ↔ Lightness (L)
> - Chroma ↔ intensity (C)
> - Hue ↔ Hue (H)
> 3. Interpolate each component
> 4. Convert the result back to a display color

pamelalozano16: Also I think it'd be more useful to have these definitions up there in "background"

proposal/analogous-sets.md
```diff
@@ -0,0 +1,126 @@
...
+at all. For those who are, the new behavior is a clear and immediate
+improvement. It more accurately represents the intended semantics of missing
+channels as having absent or templatable values, and so any behavioral
+differences are likely to be improvements rather than regressions.
+
+## Definitions
+
+### Missing Components
+
+Remove the definition of "analogous components" from [the missing components
```
pamelalozano16: I'm kind of lost in this sentence, do you mean when a user doesn't define a channel? Maybe have something like this instead?
```
In modern CSS (CSS Color 4), you can define a color but intentionally leave one or more of its channels blank by using the keyword none.

For example: lch(50% none none)
This creates a color with 50% lightness, but no defined chroma (saturation) or hue.
```

pamelalozano16 (approved):

proposal/analogous-sets.md
nex3: Analogous components are actually unrelated to mixing colors normally—when you write `color-mix(in lch, red 50%, blue 50%)`, it converts both both colors into the same space (LCH), at which point it doesn't need to know which components are "analogous" because the two converted colors have the same components. The notion is _only_ used for determining when and how to preserve missing channels through color conversions.

proposal/analogous-sets.md
nex3: This sentence is describing how to _change the specification_ when this proposal is applied; it's not a specification in and of itself. I've reworded it to be a little clearer about that.

Issue #4224, opened 2026-05-22
NodePackageImport does not load import only files when package has an `exports` field


romaricpascal:
When using `pkg:` URL with `@import` to load a stylesheet from a library that provides both a module and an import-only files, Sass will load the module rather than the import-only file if the package has an `exports` field defined in its package.json.

See https://github.com/romaricpascal/sass-pkg-import-only-repro for a reproduction.

This seems due to the NodePackageImporter only [accounting for import-only files when there's no `exports` field](https://github.com/sass/dart-sass/blob/5fd18c75e31a855476059fb6fb0c6aa829292739/lib/src/importer/node_package.dart#L141-L155). Is it intentional?

---

- [x] Specification (https://github.com/sass/sass/pull/4225)
- [x] Dart Sass (https://github.com/sass/dart-sass/pull/2772)
- [x] Tests (https://github.com/sass/sass-spec/pull/2139)

nex3:
This is intended to the extent that it is what [the specification](https://github.com/sass/sass/blob/main/spec/modules.md#export-load-paths) says, but I agree that it's not desirable, so it should be considered a spec bug.

PR #4225, opened 2026-05-23
Consistently load import-only files with the Node Package Importer

nex3:
This is a fix for a clear specification bug, so it is exempt from the usual process for changing the spec as well as the usual deprecation process. I consider it highly unlikely that anyone was depending on this clearly-undesirable behavior, so a deprecation cycle would only serve to delay fixing the issue for users who are affected.

See #4224

### DIFF
--- a/spec/modules.md
+++ b/spec/modules.md
@@ -16,7 +16,7 @@
   * [Node Package Importer](#node-package-importer)
   * [Global Importer List](#global-importer-list)
   * [Underscore-Insensitive](#underscore-insensitive)
-  * [Export Load Paths](#export-load-paths)
+  * [Candidate URLs](#candidate-urls)
   * [Basename](#basename)
   * [Dirname](#dirname)
 * [Syntax](#syntax)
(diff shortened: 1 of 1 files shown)


spec/modules.md
```diff
@@ -528,6 +558,26 @@ either another URL that's guaranteed to point to a file on disk or null.
 
 * Return null.
 
+### Resolving a `file:` URL for Imports
+
+> For historical reasons, the standard means of resolving import-only files will
+> load `foo.import.scss` even if the non-import-only file is `_foo.scss` or
+> `_foo.sass`. This means that it can't use this algorithm, which ends up only
```
Goodwine: Nit: I think it would be good to clarify what "it" means here, because it's not immediately clear to me whether "it" is the non-import-only files or the inport-only files

spec/modules.md
```diff
@@ -528,6 +558,26 @@ either another URL that's guaranteed to point to a file on disk or null.
 
 * Return null.
 
+### Resolving a `file:` URL for Imports
+
+> For historical reasons, the standard means of resolving import-only files will
```
Goodwine: Nit: what is an "import-only" file?

spec/modules.md
```diff
@@ -314,6 +316,34 @@ U+002D HYPHEN-MINUS and U+005F LOW LINE to be equal to one another.
 > underscores for backwards-compatibility. This insensitivity is generally only
 > used for Sass member names, not any other identifiers.
 
+### Candidate URLs
+
+The *candidate URLs* for a given URL is the set of URLs of equal precedence that
+this URL might be resolved as by Sass.
+
+> This does *not* include import-only or index URLs, because those have higher
```
Goodwine, line 324: nice, this comment answered a question that popped up in my mind :)

Goodwine (approved):
overal LGTM, just some comments to help clarify the "Resolving a `file:` URL for Imports" section

spec/modules.md
nex3: "it" refers to the algorithm. I've clarified and included a link.

spec/modules.md
nex3: This is a term from the Sass documentation (https://sass-lang.com/documentation/at-rules/import/#import-only-files). It's not a specification term, but this isn't normative text anyway.

Issue #739, opened 2013-05-08
Dynamic Dependencies


chriseppstein:
- [x] Proposal (https://github.com/sass/sass/pull/4243)
- [x] Flush to spec (https://github.com/sass/sass/pull/4247)
- [x] Dart Sass (https://github.com/sass/dart-sass/pull/2861)
- [x] Node.js embedded host (https://github.com/sass/embedded-host-node/pull/446)
- [x] Tests (https://github.com/sass/sass-spec/pull/2170)
- [x] Website (https://github.com/sass/sass-site/pull/1668)

---

There's a way to compute dependencies efficiently and reliably on a first pass compile that would allow for the second pass to be optimized without needing to statically analyze the sass files for their `@import` dependencies. This would also allow for generic dependencies on things like images when used with compass sprites, or image-dimension helpers.

chriseppstein:
Doing this will allow us to remove the limitations on having no scripting in `@import` directives and allow `@import` to be used in mixins and conditionals.

See #408

nex3:
...what is this way of which you speak?

chriseppstein:
I'll update with details next week :)

chriseppstein:
The only reason Sass needs to track dependencies is to answer the question "Should this sass file be re-compiled?". We use the list of imports to decide whether the file is out of date by checking its timestamp as well as the timestamp of the all transitive dependencies.

However, a change to a sass file is not the only reason why a stylesheet should need to be re-compiled. For instance, the compass `image-width` and `image-height` helper function return the dimensions of an image. If that image changes, the stylesheet is now potentially out of date and should be recompiled.

So my idea is to build a dependency graph during compilation of all the files that a sass file depends on. On first compile, we incur a compilation penalty if this dependency graph is not cached. But in most of these cases, it will be true that the css file itself is not present[^1]. When a new dependency is added, this is necessarily caused by a change to tracked dependency[^2] or the file itself, so we can still answer the question "is this file out of date".

So what I would do is introduce a Ruby API for declaring a dependency. This would be called when a file is imported, but other APIs like compass's image helpers could piggy back on it as well. Obviously, this should not be filesystem centric so that it will continue to work with `Importer` and `Cache` implementations.

[...]

PR #4243, opened 2026-07-28
[meta.load] Add a proposal

nex3:
See https://github.com/sass/sass/issues/739

### DIFF
--- a/proposal/meta-load.d.ts.md
+++ b/proposal/meta-load.d.ts.md
@@ -1,0 +1,342 @@
+# `meta.load()`: Draft 1.0
+
+*([Issue](https://github.com/sass/sass/issues/739))*
+
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+  * [Design Decisions](#design-decisions)
+    * [A New Type](#a-new-type)
+    * [Inspect Format](#inspect-format)
+    * [No JavaScript Constructor](#no-javascript-constructor)
+    * [No JavaScript Members](#no-javascript-members)
+* [Types](#types)
+  * [Operations](#operations)
+    * [Equality](#equality)
+  * [Serialization](#serialization)
+* [Functions](#functions)
+  * [`meta.type-of()`](#metatype-of)
+  * [`meta.load()`](#metaload)
+  * [`meta.get-module()`](#metaget-module)
+  * [Existing Functions](#existing-functions)
+* [Mixins](#mixins)
+  * [`meta.css()`](#metacss)
+* [JavaScript API](#javascript-api)
+  * [Types](#types-1)
+    * [`SassModule`](#sassmodule)
+      * [`assertModule`](#assertmodule)
+      * [`internal`](#internal)
+      * [Constructor](#constructor)
+* [Embedded Protocol](#embedded-protocol)
+
+## Background
+
+> This section is non-normative.
+
+The ability to load stylesheets dynamically is one of the oldest and most
+popular feature requests in Sass. Many requests have taken the form of allowing
+interpolation in `@import` and later `@use` rules, which is unfortunately a
+non-starter as it violates the [module system goal] of static analyzability.
+Instead, the [`meta.load-css()` mixin] was added, which includes CSS from a
+stylesheet based on an argument which can be dynamic.
+
+[module system goal]: ../accepted/module-system.md#low-level
+[`meta.load-css()` mixin]: ../accepted/module-system.md#metaload-css
+
+This has one notable drawback, though: it doesn't provide any means of accessing
+stylesheet *members* dynamically. `meta.load-css()` evaluates the stylesheet's
+module to CSS and then discards it, leaving any variables, mixins, or functions
+it defined inaccessible to the caller. Although the CSS is the primary use case
+of dynamic includes, this is a notable missing feature.
+
+## Summary
+
+> This section is non-normative.
+
+This proposal adds a new data type, "module", which represents a loaded Sass
+module. As with first-class functions and mixins, this type can't be constructed
+literally, but can be returned by Sass core library functions. Values of this
+type can be used as the `$module` parameter in
+[`meta.global-variable-exists()`], [`meta.module-variables()`],
+[`meta.function-exists()`], [`meta.get-function()`],
+[`meta.module-functions()`], [`meta.mixin-exists()`], [`meta.get-mixin()`], and
+[`meta.module-mixins()`], allowing the caller to access its variables,
+functions, and mixins.
+
+[`meta.function-exists()`]: https://sass-lang.com/documentation/modules/meta/#function-exists
+[`meta.get-function()`]: https://sass-lang.com/documentation/modules/meta/#get-function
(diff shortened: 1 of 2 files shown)


proposal/meta-load.d.ts.md
```diff
@@ -0,0 +1,342 @@
...
+[`meta.global-variable-exists()`]: https://sass-lang.com/documentation/modules/meta/#global-variable-exists
+[`meta.module-variables()`]: https://sass-lang.com/documentation/modules/meta/#module-variables
+[`meta.mixin-exists()`]: https://sass-lang.com/documentation/modules/meta/#mixin-exists
+[`meta.get-mixin()`]: https://sass-lang.com/documentation/modules/meta/#get-mixin
+[`meta.module-mixins()`]: https://sass-lang.com/documentation/modules/meta/#module-mixins
+
+This proposal adds two new functions:
+
+* `meta.load($url, $with: null)` takes the same arguments as `meta.load-css()`,
+  but instead of directly including the CSS it returns the loaded module.
```
Goodwine, line 79: Should this be `in addition` rather than `instead`? 

Oh, I think I am slightly confused. `load-css()` loads the CSS, but `load()` won't, is that the case?

proposal/meta-load.d.ts.md
```diff
@@ -0,0 +1,342 @@
...
+
+* `meta.get-module($module)` takes the namespace of a `@use` rule in the current
+  stylesheet and returns the module loaded by that rule. This does *not*
+  re-evaluate the module.
+
+It also adds one new mixin:
+
+* `meta.css($module)` includes a copy of the given module's CSS, like
```
Goodwine, line 87: What is the use case for loading the CSS if it was already loaded? is this meant ot be used "in tandem" with `load()` as in..

would this make `load-css()` essentially the same as `@mixin load-css{ @include meta.load(); @include meta.css() }`?

proposal/meta-load.d.ts.md
```diff
@@ -0,0 +1,342 @@
...
+
+The only operation permitted for module objects is checking for equality. All
+other operations throw an error.
+
+#### Equality
+
+Module values use identity equality.
+
+> This means that there can be two modules with the same canonical URL that are
+> nevertheless unequal, in the case that `meta.load()` was called twice to load
```
Goodwine: The `meta.load` function states that the file gets loaded via the `loading` link, which shows that it gets loaded via the "Loading Module Procedure", which then states that if a module was already loaded, then the _same_ module gets returned, so why are they unequal?

proposal/meta-load.d.ts.md
```diff
@@ -0,0 +1,342 @@
...
+
+* If `$url` isn't a string, throw an error.
+
+* If `$with` isn't null or a map, or if it's a map with any keys that aren't
+  strings, throw an error.
+
+* Let `config` be a configuration whose variable names and values are given by
+  `$with` if `$with` isn't null, or the empty configuration otherwise.
+
+* Return the result of [loading] `$url` with `config`.
```
Goodwine, line 214: Something I'm not sure that I find clear is whether `loading` implies also outputting the CSS from the module 🤔

proposal/meta-load.d.ts.md
```diff
@@ -0,0 +1,342 @@
...
+
+  [loading]: ../spec/modules.md#loading-a-module
+
+### `meta.get-module()`
+
+Add a new function to the `sass:meta` module:
+
+```
+meta.get-module($module)
```
Goodwine, line 225: If my understanding is correctly:

```scss
$mod: my-module;
@debug meta.get-module($mod); // null
@debug meta.load($mod);       // get-module(my-module)
@debug meta.get-module($mod); // null. doesn't matter it was loaded by meta.load()
@debug meta.get-module(meta.load($mod)); // get-module(my-module)
```
i.e. it exclusively works for `@use`. But why?

proposal/meta-load.d.ts.md
```diff
@@ -0,0 +1,342 @@
...
+export class SassModule extends Value {
+```
+
+##### `assertModule`
+
+Returns `this` if it's a [`SassModule`] and throws an error otherwise.
+
+[`SassModule`]: #sassmodule
+
+> The `name` parameter may be used for error reporting.
```
Goodwine, line 298: I find it slightly confusing this way. Does a module not have its own name? Is this meant to overwrite that name if so? And what is the use case for this?

spec/built-in-modules/meta.md
```diff
@@ -392,6 +392,9 @@ load-css($url, $with: null)
 
 * If `$url` isn't a string, throw an error.
 
+* If `$with` isn't null or a map, or if it's a map with any keys that aren't
```
Goodwine: It seems this is a bug fix in the spec. IMO this could be landed as a separate PR for better tracking since the current PR text does not mention this fix

proposal/meta-load.d.ts.md
nex3: `load-css()` *includes* the CSS at the point the mixin is invoked. `load()` is a function, not a mixin, so it can't directly emit CSS. Instead, users can later pass the module object it returns to the new `meta.css()` mixin to include it.

proposal/meta-load.d.ts.md
nex3: This would make `load-css` implementable as

```scss
@mixin load-css($url, $with: null) {
  $module: meta.load($url, $with: $with);
  @include meta.css($module);
}
```

There's not a specific use-case for re-including already-loaded CSS, it's just a natural consequence of allowing users to get first-class module objects for already-loaded modules.

proposal/meta-load.d.ts.md
nex3: Good catch! I was under the impression that the existing behavior for `meta.load-css()` would re-execute a module that was already loaded via `@use`, but you're right that that doesn't match the spec. In fact, my memory was wrong, and the module is only loaded once, so these would be equal.

proposal/meta-load.d.ts.md
nex3: It does not. I'll add a non-normative note.

proposal/meta-load.d.ts.md
nex3: Small nit: `meta.get-module()` throws an error if the module isn't found, rather than returning `null`. Otherwise, your example is correct, assuming there's no `@use` rule in the stylesheet.

`meta.get-module()` exists so that users can access statically-loaded modules as first-class modules. It's not strictly necessary, since you could write `meta.load()` instead with the URL used in the `@use` rule and it would get de-duplicated to the same module object. If you think it's more complexity than it's worth, I could drop it from the proposal.

proposal/meta-load.d.ts.md
nex3: This is intended to be the JS-level argument name, not the module's name. It's common to all `assert{type}` methods in the JS API.

Goodwine (approved):

Goodwine (approved):

jathak (approved):

Issue #4249, opened 2026-09-16
Improving the format for stack traces in the embedded protocol


nex3:
I'm spinning this off of https://github.com/sass/dart-sass/issues/2775 to track adding a more machine-friendly stack trace to the embedded protocol, since that's a specification change.

- [ ] Proposal (https://github.com/sass/sass/pull/4250)
- [ ] Flush to spec
- [ ] Dart Sass
- [ ] Node.js Embedded Host

nex3:
I'm going to give this a month for comment, since it's a breaking change.

PR #4250, opened 2026-09-16
[Embedded Stack Trace] Add a proposal

nex3:
See https://github.com/sass/sass/issues/4249

@ntkme Informally requesting a review from you, since you're the motivating user in this case (and very knowledgeable about the embedded protocol in general).


ntkme:
Overall look good to me. Only comment is that I would like to request an additional option to omit the pre-formatted stack trace in `formatted`, so that I can attach a host-formatted stack trace and attach it to the pretty formatted error message with context.

nex3:
If you're going to custom format the stack trace, would it also make sense to custom format the span? There's already a separate `message` field you can use to build a hand-formatted message.

ntkme:
The span does not contain any relative path that I need to worry about, and the pretty format from compiler is somewhat complex to duplicate. Therefore, I would reuse the top half of `formatted` (error message + span).

nex3:
Omitting the stack trace raises the further question of whether someone might want to omit just the span, or potentially both (since the formatted message itself isn't always identical to the raw message). I wonder if it would make sense to split the `formatted` field into something like

```proto
message FormattedMessage {
  string message = 1;
  optional string span = 2;
  string stack_trace = 3;
}
```

and recommending that users concatenate it by hand as desired. That does make the embedded host's job a bit more challenging, though.

A third option would be to emit the formatted message as we have been, but provide indices for the beginning of the span and the beginning of the stack trace so that it's easy for a host to split it apart if desired. This is a more indirect representation than I'd usually prefer, but it does have the substantial benefit of making the most desirable default behavior—presenting the fully formatted message to the user as-is—the easiest to accomplish.

ntkme:
The third option also works.

It’s a little bit informal, but I think it’s still reasonable.

ntkme:
It would be nice to clarify whether the offset at the newline character or it's after the newline character. I guess it makes sense to have the offset starting after the newline character?

proposal/embedded-stack-trace.md
```diff
@@ -0,0 +1,164 @@
...
+// The 0-based offset in `formatted` of the beginning of the formatted stack
+// trace (after the previous newline).
+//
+// This is intended to make it possible for hosts to re-order or remove segments
+// of the formatted message without being tightly coupled to its specific
+// formatting.
+uint32 formatted_stack_trace_offset = 7;
+```
+
+### `CompileFailure`
```
ntkme, line 135: I think this is a typo?

```suggestion
### `LogEvent`
```

proposal/embedded-stack-trace.md
```diff
@@ -0,0 +1,164 @@
...
+## Table of Contents
+
+* [Background](#background)
+* [Summary](#summary)
+  * [Design Decisions](#design-decisions)
+    * [Formatted Offset Fields](#formatted-offset-fields)
+* [Embedded Protocol](#embedded-protocol)
+  * [`StackFrame`](#stackframe)
+  * [`CompileFailure`](#compilefailure)
+  * [`CompileFailure`](#compilefailure-1)
```
ntkme, line 14:
```suggestion
  * [`LogEvent`](#logevent)
```

ntkme (approved):
LTGM. Thanks.

jathak (approved):

PR #4253, opened 2026-09-22
[Deprecations Except] Add a proposal

nex3:


proposal/deprecations-except.d.ts.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+In addition, we'll support the special syntax `--fatal-deprecations=dart-sass-2`
+which will select the specific deprecations that will be enabled in Dart Sass 2.
+
+## Types
+
+### `DeprecationSelector`
+
+A value that can be selects a set of [deprecations].
```
jathak:
```suggestion
A value that selects a set of [deprecations].
```

jathak (commented):
Do we want to also support this in the embedded protocol?

jathak (approved):

